PR #122, opened 2023-03-13
Update paragraph

alimpens:


packages/css/src/paragraph/paragraph.scss
```diff
@@ -5,28 +5,45 @@
...
+    overwrite that css rule with their own value.
+  */
+  font-size: var(
+    --amsterdam-paragraph-font-size,
+    calculate-fluid-style($amsterdam-typography-font-size-18, $amsterdam-typography-font-size-24)
+  );
+  font-weight: var(--amsterdam-paragraph-font-weight);
+  line-height: var(--amsterdam-paragraph-line-height);
   margin-block: 0;
+  max-width: 80ch;
```
VincentSmedinga, line 31: Interesting, but will it scale well? It could for example create unexpected empty space in a grid area. Should we create a prototype to see the effect and discuss with the designers?

VincentSmedinga: In case we keep it, let’s encode it into a token.

packages/css/src/paragraph/paragraph.scss
```diff
@@ -5,28 +5,45 @@
...
+    complex css rule with a Sass function. The variable is added here so people can
+    overwrite that css rule with their own value.
+  */
+  font-size: var(
+    --amsterdam-paragraph-font-size,
+    calculate-fluid-style($amsterdam-typography-font-size-18, $amsterdam-typography-font-size-24)
+  );
+  font-weight: var(--amsterdam-paragraph-font-weight);
+  line-height: var(--amsterdam-paragraph-line-height);
   margin-block: 0;
```
VincentSmedinga, line 30: Should this be moved into the reset mixin?

packages/css/package.json
```diff
@@ -18,6 +18,7 @@
     "clean": "rimraf dist/"
   },
   "dependencies": {
+    "@amsterdam/design-system-tokens": "workspace:*",
```
VincentSmedinga, line 21: What does this do?

packages/css/src/paragraph/paragraph.scss
```diff
@@ -5,28 +5,45 @@
 
 @import "../../utils/calculate-fluid-style";
 
+@import "../../node_modules/@amsterdam/design-system-tokens/dist/variables";
```
VincentSmedinga, line 8: Can we prevent mentioning `node_modules` here?

storybook/storybook-css/src/components/Paragraph/story-template.jsx
```diff
@@ -9,27 +9,23 @@ import clsx from 'clsx'
...
-  intro = defaultArgs.intro,
-  small = defaultArgs.small,
-}) => (
+export const Paragraph = ({ textContent = defaultArgs.textContent, size = defaultArgs.size }) => (
   <p
-    className={clsx('amsterdam-paragraph', intro && 'amsterdam-paragraph-intro', small && 'amsterdam-paragraph-small')}
+    className={clsx(
+      'amsterdam-paragraph',
+      size === 'large' && 'amsterdam-paragraph-large',
+      size === 'small' && 'amsterdam-paragraph-small',
```
VincentSmedinga, line 27:
```suggestion
      size && `amsterdam-paragraph-${size}`,
```

packages/css/src/paragraph/paragraph.scss
```diff
@@ -5,28 +5,45 @@
...
+  font-family: var(--amsterdam-paragraph-font-family);
+
+  /*
+    We don't use the amsterdam-paragraph-font-size variable, because we generate a somewhat
+    complex css rule with a Sass function. The variable is added here so people can
+    overwrite that css rule with their own value.
+  */
+  font-size: var(
+    --amsterdam-paragraph-font-size,
+    calculate-fluid-style($amsterdam-typography-font-size-18, $amsterdam-typography-font-size-24)
```
VincentSmedinga, line 26: A couple of thoughts on this function, although it’s technically out of scope of this PR.

First, I’d like to improve its name. It returns the CSS clamping function with a fluid preferred value. How about `clamped-fluid-value`?

Reading up a bit (from [here](https://css-tricks.com/snippets/css/fluid-typography/) to [there](https://css-tricks.com/simplified-fluid-typography/)) it seems that we should be able to rewrite this preferred value into the form of `(a)rem + (b)vw`. Let’s try that.

And if we get the function to return a static value in case both parameters are equal, we might even be able to get rid of the unused variable, although that is a fancy trick.

VincentSmedinga: Additionally, should we introduce component variables like `$amsterdam-paragraph-font-size-min` and `$amsterdam-paragraph-font-size-max` instead of directly using the typography tokens?

VincentSmedinga (changes requested):
Thanks! Curious about what you think of my comments.

packages/css/src/paragraph/paragraph.scss
alimpens: Hm, I see your point. However, if it creates unexpected empty space in a grid area, there's something wrong with the design of the grid area :) Text shouldn't be more than 80 characters wide, this acts as a sort of fail safe for that. 

We can test it on onderzoek.amsterdam.nl and discuss it with the designers.

packages/css/src/paragraph/paragraph.scss
alimpens: Hm, I'm not sure. A reset usually fixes differences between browsers. (Almost?) all browsers set `margin-top` and `margin-bottom` for paragraphs, so it's not different between browsers.

packages/css/package.json
alimpens: We import a design token directly in `paragraph.scss`, so it needs to be in the dependencies. However, we currently only export vanilla css files, not sass, so it should be in the `devDependencies` I think

packages/css/src/paragraph/paragraph.scss
alimpens: Apparently, sass can't import from `node_modules` natively. We could use `sass-loader` I think, but that might be overkill for a simple import...

packages/css/src/paragraph/paragraph.scss
alimpens: Regarding the name: `clamped-fluid-value` isn't very intuitive imo. Clamping is **how** you do it, not **what** you're doing. Maybe something like `generate-fluid-value`? Of even simpler: `fluid-value`?

Simplifying the function sounds like a good idea, but maybe we should wait until we know what the function should actually do. The design is still unclear on this. 

I don't quite get your third remark. If both parameters are equal, you shouldn't be using this function, right?

I thought about adding a font min and max to the design tokens, but this would sort of force external users to use fluid typography, right? By using a single design token, you can choose whether you want to use fluid typo

packages/css/src/paragraph/paragraph.scss
VincentSmedinga: We decided that we’re mainly interested in resolving the gap between browsers and our components, and less in whether something’s a ‘reset’ or a ‘normalise’. We’ll include both in a mixin called `reset`.

packages/css/src/paragraph/paragraph.scss
VincentSmedinga: Sure, let’s make a separate ticket and branch for this. And `fluid-value` is a great name.

VincentSmedinga (approved):

PR #131, opened 2023-03-15
Add stories for CSS Heading component

dlnr:


storybook/storybook-css/src/components/Heading/story-template.jsx
```diff
@@ -0,0 +1,82 @@
...
+    },
+  },
+  level: {
+    options: [1, 2, 3, 4, 5, 6],
+    control: 'select',
+    table: {
+      type: { summary: 'number' },
+      defaultValue: { summary: '1' },
+    },
+  },
```
alimpens, line 24: I think we can remove this control for the css component. It only renders a different html element, which is something you should handle outside of the component library I think.

storybook/storybook-css/src/components/Heading/story-template.jsx
```diff
@@ -0,0 +1,82 @@
...
+  textContent: 'Jouw typograaf biedt mij zulke exquise schreven!',
+  level: 1,
+  styleLevel: 1,
+}
+
+export const heading2Args = {
+  textContent: 'Jouw typograaf biedt mij zulke exquise schreven!',
+  level: 2,
+  styleLevel: 2,
+}
```
alimpens, line 45: I think we can make the `headingXArgs` objects a bit more DRY:

```suggestion
export const heading2Args = {
  ...defaultArgs,
  level: 2,
  styleLevel: 2,
}
```

storybook/storybook-css/src/components/Heading/story-template.jsx
dlnr: If you don't define these for the CSS controls will show as 'unknown' and only have a text input for value. Which would allow you to create a `H32` tag or something.

dlnr: The react variant will index the type of value, which is cooler obviously.

storybook/storybook-css/src/components/Heading/story-template.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const argTypes = {
+  textContent: {
+    table: {
+      type: { summary: 'string' },
+    },
+  },
+  level: {
+    options: [1, 2, 3, 4, 5, 6],
+    control: 'select',
```
alimpens, line 19: I think we can remove these lines without issue, right?

VincentSmedinga (approved):

VincentSmedinga:
Thanks for contributing Niels!

PR #219, opened 2023-04-19
Button Component

hcvdhaar:
- React wrapper for Utrecht Button
- new token json files for border, spacing and override file Utrecht tokens.
- Stories
- Unit tests
- Adhered to new folder structure

Note:
I think we need to write a custom class for the hover state on the secondary button. The Utrecht setup does not supports adding a wider border on hover.


packages/react/src/Button/Button.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(buttonPrimary).toBeInTheDocument()
+    expect(buttonPrimary).toHaveClass('utrecht-button--primary-action')
+    expect(buttonSecondary).toBeInTheDocument()
+    expect(buttonSecondary).toHaveClass('utrecht-button--secondary-action')
+    expect(buttonTertiary).toBeInTheDocument()
+    expect(buttonTertiary).toHaveClass('utrecht-button--subtle')
+  })
+
+  it('renders a disbabled button with a specified variant', () => {
```
alimpens, line 59:
```suggestion
  it('renders a disabled button with a specified variant', () => {
```

packages/react/src/Button/Button.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const buttonSecondary = screen.getByRole('button', {
+      name: 'secondary',
+    })
+
+    const buttonTertiary = screen.getByRole('button', {
+      name: 'tertiary',
+    })
+
+    expect(buttonPrimary).toBeInTheDocument()
+    expect(buttonPrimary).toHaveClass('utrecht-button--disabled utrecht-button--primary-action')
```
alimpens, line 87: Do we need to test this for all variants? We only need to know if setting `disabled` means that the component gets the class `utrecht-button--disabled`, right?

packages/react/src/Button/Button.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    expect(buttonSecondary).toBeInTheDocument()
+    expect(buttonSecondary).toHaveClass('utrecht-button--disabled utrecht-button--secondary-action')
+    expect(buttonSecondary).toHaveAttribute('aria-disabled', 'true')
+
+    expect(buttonTertiary).toBeInTheDocument()
+    expect(buttonTertiary).toHaveClass('utrecht-button--disabled utrecht-button--subtle')
+    expect(buttonTertiary).toHaveAttribute('aria-disabled', 'true')
+  })
+
+  it('is able to pass a (React) ref', () => {
```
alimpens, line 99: This only tests React refs, right? So the parentheses aren't necessary

```suggestion
  it('is able to pass a React ref', () => {
```

packages/react/src/Heading/Heading.test.tsx
```diff
@@ -141,7 +141,7 @@ describe('Heading', () => {
     expect(heading).toHaveClass('amsterdam-heading')
   })
 
-  it('supports ForwardRef in React', () => {
+  it('is able to pass a (React) ref', () => {
```
alimpens, line 144:
```suggestion
  it('is able to pass a React ref', () => {
```

proprietary/tokens/src/common/amsterdam/spacing.tokens.json
```diff
@@ -0,0 +1,20 @@
+{
```
alimpens, line 1: I think our spacing system will be a bit more complicated than this, for now we can just hardcode the values in the component level design tokens imo

proprietary/tokens/src/common/amsterdam/border.tokens.json
```diff
@@ -0,0 +1,15 @@
+{
```
alimpens, line 1: Is this generic enough lock down in common design tokens? Maybe we can just hardcode the values in the component level design tokens for now

proprietary/tokens/src/components/utrecht/button.tokens.json
```diff
@@ -0,0 +1,271 @@
+{
+  "utrecht": {
+    "button": {
+      "background-color": {},
```
alimpens, line 4: I don't think you need to include empty tokens, they aren't converted into css variables or anything

storybook/storybook-react/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { Button } from '@amsterdam/design-system-react/src'
+import { Meta, StoryObj } from '@storybook/react'
+
+import '@amsterdam/design-system-css/src/button/button.scss'
+
+const meta: Meta<typeof Button> = {
```
alimpens, line 11: Can you use the same structure as we do in the heading story of this [this PR](https://github.com/Amsterdam/design-system/pull/224/files#)?

packages/react/src/Button/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+    case 'tertiary':
+      return 'subtle-button'
+    default:
+      return 'primary-action-button'
+  }
+}
+
+export const Button = forwardRef(
+  (
+    { children, disabled, variant = 'primary', ...rest }: PropsWithChildren<ButtonProps>,
```
alimpens, line 30:
```suggestion
    { children, disabled, variant = 'primary', ...restProps }: PropsWithChildren<ButtonProps>,
```

packages/react/src/Button/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+    ref: ForwardedRef<HTMLButtonElement>,
+  ) => {
+    return (
+      <CommunityButton
+        {...rest}
```
alimpens, line 35:
```suggestion
        {...restProps}
```

packages/react/src/Button/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+        appearance={getAppearance(variant)}
+        ref={ref}
+        aria-disabled={disabled ?? false}
+        disabled={disabled}
```
alimpens, line 39: This is not necessary for MVP, we can finetune this in [a seperate ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-156)

packages/css/src/button/button.scss
```diff
@@ -0,0 +1,4 @@
+@import "../../node_modules/@utrecht/components/button/css/mixin";
```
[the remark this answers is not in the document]
hcvdhaar: removed in the css component branch

packages/react/src/Button/Button.test.tsx
hcvdhaar: good one, no assertions for implementation details

packages/react/src/Button/Button.tsx
hcvdhaar: done

proprietary/tokens/src/components/utrecht/button.tokens.json
hcvdhaar: For now I would like to keep it. Just for the references what the possibilities are regarding token settings. I will create a ticket to discuss this.

storybook/storybook-react/src/components/Button/Button.stories.tsx
hcvdhaar: done

proprietary/tokens/src/common/amsterdam/spacing.tokens.json
hcvdhaar: Actually I hope not. I my understanding spacings values should be used to compose other values. Preferable no values for every implementation. This can be extended of course, but I guess this should be the main reference.

proprietary/tokens/src/common/amsterdam/border.tokens.json
hcvdhaar: I consider this a shared setting and therefor it would be ok to have this in a common folder

VincentSmedinga (approved):

PR #224, opened 2023-04-20
Migrate to CSF3 and some error cleanup

dlnr:
This also seems to fix the slow rendering of docs.


storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -4,55 +4,53 @@
...
 }
 
-export const Heading2 = Template.bind({})
-
-Heading2.args = {
-  children: 'Jouw typograaf biedt mij zulke exquise schreven!',
-  level: 2,
+export const Heading2 = {
+  args: {
+    children: 'Jouw typograaf biedt mij zulke exquise schreven!',
```
alimpens, line 25: This can be a bit more DRY: 

```suggestion
    ...Heading1.args,
```

Same for the other heading levels

storybook/storybook-react/src/components/Icon/Icon.stories.tsx
```diff
@@ -4,21 +4,23 @@
...
+import { Meta, StoryFn } from '@storybook/react'
 
 import '@amsterdam/design-system-css/src/icon/icon.scss'
 
 export default {
   title: 'Icon',
   id: 'icon',
   component: Icon,
-} as ComponentMeta<typeof Icon>
+} as Meta<typeof Icon>
```
alimpens, line 16: Do we run into issues using the `satisfies` keyword here?

dlnr: No, the migrate script just doesn't use that by default.

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -4,55 +4,57 @@
...
+    level: 1,
+    styleLevel: 1,
+  },
+  tags: ['autodocs'],
+} satisfies Meta<typeof Heading>
+
+export const Heading1 = {
+  name: 'Heading 1',
```
alimpens, line 24: Can we use an empty object here? We never use this name, right?

dlnr: I believe it's how you set the title of the story

dlnr: Changed it, also tested it, `name` does nothing so I removed it from alle stories. Perhaps it was inherited from CSF2.

alimpens (approved):

PR #251, opened 2023-04-28
DES-47 Link Stories

dlnr:
Created from scratch with some hints from NLDS, as barebones as possible. Let's discuss how we extend or replace this component. The Figma and variants threw me off, the original Fabrique figma had the best resemblance to what could be developed.

The inList variant is a work in progress, when the list is a component I'll replace this markup.


packages/react/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { render } from '@testing-library/react'
+import { Link } from './Link'
+import '@testing-library/jest-dom'
+
+describe('Link', () => {
+  it('renders a link role element', () => {
```
hcvdhaar, line 6: I would avoid nested `it` statement. They should always have an assertion. This feels like composing a test case

packages/react/src/Link/Link.tsx
```diff
@@ -0,0 +1,50 @@
...
+import clsx from 'clsx'
+import { AnchorHTMLAttributes, ForwardedRef, forwardRef, PropsWithChildren } from 'react'
+
+export type LinkColor = 'default' | 'black' | 'white'
+
+export type LinkVariant = 'default' | 'bold'
+
+export interface LinkProps extends AnchorHTMLAttributes<HTMLAnchorElement> {
+  href: string
+  placeholder?: string
```
hcvdhaar, line 18: Not the same as a `placeholder` as used in inputs? I would opt for a different prop name. I guess in general I prefer different names if they are the same as the default HTML attributes. Same for `color`

packages/react/src/Link/Link.tsx
```diff
@@ -0,0 +1,50 @@
...
+  external?: boolean
+  rel?: string
```
hcvdhaar, line 20: Not needed to define a prop here. We extend the `AnchorHTMLAttributes<HTMLAnchorElement>` interface which are declared there

packages/react/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { render } from '@testing-library/react'
+import { Link } from './Link'
+import '@testing-library/jest-dom'
+
+describe('Link', () => {
```
hcvdhaar, line 5: I think we miss some tests here ;)

packages/react/src/Link/Link.tsx
hcvdhaar: From the Utrecht Link
```
export interface LinkProps extends Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'placeholder'> {
  boxContent?: boolean;
  external?: boolean;
  placeholder?: boolean;
}

File: packages/react/src/Link/Link.tsx [hunk 5]
```diff
@@ -0,0 +1,50 @@
...
+  variant?: LinkVariant
+}
+
+export const Link = forwardRef(
+  (
+    { children, href, placeholder, variant, color, className, ...restProps }: PropsWithChildren<LinkProps>,
+    ref: ForwardedRef<HTMLAnchorElement>,
+  ) => (
+    <CommunityLink
+      href={placeholder ? undefined : href}
```
hcvdhaar, line 32: Don't really get this logic. First, in the Utrecht Link `placeholder` is a boolean. Here it is a string. Even then, I don't get it :) Since `href` is required so assuming the users needs to set that prop, then this logic would not be needed right? Why would be prefer an `undefined` over a defined `href`?

dlnr: `undefined` will disable the link, which is what the Utrecht link placeholder is for.

packages/css/src/link/index.scss
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../node_modules/@amsterdam/design-system-tokens/dist/variables";
+
+@import "./mixins";
+@import "../../utils/calculate-fluid-style";
+
+.amsterdam-link {
+  @include amsterdam-link;
```
alimpens, line 12: I don't think we need these mixins yet, we can add these styles directly in the class

dlnr: Adjusted

dlnr:
@alimpens Couple of things, I've looked more carefully at the other components and dropped the token level for default, "amsterdam.link" contains all the default (base link) stuff and out variants contain only a set of tokens that we actually need.

I've also left a comment  in the stories because I can't seem to get rid of `onBackground: {() -> {}}` in storybook when left blank. Which is weird because the prop is optional.

packages/react/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it('renders with rel attribute', () => {
+    const { container } = render(
+      <Link href="#" rel="noopener">
+        {linktext}
+      </Link>,
+    )
+    const link = container.querySelector('a:only-child')
+    expect(link).toBeInTheDocument()
+    expect(link).toHaveAttribute('rel', 'noopener')
+  })
```
alimpens, line 34: Do we need to test this? These are standard html anchor attributes. Maybe we only need to test if the `Link` component renders an anchor

packages/react/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  it('renders standalone', () => {
```
alimpens, line 36:
```suggestion
  it('renders standalone variant', () => {
```

packages/react/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it('renders dark background color', () => {
+    const { container } = render(
+      <Link onBackground="dark" href="#">
+        {linktext}
+      </Link>,
+    )
+    const link = container.querySelector('a:only-child')
+    expect(link).toHaveClass('amsterdam-link amsterdam-link--background-dark')
+  })
+})
```
alimpens, line 83: There are some tests missing, if you can add an additional class and use React ref and stuff. You can get them from [Utrecht's Link tests](https://github.com/nl-design-system/utrecht/blob/main/packages/component-library-react/src/Link.test.tsx)

packages/react/src/Link/Link.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { ChevronRight } from '@amsterdam/design-system-react-icons'
+import clsx from 'clsx'
+import { AnchorHTMLAttributes, ForwardedRef, forwardRef, PropsWithChildren } from 'react'
+import { Icon } from '../Icon/Icon'
+
+export type LinkOnBackground = 'default' | 'light' | 'dark'
+export type LinkVariant = 'standalone' | 'standaloneBold' | 'inline' | 'inList'
```
alimpens, line 12: Do we want to export these types? I can't imagine why we would want to use them outside this component. I think we can define directly in `LinkProps`

packages/react/src/Link/Link.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onBackground?: LinkOnBackground
+}
+
+export const Link = forwardRef(
+  (
+    { children, href, variant, onBackground, className, ...restProps }: PropsWithChildren<LinkProps>,
+    ref: ForwardedRef<HTMLAnchorElement>,
+  ) => (
+    <a
+      href={href}
```
alimpens, line 25: I don't think we need this, right? This is already part of `restProps`

packages/react/src/Link/Link.tsx
```diff
@@ -0,0 +1,46 @@
...
+          'amsterdam-link--inline': variant === 'inline',
+          'amsterdam-link--standalone-bold': variant === 'standaloneBold',
+          'amsterdam-link--in-list': variant === 'inList',
+          'amsterdam-link--background-light': onBackground === 'light',
+          'amsterdam-link--background-dark': onBackground === 'dark',
+        },
+        className,
+      )}
+    >
+      {variant === 'inList' && <Icon svg={ChevronRight} size="level-6" />}
```
alimpens, line 40: Shouldn't the `size` be `level-6`?

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,94 @@
+{
+  "amsterdam": {
+    "link": {
+      "color": { "value": "{amsterdam.color.primary-blue}" },
+      "icon": {
+        "size": { "value": "1.2em" }
+      },
```
alimpens, line 7: This is unused I think

packages/css/src/link/link.scss
```diff
@@ -3,54 +3,82 @@
...
-  outline-offset: var(--amsterdam-focus-outline-offset);
-  text-decoration: none;
-
-  @include reset;
+  color: var(--amsterdam-link-color, blue);
+  font-family: var(--amsterdam-paragraph-font-family);
+  font-size: var(--amsterdam-link-font-size-narrow);
+  text-decoration: var(--amsterdam-link-text-decoration, underline);
+  text-decoration-color: var(--amsterdam-link-text-decoration-color, currentColor);
+  text-decoration-skip-ink: $amsterdam-link-text-decoration-skip-ink;
```
alimpens: This causes a readability issue for inline links I think

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,94 @@
...
+      "font-size": {
+        "narrow": { "value": "{amsterdam.typography.text-level.6.font-size.narrow}" },
+        "wide": { "value": "{amsterdam.typography.text-level.6.font-size.wide}" }
+      },
+      "font-weight": {
+        "value": "{amsterdam.typography.font-weight.normal}"
+      },
+      "focus": {
+        "color": { "value": "{amsterdam.color.dark-blue}" },
+        "text-decoration": { "value": "underline" }
```
alimpens, line 24: [We should add `focus-offset` here](https://amsterdam.github.io/design-system/?path=/docs/docs-designrichtlijnen-focus-indicator--docs)

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,94 @@
...
+        "text-underline-offset": { "value": "3px" },
+        "font-family": {
+          "value": "inherit"
+        },
+        "font-size": {
+          "value": "inherit"
+        }
+      },
+      "in-list": {
+        "icon-spacing": { "value": "0.5em" },
```
alimpens, line 52:
```suggestion
        "gap": { "value": "0.5em" },
```

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,94 @@
...
+        "text-decoration": { "value": "none" },
+        "text-underline-offset": { "value": "0.3em" },
+        "text-decoration-thickness": { "value": "2px" }
+      },
+      "background-dark": {
```
alimpens, line 55:
```suggestion
      "on-background-dark": {
```

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,94 @@
...
+        },
+        "hover": {
+          "color": { "value": "{amsterdam.color.primary-white}" },
+          "text-decoration": { "value": "none" }
+        },
+        "visited": {
+          "color": { "value": "{amsterdam.color.primary-white}" }
+        }
+      },
+      "background-light": {
```
alimpens, line 70:
```suggestion
      "on-background-light": {
```

storybook/storybook-css/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+import '@amsterdam/design-system-css/src/link/link.scss'
+import '@amsterdam/design-system-css/src/paragraph/paragraph.scss'
+import '@amsterdam/design-system-css/src/icon/icon.scss'
+
+type Story = StoryObj<typeof Link>
+
+const meta = {
+  title: 'CSS Components/Link',
+  id: 'link',
```
alimpens, line 17: I don't think we need this

storybook/storybook-css/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+    },
+    target: {
+      name: 'target',
+      type: { name: 'string', required: false },
+      table: {
+        category: 'HTML attribute',
+        defaultValue: { summary: '' },
+      },
+    },
+  },
```
alimpens, line 63: Do we need to add controls for this native stuff?

storybook/storybook-css/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+  args: {
+    children: 'Linktekst',
+    href: '#',
+  },
+} satisfies Meta<typeof Link>
+
+export default meta
+
+export const Standalone: Story = {
+  name: 'Standalone',
```
alimpens, line 73: We don't need these names I think, Storybook uses the variable name

storybook/storybook-css/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+  args: {
+    children: 'Linktekst',
+    href: '#',
+    variant: 'inList',
+  },
+}
+
+/**
+ * Link to use on a dark background
+ */
```
alimpens, line 151: I don't think these comments add anything, this should be described in the Storybook docs

storybook/storybook-react/src/components/Link/Link.stories.tsx
```diff
@@ -3,27 +3,145 @@
...
 
 import '@amsterdam/design-system-css/src/link/link.scss'
+import '@amsterdam/design-system-css/src/paragraph/paragraph.scss'
 import '@amsterdam/design-system-css/src/icon/icon.scss'
 
+type Story = StoryObj<typeof Link>
+
 const meta = {
   title: 'Link',
   id: 'link',
```
alimpens, line 17: We don't need this I think

storybook/storybook-react/src/components/Link/Link.stories.tsx
```diff
@@ -3,27 +3,145 @@
...
+
+export const InList: Story = {
+  name: 'InList',
+  args: {
+    children: 'Linktekst',
+    href: '#',
+    variant: 'inList',
+    onBackground: undefined,
+  },
+  decorators: [(Story) => <Story />],
```
alimpens, line 112: Is this necessary?

storybook/storybook-css/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Link to use on a light background
+ */
+export const Light: Story = {
+  name: 'Light',
+  parameters: {
+    docs: {
+      source: {
+        language: 'html',
+        code: '<a href="#" class="amsterdam-link amsterdam-link--color-white">Linktekst</a>',
```
alimpens, line 183:
```suggestion
        code: '<a href="#" class="amsterdam-link amsterdam-link--color-light">Linktekst</a>',
```

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -0,0 +1,95 @@
...
+          "value": "inherit"
+        },
+        "font-size": {
+          "value": "inherit"
+        },
+        "visited": {
+          "color": { "value": "{amsterdam.color.purple}" }
+        }
+      },
+      "in-list": {
```
alimpens, line 51: Shouldn't this have text level tokens as well? The in list variant is smaller than the default link

storybook/storybook-css/src/components/Link/Link.stories.tsx
dlnr: Added by Utrecht to storybook, I guess for testing or to have complete documentation?

packages/css/src/link/link.scss
```diff
@@ -3,54 +3,82 @@
...
+    text-decoration: $amsterdam-link-hover-text-decoration;
+    text-decoration-thickness: var(--amsterdam-link-hover-text-decoration-thickness, 1px);
+  }
+
+  &:focus {
+    color: $amsterdam-link-focus-color;
+    text-decoration: $amsterdam-link-focus-text-decoration;
+  }
+
+  &:visited {
```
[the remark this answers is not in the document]
dlnr: Added to prevent default browser purple

packages/react/src/Link/Link.test.tsx
dlnr: Added some tests

packages/react/src/Link/Link.tsx
dlnr: Wanted to prevent the inability Utrecht has to import component props. But we'll fix that if it comes around.

packages/react/src/Link/Link.tsx
dlnr: Need for forwardRef

storybook/storybook-css/src/components/Link/Link.stories.tsx
dlnr: It's actually `amsterdam-link amsterdam-link--on-background-light` forgot to change it.

alimpens (approved):

PR #252, opened 2023-04-28
Update major deps

alimpens:


packages/react/package.json
```diff
@@ -72,8 +72,7 @@
     "rollup-plugin-scss": "4.0.0",
     "rollup-plugin-typescript2": "0.34.1",
     "sass": "1.62.0",
-    "tslib": "2.5.0",
-    "typescript": "4.9.5"
```
dlnr, line 76: Removing typescript correct?

proprietary/react-icons/package.json
```diff
@@ -35,8 +35,7 @@
     "react": "18.2.0",
     "rollup": "2.79.1",
     "rollup-plugin-peer-deps-external": "2.2.4",
-    "rollup-plugin-typescript2": "0.34.1",
-    "typescript": "4.9.5"
```
dlnr, line 39: Removing typescript correct?

storybook/storybook-react/package.json
```diff
@@ -50,7 +50,6 @@
     "react-dom": "18.2.0",
     "require-from-string": "2.0.2",
     "storybook": "7.0.7",
-    "typescript": "4.9.5",
```
dlnr, line 53: Removing typescript correct?

dlnr (commented):
Seems fine except for the removal of typescript package. Or do you rely on locally installed TS?

packages/react/package.json
alimpens: [Typescript is defined in the root package.json](https://github.com/Amsterdam/design-system/blob/main/package.json#L47), I don't think we need to define it in all the sub packages, right?

dlnr (approved):

PR #311, opened 2023-05-26
Feature/storybook-theme

dlnr:


storybook/storybook-theme/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
alimpens, line 1: I'm not sure we should make a seperate package for this. If we add `@storybook/theming` to both storybooks and just use this directory to store a generic theme config, we could keep it a lot more simple. Or maybe even easier: I think we only need to define the theme in the main storybook (storybook-css), right? Then we could do everything in that directory.

storybook/storybook-theme/amsterdam.js
```diff
@@ -0,0 +1,7 @@
+import { create } from '@storybook/theming/create'
+import logo from './logo.svg'
```
alimpens, line 2: Can't we get the logo from `@amsterdam/icons`?

alimpens: Sure

storybook/storybook-theme/package.json
dlnr: I had to add npm to be able to use theming, the alternative is the storybook directories having their own theme config.

dlnr: I can remove main though

alimpens: Maybe we can just store the config in `/storybook` (the object with `brandTitle` and `brandImage`), and create the themes seperately in both storybooks.

proprietary/icons/svgo.config.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = {
+  plugins: [
+    {
+      name: 'preset-default',
+      params: {
+        overrides: {
+          inlineStyles: false,
+          convertStyleToAttrs: false,
+          removeStyleElement: false,
+          removeDoctype: false,
```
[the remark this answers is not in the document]
dlnr: Removed

storybook/storybook-theme/package.json
dlnr: Changed it to suggested solution

alimpens: One suggestion (last one, I swear ;) ) : you suggest a `/storybook-theme` directory with one file in it. Couldn't we just place a file directly in the `/storybook` directory, and call it `themeConfig.js` or something?

alimpens (approved):

alimpens (approved):

PR #333, opened 2023-06-01
Add remaining docs to css and React stories

alimpens:


hcvdhaar:
Lijkt me prima

alimpens:
> Lijkt me op zich prima, lijkt wel of we zelfde docs files aanmaken voor css en react

They're almost the same, the css docs import extra documentation that's specific to the css 'components' ([here](https://github.com/Amsterdam/design-system/pull/333/files#diff-7199f0c8eddee75882bf51c54bdc3477ea14d9ee5ee589d39c55f0f964643cffR10), for example).

If the PR is fine as far as you're concerned, can you approve it?

hcvdhaar (approved):

PR #340, opened 2023-06-07
Add and implement text levels

alimpens:


packages/css/src/button/button.scss
```diff
@@ -1,4 +1,13 @@
 @import "../../node_modules/@utrecht/components/button/css";
+@import "../../utils/breakpoint";
+
+.amsterdam-button {
+  font-size: var(--amsterdam-button-font-size-narrow);
+
+  @media screen and (width > $amsterdam-breakpoint) {
```
hcvdhaar, line 7: Could this be  a candidate for a `mixin` with some params? We use in multiple places

packages/css/utils/breakpoint.scss
```diff
@@ -4,7 +4,7 @@
  */
 
 /*
-  The breakpoint is 820px / 16 = 51.25rems
+  The breakpoint is 854px / 16 = 53.375rems
   https://developer.mozilla.org/en-US/docs/Web/CSS/@media#accessibility_concerns
 */
-$amsterdam-breakpoint: 51.25rem;
+$amsterdam-breakpoint: 53.375rem;
```
hcvdhaar, line 10: Looks like an arbitrary value. maybe a descriptive comment?

hcvdhaar:
Nice job! 🦄

packages/css/src/button/button.scss
alimpens: Sure, what do you think of something like this:

```css
@mixin on-wide-screen {
  @media screen and (width > $amsterdam-breakpoint) {
    @content;
  }
}
```

Usage would be:

```css
  @include on-wide-screen {
    font-size: var(--amsterdam-button-font-size-wide);
  }
```

packages/css/utils/breakpoint.scss
alimpens: Hm, I'm not sure what to add here aside from the comments at the start of the file and the breakpoint docs in [this PR](https://github.com/Amsterdam/design-system/pull/342). Any suggestions?

hcvdhaar (approved):

PR #341, opened 2023-06-07
Add focus docs

alimpens:


storybook/storybook-css/src/docs/focus.docs.mdx
```diff
@@ -0,0 +1,16 @@
+import { Meta } from "@storybook/blocks";
+
+<Meta title="Docs/Designrichtlijnen/Focus staat" />
+
+# Focus staat
```
VincentSmedinga, line 5: This is a rather literal translation. I’d propose ‘focus-indicator’, which highlights the visual aspect of _indicating_ a certain state, not the state itself.

storybook/storybook-css/src/docs/focus.docs.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+De toegankelijkheidsrichtlijnen eisen dat het element op de pagina dat focus heeft visueel herkenbaar is. Dit geldt ook voor formuliervelden. Houd rekening bij het ontwerp en de ontwikkeling van de website of applicatie met de focus staat.
```
VincentSmedinga, line 7: Here comes a more general point, but this seems a good moment to make it. I think it would be stronger to demand indicating a focused element ourselves, rather than ‘hide behind’ a law or regulation. How about just “Een element dat focus heeft moet visueel herkenbaar zijn. Dit geldt voor interactieve elementen zoals links, knoppen en formuliervelden.”

The last sentence may be unnecessary, as we implement focus state in our components, so that implementers don’t need to think about it.

storybook/storybook-css/src/docs/focus.docs.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+Wij gebruiken grotendeels de standaard focus staat van de webbrowsers. De enige aanpassing die we doen is de focus `outline` een `offset` geven van 2 pixels, zodat er ruimte zit tussen het element en de focus staat. Hierdoor heeft de focus staat ook voldoende contrast met donkere elementen.
+
+## Relevante WCAG eisen
```
VincentSmedinga, line 11: More general as well: how can we manage the order of sections on a docs page? I’d say the example and API table should be almost at the top of the page. Instructions for usage, or guidelines to deal with can come below. No problem to fix this later.

VincentSmedinga (changes requested):
Thanks for taking the time to add this page!

storybook/storybook-css/src/docs/focus.docs.mdx
alimpens: I agree. I got this text from the [current design system page](https://amsterdam.nl/digitalservices/designrichtlijnen/focus-staat/), but a more active voice is preferable.

I think we need to keep that last sentence though, the reason I added this doc now is because our devs weren't aware of how we handle the focus indicator.

storybook/storybook-css/src/docs/focus.docs.mdx
alimpens: I agree, the general structure of docs pages can be improved. I'll add a ticket to our board. I think it's best to do this later though, when we have some more documentation pages. For now, our focus should be on making components, I think.

VincentSmedinga (approved):

PR #389, opened 2023-06-29
feat/DES 54 breadcrumb

hcvdhaar:
Not sure about the logging when an invalid element type is provided as a child.


packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+  text-decoration: none;
+
+  &:focus,
+  &:hover,
+  &:active {
+    box-shadow: var(--amsterdam-breadcrumb-item-link-hover-box-shadow);
+    color: var(--amsterdam-breadcrumb-hover-color);
+  }
+
+  &:visited {
```
alimpens, line 70: Should breadcrumbs have a visited state? I don't see it in the designs.

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+      return React.Children.toArray(children).filter((child, index) => {
+        if (isReactFunctionCom(child) && child.type.displayName !== 'BreadcrumbItem') {
+          // TODO: Do we want to log this as a warning?
+          console.warn(`Breadcrumb: ${child.type.displayName ?? child.type} is not a valid child`)
+          return false
+        }
+
+        return index < breadcrumbLimit
+      })
+    }, [children])
```
alimpens, line 32: As discussed: for now we won't check if components have the right subcomponents, that's the consumers job

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface BreadcrumbComponent
+  extends React.ForwardRefExoticComponent<
+    PropsWithChildren<HTMLAttributes<HTMLElement>> & React.RefAttributes<HTMLElement>
+  > {
+  Item: typeof BreadcrumbItem
+}
+
+// We're validating against an index. Maxiumum children are 8.
+const breadcrumbLimit = 8
```
alimpens, line 18: As discussed: for now we won't check children amount

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+          return false
+        }
+
+        return index < breadcrumbLimit
+      })
+    }, [children])
+
+    return (
+      <nav {...restProps} className="amsterdam-breadcrumb" ref={ref}>
+        <ol className="amsterdam-breadcrumb-list">{allowedBreadcrumbItems}</ol>
```
alimpens, line 36: Should have double underscores in BEM I think?

```suggestion
        <ol className="amsterdam-breadcrumb__list">{allowedBreadcrumbItems}</ol>
```

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+Breadcrumb.displayName = 'Breadcrumb'
+
+interface BreadcrumbItemProps extends PropsWithChildren {
+  href: string
+}
+
+// TODO: We might wanna wrap this in a forwardRef as well.
+const BreadcrumbItem = ({ children, href }: BreadcrumbItemProps) => {
+  return (
+    <li className="amsterdam-breadcrumb-item">
```
alimpens, line 51:
```suggestion
    <li className="amsterdam-breadcrumb__item">
```

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <a className="amsterdam-breadcrumb-link" href={href}>
```
alimpens, line 52:
```suggestion
      <a className="amsterdam-breadcrumb__link" href={href}>
```

packages/react/src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,60 @@
...
+  },
+) as BreadcrumbComponent
+
+Breadcrumb.displayName = 'Breadcrumb'
+
+interface BreadcrumbItemProps extends PropsWithChildren {
+  href: string
+}
+
+// TODO: We might wanna wrap this in a forwardRef as well.
```
alimpens, line 48: Yeah, I think we should

packages/react/src/Breadcrumb/index.ts
```diff
@@ -0,0 +1 @@
+export { Breadcrumb } from './Breadcrumb'
```
alimpens, line 1: Do we want to export types here as well?

storybook/storybook-css/src/components/Breadcrumb/Breadcrumb.docs.mdx
```diff
@@ -0,0 +1,12 @@
+import { Markdown, Meta, Primary } from "@storybook/blocks";
+import * as Breadcrumb from "./Breadcrumb.stories.tsx";
+import README from "../../../../../packages/css/src/breadcrumb/README.md?raw";
+import css from "../../../../../packages/css/src/breadcrumb/breadcrumb-css.md?raw";
```
alimpens, line 4: We don't need this if it's empty, right?

storybook/storybook-css/src/components/Breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+} satisfies Meta<typeof Breadcrumb>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const BreadcrumbStory: Story = {
+  render: (args) => (
+    <Breadcrumb {...args}>
+      <Breadcrumb.Item href="/some-url.com">home</Breadcrumb.Item>
```
alimpens, line 23: Maybe we should link to the same page, so you don't accidentally leave storybook

```suggestion
      <Breadcrumb.Item href="#">home</Breadcrumb.Item>
```

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+@import "../../utils/breakpoint";
+
+@mixin reset {
+  box-sizing: border-box;
+  break-after: avoid;
+  break-inside: avoid;
+  margin-block: 0;
+  margin-inline: 0;
```
alimpens, line 13: `margin-inline` is 0 by default, right?

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+  padding-inline: 0;
+}
+
+.amsterdam-breadcrumb {
+  color: var(--amsterdam-breadcrumb-color);
```
alimpens, line 18: Is color only used for `.amsterdam-breadcrumb-link`? Why not just set it there?

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+.amsterdam-breadcrumb-link {
+  color: inherit;
+  cursor: pointer;
+  outline-offset: var(--amsterdam-breadcrumb-item-link-outline-offset);
+  text-decoration: none;
+
+  &:focus,
+  &:hover,
+  &:active {
+    box-shadow: var(--amsterdam-breadcrumb-item-link-hover-box-shadow);
```
alimpens, line 66: Shouldn't we use `text-decoration: underline` for this? Similar to the link component?

packages/css/src/breadcrumb/breadcrumb.scss
hcvdhaar: It is, copied the `reset` from some other file. `<ol>` doesn't need a  `margin-inline: 0` Because that is the default.

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+  margin-inline: 0;
+  padding-inline: 0;
+}
+
+.amsterdam-breadcrumb {
+  color: var(--amsterdam-breadcrumb-color);
+  font-family: var(--amsterdam-breadcrumb-font-family, inherit);
+  font-size: var(--amsterdam-breadcrumb-font-size-narrow);
+  line-height: var(--amsterdam-breadcrumb-line-height);
+  text-transform: lowercase;
```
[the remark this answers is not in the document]
hcvdhaar: hmm yeah you are right :)

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,73 @@
...
+.amsterdam-breadcrumb-item {
+  align-items: center;
+  display: flex;
+}
+
+.amsterdam-breadcrumb-item:not(:last-child)::after {
+  background-color: var(--amsterdam-breadcrumb-color);
+  content: "";
+  display: inline-block;
+  font-size: 16px;
```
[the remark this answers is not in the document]
hcvdhaar: Nope, used for mocking the icon

packages/css/src/breadcrumb/breadcrumb.scss
hcvdhaar: I think you can not increase the space between the underline and the text. Also I think this is a different implementation of a link then the actual `<Link />` component

packages/css/src/breadcrumb/breadcrumb.scss
hcvdhaar: Not in the designs. Will remove it.

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: Since this is new component, it should be considered as a new `block`. Therefore, I named it `amsterdam-breadcrumb-item`, the link would/should be `amsterdam-breadcrumb-item__link`

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: done

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: done

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: done

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: done

packages/react/src/Breadcrumb/Breadcrumb.tsx
hcvdhaar: see other related comment

packages/react/src/Breadcrumb/index.ts
hcvdhaar: Not sure if we want or need to export the types. I guess something we might want to discuss. For now I will keep it consistent, so I exported it.

storybook/storybook-css/src/components/Breadcrumb/Breadcrumb.docs.mdx
hcvdhaar: removed it.

storybook/storybook-css/src/components/Breadcrumb/Breadcrumb.stories.tsx
hcvdhaar: done

packages/css/src/breadcrumb/breadcrumb.scss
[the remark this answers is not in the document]
alimpens: Your current fix would capitalize all words in a link I think. You can use `&:first-letter {text-transform: uppercase}` I think

packages/css/src/breadcrumb/breadcrumb.scss
```diff
@@ -0,0 +1,67 @@
...
+  -webkit-mask-size: 16px;
+  mask-size: 16px;
+  width: 16px;
+}
+
+.amsterdam-breadcrumb-item__link {
+  color: inherit;
+  cursor: pointer;
+  outline-offset: var(--amsterdam-breadcrumb-item-link-outline-offset);
+  text-decoration: none;
```
alimpens, line 59: Shouldn't we give these links some padding? Currently, the hit area on small screens is ~25px, which is smaller than the [44px required by the WCAG](https://www.w3.org/TR/WCAG21/#target-size).

The `inList` Link variant has the same problem, maybe we should discuss this with design as well.

packages/css/src/breadcrumb/breadcrumb.scss
hcvdhaar: done

packages/css/src/breadcrumb/breadcrumb.scss
hcvdhaar: done

VincentSmedinga (approved):

PR #391, opened 2023-06-30
DES-63 Unordered List

dlnr:
Just the default UL to get started, checklist and errorlist will come after as discussed


packages/react/src/UnorderedList/UnorderedListItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, LiHTMLAttributes, PropsWithChildren } from 'react'
+
+export type UnorderedListItemProps = LiHTMLAttributes<HTMLLIElement>
+
+export const UnorderedListItem = forwardRef(
```
alimpens, line 11: Don't we want to use the same compound component structure Hee Chan suggested here?

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.amsterdam-unordered-list {
+  color: var(--amsterdam-unordered-list-color);
+  font-family: var(--amsterdam-unordered-list-font-family);
+  font-size: var(--amsterdam-unordered-list-font-size-narrow);
+  font-weight: var(--amsterdam-unordered-list-font-weight);
+  line-height: var(--amsterdam-unordered-list-line-height);
+  list-style: var(--amsterdam-unordered-list-list-style);
+  margin-block: var(--amsterdam-unordered-list-margin-block);
+  padding-inline-end: var(--amsterdam-unordered-list-padding-inline-end);
```
alimpens, line 16: Do we need this?

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -0,0 +1,31 @@
...
+@import "../../utils/breakpoint";
+
+.amsterdam-unordered-list {
+  color: var(--amsterdam-unordered-list-color);
+  font-family: var(--amsterdam-unordered-list-font-family);
+  font-size: var(--amsterdam-unordered-list-font-size-narrow);
+  font-weight: var(--amsterdam-unordered-list-font-weight);
+  line-height: var(--amsterdam-unordered-list-line-height);
+  list-style: var(--amsterdam-unordered-list-list-style);
+  margin-block: var(--amsterdam-unordered-list-margin-block);
```
alimpens, line 15: This should be 0, right? Components shouldn't set their own margins

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -0,0 +1,31 @@
...
+  padding-inline-end: var(--amsterdam-unordered-list-padding-inline-end);
+
+  @media screen and (width > $amsterdam-breakpoint) {
+    font-size: var(--amsterdam-unordered-list-font-size-wide);
+  }
+}
+
+.amsterdam-unordered-list__item {
+  padding-inline-start: var(--amsterdam-unordered-list-item-padding-inline-start);
```
alimpens, line 24: This padding seems to be larger than the design

proprietary/tokens/src/components/amsterdam/unordered-list.tokens.json
```diff
@@ -0,0 +1,22 @@
...
+      "font-size": {
+        "narrow": { "value": "{amsterdam.typography.text-level.6.font-size.narrow}" },
+        "wide": { "value": "{amsterdam.typography.text-level.6.font-size.wide}" }
+      },
+      "font-weight": { "value": "{amsterdam.typography.font-weight.normal}" },
+      "line-height": { "value": "{amsterdam.typography.text-level.6.line-height}" },
+      "list-style": { "value": "square" },
+      "margin-block": { "value": "0.5rem" },
+      "padding-inline-end": { "value": "11px" },
+      "item": {
```
alimpens, line 15: This naming (amsterdam-unordered-list-item-list-style-type) is a bit odd. Maybe we should add `level-1` (with `margin-block` and `list-style-type`) and `level-2` (with the same)?

storybook/storybook-css/src/components/UnorderedList/UnorderedList.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+import '@amsterdam/design-system-css/src/unordered-list/unordered-list.scss'
+
+const meta = {
+  title: 'CSS Components/UnorderedList',
+  component: UnorderedList,
+  argTypes: {
+    children: {
+      table: {
+        type: { summary: 'string' },
```
alimpens, line 17: This doesn't seem to work

storybook/storybook-css/src/components/UnorderedList/UnorderedList.stories.tsx
```diff
@@ -0,0 +1,42 @@
+/**
```
alimpens, line 1: We're missing the React stories

packages/css/src/unordered-list/unordered-list.scss
dlnr: The design shows an 8px margin top and bottom, shall I change it to this?

```suggestion
  padding-block: var(--amsterdam-unordered-list-padding-block);
```

packages/css/src/unordered-list/unordered-list.scss
dlnr: Also something from Figma

packages/css/src/unordered-list/unordered-list.scss
dlnr: If you change a UL margin and padding to 0 the bullet will be centered on the start of the window, I've taken the liberty to visualy put the list on the same position.

proprietary/tokens/src/components/amsterdam/unordered-list.tokens.json
dlnr: Like a variant?


```suggestion
      "amsterdam": {
    	"unordered-list": {
			"level-1": {

			},
			"level-2": {
				
			}
		}
	  }
```

NLDS doesn't seem to be doing this.
see: utrecht/proprietary/design-tokens/src/component/utrecht/unordered-list.tokens.json

alimpens: Hm, maybe the variant approach isn't quite correct either. But `item` isn't either, you're using it to style a `ul`, not `li`. Maybe something like this?

```
"amsterdam": {
  "unordered-list": {
    "unordered-list": {
      "list-style-type"
      etc.
    }
  }
}
```

packages/css/src/unordered-list/unordered-list.scss
alimpens: No, it should be zero, components shouldn't set their own margins

packages/css/src/list/list.scss
```diff
@@ -1,16 +0,0 @@
-/**
```
alimpens, line 1: We want to keep a seperate List component, right? So maybe we should keep this file, the React component and the story

packages/css/src/unordered-list/unordered-list.scss
dlnr: So box sizing to realise the design? If you need to add `ul` styling to every div that can contain it you'll create a lot of rubish.

dlnr: Not to mention the places that you won't think of where developers can use this component.

proprietary/tokens/src/components/amsterdam/unordered-list.tokens.json
dlnr: `"item"` does give you a logical var name, one that resembles the component name `--amsterdam-unordered-list-item-list-style-type`

dlnr: `--amsterdam-unordered-list-unordered-list-list-style-type` seems a bit excessive.

packages/css/src/list/list.scss
dlnr: Sure, perhaps we should deal with it story DES-23

alimpens (approved):

alimpens (approved):

PR #392, opened 2023-06-30
Rename visual size prop of headings

VincentSmedinga:
This renames `styleLevel` to `size` and connects it to the concept of levels through its values, e.g. `<Heading level="2" size="level-4">`.


storybook/storybook-css/src/components/Heading/Heading.stories.tsx
```diff
@@ -10,26 +10,21 @@ import '@amsterdam/design-system-css/src/heading/heading.scss'
...
     },
-    styleLevel: {
-      options: [1, 2, 3, 4],
-      control: 'select',
-      table: {
-        type: { summary: 'number' },
-        defaultValue: { summary: '1' },
-      },
+    size: {
+      options: ['level-1', 'level-2', 'level-3', 'level-4'],
```
alimpens, line 25: Don't we want to keep this a `select`?

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -14,7 +14,17 @@ const meta = {
...
     level: 1,
-    styleLevel: 1,
+    size: 'level-1',
+  },
+  argTypes: {
+    level: {
+      table: {
+        disable: true,
+      },
+    },
```
alimpens, line 24: In the React story we do want to offer the `level` control, right?

storybook/storybook-css/src/components/Heading/Heading.stories.tsx
VincentSmedinga: I think I prefer radio buttons for about five options or less – we have four here. Saves the user a click and clearly shows the options up front.

We should create a Storybook guideline for this.

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
VincentSmedinga: I copied this from the CSS Story. In both cases, we show separate pages for Heading 1 – 4. I don’t think it makes sense to allow changing the level for one of those if all levels have their own page.


We could combine all headings into one Story which does expose the `level` prop, but I think I prefer the current setup.

alimpens: For CSS stories the level prop is removed because the user has to generate the HTML tags themselves. For the React component the component does that for you. Generally, we show all controls with all options for a component in the first story, with specific settings of props in the subsequent stories. I think that makes sense, if you want to play with the different props you can do so in the first story.

VincentSmedinga: Alright, I’ve restored it.

VincentSmedinga:
I’ve rebased this branch, updating for the Accordion changes, and force-pushed onto `main`.

VincentSmedinga:
Next to making type names singular (the value of a `headingLevel` prop is one heading level, not one heading levels), I propose we require prefixing types and other exports from component files with their component name, hence `Levels` -> `Level` -> `HeadingLevel`.

packages/react/src/Heading/index.ts
```diff
@@ -1,2 +1,3 @@
-export { Heading } from './Heading'
-export type { HeadingProps } from './Heading'
+export type { HeadingLevel, HeadingProps } from './Heading'
```
alimpens, line 1: Do we want to expose `HeadingLevel` seperately to the outside world? It feels more like an internal type imo. When would a consumer need it?

packages/react/src/Heading/index.ts
```diff
@@ -1,2 +1,3 @@
...
+
+export { Heading, getHeadingElement } from './Heading'
```
alimpens, line 3: Same thing for `getHeadingElement`, when would a consumer need this?

packages/react/src/Heading/index.ts
VincentSmedinga: Accordion does ;) That was one of the things I wanted to reconsider.

Although in general I don’t think component exporting their essential related context is a bad thing.

packages/react/src/Heading/index.ts
VincentSmedinga: Not exporting made the unit tests for Accordion fail – although I must say I didn’t investigate very deeply _why_ these tests seem to need access to this function.

packages/react/src/Heading/index.ts
VincentSmedinga: Updated according to our discussion leading to #409.

packages/react/src/Heading/index.ts
VincentSmedinga: Updated according to our discussion leading to #409. Test still working.

alimpens (approved):

PR #398, opened 2023-07-05
Add Accordion component

alimpens:
Couple of questions:

- do we want the `headingLevel` prop on `<Accordion />` to have a default? Or do we want to force users to think about the correct heading level?
- There's currently no check on the amount of children of  `<Accordion />`. The docs say it should have a minimum of 3 and a maximum of 10 children. Do we want to enforce this in code?
- Do we want to ensure `<Accordion />` can only have `<Accordion.Section />` as children? Do we have a pattern for that yet?


packages/css/src/accordion/accordion.scss
```diff
@@ -0,0 +1,63 @@
...
+  padding-block: 0.75rem;
+  padding-inline: 1rem;
+  width: 100%;
+
+  @media screen and (width > $amsterdam-breakpoint) {
+    font-size: var(--amsterdam-accordion-button-font-size-wide);
+  }
+
+  &:focus {
+    outline-offset: var(--amsterdam-accordion-button-focus-outline-offset);
```
dlnr, line 33: Not sure what was agreed on focus outline but a blue border with radius (in Chrome) looks weird around a grey border without radius.

packages/css/src/accordion/accordion.scss
```diff
@@ -0,0 +1,63 @@
...
+  transform: rotate(180deg);
+}
+
+.amsterdam-accordion__panel {
+  display: none;
+  padding-block: 1rem;
+  padding-inline: 1rem;
+}
+
+.amsterdam-accordion__panel--expanded {
```
dlnr, line 61: A nice prefers reduced motion query for an animated icon but nog for the expanding of the content. Seems like a missed opportunity.

packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { ChevronDown } from '@amsterdam/design-system-react-icons'
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren, useContext, useId, useState } from 'react'
+import AccordionContext from './AccordionContext'
+import { getElement } from '../Heading/Heading'
```
dlnr, line 10: Should we name this better and add an index to Heading?
 
```suggestion
import { getHeadingElement } from '../Heading'
```

packages/react/src/Accordion/useFocusWithArrows.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+        default:
+      }
+
+      if (
+        (e.key === KeyboardKeys.ArrowDown ||
+          e.key === KeyboardKeys.ArrowUp ||
+          e.key === KeyboardKeys.ArrowLeft ||
+          e.key === KeyboardKeys.ArrowRight ||
+          e.key === KeyboardKeys.Home ||
```
dlnr, line 82: Doesn't play nice on a mac but nice addition. Maybe add escape to clear focus?

packages/react/src/Accordion/useFocusWithArrows.tsx
```diff
@@ -0,0 +1,97 @@
...
+          }
+
+          break
+        }
+
+        case previous: {
+          if (getIndex(activeElement) !== 0) {
+            el = focusableEls[getIndex(activeElement) - 1]
+          } else if (rotating) {
+            el = focusableEls[focusableEls.length - 1]
```
dlnr, line 59: Perhaps useLast can be used?

```suggestion
            el = focusableEls.findLast((element) => element.classList.contains('amsterdam-accordion__button'))
```

Or some other condition to test if is a focusable element

packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    const HeadingX = getElement(headingLevel)
+    const id = useId()
+    const buttonId = `button-${id}`
+    const panelId = `panel-${id}`
+
+    return (
+      <div className={clsx('amsterdam-accordion__section', className)} ref={ref} {...otherProps}>
+        <HeadingX className={'amsterdam-accordion__header'}>
+          <button
```
dlnr, line 34: Storybook a11y is showing a problem: "Element has focusable descendants", "Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies". I've test Apple Voiceover and could not get the accordion content to focus (and read to me).

packages/css/src/accordion/accordion.scss
alimpens: Agreed, I'll remove the `focus-visible` style

packages/css/src/accordion/accordion.scss
```diff
@@ -0,0 +1,63 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+@import "../../utils/breakpoint";
+
+.amsterdam-accordion__header {
+  display: flex;
+  margin-block: 0;
```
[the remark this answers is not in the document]
alimpens: As discussed, components shouldn't set space on their outside edges

packages/css/src/accordion/accordion.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  &:focus-visible,
+  &:hover {
+    box-shadow: var(--amsterdam-accordion-button-hover-box-shadow);
+  }
+
+  svg {
+    transform: rotate(0deg);
+    transition: transform 0.3s ease;
```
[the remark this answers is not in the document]
alimpens: Good idea, [I made a ticket for this](https://gemeente-amsterdam.atlassian.net/browse/DES-251)

packages/css/src/accordion/accordion.scss
alimpens: I really dislike animating the expanding of content for accordions, it just means you have to wait longer before you see the text.

packages/react/src/Accordion/AccordionSection.tsx
alimpens: I like the rename. I don't think we should add the function to the index of Heading though, we don't really need to expose this function to our consumers

packages/react/src/Accordion/AccordionSection.tsx
alimpens: Good catch, I accidentally had nested interactive elements. Not sure why VoiceOver can't access the accordion content, we have to look into that

packages/react/src/Accordion/useFocusWithArrows.tsx
alimpens: You mean you want to check if `focusableEls.length !== 0`? If `Accordion` doesn't have children, you can't focus on it and `useFocusWithArrows` is never called, so it doesn't seem to be a problem

packages/react/src/Accordion/useFocusWithArrows.tsx
alimpens: What doesn't work?

proprietary/tokens/src/components/amsterdam/accordion.tokens.json
```diff
@@ -0,0 +1,24 @@
...
+        "font-family": { "value": "{amsterdam.typography.font-family}" },
+        "font-size": {
+          "narrow": { "value": "{amsterdam.typography.text-level.5.font-size.narrow}" },
+          "wide": { "value": "{amsterdam.typography.text-level.5.font-size.wide}" }
+        },
+        "font-weight": { "value": "{amsterdam.typography.font-weight.bold}" },
+        "line-height": { "value": "{amsterdam.typography.text-level.5.line-height}" },
+        "hover": {
+          "box-shadow": {
+            "value": "inset 0 0 0 2px {amsterdam.color.neutral-grey3}"
```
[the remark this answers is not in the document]
alimpens: Sure, but this does the same in a single line of css

packages/react/src/Accordion/useFocusWithArrows.tsx
dlnr: Home just moves op to the start of the page, so the keys are already set by the browsers and with Mac that wins over js in the page I guess.

dlnr (approved):

PR #408, opened 2023-07-08
Add initial grid components

VincentSmedinga:


packages/css/src/grid/page-grid.scss
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../utils/breakpoint";
+
+.amsterdam-page-grid {
+  column-gap: 2%;
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  max-width: 1440px;
```
VincentSmedinga, line 12: Should we extract to a constant, like the breakpoint? If so, which of the other values here as well?

packages/react/src/Paragraph/Paragraph.tsx
```diff
@@ -6,19 +6,26 @@
 
 import clsx from 'clsx'
 import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { gridCellClassNames, GridCellProps } from '../grid'
 
-export interface ParagraphProps extends HTMLAttributes<HTMLParagraphElement> {
+export type ParagraphProps = {
```
VincentSmedinga, line 11: Is there a reason to prefer `interface` over `type`? Extending _and_ intersecting doesn’t work, intersecting `HTMLAttributes` does and seems less complex?

packages/react/src/grid/GridCell.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { ReactNode } from 'react'
+import { PageGridColumnNumber } from './PageGrid'
+
+export type GridCellProps = {
+  children: ReactNode
+  gridColumns?: [PageGridColumnNumber, PageGridColumnNumber]
```
VincentSmedinga, line 12: To do: document props, functions, add stories, everything.

packages/react/src/grid/GridCell.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (!gridColumns) {
+    return undefined
+  }
+
+  const [start, end] = gridColumns
+
+  return clsx(start && `amsterdam-grid-start-${start}`, end && `amsterdam-grid-end-${end}`)
+}
+
+export const GridCell = ({ children, gridColumns }: GridCellProps) => (
```
VincentSmedinga, line 25: Decide whether we need this component if many (if not all) other components will include `GridColumnProps` themselves.

packages/react/src/grid/PageGrid.tsx
```diff
@@ -0,0 +1,16 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { ReactNode } from 'react'
+
+export type PageGridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
```
VincentSmedinga, line 8: Find a way to have just four options for the grid on narrow screens.

storybook/storybook-css/src/docs/stories.scss
VincentSmedinga, line 1: Discuss what we should do with these.

We could use some components or class names to use in our stories. E.g. the pink box to highlight otherwise invisible layout. The class names for figure and image seem necessary for a grid example using images, but should be replaced by a reset stylesheet or, better even, our own future components.

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+      layout: 'centered',
+    },
+  },
+}
+
+export const ThreeBlocks: Story = {
+  args: {
+    children: (
+      <>
+        <GridCell gridColumns={[1, 4]}>
```
VincentSmedinga, line 50: This is a compact API. Is it readable enough? Can people infer what it means and get used to it?

The obvious alternative is `<GridCell start={1} end={4}>`. Probably even better. But to allow all components to be placed on a grid, these prop names are too generic. I’d say we’d prefer something like `<Component grid-column-start={1} grid-column-end={4}>` or `<Component grid-column={[1, 4]}>`.

Additionally, we must consider the need to specify grid placement for all breakpoints, of which we currently have two, e.g. `<Component grid-column={{narrow: [1, 1], wide: [1, 4]}>`. Is this acceptable?

storybook/storybook-css/src/docs/stories.scss
```diff
@@ -0,0 +1,15 @@
+.amsterdam-docs-figure {
```
VincentSmedinga, line 1: Discuss what we should do with these styles.


The CSS for figures seemed necessary in one of the grid examples. A reset style sheet could take its place, and the future usage of our own components would obsolete them.

The pink highlighted box is useful to indicate otherwise invisible layout. Could still be useful, but it does clutter the examples. A preset in the toolbar for this story could have the same result, but I couldn’t find complete API docs for a story’s `parameters` very easily.

storybook/storybook-react/src/components/grid/PageGrid.docs.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+### A middle block
+
+Spanning ten columns, numbers 2 up to 11.
+
+<Canvas of={PageGridStories.MiddleBlock} />
+
+### Without the grid cell component
+
+Paragraphs can be used directly on the grid, as they expose the grid cell props as well.
```
VincentSmedinga, line 29: If we like this approach, generalise it for all components.

packages/react/src/grid/PageGrid.tsx
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { PropsWithChildren } from 'react'
+
+export type PageGridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
+
+export type PageGridProps = PropsWithChildren<{}>
+
+export const PageGrid = ({ children }: PageGridProps) => <div className="amsterdam-page-grid">{children}</div>
```
VincentSmedinga, line 12: I’ve called it `PageGrid` because I expect a separate `Grid` component in the future with more props.

This component will always produce the full-page grid of 1440 (1920) pixels, 4 or 12 columns, etc.

packages/react/src/grid/GridCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { PropsWithChildren } from 'react'
+import { PageGridColumnNumber } from './PageGrid'
+
+export type GridCellProps = PropsWithChildren<{
+  gridColumns?: [PageGridColumnNumber, PageGridColumnNumber]
```
VincentSmedinga, line 11: This is a compact API. Is it readable enough? Can people infer what it means and get used to it?

The obvious alternative is `<GridCell start={1} end={4}>`. Probably even better. But to allow all components to be placed on a grid, these prop names are too generic. I’d say we’d prefer something like `<Component grid-column-start={1} grid-column-end={4}>` or `<Component grid-columns={[1, 4]}>`.

(If we decide to keep a grid cell component, we could use unprefixed props for that component only: `<GridCell columns={[1, 4]}>`.

Additionally, we must consider the need to specify grid placement for all breakpoints, of which we currently have two, e.g. `<Component grid-columns={{narrow: [1, 1], wide: [1, 4]}>`. Is this acceptable?

I’ve kept `grid-span` out of scope for now, and named grid cells as well. We might use the latter for page templates components.

VincentSmedinga: Hmm, looking at the example with paragraphs, using spans could offer more simplicity, e.g.

```
  <Paragraph grid-columns={3}>…</Paragraph>
  <Paragraph grid-columns={9}>…</Paragraph>
```

The user would then be responsible to find numbers correctly adding up to 12. And for more creative layouts, we would need to combine props, e.g.:

```
  <Paragraph grid-column-start={2} grid-columns={3}>…</Paragraph>
  <Paragraph grid-columns={7}>…</Paragraph>
```

VincentSmedinga:
To do: documentation and tests.

storybook/storybook-css/src/docs/stories.scss
dlnr: Or perhaps a debug option for the grid that creates a pseudo element that is pink? I'm also curious to why NLDS doesn't contain a grid element.

dlnr (commented):
Just some comments that came up at first glance.

packages/react/src/Paragraph/Paragraph.tsx
VincentSmedinga: I do think it will prove useful to assign a combination of reusable prop objects (interfaces or type aliases) to a component like `Paragraph`.

In general, both concepts support combinations: interfaces can be [extended](https://www.typescriptlang.org/docs/handbook/interfaces.html#extending-interfaces), type aliases can be [intersected](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types).

The reason for my comment is that I wanted to add `& GridColumnProps` to this `ParagraphProps` interface to make a paragraph accept props like `grid-column-start`, and found that intersections aren’t for interfaces. That’s why I converted it to a type alias for now. For the Amsterdam app, we use them exclusively instead of interfaces. Made me curious after the difference and why someone chose interfaces over type aliases here.

While researching these terms I learned:
* “Differences in how conflicts are handled is the main reason to pick one over the other between an interface and a type alias of an intersection type.”
* “Type aliases, unlike interfaces, can describe more than just object types.”
* “Type aliases and interfaces are very similar, and in many cases you can choose between them freely. Almost all features of an interface are available in type, the key distinction is that a type cannot be re-opened to add new properties vs an interface which is always extendable.”

”For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type” ([source](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)).

 We could go for interfaces here as we already started out with them and the TS docs suggest so. I’ll check if I can use them in my grid work as well.

storybook/storybook-css/src/docs/stories.scss
VincentSmedinga: No-one implementing NLDS has started with an any specific layout as far as I know. Not sure why. It’s one of my focus points in design systems, so I guess we’ll be taking this for the team then.

A pink highlight will be useful in other layout components than grid alone, e.g. in our flex rows and columns, in size or aspect ratio components, or to document sizing design tokens.

I’ve created quite the arsenal of component specific for the design system docs in an earlier project, that’s why I’m kicking off with this approach here as well.

Instead of class names, I could also make a `<PinkBox>` component somewhere in a directory like `components/docs`. But we should probably check what Storybook offers us here first.

packages/react/src/Paragraph/Paragraph.tsx
```diff
@@ -6,19 +6,26 @@
...
   size?: 'small' | 'large'
-}
+} & GridCellProps &
+  HTMLAttributes<HTMLParagraphElement>
 
 export const Paragraph = forwardRef(
   (
-    { children, size, className, ...otherProps }: PropsWithChildren<ParagraphProps>,
+    { children, className, gridColumns, size, ...otherProps }: PropsWithChildren<ParagraphProps>,
```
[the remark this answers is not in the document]
VincentSmedinga: I’m not sure why we would limit the type of components or elements that could be grid items. Why not use a grid to line up two paragraphs next to each other?

What’s funny is that I would have expected this discussion to go the other way around :) I’m all for single responsibility, so I don’t principally object against requiring `Grid.Cell` components with `start`, `end` and `span` props as children of `Grid`. I would have expected you to frown upon this ‘unnecessary div`. Am I mistaken?

packages/css/src/grid/page-grid.scss
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../utils/breakpoint";
+
+.amsterdam-page-grid {
+  column-gap: 2%;
```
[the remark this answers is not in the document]
dlnr: I like the last one with em's but clamp could prevent the gap from "exploding" when people have their font size maxed out.

I came across this article, it might help 

https://moderncss.dev/contextual-spacing-for-intrinsic-web-design/#gap

packages/react/src/Paragraph/Paragraph.tsx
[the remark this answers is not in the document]
dlnr: Sure the example in storybook seems valid, it's just that for this case i would prefer:

```
columns: 2;
```

packages/react/src/grid/GridCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+}>
+
+export const gridCellClassNames = (gridColumns?: [PageGridColumnNumber, PageGridColumnNumber]) => {
+  if (!gridColumns) {
```
hcvdhaar, line 15: just a shorthand return would be enough. But not sure what the code style convention we have.
```
if (!gridColumns) return
```

packages/react/src/grid/GridCell.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return undefined
+  }
+
+  const [start, end] = gridColumns
```
hcvdhaar, line 19: Could start be 10 and end 5? Otherwise we might need some validation?

packages/react/src/grid/PageGrid.tsx
```diff
@@ -0,0 +1,14 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { PropsWithChildren } from 'react'
+
+export type PageGridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
```
hcvdhaar, line 8: Is no being use here. but in the `GridCell` component. Would opt for declaring the type there (to keep stuff close to where it is being used) or create separate type file for it.

packages/react/src/grid/PageGrid.tsx
hcvdhaar: Could it have other children than `GridCell`? If they are closely related then having it as a sub component might be an option. This does not restrict the client from providing other elements by the API is more clear.
```
<PageGrid>
  <PageGrid.Cell>
  <PageGrid.Cell>
</PageGrid>

File: packages/react/src/grid/GridCell.tsx (hunk 10 above)
Comment (alimpens) [reply packages/react/src/grid/GridCell.tsx:11]: > `gridColumns={[1, 4]}`

I'm not sure about this API, it doesn't feel very intuitive to me. Wouldn't using an object there work?

Something like:

`gridColumns={{ start: 1, end: 4 }}`

And for above and below breakpoint:

`gridColumns={{ narrow: { start: 1, end: 4 }, wide: { start: 1, end: 8 }}}`

I think Typescript will be more of a help with objects as well. 

I'd personally opt for the `span` approach over the `end` approach, it's easier for me to say 'start at column 3 and span 6 columns' than 'start at column 3 and then stop at 3 plus 6 is 9, so column 9'.

File: storybook/storybook-css/src/docs/stories.scss (hunk 7 above)
Comment (alimpens) [reply storybook/storybook-css/src/docs/stories.scss:1]: Prefixing this with `amsterdam-` was actually a little confusing to me 😅 I thought it was styling that had to do with a component, and I didn't understand why it was in `/storybook`. Maybe just use `docs-`? And call the Sass file `docs.scss` as well?

File: storybook/storybook-react/src/components/grid/PageGrid.stories.tsx [hunk 16]
```diff
@@ -0,0 +1,119 @@
...
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    children: (
+      <>
+        {Array.from(Array(12).keys()).map((i) => (
+          <GridCell key={i}>
+            <div className="amsterdam-docs-pink-box">
```
alimpens, line 31: I think we can add the `amsterdam-docs-pink-box` class directly on the `Paragraph` component, no? Cleans up the code view in SB a bit

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+} satisfies Meta<typeof PageGrid>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    children: (
+      <>
```
alimpens, line 28: I don't think this fragment is necessary, Storybook doesn't render it well either

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+    docs: {
+      layout: 'centered',
+    },
+  },
+}
+
+export const ThreeColumns: Story = {
+  args: {
+    children: (
+      <>
```
alimpens, line 49: We can lose the fragment here and wrap the `GridCell`s in an array, to improve the Storybook code view. [See the Unordered List stories for an example](https://github.com/Amsterdam/design-system/blob/main/storybook/storybook-react/src/components/UnorderedList/UnorderedList.stories.tsx)

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </GridCell>
+      </>
+    ),
+  },
+}
+
+export const ThreeImages: Story = {
+  args: {
+    children: (
+      <>
```
alimpens, line 73: Same, fragment isn't necessary

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </div>
+      </GridCell>
+    ),
+  },
+}
+
+export const Paragraphs: Story = {
+  args: {
+    children: (
+      <>
```
alimpens, line 109: Same, fragment isn't necessary

packages/react/src/grid/PageGrid.tsx
alimpens: It's just two different types, right? `PageGridColumnNumberWide` and `PageGridColumnNumberNarrow`?

storybook/storybook-css/src/docs/stories.scss
VincentSmedinga: Yea something like that – the prefix was necessary because of an ESLint rule 😏

packages/react/src/grid/PageGrid.tsx
hcvdhaar: Could be, but how would the component know on render when type to use or what the restriction is?

packages/css/src/grid/page-grid.scss
VincentSmedinga: I’ve extracted tokens for `column-gap`, `max-width`, and `column-count`.

packages/react/src/Paragraph/Paragraph.tsx
VincentSmedinga: I’ll revert this change from `interface` to `type` for paragraph as the subject of this branch is grids. Interfaces can be extended from multiple types. Let’s continue this discussion later.

packages/react/src/grid/PageGrid.tsx
VincentSmedinga: I think the usual approach is to specify grid config per breakpoint, e.g.

```
export type GridCellProps = PropsWithChildren<{
  gridColumns?: { 
    narrow: [PageGridColumnNumberNarrow, PageGridColumnNumberNarrow] 
    wide: [PageGridColumnNumberWide, PageGridColumnNumberWide] 
  }
}>
```

To allow merging this PR today, I created a separate story in Jira for the narrow screens API.

packages/react/src/grid/GridCell.tsx
VincentSmedinga: Let’s start with just `Paragraph` extending `GridCellProps` and add more later.

storybook/storybook-css/src/docs/stories.scss
VincentSmedinga: Later.

packages/react/src/grid/PageGrid.tsx
VincentSmedinga: While I like that pattern in general, I feel we’ll have multiple `*Grid` components, but just one `GridCell` component. Plus, we’ll allow all components that implement `GridCellProps` as children of `PageGrid` and other grid components. So I’ll leave it like this for now.

packages/react/src/grid/GridCell.tsx
VincentSmedinga: Yes, touched on the multiple-breakpoint API above, moved out of scope for this PR.

The object notation is a bit verbose but much clearer, certainly after supporting multiple breakpoints.

I agree with starting with `start` and `span` rather than `start` and `end`.

VincentSmedinga: I even made `start` optional and a single `span` the default, so we can abbreviate `<Component gridColumns={{start: 1, span: 4}}>` to `<Component gridColumns={4}>`.

packages/react/src/Paragraph/Paragraph.tsx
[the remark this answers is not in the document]
VincentSmedinga: I still think we can get by with a reusable set of specific grid column props, if we also take care that they have exclusivity, i.e. no other component tampers with grid column settings.

That polymorphic approach looks promising though, let’s experiment with it.

Not sure about opening up utility class names on React components, I think that bestows just a bit too much freedom on our consumers.

packages/react/src/grid/GridCell.tsx
VincentSmedinga: We’ve avoided this now that we use `start` and `span` rather than `end`.

packages/css/src/grid/grid-cell.scss
```diff
@@ -0,0 +1,104 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../utils/breakpoint";
+
+@media (min-width: $amsterdam-breakpoint) {
```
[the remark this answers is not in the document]
VincentSmedinga: Alright!

packages/react/src/grid/PageGrid.tsx
VincentSmedinga: My reasoning: the grid cell is asking the grid how many columns it consists of.

The type is indeed not used in the grid component, but the essence is that the number 12 is the grid’s responsibility.

I’m keeping it here.

packages/css/src/grid/grid-cell.scss
```diff
@@ -0,0 +1,104 @@
...
+  .amsterdam-grid-start-1 {
```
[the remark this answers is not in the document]
VincentSmedinga: We could, and we will!

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
VincentSmedinga: Should work but it doesn’t? Could you try next week?

packages/react/src/grid/GridCell.tsx
VincentSmedinga: Changed it to return an empty string – I missed that this wasn’t a component but a function, and our [consistent-return](https://eslint.org/docs/latest/rules/consistent-return) rule of ESLint wants all return values to be of the same type.

storybook/storybook-react/src/components/grid/PageGrid.stories.tsx
VincentSmedinga: The class name sets a padding, so it’s more suitable for a container than for a paragraph. We’ll improve on this concept in the near future.

alimpens (approved):

PR #409, opened 2023-07-10
Add React code convention docs

alimpens:


packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
+# Coding Conventions
+
+## Barrel files
+
+In our React package we use barrel files (`index.ts` in the root and in every component folder) to simplify the import of React components.
```
VincentSmedinga, line 5:
```suggestion
We use barrel files (`index.ts` in both the package root and each component directory) to make it easier for consuming applications to import components from our React library.
```

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
...
+Barrel files allow a consumer to do this:
```
VincentSmedinga, line 6: While ‘consumer’ is an explicit and accurate term, would we be equally clear if we use ‘application’ everywhere and mention ourselves being a ‘library’ as an opposition where useful?

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
...
+
+`import { Heading, Link, Paragraph } from '@amsterdam/design-system-react'`
+
+Instead of this:
+
+```js
+import Heading from "@amsterdam/design-system-react/dist/Heading/Heading";
+import Link from "@amsterdam/design-system-react/dist/Link/Link";
+import Paragraph from "@amsterdam/design-system-react/dist/Paragraph/Paragraph";
```
VincentSmedinga, line 15:
```suggestion
import Heading from "@amsterdam/design-system-react/dist/heading/Heading";
import Link from "@amsterdam/design-system-react/dist/link/Link";
import Paragraph from "@amsterdam/design-system-react/dist/paragraph/Paragraph";
```

I think directory names should be lower snake case and will submit a PR for this asap. We could already apply this here.

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
+# Coding Conventions
+
+## Barrel files
+
+In our React package we use barrel files (`index.ts` in the root and in every component folder) to simplify the import of React components.
+Barrel files allow a consumer to do this:
+
+`import { Heading, Link, Paragraph } from '@amsterdam/design-system-react'`
+
+Instead of this:
```
VincentSmedinga, line 10: Oh wait, I hadn’t realised the `dist` directory being necessary below. We could highlight the undesirability of that approach:

```suggestion
Instead of a separate import for each component, while even reaching into the `dist` directory:
```

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
...
+```js
+import Heading from "@amsterdam/design-system-react/dist/Heading/Heading";
+import Link from "@amsterdam/design-system-react/dist/Link/Link";
+import Paragraph from "@amsterdam/design-system-react/dist/Paragraph/Paragraph";
+```
+
+However, barrel files have 2 potential pitfalls:
+
+- Without proper treeshaking, importing from a barrel file can mean that a consumer imports the entire library, instead of just the components they need. To avoid this, the consumer needs to set up treeshaking in their project (TODO: add documentation on this? Or at least describe both methods of importing?)
+- When a component imports from a sibling component, using a barrel file for this import can lead to cyclical dependencies.
```
VincentSmedinga, line 21:
```suggestion
However, barrel files have two potential pitfalls:

1. Without proper treeshaking, importing from a barrel file can mean that a consumer imports the entire library, instead of just the components they need. To avoid this, the consumer needs to set up treeshaking in their project (TODO: add documentation on this? Or at least describe both methods of importing?)
2. If a component imports from a sibling component, using a barrel file can lead to cyclical dependencies.
```

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
...
+
+We can avoid this last pitfall by adhering to the following code convention: barrel file imports should only be used to import from a package, not within one.
+
+To illustrate: If the `Accordion` React component needs the `HeadingLevel` type from the `Heading` React component, it should import it like this:
+
+`import { HeadingLevel } from '../Heading/Heading'`
```
VincentSmedinga, line 27:
```suggestion
`import { HeadingLevel } from '../heading/Heading'`
```

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
...
+
+Instead of this:
+
+`import { HeadingLevel } from '../Heading'` or `import { HeadingLevel } from '../'`
```
VincentSmedinga, line 31:
```suggestion
`import { HeadingLevel } from '../heading'` or `import { HeadingLevel } from '../'`
```

packages/react/documentation/coding-conventions.md
```diff
@@ -0,0 +1,31 @@
+# Coding Conventions
+
+## Barrel files
```
VincentSmedinga, line 3: Using the guideline itself for the title makes it easier to find.

```suggestion
## Import components through barrel files only from outside the package
```

VincentSmedinga (changes requested):

packages/react/documentation/coding-conventions.md
alimpens: I think having directory names for React components starting with a capital is a more common convention than snake case, any reason to change this?

packages/react/documentation/coding-conventions.md
alimpens: [Numbers are more easy to understand as numbers vs text](https://www.amsterdam.nl/schrijfwijzer/heldere-taal-basis-onze-huisstijl/tekstonderdelen-heldere-taal/getallen-bedragen-breuken-percentages/).

Having an ordered list makes sense though, I'll change it!

packages/react/documentation/coding-conventions.md
VincentSmedinga: I understand. The ambiguousness of ‘consumer’ referring to either an application or its developer seems a bit helpful here.

I still think the term is a bit difficult, but let’s postpone for now.

packages/react/documentation/coding-conventions.md
VincentSmedinga: I wasn't aware of the convention and I understand it, especially for two-word components which would have different spelling in their directory and file name, so let’s follow along.

packages/react/documentation/coding-conventions.md
VincentSmedinga: I updated the suggestion to return the 2.

packages/react/documentation/coding-conventions.md
VincentSmedinga: Sure, but in the context of how to import files right?

VincentSmedinga (approved):

PR #442, opened 2023-07-19
Link tweaks

alimpens:
- I did some refactoring of the css and tokens, I removed some unused tokens and restructured the css a bit.
- I removed the standalone bold variant, because that doesn't seem to exist in the designs? I think there was a mixup between standalone and menu links, menu links have a bold variant. Menu links currently don't exist, we should make a seperate ticket for adding those
- I updated the readme
- I did some other stuff [described in the ticket](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378?modal=detail&selectedIssue=DES-256)


packages/css/src/link/link.scss
```diff
@@ -6,45 +6,41 @@
...
+  color: var(--amsterdam-link-color);
   font-family: var(--amsterdam-paragraph-font-family);
   font-size: var(--amsterdam-link-font-size-narrow);
+  font-weight: var(--amsterdam-link-font-weight);
   line-height: var(--amsterdam-link-line-height);
   outline-offset: var(--amsterdam-link-outline-offset);
-  text-decoration: var(--amsterdam-link-text-decoration, underline);
-  text-decoration-color: var(--amsterdam-link-text-decoration-color, currentColor);
-  text-decoration-thickness: var(--amsterdam-link-text-decoration-thickness, 1px);
+  padding-block: 0.5rem;
```
dlnr, line 15: This seems dangerous as default

packages/css/src/link/link.scss
```diff
@@ -6,45 +6,41 @@
...
-
+  &:active,
   &:focus {
     color: var(--amsterdam-link-focus-color);
-    text-decoration: var(--amsterdam-link-focus-text-decoration);
-  }
-
-  &:visited {
-    color: var(--amsterdam-link-visited-color);
+    text-decoration-thickness: var(--amsterdam-link-focus-text-decoration-thickness);
```
dlnr, line 23: This says focus thickness, not visited

packages/react/src/Link/Link.tsx
```diff
@@ -28,15 +40,15 @@ export const Link = forwardRef(
...
           'amsterdam-link--in-list': variant === 'inList',
+          'amsterdam-link--in-list__chevron': variant === 'inList' && !icon,
           'amsterdam-link--on-background-light': onBackground === 'light',
           'amsterdam-link--on-background-dark': onBackground === 'dark',
         },
         className,
       )}
     >
-      {variant === 'inList' && <Icon svg={ChevronRight} size="level-7" />}
+      {variant === 'inList' && <Icon svg={icon ? icon : ChevronRight} size="level-7" />}
```
dlnr, line 51: Perhaps inlist icon size could become a token? I thought you mentioned it could be one of 3 sizes.

proprietary/tokens/src/components/amsterdam/link.tokens.json
```diff
@@ -73,18 +65,6 @@
       "on-background-light": {
         "color": { "value": "{amsterdam.color.primary-black}" },
         "text-decoration": { "value": "underline" },
-        "text-decoration-thickness": { "value": "2px" },
```
dlnr, line 76: Mixin these up with inline could cause confusion later. For example I would have opted for the standalone to use on background.

storybook/storybook-react/src/components/Link/Link.docs.mdx
```diff
@@ -18,10 +18,6 @@ Op zichzelf staande koppelingen staan direct na inhoud. Ze mogen niet in zinnen
 
 <Canvas of={LinkStories.Standalone} />
 
-### Standalone Bold
-
-<Canvas of={LinkStories.StandaloneBold} />
```
dlnr, line 23: Perhaps Figma should be updated?

storybook/storybook-react/src/components/Link/Link.stories.tsx
```diff
@@ -106,6 +100,16 @@ export const InList: Story = {
   },
 }
 
+export const InListWithIcon: Story = {
+  args: {
+    children: 'Linktekst',
+    href: '#',
+    variant: 'inList',
+    onBackground: undefined,
+    icon: Email,
```
dlnr, line 109: No Chevron?

packages/css/src/link/link.scss
alimpens: It doesn't apply to the `visited` state, only to `hover`, `active` and `focus`. The standalone link doesn't have `visited` state styling.

packages/react/src/Link/Link.tsx
alimpens: For `inList` links it can only be 16x16px (for the default chevron) or text level 7 (for all other icons). Icon sizing is still a bit messy though, we should discuss it with design

proprietary/tokens/src/components/amsterdam/link.tokens.json
alimpens: Are you refering to the `.amsterdam-link--inline.amsterdam-link--on-background-dark` selector? Inline links are the only links that have a different styling on light or dark background, other than the standard color change. It makes more sense to me to handle that difference seperately, instead of applying these styles to all link variants, even the ones that don't need it.

Also, I'm kind of hoping the difference in styling (other than color) between inline links on dark/light background and the regular inline link styling will be removed in the designs. Inline links should have an underline in all cases imo

storybook/storybook-react/src/components/Link/Link.docs.mdx
alimpens: Yeah, the documentation on link variants are a little confusing at the moment. I've added a comment in the Figma file.

storybook/storybook-react/src/components/Link/Link.stories.tsx
alimpens: Do you mean we should disallow chevrons in the `icons` prop?

packages/css/src/link/link.scss
dlnr: You don't yet know where links are going to be used, the padding could become unwanted in certain use cases. Developers will create overrides.

packages/css/src/link/link.scss
dlnr: ah I missed the red diff

storybook/storybook-react/src/components/Link/Link.stories.tsx
dlnr: No, I mean the inList example in Figma has a chevron

proprietary/tokens/src/components/amsterdam/link.tokens.json
dlnr: Yes the selector, I've tried to match the selectors with token names to prevent confusion, and treat them as isolated states/components. Mixing it up to save on a couple of lines of CSS seems wrong.

packages/css/src/link/link.scss
alimpens: > Developers will create overrides.

They shouldn't ;) If the component height is based only on the `line-height` of its text, it's not large enough to comply with [this WCAG criterion](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html). Same as the Button component for example, this component needs padding as part of its intrinsic height. Similar to Button, devs can mess with this padding, but that would mean they'd be 'breaking' the component

proprietary/tokens/src/components/amsterdam/link.tokens.json
alimpens: It's not really about saving a couple of lines of css. In the current version the `text-decoration` stuff is applied to all variants, but it's only relevant for the `inline` variant. I could add `on-background-dark/light-inline-text-decoration` tokens? The selector would be the same though, I think. It's styling that only applies if the component has both `amsterdam-link--inline` and `amsterdam-link--on-background-dark` classes. Although I guess we could nest the selector, so something like:

```css
.amsterdam-link--on-background-light {
  &.amsterdam-link--inline {
  }
}
```

storybook/storybook-react/src/components/Link/Link.stories.tsx
alimpens: An `inList` link has a chevron by default, so the `InList` story already shows the chevron

proprietary/tokens/src/components/amsterdam/link.tokens.json
VincentSmedinga: The design has been updated to always underline all inline links, so this will no longer be an issue.

packages/css/src/link/link.scss
```diff
@@ -6,45 +6,41 @@
 @import "../../utils/breakpoint";
 
 .amsterdam-link {
-  color: var(--amsterdam-link-color, blue);
-  font-family: var(--amsterdam-paragraph-font-family);
+  color: var(--amsterdam-link-color);
+  font-family: var(--amsterdam-link-font-family);
```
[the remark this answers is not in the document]
VincentSmedinga: Discussed this. Conclusion: yes, components must only use their own tokens to make reuse easier. E.g. another implementer of our components may want to have different font families for paragraph and link and this would not be possible if links use the paragraph font tokens.

storybook/storybook-react/src/components/Link/Link.stories.tsx
VincentSmedinga: The chevron icon is the default, but it can be overridden.

storybook/storybook-react/src/components/Link/Link.stories.tsx
```diff
@@ -102,6 +96,16 @@ export const InList: Story = {
...
```
VincentSmedinga, line 105: I can’t change the icon in Storybook.

packages/react/src/Link/Link.tsx
VincentSmedinga, line 51: Let’s use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) in cases like this.

```suggestion
      {variant === 'inList' && <Icon svg={icon ?? ChevronRight} size="level-7" />}
```

VincentSmedinga (changes requested):

packages/css/src/link/link.scss
alimpens: Using `min-height` causes issues with the alignment of icons to text in links. For now we'll use padding.

VincentSmedinga (approved):

PR #443, opened 2023-07-19
Export both styled and unstyled components

alimpens:


packages/react/rollup.config.mjs
```diff
@@ -39,6 +39,47 @@ export default [
...
+        globals: outputGlobals,
+      },
+      {
+        file: './unstyled/index.js',
+        format: 'esm',
+        sourcemap: true,
+        globals: outputGlobals,
+      },
+    ],
+    external: [/@babel\/runtime/, 'react-dom', 'react'],
```
hcvdhaar, line 82: Duplicate?

hcvdhaar:
Not sure about the name `unstyled`, would opt for something like `base` or `raw`.

packages/react/rollup.config.mjs
hcvdhaar: nevermind, not on the same level

alimpens:
> Not sure about the name `unstyled`, would opt for something like `base` or `raw`.

I thought about this a bit. Names that are often used for unstyled components are 'headless', 'whitelabel' or 'primitive'. Those options all feel a bit cryptic to me though, 'unstyled' feels like a more accurate description of what it is. I'll start a discussion in Slack

dlnr (dismissed):

PR #444, opened 2023-07-21
DES-23 Add `markers` prop to lists, correct whitespace, add documentation

dlnr:


storybook/storybook-css/src/components/UnorderedList/UnorderedList.docs.mdx
```diff
@@ -9,3 +9,10 @@ import README from "../../../../../packages/css/src/unordered-list/README.md?raw
 <Primary />
 
 <Controls />
+
+### No Marker
+
+Voor een lijst zonder marker gebruik je de `no-marker` modifier.
+Deze wordt gebruikt voor onder andere de navigatie in de header of footer maar ook voor zoekresultaten.
```
VincentSmedinga, line 16: How do you like this?

```suggestion
### Zonder opsommingsteken

Gebruik dit component niet alleen voor tekstuele opsommingen, maar ook voor lijsten van grotere secties zoals een overzicht van artikelen of een verzameling zoekresultaten. Semantisch zijn dit immers ook lijsten.

Gebruikers van schermlezers krijgen op deze manier ook informatie over de lengte van de lijst en het volgnummer van het item waarop ze focus zetten.

Je maakt zo’n ‘onzichtbare’ lijst via de `no-marker` modifier.
```

VincentSmedinga: The last piece can become ‘de `noMarker` prop’ in the docs for the React component.

VincentSmedinga (approved):

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -29,7 +32,27 @@
...
 }
 
-.amsterdam-unordered-list__item {
+.amsterdam-unordered-list .amsterdam-unordered-list__item {
+  padding-inline: var(--amsterdam-unordered-list-unordered-list-item-padding-inline);
+}
+
+.amsterdam-unordered-list--no-marker {
+  list-style-type: none;
+  margin-block: 0;
```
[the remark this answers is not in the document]
dlnr: Pushed

VincentSmedinga (dismissed):

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -29,7 +38,24 @@
   @include reset;
 }
 
-.amsterdam-unordered-list__item {
+.amsterdam-unordered-list > .amsterdam-unordered-list__item {
```
alimpens, line 41: Why was the `>` selector necessary again? An `unordered-list__item` always has an `unordered-list` as a parent, no?

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -29,7 +38,24 @@
...
+  padding-inline: var(--amsterdam-unordered-list-unordered-list-item-padding-inline);
```
alimpens, line 42: This padding should only apply to list items in nested lists, right? I think it applies to all list items now

packages/react/src/UnorderedList/UnorderedList.tsx
```diff
@@ -13,13 +13,21 @@ interface UnorderedListComponent
...
 }
 
+export interface UnorderedListProps extends HTMLAttributes<HTMLUListElement> {
+  noMarker?: boolean
+}
+
 export const UnorderedList = forwardRef(
   (
-    { children, className, ...restProps }: PropsWithChildren<HTMLAttributes<HTMLUListElement>>,
+    { children, noMarker = false, className, ...restProps }: PropsWithChildren<UnorderedListProps>,
```
alimpens, line 22: I thought we opted for `noMarkers`? We would have to change this in the classes and story names as well

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
```diff
@@ -9,3 +9,10 @@ import README from "../../../../../packages/css/src/unordered-list/README.md?raw
 <Primary />
 
 <Controls />
+
+### No Marker
+
+Voor een lijst zonder marker gebruik je de `no-marker` modifier.
```
alimpens, line 15:
```suggestion
Voor een lijst zonder marker gebruik je de `noMarker` _property_.
```

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
```diff
@@ -9,3 +9,10 @@ import README from "../../../../../packages/css/src/unordered-list/README.md?raw
...
+Deze wordt gebruikt voor onder andere de navigatie in de header of footer.
```
alimpens, line 16: The css stories will be removed soon, I would move all documentation you've added there to the React story

packages/css/src/unordered-list/unordered-list.scss
VincentSmedinga: Otherwise, list items of a nested list will be affected. The intention is to aim for the items of this specific list only, the selector should express that.

alimpens: The current selector doesn't meet that goal though, I think? Nested lists have the following class structure:

```css
amsterdam-unordered-list
  unordered-list__item <- this is selected
  amsterdam-unordered-list
    unordered-list__item <- but this is selected as well
```

VincentSmedinga: Yes but that is accidental inheritance, right? Compare how we also assign a font family to an inline link.

Additionally, if we want to support mixing ordered and unordered lists, we need to make sure nothing gets mixed up between the two variations of list item.

Come to think of it, shouldn’t we have `amsterdam-ordered-list__item` and `amsterdam-unordered-list__item`?

packages/css/src/unordered-list/unordered-list.scss
VincentSmedinga: Let’s fix all padding when we update the list gap in DES-240.

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
VincentSmedinga: Ik heb deze alinea al iets anders verwoord.

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
VincentSmedinga: This has already been done.

packages/react/src/UnorderedList/UnorderedList.tsx
VincentSmedinga: Yes, I’ll update this.

packages/css/src/unordered-list/unordered-list.scss
VincentSmedinga: Discussed and agreed. We’ll improve this soon.

VincentSmedinga (dismissed):

VincentSmedinga:
This applies correct whitespace to indent list items, aligns the bullets and numbers, uses the square marker from the font.

Design tokens are expressed in rems, 1rem being 16px, and the requested indentation at level 1 being 40px. The indentation of level 2 is tricky for the unordered list as the dashes must align with the text of the parent item (while the level 1 bullet is centred).

I’ve renamed the `noMarkers` prop to `markers` to prevent 🤯 for `noMarkers={true}` or `{noMarkers=false}`.

Used representative content for all examples and added a few more, e.g. for the `start` and `reversed` HTML attributes. Removed the `README.md` as it didn’t add much any more and I’d like more freedom in the sections of a docs page. I’ll prepare a guideline to discuss.

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
```diff
@@ -1,15 +1,52 @@
-import { Canvas, Controls, Markdown, Meta, Primary } from "@storybook/blocks";
+import { Canvas, Meta } from "@storybook/blocks";
 import * as OrderedListStories from "./OrderedList.stories.tsx";
-import README from "../../../../../packages/css/src/ordered-list/README.md?raw";
 
 <Meta of={OrderedListStories} />
 
-<Markdown>{README}</Markdown>
+# Ordered list
```
alimpens, line 6: There are / were two reasons for importing the readme here:

1. If you have the same documentation in multiple places, it's easier / more maintainable to define it once and use it multiple times. We currently don't have multiple storybooks anymore, but in the future we will, so this might become relevant again.
2. If you look up the component directly on github, you would have (part of) the docs there as well. 

I agree that we should rethink this approach, but maybe we should make a decision on that first, before implementing a different approach in this PR?

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
```diff
@@ -1,15 +1,52 @@
...
 
-<Primary />
+Een geordende lijst is een verticale groep gerelateerde inhoud met een volgorde.
+
+## Voorbeelden
+
+### Basis
+
+De lijst heeft oplopende getallen als opsommingstekens.
+Er is genoeg ruimte voor getallen tot 99.
```
alimpens, line 15:
```suggestion
Er is genoeg ruimte voor getallen tot en met 99.
```

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
```diff
@@ -1,15 +1,52 @@
...
+Tussen de items staat extra witruimte.
+Dat geeft een beter onderscheid tussen de items, zeker als ze uit meerdere regels tekst bestaan.
+
+<Canvas of={OrderedListStories.Basis} />
 
 ### Twee niveaus
 
-<Canvas of={OrderedListStories.TwoLevels} />
+Een lijst mag één dieper niveau hebben.
+In dit geval worden onderkast letters gebruikt als opsommingsteken.
```
alimpens, line 24:
```suggestion
In dit geval worden kleine letters gebruikt als opsommingsteken.
```

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
```diff
@@ -1,15 +1,52 @@
...
+
+<Canvas of={OrderedListStories.AflopendeNummering} />
+
+### Lijst zonder opsommingstekens
+
+Een overzicht van artikelen of zoekresultaten is ook een geordende lijst.
+Gebruik in dit soort gevallen een lijst zonder opsommingstekens om die semantiek aan te brengen.
+Daardoor kunnen schermlezers helpen de lijst te begrijpen en te navigeren.
+
+In dit geval verwacht de lijst dat componenten in de items de typografie definiëren.
```
alimpens, line 48: This sentence is a bit tricky to parse. Maybe something like this?

```suggestion
Let op: deze lijst heeft geen stijling (zoals typografie en dergelijke) voor zijn items. Deze stijling moet je in de componenten in de items definiëren.
```

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
```diff
@@ -1,11 +1,38 @@
...
-<Controls />
+## Voorbeelden
+
+### Basis
+
+Een ongeordende lijst heeft vierkante opsommingstekens.
+Tussen de items staat extra witruimte.
+Dat geeft een beter onderscheid tussen de items, zeker als ze uit meerdere regels tekst bestaan.
+
+<Canvas of={UnorderedListStories.Basis} />
```
alimpens, line 18: The controls are missing here

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
```diff
@@ -1,15 +1,52 @@
...
+## Voorbeelden
+
+### Basis
+
+De lijst heeft oplopende getallen als opsommingstekens.
+Er is genoeg ruimte voor getallen tot 99.
+Tussen de items staat extra witruimte.
+Dat geeft een beter onderscheid tussen de items, zeker als ze uit meerdere regels tekst bestaan.
+
+<Canvas of={OrderedListStories.Basis} />
```
alimpens, line 19: The controls are missing here

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
+  li {
+    margin-inline-start: var(--amsterdam-ordered-list-item-margin-inline-start);
+    padding-inline-start: var(--amsterdam-ordered-list-item-padding-inline-start);
+  }
 }
 
+/** A nested list has a different marker and less spacing for correct alignment. */
 .amsterdam-ordered-list .amsterdam-ordered-list {
   list-style-type: var(--amsterdam-ordered-list-ordered-list-list-style-type);
-  padding-block: var(--amsterdam-ordered-list-ordered-list-padding-block);
```
alimpens, line 37: Shouldn't there be `0.5rem` spacing between an item's text and its nested list as well?

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
 
 @import "../../utils/breakpoint";
 
+/**
+ * - Applies border box sizing.
+ * - Resets whitespace and list styles.
+ */
```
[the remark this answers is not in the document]
VincentSmedinga: Discussed: we’ll continue documenting CSS blocks like this, even if they seem silly, to explain the implementation choices we made.

Things can be implemented in various ways (e.g. flex or grid, list `::marker` or `::before` pseudo element) and we want to document the choices we made there for later. We prefer documenting this in JSDoc near the code itself over having a separate `architecture.md` file.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
+
+  display: grid;
+  gap: var(--amsterdam-ordered-list-gap);
+}
+
+/**
+ * A list with markers:
+ * - sets typography for its text content
+ * - applies correct spacing at the start of items.
+ */
```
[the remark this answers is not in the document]
VincentSmedinga: See above.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
   @media screen and (width > $amsterdam-breakpoint) {
     font-size: var(--amsterdam-ordered-list-wide-font-size);
   }
 
-  @include reset;
-}
-
-.amsterdam-ordered-list__item {
-  padding-inline: var(--amsterdam-ordered-list-item-padding-inline);
+  li {
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, most certainly. I’ve made a note for a guideline not to style elements directly.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
   margin-block: 0;
+  padding-inline-start: 0;
   -webkit-text-size-adjust: 100%;
 }
 
+/** A list is a grid column with a gap between items. */
 .amsterdam-ordered-list {
+  @include reset;
+
+  display: grid;
```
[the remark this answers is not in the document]
VincentSmedinga: It’s a remnant of another approach I tried but reverted, where the list marker boxes had their own grid cells through a `::before` pseudo-element. Could have replaced `display: grid` with `display: flex; flex-direction: column` again but this is equivalent (and one less line) so I left it like this.

An actually substantive reason is that we don’t need the growing and shrinking behaviour of flex, just the track and gap that grid also provides.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
   }
 
-  @include reset;
-}
-
-.amsterdam-ordered-list__item {
-  padding-inline: var(--amsterdam-ordered-list-item-padding-inline);
+  li {
+    margin-inline-start: var(--amsterdam-ordered-list-item-margin-inline-start);
+    padding-inline-start: var(--amsterdam-ordered-list-item-padding-inline-start);
```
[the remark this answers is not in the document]
VincentSmedinga: Why do you think so? It provides the entire indentation of list items and positions the marker.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,54 @@
...
+/** A nested list has a different marker and less spacing for correct alignment. */
 .amsterdam-ordered-list .amsterdam-ordered-list {
   list-style-type: var(--amsterdam-ordered-list-ordered-list-list-style-type);
-  padding-block: var(--amsterdam-ordered-list-ordered-list-padding-block);
-  padding-inline: var(--amsterdam-ordered-list-ordered-list-padding-inline);
-}
 
-.amsterdam-ordered-list .amsterdam-ordered-list .amsterdam-ordered-list__item {
-  padding-inline: var(--amsterdam-ordered-list-ordered-list-item-padding-inline);
+  li {
```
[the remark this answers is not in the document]
VincentSmedinga: Yes replaced both.

storybook/storybook-react/src/components/UnorderedList/UnorderedList.docs.mdx
VincentSmedinga: I think we should offer controls on the individual story pages only, to keep the docs page easier to read. That’s also why I’m initially hiding the code now.

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
VincentSmedinga: I think we should offer controls on the individual story pages only, to keep the docs page easier to read. That’s also why I’m initially hiding the code now.

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
VincentSmedinga: Yes, very valid points.

I’ve reinstated a minimal `README.md` and added a high priority Story to the backlog to improve the documentation approach.

packages/css/src/unordered-list/unordered-list.scss
```diff
@@ -5,40 +5,54 @@
 
 @import "../../utils/breakpoint";
 
+/**
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, the only code change for both was replacing the `li` selector in CSS.

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
VincentSmedinga: Fair point. I did choose still another two sentences though.

packages/css/src/ordered-list/ordered-list.scss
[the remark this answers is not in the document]
alimpens: You set the value for `padding-inline-start` as `0` here, but that is the default value. Removing this line doesn't seem to change the indentation or marker position. Or am I missing a use case here?

storybook/storybook-react/src/components/OrderedList/OrderedList.docs.mdx
alimpens: Ah, I kinda disagree on that one, but let's discuss that in the seperate Storybook docs ticket

packages/css/src/ordered-list/ordered-list.scss
VincentSmedinga: No, I don’t think so. That would position a nested list exactly between its parent item and the next parent item.

Keeping it a bit closer to its parent further indicates where it belongs.

I’ll add this to a task to feedback the designers.

alimpens: Hm, I'm not a huge fan of how that looks, but let's discuss it with the designers

packages/css/src/ordered-list/ordered-list.scss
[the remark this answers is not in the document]
VincentSmedinga: Ah, just that one line, not both. And only for ordered list, not unordered.

I gave this some more thought and experimentation. One important conclusion is that the browser draws the marker in the margin area.

So we need to:

1. Assign a `margin` to a list item to position its marker horizontally. The marker is drawn at the end of the margin box.
2. Assign a `padding` to a list item with the rest of the desired indentation width. This number is generally higher for glyphs than for decimals.
3. Have less indentation for nested list items to align these markers with their parent text.
4. If so desired, keep the total indentation widths equal between ordered and unordered lists.

I’ve addressed and documented these in the next commit.

Ordered lists now have a small padding as well, to have a bit more space between the item number and its text, and to have equal indentation between ordered and unordered lists. So this line is not to be removed anymore.

proprietary/tokens/src/components/amsterdam/ordered-list.tokens.json
```diff
@@ -14,13 +14,26 @@
...
+          "comment": "The total level 1 indentation for Amsterdam is 40 pixels, or 2.5rem."
+        }
       },
       "ordered-list": {
         "list-style-type": { "value": "lower-alpha" },
         "item": {
-          "margin-inline-start": { "value": "1.5rem" }
+          "margin-inline-start": {
+            "value": "1.5rem",
+            "comment": "Indent less than the parent to start-align the child’s marker with the parent text."
```
alimpens, line 31: I think start-aligning the child's marker to the parent isn't really possible with static values, because we have fluid font-sizes. We could use a fluid value here, but I'm also fine with it not being perfectly aligned.

packages/css/src/ordered-list/ordered-list.scss
```diff
@@ -5,39 +5,55 @@
...
   @media screen and (width > $amsterdam-breakpoint) {
     font-size: var(--amsterdam-ordered-list-wide-font-size);
   }
 
-  @include reset;
-}
-
-.amsterdam-ordered-list__item {
-  padding-inline: var(--amsterdam-ordered-list-item-padding-inline);
+  > .amsterdam-ordered-list__item {
```
alimpens, line 45: Is the child combinator (`>`) used here to prevent this styling from seeping through to nested list list items? Because that doesn't work, a nested list has the exact same structure as a regular list. Both regular list list items and nested list list items get this styling now.

proprietary/tokens/src/components/amsterdam/ordered-list.tokens.json
VincentSmedinga: Wait yes, I thought I covered that as I considered rems to be dynamic, but they aren’t of course. We could do the clamping trick there, but the misalignment on narrow screens does seem very minimal.

And hopefully we can clamp the font-size on the root element, so that 1rem becomes viewpport-dependent.

So I’d propose leaving this as-is.

packages/css/src/ordered-list/ordered-list.scss
VincentSmedinga: Yes it is. And to prevent unordered list items being affected by ordered list item styles and vice versa.

Nested list items get these styles applied through their own list parent, or am I missing something?

alimpens: > to prevent unordered list items being affected by ordered list item styles and vice versa

But that's already taken care of by the BEM classes, right? Ordered list items get `.amsterdam-ordered-list__item`, unordered list items get `.amsterdam-unordered-list__item`. 

> Nested list items get these styles applied through their own list parent

Yes, exactly. I guess my point is: currently `> .amsterdam-ordered-list__item` does the exact same thing as `.amsterdam-ordered-list__item`, so why do we need the child combinator? 

For nested lists, the `li` style is currently first set by lines 46-47, and then overwritten by 56-67.

proprietary/tokens/src/components/amsterdam/ordered-list.tokens.json
alimpens: > And hopefully we can clamp the font-size on the root element, so that 1rem becomes viewpport-dependent.

I think this would overwrite the user preference for the root font-size, no? We should be careful with that. 

Leaving this as-is for now is fine by me!

packages/css/src/ordered-list/ordered-list.scss
VincentSmedinga: Sharp as a knife, you! We don’t even need the parent selector.

alimpens (approved):

packages/css/src/ordered-list/ordered-list.scss
VincentSmedinga: I’ve removed the child combinator.

And your mention of overriding styles made me realise that we didn’t indent unordered list items nested in an ordered list and vice versa. Now we do.

alimpens (approved):

PR #472, opened 2023-08-18
Add Checkbox component

alimpens:
- We should cover these 18 states:

|                   | unchecked | checked | indeterminate |
|-------------------|-----------|---------|---------------|
| default               |     x      |   x      |  x             |
| default:hover     |       x    |     x    |       x        |
| invalid           |    x       |    x     |      x         |
| invalid:hover     |     x      |    x     |       x        |
| disabled          |      x     |    x     |        x       |
| disabled:invalid |     x      |    x     |       x        |

- Should hover states be the same as focus states? We use that pattern for links now -> No, these should be seperate


storybook/storybook-react/src/components/Checkbox/Checkbox.docs.mdx
```diff
@@ -0,0 +1,17 @@
...
+import * as CheckboxStories from "./Checkbox.stories.tsx";
+import README from "../../../../../packages/css/src/checkbox/README.md?raw";
+
+<Meta of={CheckboxStories} />
+
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
```
dlnr, line 11: Like you mentioned in @VincentSmedinga changes in the no marker PR, I think that these controls are in the storybook example for MDX but feel double/unnecessary because the default variant (story) will have the same controls and they distract from the actual docs.

packages/css/src/checkbox/checkbox.scss
```diff
@@ -0,0 +1,145 @@
...
+}
+
+.amsterdam-checkbox__input:focus-visible + .amsterdam-checkbox__label,
+:not(.amsterdam-checkbox__input:disabled) + .amsterdam-checkbox__label:hover {
+  color: var(--amsterdam-checkbox-hover-color);
+  text-decoration: underline;
+  text-underline-offset: 3px;
+}
+
+.amsterdam-checkbox__label__checkmark {
```
dlnr, line 48: The double underscore is never nested in BEM, or at least I've never seen it. Usualy only one level, so something like

```suggestion
.amsterdam-checkbox__label-checkmark {
```

or

```suggestion
.amsterdam-checkbox__checkmark {
```

See: https://cssguidelin.es/#bem-like-naming

packages/css/src/checkbox/checkbox.scss
```diff
@@ -0,0 +1,145 @@
...
+
+.amsterdam-checkbox__label__checkmark::after {
+  box-shadow: var(--amsterdam-checkbox-checkmark-box-shadow);
+  content: "";
+  height: 1.5rem;
+  width: 100%;
+}
+
+// Checkmark checked state
+.amsterdam-checkbox__input:checked + .amsterdam-checkbox__label .amsterdam-checkbox__label__checkmark::after {
```
dlnr, line 68: Parent selector not really needed unless something is conflicting.

```suggestion
.amsterdam-checkbox__input:checked + .amsterdam-checkbox__label__checkmark::after {
```

storybook/storybook-react/src/components/Checkbox/Checkbox.docs.mdx
alimpens: The `Controls` component is used here to add controls to the default variant? But let's discuss this in a seperate PR

packages/react/src/Checkbox/README.md
```diff
@@ -0,0 +1,3 @@
+# React Checkbox component
```
dlnr, line 1: Like the css readme comment, the mdx will suffice.

storybook/storybook-react/src/components/Checkbox/Checkbox.docs.mdx
dlnr: No these controls add knobs to the docs story, the stories for each variant contain the controls. You don't really need controls here

VincentSmedinga (changes requested):

packages/react/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,65 @@
...
+          className="amsterdam-checkbox__input"
+          ref={innerRef}
+          id={id}
+          aria-invalid={invalid || undefined}
+        />
+        <label className="amsterdam-checkbox__label" htmlFor={id}>
+          <span
+            className={clsx(
+              'amsterdam-checkbox__label__checkmark',
+              invalid && 'amsterdam-checkbox__label__checkmark--invalid',
```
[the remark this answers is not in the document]
dlnr: So no native `input:invalid`?

VincentSmedinga:
> * We should cover these 15 states: [checkbox.xlsx](https://github.com/Amsterdam/design-system/files/12431135/checkbox.xlsx)

There are 3 more states: invalid + disabled. Curious, I know, but possible. I wasn't able to update the spreadsheet. I did update my [comment in Figma](https://www.figma.com/file/9IGm6IdPUYizBNGsUnueBd?node-id=4442:25147&mode=dev#536477972).

> * Should hover states be the same as focus states? We use that pattern for links now

No, the focus indicator is a ring around the interactive area (checkbox + label). The focus indicator is visible when the element has keyboard focus, the hover visualisation while the pointer hovers above it. Both can happen simultaneously.

alimpens:
> There are 3 more states: invalid + disabled. Curious, I know, but possible. I wasn't able to update the spreadsheet. I did update my [comment in Figma](https://www.figma.com/file/9IGm6IdPUYizBNGsUnueBd?node-id=4442:25147&mode=dev#536477972).

Should these have seperate styling though? A disabled checkbox is '[not mutable, focusable, or even submitted with the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled)', so the invalid state doesn't 'do' anything when combined with disabled. OTOH, screenreaders can read both states present on the checkbox. We could choose to make them mutually exclusive though. 

> No, the focus indicator is a ring around the interactive area (checkbox + label). The focus indicator is visible when the element has keyboard focus, the hover visualisation while the pointer hovers above it. Both can happen simultaneously.

Good point. I though the pattern of adding the hover state to the focus state was quite common, but I checked in several design systems, and these are usually distinct. Should we remove this pattern from the Link and Button components as well?

VincentSmedinga:
> Should these have seperate styling though? A disabled checkbox is '[not mutable, focusable, or even submitted with the form](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled)', so the invalid state doesn't 'do' anything when combined with disabled.

I think you’re describing the reason not to have a separate visual design for invalid + disabled + hover.

An invalid + disabled checkbox should still have an initial, static visual design to indicate the combination of both states. Indeed, while being not mutable, focusable etc.

> Should we remove this pattern from the Link and Button components as well?

Yes please.

packages/react/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,62 @@
...
+          {...restProps}
+          type="checkbox"
+          className="amsterdam-checkbox__input"
+          ref={innerRef}
+          id={id}
+          aria-invalid={invalid || undefined}
+        />
+        <label className="amsterdam-checkbox__label" htmlFor={id}>
+          <span
+            className={clsx('amsterdam-checkbox__checkmark', invalid && 'amsterdam-checkbox__checkmark--invalid')}
```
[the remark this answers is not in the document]
alimpens: That would be a little odd in this case though, the outer container doesn't 'hold' any of the checkbox states, the `input` does. I think it would be better to replace `amsterdam-checkbox__checkmark--invalid` with `input:invalid` and `input[aria-invalid="true"]` selectors, so all state will be managed in the same place

proprietary/tokens/src/components/amsterdam/checkbox.tokens.json
```diff
@@ -0,0 +1,79 @@
+{
+  "amsterdam": {
+    "checkbox": {
+      "color": { "value": "{amsterdam.color.primary-black}" },
+      "checkmark": {
+        "box-shadow": {
```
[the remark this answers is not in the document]
alimpens: That would be a little confusing I think. `border-width` and `border-color` are existing css properties, we shouldn't use those names if we aren't actually setting those properties. 

> we’re setting the border color here, not a shadow

I disagree, we're setting the `box-shadow` here, it just happens to look like a border. Because 'border' means something in CSS-land, which shouldn't mix up those terms imo.

alimpens:
> Yes please.

[I've added a ticket to our board](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378/backlog?selectedIssue=DES-296)

packages/react/src/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
alimpens: Good call, I forgot that one

VincentSmedinga (changes requested):
I’ve resolved some conversations and replied to others.

VincentSmedinga (approved):

PR #473, opened 2023-08-23
Add inverse color prop to Heading (#449)

alimpens:
* Add Heading inverse color, run prettier

* Use double dashes for modifiers

* Use double dashes in React component as well

* Update story

* Add stories, clean up existing stories

* Add and update tests


storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -11,21 +11,51 @@ const meta = {
   component: Heading,
   args: {
     children: 'Jouw typograaf biedt mij zulke exquise schreven!',
-    level: 1,
-    size: 'level-1',
+    level: undefined,
```
VincentSmedinga, line 14: Why do we need these?

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -11,21 +11,51 @@ const meta = {
...
+    size: undefined,
+    inverseColor: false,
   },
   argTypes: {
+    level: {
```
VincentSmedinga, line 19: If we make `size` a radio again, we should make this prop a radio as well.

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -11,21 +11,51 @@ const meta = {
...
 type Story = StoryObj<typeof meta>
 
-export const Heading1: Story = {}
+export const Default: Story = {
+  decorators: [
+    (Story, context) => (
+      <div
+        style={{
+          background: context.args.inverseColor ? '#004699' : 'transparent',
+          padding: context.args.inverseColor ? '16px' : 0,
```
VincentSmedinga, line 44: This prevents a layout change, and the `px` aren’t necessary:

```suggestion
          margin: context.args.inverseColor ? -16 : 0,
          padding: context.args.inverseColor ? 16 : 0,
```

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
```diff
@@ -11,21 +11,51 @@ const meta = {
...
+        <Story />
+      </div>
+    ),
+  ],
+}
+
+export const Heading1: Story = {
+  args: {
+    level: 1,
+    size: 'level-1',
```
VincentSmedinga, line 56: I think we shouldn’t specify the size if it’s the same as the level.

And maybe even remove level for heading 1?

VincentSmedinga (approved):

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
alimpens: I thought it was a way to keep the code view clean of props, but it doesn't actually seem to be necessary. Not sure why I thought it was 😅 I'll remove it from the main and `InvertedColor` stories

storybook/storybook-react/src/components/Heading/Heading.stories.tsx
alimpens: Hm, I chose this approach so you would have code examples of all levels and all sizes, even though it's not necessary to specify the size when it's the same as the level. But I can see why this would be confusing.

The verbose approach would be to show all possible sizes for every level, but that would be a bit much. 

We could show the four sizes in order with different levels? Or, if we just want a visual example of the four sizes,  we could show the four levels in order?

alimpens: Yeah, what you expect is what I tried to describe with my second option 😄 I'll do that for now.

I guess the approach depends on whether you see the examples as something developers can quickly copy-paste, or just visual examples. But, you're right, that's for a different PR

VincentSmedinga: 👍🏼 Developers can quickly copy-paste after clicking ‘Show code’. That will happen a couple of times per month. Viewing hopefully a hundredfold more.

VincentSmedinga (approved):

PR #475, opened 2023-08-23
Add component scaffolding tool

alimpens:


plop-templates/react.docs.md.hbs
```diff
@@ -0,0 +1,3 @@
+# React {{sentenceCase name}} component
+
+[{{sentenceCase name}} documentation](../../../css/src/{{kebabCase name}}/README.md)
```
dlnr, line 3: We're dropping the readme in favour of the mdx file right?

VincentSmedinga: We’re considering it, or having even multiple separate markdown files.

In #444, Aram noted that the `README.md` file is being rendered on GitHub, which would a strong reason to maintain it.

Not sure yet how we can efficiently have both a README and a Storybook docs file.

I would say leave this as it is for now.

VincentSmedinga (approved):
Great idea to use this and a good start setting it up 👍🏼

PR #476, opened 2023-08-23
Add page header component

alimpens:


packages/css/src/hero/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Gebruik dit component om de aandacht van de gebruiker te trekken.
+- Gebruik dit component maar 1 keer per pagina, vaak alleen op de voorpagina.
+- Niet gebruiken als je lange stukken tekst wilt weergeven, de tekst moet kort en duidelijk zijn.
+
+## Relevante WCAG eisen
+
+Het hero component is een variant van een heading level 1, met een afwijkende stijl. Let bij gebruik van dit component goed op dat de heading hiërarchie van de pagina logisch blijft.
+
+Voor dit component gelden dezelfde WCAG eisen als voor [het heading component](/docs/heading--docs).
```
VincentSmedinga, line 16: This link seems to have changed. Does a more robust approach to link between docs files exist?

```suggestion
Voor dit component gelden dezelfde WCAG eisen als voor [het heading component](/docs/react_heading--docs).
```

packages/css/src/hero/hero.scss
```diff
@@ -0,0 +1,32 @@
...
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../utils/breakpoint";
+
+@mixin reset {
+  box-sizing: border-box;
+  break-after: avoid;
+  break-inside: avoid;
```
VincentSmedinga, line 11: Are these two and `margin-inline` resetting default UA styles?

packages/react/src/Hero/Hero.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * De kleur van de titel
+   * Gebruik deze property om de titel in tegenovergestelde kleur te tonen.
+   */
+  inverseColor?: boolean
+}
+
+export const Hero = forwardRef(
+  (
+    { children, className, inverseColor, ...restProps }: PropsWithChildren<HeroProps>,
```
VincentSmedinga, line 19: We should replace rendering the `children` with a `text` string prop. We only want text in a hero, right? Also, e.g. the `inverseColor` prop may have no effect on child components.

I had already added a story to Jira to do this for `Heading` and `Button`.

VincentSmedinga (changes requested):

packages/react/src/Hero/Hero.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  it('renders a design system BEM class name', () => {
+    const { container } = render(<Hero />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('amsterdam-hero')
+  })
+
+  it('can have a custom class name', () => {
```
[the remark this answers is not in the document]
alimpens: I think maybe this test is a bit redundant. What you're actually testing here is the same as the next test, whether a component can have an additional class. Our original classes are never overwritten. 

I do think we should allow additional classes, otherwise our users can't restyle our components.

packages/react/src/Hero/Hero.tsx
alimpens: I saw the story, I'm not quite sure about that though, for a couple of reasons:

- a bit arbitrary, but as a developer, I prefer writing `<Button>Label</Button>` vs `<Button label="label" />
- if we only allow text, our users aren't able to customize the contents of these containers at all. A button with a loading state for example, which we don't have now, would not be able to be built from our components by a user. And neither would, and that's the most important reason imo, a composition of components we're currently not aware of, but which is necessary in a users project.

packages/css/src/hero/README.md
VincentSmedinga: Ah, yes, I’m generally using the React Storybook only.

packages/css/src/hero/README.md
```diff
@@ -0,0 +1,16 @@
+# Hero
+
+Het hero component is een opvallende tekst aan het begin van een pagina, direct na de header.
+Gebruik dit component om een kernboodschap over te brengen of duidelijk te maken waar de pagina over gaat.
```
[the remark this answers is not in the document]
VincentSmedinga: Great

packages/css/src/hero/hero.scss
VincentSmedinga: I would say the reset mixin groups declarations that override unwanted values in the UA defaults. If we didn't have the mixin, we would be inclined to add a comment `/* UA override */` to each of these lines. Through the mixin, we move the ‘reset’ indicator higher up and group everything.

Our own design choices, for which the UA stylesheet specifies nothing or accepts the default value, should go in the declaration block for the component class name.

packages/react/src/Hero/Hero.tsx
```diff
@@ -0,0 +1,32 @@
...
+      {...restProps}
+      ref={ref}
+      className={clsx('amsterdam-hero', inverseColor && 'amsterdam-hero--inverse-color', className)}
+    >
+      {children}
+    </h1>
+  ),
+)
+
+Hero.displayName = 'Hero'
```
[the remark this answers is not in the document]
VincentSmedinga: Aha, thanks.

packages/react/src/Hero/Hero.tsx
VincentSmedinga: I understand your second point. But I don’t think we can carry the responsibility of preparing our component with many unknown compositions.

E.g. if people need a spinner, we should add it to the component as a feature, and make sure it works well with all existing features.

Let’s not change it now, add it to that story, discuss more later.

VincentSmedinga (approved):

VincentSmedinga (approved):

PR #481, opened 2023-08-24
Generate bundled css

alimpens:


packages/css/src/index.scss
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+/* Append here */
+@import "./accordion/accordion";
```
VincentSmedinga, line 7: Is this a good time to sort imports in barrel files alphabetically?

VincentSmedinga (approved):
Please explain why this is necessary again. Was this a bug?

alimpens:
> Please explain why this is necessary again. Was this a bug?

No, not really, just an extra feature. With this, you can use unstyled components (or your own components) and style them all in one go. 

It's also in preparation of something I want to discuss. We export styled components now, which render all styles in js strings and push them to the document `head` on load. Loading one component from the styled index file means loading all styles from all components in the DS. Which is pretty much a less efficient way of doing what a single stylesheet does as well.

So my proposition is to only export unstyled components, and use this css file to style them. [See this discussion for more information](https://datalab-amsterdam.slack.com/archives/C01DAT4TRPF/p1691568231239299)

packages/css/src/index.scss
VincentSmedinga: Hmm that’s a downside. For later.

VincentSmedinga (approved):

PR #483, opened 2023-08-24
Add toptask link

alimpens:
To discuss:

- currently, when you select the toptask with a screenreader, it reads 'link title, link description', implying there are 2 links when there's only one. I think we should change the structure to match https://inclusive-components.design/cards/ -> we've changed it to two `span`s, which makes a screenreader read it as one. Also, we've added an `aria-label` with `${label}, ${description}`
- should you be able to add an icon in the mvp? -> not mvp
- do we want to cut off the title and description at 2 lines in the component? [Like so](https://gitlab.com/os-amsterdam/website-onderzoek-en-statistiek/-/blob/develop/website/components/Card/Card.style.js?ref_type=heads#L100-105) -> Vincent will check with Inge

### DIFF
--- a/packages/css/src/top-task-link/README.md
+++ b/packages/css/src/top-task-link/README.md
@@ -1,0 +1,1 @@
+# Top task link
--- a/packages/css/src/top-task-link/README.md
+++ b/packages/css/src/top-task-link/README.md
@@ -1,1 +1,15 @@
 # Top task link
+
+De toptask link is een navigatie-element dat opvallender is dan een link, om de aandacht te trekken voor de meest belangrijke informatie.
+Zet de toptask link in zodat de gebruiker snel de juiste informatie vindt én een bijbehorende taak makkelijk kan uitvoeren.
+
+## Richtlijnen
+
+<!-- TODO: moet er een max van 2 regels in het component worden ingebouwd voor title en description? -->
+
+- Een toptask link moet een titel hebben. Hou deze titel zo kort mogelijk, hij mag maximaal over 2 regels lopen.
+- Een toptask link kan een omschrijving hebben, maar dat is niet verplicht. De omschrijving staat onder de titel en is ook maximaal 2 regels.
+
+## Relevante WCAG eisen
+
+Voor de toptask link gelden dezelfde eisen als voor [reguliere links](/docs/link--docs).
--- a/packages/css/src/top-task-link/README.md
+++ b/packages/css/src/top-task-link/README.md
@@ -5,8 +5,6 @@ Zet de toptask link in zodat de gebruiker snel de juiste informatie vindt én ee
 
 ## Richtlijnen
 
-<!-- TODO: moet er een max van 2 regels in het component worden ingebouwd voor title en description? -->
-
 - Een toptask link moet een titel hebben. Hou deze titel zo kort mogelijk, hij mag maximaal over 2 regels lopen.
 - Een toptask link kan een omschrijving hebben, maar dat is niet verplicht. De omschrijving staat onder de titel en is ook maximaal 2 regels.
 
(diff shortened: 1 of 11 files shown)


packages/css/src/top-task-link/README.md
```diff
@@ -0,0 +1,13 @@
+# Top task link
```
VincentSmedinga, line 1: Good call on the rename.

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
+
+### Three columns
+
+Drie toptask links naast elkaar in een rij.
```
VincentSmedinga, line 15: For me they’re below each other. The `.amsterdam-page-grid` class name doesn’t do anything?

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+<Canvas of={TopTaskLinkStories.ThreeColumns} />
+
+### Four columns
```
VincentSmedinga, line 19: One of these two examples may suffice.

VincentSmedinga (approved):

VincentSmedinga:
> currently, when you select the toptask with a screenreader, it reads 'link title, link description', implying there are 2 links when there's only one. I think we should change the structure to match https://inclusive-components.design/cards/

It is one link so indeed it should be read out as one: ‘(title), (description), link’. I would say the combination of both texts should be still short enough to be read out as a whole. The title alone will often bring too little context. Users of screen readers often set a high reading speed which makes it it time-efficient.

> should you be able to add an icon in the mvp?

Not in this PR, I created a ticket for it.

> do we want to cut off the title and description at 2 lines in the component? [Like so](https://gitlab.com/os-amsterdam/website-onderzoek-en-statistiek/-/blob/develop/website/components/Card/Card.style.js?ref_type=heads#L100-105)

Personally I’m not at all in favour of automatic truncation. Am waiting for confirmation by Inge.

packages/css/src/top-task-link/README.md
alimpens: Yeah, makes more sense imo

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
alimpens: Hm, it works for me. Maybe the css wasn't loaded when switching branches? Hard refresh?

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
alimpens: In the Figma file an example is given for a 4 column and 3 column top task link, implying those are the only 2 sizes they can have. I'll check that assumption with design. If that's true, I'll add that to the docs as well

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
VincentSmedinga: Had tried that yesterday, to no avail, probably local then.

storybook/storybook-react/src/components/TopTaskLink/TopTaskLink.docs.mdx
VincentSmedinga: Would surprise me. In the app we use a list of them – as we’ll do on a website visited on a phone – and also separate instances.

VincentSmedinga (approved):

PR #484, opened 2023-08-25
Increase whitespace between list items

VincentSmedinga:


alimpens:
Why are we doing this again? This is currently not in the designs, right?

alimpens:
> Why are we doing this again? This is currently not in the designs, right?

It's a visual improvement, will be added to the designs later

alimpens (approved):

PR #488, opened 2023-08-30
Remove padding from links and some other improvements

alimpens:
A couple of changes in this PR:

- I've remove the padding from standalone and in list links, checkboxes and breadcrumbs. [The new WCAG states that hit areas only have to be 24x24px](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html), so the padding isn't necessary.
- I've fixed a typo in a token reference
- I've changed the display for standalone link to `inline-block`, so the hit area is the same as the line-height.

### DIFF
--- a/packages/css/src/link/link.scss
+++ b/packages/css/src/link/link.scss
@@ -19,9 +19,9 @@
 }
 
 .amsterdam-link--standalone {
+  display: inline-block;
   font-size: var(--amsterdam-link-standalone-narrow-font-size);
-  line-height: var(--amsterdam-link-standalon-line-height);
-  padding-block: 0.5rem;
+  line-height: var(--amsterdam-link-standalone-line-height);
   text-decoration-thickness: var(--amsterdam-link-standalone-text-decoration-thickness);
   text-underline-offset: var(--amsterdam-link-standalone-text-underline-offset);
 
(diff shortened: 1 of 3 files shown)


packages/css/src/link/link.scss
```diff
@@ -19,9 +19,9 @@
 }
 
 .amsterdam-link--standalone {
+  display: inline-block;
   font-size: var(--amsterdam-link-standalone-narrow-font-size);
-  line-height: var(--amsterdam-link-standalon-line-height);
-  padding-block: 0.5rem;
+  line-height: var(--amsterdam-link-standalone-line-height);
```
VincentSmedinga, line 24: Can we configure [this Stylelint rule](https://stylelint.io/user-guide/rules/no-unknown-custom-properties/) to prevent this in the future?

VincentSmedinga (dismissed):

packages/css/src/link/link.scss
alimpens: I tried it out, it only seems to work for local css vars. Our component css doesn't have a direct import of the design token css vars, so it can't 'know' whether a css var exists.

VincentSmedinga (approved):

PR #491, opened 2023-09-01
Add PageMenu Component

dlnr:
In order te re-use the horizontal menu in both the Header and Footer paterns we could use something like this. I've kept the amount of props like text labels or icons to a minimum.

### DIFF
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -9,7 +9,6 @@
   box-sizing: border-box;
   margin-block: 0;
   padding-inline: 0;
-  -webkit-text-size-adjust: 100%;
 }
 
 @mixin reset-item {
@@ -24,6 +23,7 @@
   flex-direction: row;
   flex-wrap: wrap;
   list-style: none;
+  row-gap: var(--amsterdam-page-menu-list-row-gap);
 
   @include reset-list;
 }
@@ -62,8 +62,8 @@
   @include reset-item;
 }
 
-.amsterdam-page-menu__link a:hover,
-.amsterdam-page-menu__button button:hover {
+.amsterdam-page-menu__link:hover,
+.amsterdam-page-menu__button:hover {
   color: var(--amsterdam-page-menu-item-hover-color);
 }
 
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -16,6 +16,13 @@
   -webkit-text-size-adjust: 100%;
 }
 
+@mixin reset-button {
+  background-color: transparent;
+  border: 0;
+  margin-block: 0;
+  padding-inline: 0;
+}
+
 .amsterdam-page-menu {
   align-items: center;
   column-gap: var(--amsterdam-page-menu-column-gap);
@@ -52,14 +59,11 @@
 }
 
 .amsterdam-page-menu__button {
-  background-color: transparent;
-  border: 0;
   cursor: pointer;
-  margin-block: 0;
-  padding-inline: 0;
 
-  @include page-menu-item;
   @include reset-item;
+  @include reset-button;
+  @include page-menu-item;
 }
 
 .amsterdam-page-menu__link:hover,
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -16,14 +16,14 @@
   -webkit-text-size-adjust: 100%;
 }
 
-.amsterdam-page-menu__list {
+.amsterdam-page-menu {
   align-items: center;
-  column-gap: var(--amsterdam-page-menu-list-column-gap);
+  column-gap: var(--amsterdam-page-menu-column-gap);
   display: flex;
   flex-direction: row;
   flex-wrap: wrap;
   list-style: none;
-  row-gap: var(--amsterdam-page-menu-list-row-gap);
+  row-gap: var(--amsterdam-page-menu-row-gap);
 
   @include reset-list;
 }
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -1,0 +1,8 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-page-menu {
+  /* Add styles here */
+}
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -3,6 +3,56 @@
  * Copyright (c) 2023 Gemeente Amsterdam
  */
 
-.amsterdam-page-menu {
-  /* Add styles here */
+@import "../../utils/breakpoint";
+
+@mixin reset {
+  border: 0;
+  box-sizing: border-box;
+  margin-block: 0;
+  padding-inline: 0;
+  -webkit-text-size-adjust: 100%;
+}
+
+.amsterdam-page-menu__list {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: var(--amsterdam-page-menu-gap);
+  list-style: none;
+
+  @include reset;
+}
+
+.amsterdam-page-menu__link a,
+.amsterdam-page-menu__button button {
+  align-items: center;
+  background: none;
+  color: var(--amsterdam-page-menu-item-color);
+  cursor: pointer;
+  display: inline-flex;
+  flex-direction: row;
+  font-family: var(--amsterdam-page-menu-font-family);
+  font-size: var(--amsterdam-page-menu-item-narrow-font-size);
+  gap: var(--amsterdam-page-menu-item-gap);
+  line-height: var(--amsterdam-page-menu-item-line-height);
+  text-align: center;
+  text-decoration: var(--amsterdam-page-menu-item-text-decoration);
+  touch-action: manipulation;
+
+  @media screen and (width > $amsterdam-breakpoint) {
+    font-size: var(--amsterdam-page-menu-item-wide-font-size);
+  }
+
+  @include reset;
+}
+
+.amsterdam-page-menu__link a:hover,
+.amsterdam-page-menu__button button:hover {
+  color: var(--amsterdam-page-menu-item-hover-color);
+}
+
+.amsterdam-page-menu__link svg,
+.amsterdam-page-menu__button svg {
+  color: currentColor;
 }
--- a/packages/css/src/page-menu/page-menu.scss
+++ b/packages/css/src/page-menu/page-menu.scss
@@ -5,35 +5,36 @@
 
 @import "../../utils/breakpoint";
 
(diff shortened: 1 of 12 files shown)


packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,58 @@
...
+  gap: var(--amsterdam-page-menu-gap);
+  list-style: none;
+
+  @include reset;
+}
+
+.amsterdam-page-menu__link a,
+.amsterdam-page-menu__button button {
+  align-items: center;
+  background: none;
```
alimpens, line 30: I think you want to set `background-color` here? That's a bit more specific

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+  ButtonHTMLAttributes,
+  ForwardedRef,
+  forwardRef,
+  HTMLAttributes,
+  PropsWithChildren,
+} from 'react'
+import { Icon } from '../Icon'
+
+interface PageMenuComponent
+  extends React.ForwardRefExoticComponent<
```
alimpens, line 17: You can import these types directly, like so:

```suggestion
  extends ForwardRefExoticComponent<
```

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+    PropsWithChildren<HTMLAttributes<HTMLElement>> & React.RefAttributes<HTMLElement>
```
alimpens, line 18:
```suggestion
    PropsWithChildren<HTMLAttributes<HTMLElement>> & RefAttributes<HTMLElement>
```

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+  > {
+  Link: typeof PageMenuLink
+  Button: typeof PageMenuButton
+}
+
+export const PageMenu = forwardRef(
+  ({ children, ...restProps }: PropsWithChildren<HTMLAttributes<HTMLElement>>, ref: ForwardedRef<HTMLElement>) => {
```
alimpens, line 25: We've already defined this type earlier, can we extract it into a variable?

packages/react/src/PageMenu/PageMenu.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+import '@testing-library/jest-dom'
+import { Icon } from '../Icon'
+
+describe('page menu', () => {
+  it('renders a page menu with children', () => {
+    const { container } = render(
+      <PageMenu>
+        <PageMenu.Link href="#">English</PageMenu.Link>
+        <PageMenu.Link href="#">
+          Inloggen Mijn Amsterdam <Icon svg={Login} size="level-7" />
```
alimpens, line 14: The icon interface changed, right? It should be `<PageMenu.Link href="#" icon={Login}>`

storybook/storybook-react/src/components/PageMenu/PageMenu.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { PageMenu } from '@amsterdam/design-system-react'
+import { Login, Menu } from '@amsterdam/design-system-react-icons'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Page menu',
+  component: PageMenu,
+  args: {
+    children: 'Nieuw component',
+  },
```
alimpens, line 15: This isn't necessary, right?

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,58 @@
...
+  padding-inline: 0;
+  -webkit-text-size-adjust: 100%;
+}
+
+.amsterdam-page-menu__list {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: var(--amsterdam-page-menu-gap);
```
alimpens, line 21: I think we should set different gaps for columns and rows. The list wraps, so the row gap is also 2.5rem now, which is a bit much.

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,58 @@
...
+  list-style: none;
+
+  @include reset;
+}
+
+.amsterdam-page-menu__link a,
+.amsterdam-page-menu__button button {
```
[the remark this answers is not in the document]
dlnr: Sure, I've opted for this solution to create the least amount of css line but I could combine stuff in an include?

packages/css/src/page-menu/page-menu.scss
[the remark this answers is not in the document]
dlnr: This would introduce an extra node in the tokens, i'll give it a go.

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,73 @@
...
+  cursor: pointer;
+  margin-block: 0;
+  padding-inline: 0;
+
+  @include page-menu-item;
+  @include reset-item;
+}
+
+.amsterdam-page-menu__link a:hover,
+.amsterdam-page-menu__button button:hover {
```
alimpens, line 66: This isn't correct anymore, right?

```suggestion
.amsterdam-page-menu__link:hover,
.amsterdam-page-menu__button:hover {
```

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+@mixin reset-item {
+  box-sizing: border-box;
+  -webkit-text-size-adjust: 100%;
+}
+
+.amsterdam-page-menu__list {
+  align-items: center;
+  column-gap: var(--amsterdam-page-menu-list-column-gap);
```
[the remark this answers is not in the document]
dlnr: Sure but should the component be able to wrap at all?

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface PageMenuComponent extends ForwardRefExoticComponent<PageMenuProps & RefAttributes<HTMLElement>> {
+  Link: typeof PageMenuLink
+  Button: typeof PageMenuButton
+}
+
+export const PageMenu = forwardRef(
+  ({ children, className, ...restProps }: PageMenuProps, ref: ForwardedRef<HTMLElement>) => {
+    return (
+      <nav {...restProps} className={clsx('amsterdam-page-menu', className)} ref={ref}>
```
alimpens, line 29: For the footer the list shouldn't be wrapped in a `nav`, I think. Maybe PageMenu should just be `ul` -> `li` -> `link` / `button`, and the `nav` should be wrapped around it when composing the Header?

dlnr: Good point, I'll take it out.

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,73 @@
...
+  @include page-menu-item;
+  @include reset-item;
+}
+
+.amsterdam-page-menu__button {
+  background-color: transparent;
+  border: 0;
+  cursor: pointer;
+  margin-block: 0;
+  padding-inline: 0;
```
VincentSmedinga, line 59: All of these, except for the `pointer`, could be in a `reset-button` mixin?

packages/react/src/PageMenu/PageMenu.tsx
VincentSmedinga: I agree. See this quote from the spec: “In particular, it is common for footers to have a list of links to various key parts of a site, but the footer element is more appropriate in such cases, and no nav element is necessary for those links.” ([source](http://html5doctor.com/nav-element/))

Please add documentation that implementers must wrap a Page Menu in the page header in a `nav` element, and they must not do so in the page footer.

VincentSmedinga (changes requested):

packages/css/src/page-menu/page-menu.scss
```diff
@@ -0,0 +1,73 @@
...
+
+  @include page-menu-item;
+  @include reset-item;
+}
+
+.amsterdam-page-menu__link:hover,
+.amsterdam-page-menu__button:hover {
+  color: var(--amsterdam-page-menu-item-hover-color);
```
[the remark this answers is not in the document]
dlnr: Onderzoek site has the blue color, also me and @alimpens were not happy with the bold hover in the design. I've commented the Figma file and made a PR anyway because we need the menu and can change it later.

packages/css/src/page-menu/page-menu.scss
dlnr: Sure, but a mixin is for re-usability and it wouldn't be needed elsewhere.

dlnr: Also I personally feel that the [css was cleaner](https://github.com/Amsterdam/design-system/blob/32ccadd4c9d29844b09eaf0f64f3a4169be08557/packages/css/src/page-menu/page-menu.scss) before trying to resolve feedback, the rules applied to both the link and button were not a cause for concern.

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface PageMenuButtonProps extends PropsWithChildren<ButtonHTMLAttributes<HTMLButtonElement>> {
+  icon?: Function
+}
+
+const PageMenuLink = forwardRef(
+  ({ children, icon, ...restProps }: PageMenuLinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
+    return (
+      <li className="amsterdam-page-menu__item">
```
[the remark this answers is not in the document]
VincentSmedinga: Alright, ‘menu item’ is indeed very common.

packages/react/src/PageMenu/README.md
```diff
@@ -0,0 +1,3 @@
+# React Page menu component
```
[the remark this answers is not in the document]
alimpens: This is a change in the scaffolding, which means that if we change it, we should change it for all components and in the template.

Maybe we should do that in a seperate PR?

packages/css/src/page-menu/page-menu.scss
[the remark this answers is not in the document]
VincentSmedinga: Alright, although I would still have been a bit more logical to derive an alternative design from the rest of the system than from another website.

packages/css/src/page-menu/page-menu.scss
VincentSmedinga: We currently reuse none of our mixins.

We use reset mixins to separate reset styles from design choices, to prevent having multiple `/* UA override */` comments in a declaration block. See them as the original ‘reset.css’ but then a single one per component, as we avoid tight coupling of all components with a shared stylesheet.

VincentSmedinga: I’ve applied this.

VincentSmedinga (approved):

PR #492, opened 2023-09-01
Add initial Jira guidelines

VincentSmedinga:


documentation/jira-guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+  - Check status huidige component: Fabrique / onze Figma, onze Design System website.
+  - Kijk wat er beschikbaar is vanuit NL Design System.
+  - Zoek vergelijkbare ontwerpen in bijv. The Component Gallery
+  - Zet op een rij wat het component allemaal kan en hoe het eruit ziet.
+  - Ga na wat er aan a11y nodig is, bijv. Amsterdam TamTam, Fronteers Slack
+  - Betrek het perspectief van een front-ender en wellicht andere rollen (huisstijl, redactie?).
+  - Bepaal de naamgeving.
+- Design (eventueel meerdere Subtaken)
+  - Werk alle varianten van het component uit in Figma en zet ze overzichtelijk op een rij.
+  - Let op hover, focus states.
```
alimpens, line 55: Focus states aren't currently part of the design, right?

documentation/jira-guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+### Subtasks
+
+- Research
+  - Gebruiken we het NLDS community component?
+  - Vind inspiratie online, bijv. The Component Gallery.
+- Development (eventueel meerdere Subtaken)
+  - Design tokens
+  - Class names
+  - React component
+  - Accessibility
```
alimpens, line 114: Might be nitpicking, but I'd like to put a11y first in this list. A11y and semantics should define the structure of the component, then you should setup the component in React. Only after the component is functional, you should start styling it. So maybe something like:

- Accessibility
- React component
- Class names
- Design tokens
- Unit tests
- Documentation

VincentSmedinga: Well by now you know I love nitpicking :)

I think setting these items op will be iterative anyway, but I can’t be against putting accessibility on top. Applied your suggestion.

documentation/jira-guidelines.md
VincentSmedinga: Hmm… alright.

PR #495, opened 2023-09-11
Add checkbox docs

alimpens:


packages/css/src/checkbox/README.md
```diff
@@ -1 +1,27 @@
...
+- Een selectievakje kan een antwoord op een gesloten vraag zijn, als er meerdere antwoorden mogelijk zijn.
+- Een lijst van selectievakjes is verticaal gestapeld om leesbaarheid en bruikbaarheid te bevorderen.
+- Een lijst van selectievakjes moet in een `fieldset` zitten, met een `legend` die beschrijft waar de lijst over gaat. Als de selectievakjes bijvoorbeeld worden gebruikt om antwoord op een vraag te krijgen, dan is de `legend` de vraag.
+
+## Checkbox labels
+
+Een label:
+
+- ... begint met een hoofdletter
+- ... heeft geen interpunctie aan het einde als het een enkele zin, woord of fragment is.
```
dlnr, line 20: `text-wrap: balance` could be a nice addition to the longer labels, it make the block of text more of a whole (preventing orphaned words on a new line) and thus more tightly connected to the checkbox.

alimpens: That does look better. There is a downside to using it for interactive elements though. The hit area stays the same size, so you get a nice block of text with (imo unexpected) clickable space next to it. Let's discuss this tomorrow as well.

VincentSmedinga: Nice feature to consider, but out of scope for this documentation PR.

VincentSmedinga (approved):

PR #498, opened 2023-09-11
Underline page menu items on hover

VincentSmedinga:


packages/css/src/page-menu/page-menu.scss
```diff
@@ -69,6 +69,7 @@
 .amsterdam-page-menu__link:hover,
 .amsterdam-page-menu__button:hover {
   color: var(--amsterdam-page-menu-item-hover-color);
```
alimpens, line 71: I don't think this is necessary anymore, hover doesn't change the color by default

alimpens:
The underline looks a bit out of place imo, shouldn't this be the same underline as an inline link? So `text-underline-offset: 3px;` and `text-decoration-thickness: 2px;`?

packages/css/src/page-menu/page-menu.scss
VincentSmedinga: I thought we could leave it in for other organisations wanting to use our components.

alimpens:
> > The underline looks a bit out of place imo, shouldn't this be the same underline as an inline link? So `text-underline-offset: 3px;` and `text-decoration-thickness: 2px;`?
> 
> Inline link doesn’t have an underline offset, standalone link does. I agree that the page menu links may be more like standalone ones. But it’s so not pretty…

No, inline link does as well, 3px. Maybe we should wait on design before merging this PR? Doesn't feel very efficient to replace one visual inconsistency with the other imo 😅

alimpens:
> > No, inline link does as well, 3px. Maybe we should wait on design before merging this PR? Doesn't feel very efficient to replace one visual inconsistency with the other imo 😅
> 
> I’m already consulting Inge about this.
> 
> And what’s with this weird quote-replying, let’s try commenting on lines of code where possible.

Yeah, sorry, Githubs general comment system isn't great. I originally posted it as a comment because it's more of a general comment, not really related to a specific line of code

packages/css/src/page-menu/page-menu.scss
```diff
@@ -69,6 +69,7 @@
 .amsterdam-page-menu__link:hover,
```
dlnr, line 69: I realise after reading the interactive elements that I did not include a focus-visible state with the token --amsterdam-focus-outline-offset.

dlnr (commented):
Seeing this discussion and the pending Link adjustments I wonder if it would be better to have common link tokens (like focus and color) to prevent a flurry of changes that we have to do (and not forget) everytime we change one link to have an underline or offset. To prevent some inconstencies.

dlnr:
> To prevent some inconstencies.

Like forgetting the focus outline offset.

packages/css/src/page-menu/page-menu.scss
```diff
@@ -30,6 +30,7 @@
   flex-direction: row;
   flex-wrap: wrap;
   list-style: none;
+  outline-offset: var(--amsterdam-focus-outline-offset);
```
alimpens, line 33: This shouldn't be on `amsterdam-page-menu`, right? That's a `ul`. It should be on `amsterdam-page-menu__link` and `.amsterdam-page-menu__button` I think

alimpens (approved):

PR #499, opened 2023-09-11
Restructure table of contents in Storybook

VincentSmedinga:
Now that we have more and more components, it makes sense to group them together.


VincentSmedinga:
No this is the wrong branch.

PR #500, opened 2023-09-11
Restructure table of contents in Storybook

VincentSmedinga:
Now that we have more and more components, it makes sense to group them together. We can improve the visual design of e.g. the group labels later.

I applied the grouping in every package and although I have tested, please verify that I didn’t forget or break anything. Tests are succeeding.

Is this the right time to add one or more aliases to clean up our imports?


VincentSmedinga:
I am strongly in favour of grouping components in the table of contents of Storybook. Not far from now, we will have dozens, and listing Accordion, Breadcrumb, Button, Card, Checkbox, Details in alphabetical order will just not. be very helpful. Only if you know what you’re looking for _and_ which name we chose for it will you find it directly.

I admit that this requires two steps to find a component: first, the group it’s in, then the component itself. But this is a good thing – same reason why the Avocados, Bread, Champagne,  Dropjes and Eggs aren't adjacent in the supermarkets either.

Let’s keep the groups open, showing all components at once. That helps scan the groups and lists, and scrolling is not that expensive. Also if you’re not sure what to use (a radio button, a checkbox, or a switch?) it makes sense to encounter them all in the same group. Even though in this specific example, the group name ‘forms’ isn’t entirely accurate yet.

People have other options to find what they’re looking for: there’s a search field, and I want to link each mention of a component in prose as well to help navigating through the content.

One thing I kept for later is to have the `Footer` in a ‘Patterns’ group. Patterns is on the same level as Components, so including that now would mean introducing two new levels at once and that may be a bit much. I’ve tucked it into the ‘Components’ group for now.

I think we should apply the same grouping in code. We’ll get used to the groups soon enough and we can use shortcuts (Cmd+P in VS Code I believe, and Shift+Shift in WebStorm) to quickly open a a file by fuzzy-searching the name of files, functions, and types etc.

dlnr:
This seems like a good way to go but I would like to argue that this produces very long and confusing import/export paths (like `@import "./buttons/button/button";`) while all that I think is that you wanted to achieve is hierarchy in Storybook. Which you can easily do in the meta title of the stories. ie. `title: 'Buttons/Button'`.

This also breaks with common practise in the design systems I have seen thisfar. Component folders are usually pretty 1 dimensional and for what it's worth [Den Haag](https://nl-design-system.github.io/denhaag) does a pretty good job of keeping storybook clean and easy to read. And yes I'm still a fan of having a CSS section in storybook, a design system is more than just react components.

alimpens:
I agree with Niels that grouping the directories causes unnecessary complexity, and I think it makes things harder to find (which I guess is the opposite of what you're trying to achieve?)

I think even grouping in Storybook can have an adverse effect on findability. Try finding a Tabs component without using the search box [here ](https://style.monday.com/?path=/story/welcome--page) vs [here](https://design.gs.com/components). While the Vibe DS Storybook arguably looks nicer, the UX is worse imo.

dlnr:
The second example is not a storybook and has no search. 

I do think some hierarchy in storybook will benefit the user, and it doesn't affect searching. In fact if we mention the Header for example in a pattern or layout context and the component itself in different pages I think it might help a lot when you search for Header.

VincentSmedinga:
We discussed offline and decided to keep the grouping in Storybook, not in code.

The Storybook package does need the directories for the sorting both of and within groups to work.

PR #511, opened 2023-09-15
Add initial Switch component

hcvdhaar:
- Some files has changed after running `prettier`.
   This is just formatting.

TODO: Need to discuss.
- It seems the Hero component was is not on `main`?
   Might need to remove that.


packages/css/src/switch/switch.scss
```diff
@@ -0,0 +1,68 @@
...
+  appearance: none;
+  margin-block: 0;
+  margin-inline: 0;
+  opacity: 0%; /* to hide native input field in older iOS */
+}
+
+.amsterdam-switch__toggle {
+  background-color: var(--amsterdam-switch-background-color);
+  border-radius: 50px;
+  box-sizing: border-box;
```
alimpens, line 21: This is sort of a reset, right? The convention is to wrap resets in a `reset` mixin, so we know what its function is

packages/css/src/switch/switch.scss
```diff
@@ -0,0 +1,68 @@
...
+  display: inline-block;
+  padding-block: 1px;
+  padding-inline: 1px;
+  transition: background-color 0.2s ease-in-out;
+  width: var(--amsterdam-switch-toggle-width);
```
alimpens, line 26: The clickable area should get `cursor: pointer` I think

packages/css/src/switch/switch.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.amsterdam-switch__toggle-inner {
+  background-color: var(--amsterdam-switch-toggle-background-color);
+  border-radius: 50%;
+  display: block;
+  height: var(--amsterdam-switch-toggle-inner-height);
```
alimpens, line 33: Do we want to hook this up to our text levels as well? So the switch will be as tall as the line height of a text level?

packages/css/src/switch/switch.scss
```diff
@@ -0,0 +1,68 @@
...
+  transform: translate(0);
```
alimpens, line 34: Is this necessary?

packages/react/src/Switch/README.md
```diff
@@ -0,0 +1,5 @@
+# React Switch Component
+
+The React `<Switch />` component is a wrapper around the `<input type="checkbox" />`. It can be used as a standalone component or as a form control element. When used as a standalone component provide a `aria-labeledby` prop to keep it accessible. When used as a formcontrol, make sure to add the associated `<Label>` component.
```
alimpens, line 3:
```suggestion
The React `<Switch />` component is a wrapper around the `<input type="checkbox" />`. It can be used as a standalone component or as a form control element. When used as a standalone component provide an `aria-label` or `aria-labelledby` prop to keep it accessible. When used as a formcontrol, make sure to add the associated `<FormLabel>` component.
```

packages/react/src/Switch/README.md
```diff
@@ -0,0 +1,5 @@
...
+
```
alimpens, line 4: I think the docs on how developers should use this component should be in the story, no?

packages/react/src/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    expect(switchElement).toBeChecked()
+  })
+
+  it('renders a design system BEM class name', () => {
+    const { container } = render(<Switch />)
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('amsterdam-switch')
+  })
+
```
alimpens, line 38: We can add a couple more test cases here:

- it can have a additional class name
- it is not disabled by default
- it can have a disabled state
- it can trigger a change event
- it does not trigger a change event when disabled
- it triggers a change event by clicking on label

You can mostly copy-paste these from the Checkbox component I think

packages/react/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <input
+          {...restProps}
+          className="amsterdam-switch__input"
+          id={inputId}
+          ref={ref}
+          type="checkbox"
+          role="switch"
+        />
+
+        <label className="amsterdam-switch__label" htmlFor={inputId} aria-hidden="true">
```
alimpens, line 21: Is `aria-hidden` necessary here? My screen reader doesn't announce anything for empty labels.

proprietary/tokens/src/components/amsterdam/switch.tokens.json
```diff
@@ -0,0 +1,46 @@
...
+      },
+      "background-color": {
+        "value": "{amsterdam.color.neutral-grey3}"
+      },
+      "toggle": {
+        "background-color": {
+          "value": "{amsterdam.color.primary-white}"
+        },
+        "width": {
+          "value": "58px"
```
alimpens, line 15: This should be in rems I think, and probably related to the circle size. If we use a text level line-height for the circle size, this should be twice that, I think?

proprietary/tokens/src/components/amsterdam/top-task-link.tokens.json
```diff
@@ -2,33 +2,61 @@
   "amsterdam": {
     "top-task-link": {
       "description": {
-        "color": { "value": "{amsterdam.color.primary-black}" },
-        "font-family": { "value": "{amsterdam.typography.font-family}" },
-        "font-weight": { "value": "{amsterdam.typography.font-weight.normal}" },
-        "line-height": { "value": "{amsterdam.typography.text-level.7.line-height}" },
+        "color": {
```
alimpens, line 5: Do you have local prettier settings that differ from [our prettier config](https://github.com/Amsterdam/design-system/blob/main/.prettierrc.json)? I don't think these line breaks are necessary?

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,24 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { Switch } from '@amsterdam/design-system-react/src'
+import { Meta, StoryObj } from '@storybook/react'
+
+import '@amsterdam/design-system-css/src/switch/switch.scss'
```
alimpens, line 9: This isn't necessary anymore

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  component: Switch,
+} satisfies Meta<typeof Switch>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    disabled: false,
```
alimpens, line 22: This is default, no?

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  },
+}
```
alimpens, line 24: Let's show the 2 cases described in `packages/react/src/Switch/README.md` in Storybook, so the default can be a Switch with `aria-label` and another Story can be the Switch with a FormLabel. That way, the examples in storybook show the best practice

package.json
```diff
@@ -58,7 +58,7 @@
     "lint:js": "eslint . --ext .js,.json,.jsx,.mdx,.ts,.tsx --report-unused-disable-directives",
     "lint:md": "markdownlint **/*.md --ignore node_modules",
     "lint:package-json": "npmPkgJsonLint **/package.json",
-    "lint:package-lock": "pnpm ls --recursive",
+    "lint:package-lock": "pnpm xls --recursive",
```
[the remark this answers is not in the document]
hcvdhaar: I guess this from the merge with develop I guess. Will revert it to ` "lint:package-lock": "pnpm ls --recursive",`

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
hcvdhaar: removed it

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
hcvdhaar: Indeed, but is for show the controls

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
hcvdhaar: I guess you mean something like this?
```
   <label htmlFor="switch-label">
        Switch
        <Switch {...args} id="switch-label" />
      </label>
```

how would you style the `label`? I would assume to do this with css modules. But this give me a typescript error for unknown module. Styling is applied, but we need to update the tsconfig. Are there any other options?

packages/react/src/Switch/README.md
alimpens: What I meant is: how developers should use this component should be part of the story, so we can add it there and remove this part of the readme:

> It can be used as a standalone component or as a form control element. When used as a standalone component provide an `aria-label` or `aria-labelledby` prop to keep it accessible. When used as a formcontrol, make sure to add the associated `<FormLabel>` component.

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  argTypes: {
+    disabled: {
+      control: 'radio',
+      options: [true, false],
+      defaultValue: false,
+    },
```
alimpens, line 24:
```suggestion
    disabled: { control: 'boolean' },
```

packages/react/src/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+describe('Switch', () => {
+  it('renders an element with role switch', () => {
+    render(<Switch />)
+    const switchElement = screen.getByRole('switch')
+
+    expect(switchElement).toBeInTheDocument()
+  })
+
+  it('renders the Switch with the default state unchecked', () => {
```
alimpens, line 14:
```suggestion
  it('is not checked by default', () => {
```

All these tests are already grouped in the 'Switch' `describe`, so it's not necessary to repeat that here imo

packages/react/src/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    render(<Switch />)
+    const switchElement = screen.getByRole('switch')
+
+    expect(switchElement).not.toBeChecked()
+  })
+
+  it('renders the Switch with the default state disabled', () => {
```
alimpens, line 21:
```suggestion
  it('is not disabled by default', () => {
```

packages/css/src/switch/switch.scss
```diff
@@ -0,0 +1,63 @@
+@mixin reset {
+  box-sizing: border-box;
+}
+
+.amsterdam-switch {
+  display: inline-block;
```
dlnr, line 6: This is not needed

packages/react/src/Switch/README.md
dlnr: There is actually an a empty label in the switch. How should we solve this? I don't think two labels is desirable.

packages/react/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <input
+          {...restProps}
+          className="amsterdam-switch__input"
+          id={inputId}
+          ref={ref}
+          type="checkbox"
+          role="switch"
+        />
+
+        <label className="amsterdam-switch__label" htmlFor={inputId}></label>
```
dlnr, line 22: @alimpens This label is not going to be filled as a formcontrol any time soon. Normally a switch is wrapped in a label in order to make it clickable while hiding the native checkbox. And before you say a it should be able to be used without a label. The label was already here.

storybook/storybook-react/src/components/Switch/Switch.stories.tsx
dlnr: No we can't, currently the Switch has a label enclosed to allow the click on a hidden input element. I would prefer to always wrap the input in a label but I recall you wanted to allow a switch without a label?

dlnr: Your switch component needs to look like this:
```
<label for="switch-1" class="amsterdam-switch">
          Default
          <input type="checkbox" role="switch" id="switch-1">
</label>
```

packages/react/src/Switch/README.md
alimpens: If we want to use this both as a standalone component and a form control element, we do need 2 labels in the second case, right?

Standalone: the empty label in Switch is used for the visual representation of the switch, and if you click it it should change the checkbox. It's possible to do this in other ways, but I think this is one of the most concise ways. 

Form control: if you want to use the Switch with a visible text label, you can add a `FormLabel` to it. This means the input has 2 labels, but that is valid html. We should check how screenreaders handle this though, [this used to cause issues on Safari + VoiceOver](https://scottaohara.github.io/tests/html-inputs/double-label.html)

packages/react/src/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+describe('Switch', () => {
+  it('renders an element with role switch', () => {
+    render(<Switch />)
+    const switchElement = screen.getByRole('switch')
+
+    expect(switchElement).toBeInTheDocument()
+    expect(switchElement).not.toBeChecked()
+  })
+
+  it('s not disabled by default', () => {
```
alimpens, line 15:
```suggestion
  it('is not disabled by default', () => {
```

packages/css/src/switch/README.md
```diff
@@ -0,0 +1,19 @@
...
+## Richtlijnen
+
+- Schakel alleen tussen twee verschillende statussen. Het is een binaire actie.
+- Gebruik labels met een schakelaar zodat de actie duidelijk is.
+- De actie vindt direct plaats zodra de gebruiker de schakelaar gebruikt.
+
+## Relevante WCAG eisen
+
+- [WCAG 1.3.5](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html): het is zowel voor een gebruiker als programmatisch duidelijk wat het doel van een formulierveld is.
+- [Interactieve elementen](https://amsterdam.github.io/design-system/?path=/docs/docs-designrichtlijnen-interactieve-elementen--docs)
```
alimpens, line 14:
```suggestion

Switch is een interactief element, hier gelden [de algemene eisen en richtlijnen voor interactieve elementen](https://amsterdam.github.io/design-system/?path=/docs/docs-designrichtlijnen-interactieve-elementen--docs) voor.
```

alimpens (approved):

PR #580, opened 2023-09-29
Add Alert component

dlnr:
Note that the close button is to be replaced with the ActionButton component.


packages/react/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,67 @@
...
+        {...restProps}
+        ref={ref}
+        className={clsx(
+          'amsterdam-alert',
+          type === 'error' && 'amsterdam-alert--error',
+          type === 'success' && 'amsterdam-alert--success',
+          className,
+        )}
+      >
+        {alertIcon() && <div className="amsterdam-alert__icon">{alertIcon(!!title)}</div>}
```
VincentSmedinga, line 55: We can prevent evaluating the function twice if the suggestion above is applied:

```suggestion.tsx
        {!!alertIcon && <div className="amsterdam-alert__icon">{alertIcon}</div>}
```

packages/react/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <div className="amsterdam-alert__content">
+          {title && <span className="amsterdam-alert__title">{title}</span>}
+          <div className="amsterdam-alert__text">{children}</div>
```
VincentSmedinga, line 58: We shouldn’t wrap children in a text class name. The children must be made responsible for their own appearance. Implementers must use components as children, e.g. a `Paragraph`.

packages/react/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+        {closeable && <AlertClose size={title ? 'level-5' : 'level-6'} onClick={onClose} />}
```
VincentSmedinga, line 60: Let’s extract this size function, we’re using it twice.

VincentSmedinga, line 60: ChatGPT :wink: suggested creating a `closeButtonRef` to get a reference to the close button element. This can be useful if you need to programmatically interact with the close button in the future.

`const closeButtonRef = useRef(null)` and `<AlertClose ref={closeButtonRef} />`

VincentSmedinga, line 60: The `onClose` callback is optional so it may be undefined – we can use [optional chaining with function calls](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls): `onClick={onClose?.()}`.

storybook/storybook-react/src/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,108 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { Alert, Link, Paragraph, UnorderedList } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Feedback/Alert',
```
VincentSmedinga, line 10: Please add `'Feedback'` to the `storySort` array in `storybook/storybook-react/config/preview.tsx` to get correct sorting in the sidebar.

storybook/storybook-react/src/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+  },
+}
+
+export const InformationWithInlineLink: Story = {
+  args: {
+    type: undefined,
+    children: (
+      <Paragraph>
+        Tijdens Koningsdag zijn <Link variant="inline">alle Stadsloketten</Link> gesloten, storingen op formulieren,
+        enz.
```
VincentSmedinga, line 67:
```suggestion
        Tijdens Koningsdag zijn <Link variant="inline">alle Stadsloketten</Link> gesloten. Ook 14 020 en alle andere telefoonnummers van de gemeente zijn niet bereikbaar.
```

([source](https://www.amsterdam.nl/veelgevraagd/op-welke-feestdagen-is-de-gemeente-gesloten-of-eerder-dicht-92eb6-kp))

storybook/storybook-react/src/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ),
+  },
+}
+
+export const Error: Story = {
+  args: {
+    type: 'error',
+    icon: true,
+    closeable: true,
+    children: <Paragraph>U bent verplichte velden vergeten in te vullen</Paragraph>,
```
VincentSmedinga, line 78: Please end paragraphs with a full stop, or a colon if a list follows (multiple occurrences).

packages/css/src/alert/alert.scss
```diff
@@ -0,0 +1,78 @@
...
+  gap: var(--amsterdam-alert-gap);
+  justify-content: space-between;
+  padding-block-end: var(--amsterdam-alert-padding-block-end);
+  padding-block-start: var(--amsterdam-alert-padding-block-start);
+  padding-inline-end: var(--amsterdam-alert-padding-inline-end);
+  padding-inline-start: var(--amsterdam-alert-padding-inline-start);
+}
+
+.amsterdam-alert__content {
+  flex-basis: 100%;
```
VincentSmedinga, line 25: Prefer using the shorthand property, as [advised in the spec](https://drafts.csswg.org/css-flexbox/#flex-components).

```suggestion
  flex: auto;
```

packages/css/src/alert/alert.scss
```diff
@@ -0,0 +1,78 @@
...
+  font-family: var(--amsterdam-alert-text-font-family);
+  font-size: var(--amsterdam-alert-text-narrow-font-size);
+  font-weight: var(--amsterdam-alert-text-font-weight);
+  line-height: var(--amsterdam-alert-text-line-height);
+  row-gap: var(--amsterdam-alert-text-row-gap);
+
+  @media screen and (width > $amsterdam-breakpoint) {
+    font-size: var(--amsterdam-alert-text-wide-font-size);
+  }
+}
```
VincentSmedinga, line 51: All of this should be removed. If `children` are used, they are responsible for their layout and appearance.

VincentSmedinga (changes requested):

packages/react/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,67 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import { Alert as AlertSvg, Checkmark, Close } from '@amsterdam/design-system-react-icons'
```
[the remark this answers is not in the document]
dlnr: I've already added a ticket to rename the icons to prevent this issue: https://gemeente-amsterdam.atlassian.net/browse/DES-389

packages/react/src/Alert/Alert.tsx
dlnr: Will be replacing this when ACtionButton becomes a thing.

dlnr: `Type 'void | undefined' is not assignable to type 'MouseEventHandler<HTMLButtonElement> | undefined'`
Also this is method is from the documentation and also used in other NLDS components.

packages/css/src/alert/README.md
```diff
@@ -0,0 +1,18 @@
...
+- Geel: informatie
+- Rood: fout
+- Groen: goed
+
+## Richtlijnen
+
+- Standaard heeft het element geen sluiten knop. Optioneel kan deze toegevoegd worden.
+- Optioneel kan een icoon worden weggelaten. Een geel alarm heeft nooit een icoon.
+- Optioneel kan een titel worden weggelaten.
+- Aangeraden wordt om het alarm in ieder geval van een paragraaf text te voorzien en deze eventueel te ondersteunen met een link, opsomming of knop.
```
alimpens, line 18: If we choose not to translate Alert:

```suggestion
- Standaard heeft het element geen sluiten knop. Optioneel kan deze toegevoegd worden.
- Optioneel kan een icoon worden weggelaten. Een geel Alert heeft nooit een icoon.
- Optioneel kan een titel worden weggelaten.
- Aangeraden wordt om een Alert in ieder geval van een paragraaf text te voorzien en deze eventueel te ondersteunen met een link, opsomming of knop.
```

packages/css/src/alert/alert.scss
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+@import "../../utils/breakpoint";
+
+.amsterdam-alert {
+  align-items: flex-start;
+  background-color: var(--amsterdam-alert-background-color);
+  border-color: var(--amsterdam-alert-border-color);
+  border-style: var(--amsterdam-alert-border-style);
+  border-width: var(--amsterdam-alert-border-width);
```
alimpens, line 13: Is there a reason for not using the shorthand here?

packages/css/src/alert/alert.scss
```diff
@@ -0,0 +1,64 @@
...
+  padding-block-start: var(--amsterdam-alert-padding-block-start);
+  padding-inline-end: var(--amsterdam-alert-padding-inline-end);
+  padding-inline-start: var(--amsterdam-alert-padding-inline-start);
+}
+
+.amsterdam-alert__content {
+  flex: auto;
+}
+
+.amsterdam-alert__title {
```
alimpens, line 28: We generally specify the text color as well for text elements. Do we want to do that here as well?

packages/react/src/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  it('supports ForwardRef in React', () => {
+    const ref = createRef<HTMLDivElement>()
+
+    const { container } = render(<Alert ref={ref} />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(ref.current).toBe(component)
+  })
```
alimpens, line 70: These are duplicate, you already test these on lines 16-42

packages/react/src/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  it('renders the icon if the type is success', () => {
```
alimpens, line 72:
```suggestion
  it('renders the icon if icon is true and the type is success', () => {
```

packages/react/src/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    const component = container.querySelector(':only-child')
+
+    const icon = component?.querySelector('.amsterdam-alert__icon')
+
+    expect(icon).toBeInTheDocument()
+    expect(icon).toBeVisible()
+  })
+
+  it('does not render the icon if the type is default', () => {
```
alimpens, line 83:
```suggestion
  it('does not render the icon if icon is true and the type is default', () => {
```

packages/react/src/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    expect(ref.current).toBe(component)
+  })
+
+  it('renders a design system BEM class name', () => {
+    const { container } = render(<Alert />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('amsterdam-alert')
```
alimpens, line 49: Do we want to check here if `amsterdam-alert__content` is rendered as well?

packages/react/src/Alert/Alert.tsx
```diff
@@ -0,0 +1,67 @@
...
+  onClose?: () => void
+}
+
+interface AlertCloseProps extends HTMLAttributes<HTMLButtonElement> {
+  size?: 'level-5' | 'level-6'
+}
+
+const AlertClose = forwardRef(
+  ({ className, size, ...restProps }: AlertCloseProps, ref: ForwardedRef<HTMLButtonElement>) => (
+    <button {...restProps} ref={ref} className={clsx('amsterdam-alert__close', className)} title="close">
```
alimpens, line 26: We don't use the `title` html attribute, see [the docs on interactive elements](https://amsterdam.github.io/design-system/?path=/docs/docs-designrichtlijnen-interactieve-elementen--docs#hover).

```suggestion
    <button {...restProps} ref={ref} className={clsx('amsterdam-alert__close', className)}>
```

storybook/storybook-react/src/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+  component: Alert,
+  args: {
+    title: 'Let op',
+    closeable: false,
+    icon: false,
+  },
+  argTypes: {
+    type: {
+      control: {
+        type: 'select',
```
alimpens, line 20: Let's make this a radio group, so you can see all options at once.

storybook/storybook-react/src/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    ),
+  },
+}
+
+export const InformationWithInlineLink: Story = {
+  args: {
+    type: undefined,
+    children: (
+      <Paragraph>
+        Tijdens Koningsdag zijn <Link variant="inline">alle Stadsloketten</Link> gesloten. Ook 14 020 en alle andere
```
alimpens, line 69:
```suggestion
        Tijdens Koningsdag zijn <Link variant="inline" onBackground="light" href="#">alle Stadsloketten</Link> gesloten. Ook 14 020 en alle andere
```

plop-templates/react.tsx.hbs
```diff
@@ -8,8 +8,7 @@ import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'rea
 
 export interface {{pascalCase name}}Props extends PropsWithChildren<HTMLAttributes<HTMLElement>> {}
 
-export const {{pascalCase name}} = forwardRef(
-  ({ children, className, ...restProps }: {{pascalCase name}}Props, ref: ForwardedRef<HTMLElement>) => (
+export const {{pascalCase name}} = forwardRef(({ children, className, ...restProps }: {{pascalCase name}}Props, ref: ForwardedRef<HTMLElement>) => (
```
alimpens, line 11: This is only correct for components with names of 5 characters or shorter. I think most component names will be longer than that, so I wouldn't change this.

storybook/storybook-react/src/Alert/Alert.stories.tsx
dlnr: Linking to # causes strange behavior in Storybook. It wil open the story iframe.

packages/react/src/Alert/Alert.tsx
dlnr: It caused a accessibility error in storybook.

packages/react/src/Alert/Alert.test.tsx
dlnr: That div is actually always there.

packages/css/src/alert/alert.scss
alimpens: Can't we use the shorthand in the tokens as well? So:

`border: var(--amsterdam-alert-border);`

and

```
{
  "amsterdam": {
    "alert": {
      "border": { "value": "4px solid {amsterdam.color.yellow}" },
    }
  }
}
```

packages/react/src/Alert/Alert.test.tsx
alimpens: But we do want to check if it's there, no? If we remove it in future MRs, we're introducing a breaking change in the public facing interface.

packages/css/src/alert/alert.scss
dlnr: The reason i did not do this is because error and success just set the border color and it might become confusing, especialy if you want to reuse our component.

Also, we realy need a semantic layer for color, something like `--amsterdam-color-reactive-error`.

packages/react/src/Alert/Alert.test.tsx
alimpens: Let's leave this as is for now, and discuss our testing strategy with the other devs.

PR #581, opened 2023-09-29
Add Card component

alimpens:


packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+  AnchorHTMLAttributes,
+  ForwardedRef,
+  forwardRef,
+  ForwardRefExoticComponent,
+  HTMLAttributes,
+  PropsWithChildren,
+  RefAttributes,
+} from 'react'
+
+export const CardHeadingGroup = forwardRef(
```
VincentSmedinga, line 17: Will a `children` prop not provide too much freedom?

I’m inclined to replace it with `heading` and `tagline` props.

storybook/storybook-react/src/Card/Card.docs.mdx
```diff
@@ -0,0 +1,21 @@
...
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
+
+## Card with tagline
+
+Een card kan een _tagline_ hebben, tekst voor de titel.
+Om de volgorde voor niet-visuele _user agents_ (zoals een schermlezer) logisch te houden,
```
VincentSmedinga, line 16: Is it too imprecise to just mention screen readers everywhere? “niet-visuele _user agents_” feels a bit too technical.

```suggestion
Om de volgorde voor schermlezers logisch te houden,
```

packages/css/src/card/card.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  /*
+    Remove link focus outline only if :has is supported,
+    which means the outline on the whole card is set
+  */
+  &:has(:focus-visible) a:focus {
+    outline: none;
+  }
+
+  /*
```
VincentSmedinga, line 28: Clever! I do wonder whether it’s going to hold up with more advanced cards, but let’s see how far we get.

VincentSmedinga (changes requested):

packages/css/src/card/README.md
```diff
@@ -0,0 +1,12 @@
+# Card
+
+Je gebruikt een card om inhoud te groeperen die te maken heeft met een enkele link. Met een card kun je extra aandacht vestigen op deze link, en wat context bieden. Een card kan uit meerdere elementen bestaan, de link is verplicht.
```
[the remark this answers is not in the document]
alimpens: > Should we focus a bit more on a card representing an article

An html article you mean? So a 'self-contained [...] independently distributable or reusable [...] composition'. Feels a bit generic imo. The way we use cards is strongly tied to the link they contain, I think?

We could remove the first two mentions of 'link', but I don't think that makes the text clearer.

packages/css/src/card/card.scss
alimpens: Yeah, I think we should be fine as long as we don't wrap the link in an element that has `position: relative`

packages/css/src/card/card.scss
```diff
@@ -0,0 +1,59 @@
...
+    Taken from https://inclusive-components.design/cards/
+  */
+  a::after {
+    bottom: 0;
+    content: "";
+    display: block;
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: 0;
```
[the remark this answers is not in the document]
alimpens: [Inset is pretty well supported](https://caniuse.com/?search=inset), and the nice thing about this solution is that, if it doesn't work, you still have a regular link you can click on. So let's go for it!

packages/react/src/Card/Card.tsx
alimpens: So something like this?

```
<Card.HeadingGroup key={1}
  heading="Monitor Attracties MRA"
  tagline="Dossier"
>
```

Could work, but we would also have to pass `size` and `level` (and all regular heading props?) for the Heading, and `href` (and all regular link props?) for the Link. Oh, and the props for Paragraph. Together with the regular arguments about inline styling of text, parts of the text being in a different language etc. 

Maybe leave it as is for now?

storybook/storybook-react/src/Card/Card.docs.mdx
alimpens: I've chosen these words because I wanted to include braille readers and any non-visual user agent I don't know about :D 

We could do something like this, to make it clear we don't consider screen readers the only non-visual UA:

```suggestion
Om de volgorde voor bijvoorbeeld schermlezers logisch te houden,
```

But if that's too nitpicky, we can go with your version.

packages/css/src/card/README.md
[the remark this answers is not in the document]
VincentSmedinga: Well, a conceptual article – news, a blog post, a job posting, an event, a building project, info on a waste guide fraction…

I’ll give it a try:

```suggestion
Gebruik een card op een overzichtspagina om te verwijzen naar de detailpagina van een artikel zoals een nieuwsbericht, een vacature of een zoekresultaat. De card bevat meestal de titel, introductie en afbeelding van het artikel. De link is verplicht.
```

packages/react/src/Card/Card.tsx
VincentSmedinga: Depends on the amount of variety in the designs. I would expect the tagline to always have the same appearance. The level and size of the heading may well be variable.

What if people add three child elements? They will be surprised to find them all in reverse order.

Combining everything, we could make `tagline` a prop, render its value first, then all children in their own order. Makes it easy to configure the heading. Only if we get cards with a dark background will we need to pass an `inverseColor` prop from `HeadingGroup` to its tagline `Paragraph`, but that’s not very probale. Good idea?

storybook/storybook-react/src/Card/Card.docs.mdx
VincentSmedinga: If you’re naming it a ‘braille reader’, that’s a member of the ‘screen reader’ group as well, right? :)

VincentSmedinga (changes requested):

packages/react/src/Card/Card.tsx
alimpens: > we could make tagline a prop, render its value first

That would cause a wrong DOM order, the tagline comes after the title in the DOM. I think we can make `tagline` a prop, but the order of the DOM elements still needs to be reversed visually.

> What if people add three child elements? They will be surprised to find them all in reverse order.

I don't think this will be a huge issue, if you're adding more elements than just the link in the heading you're probably creating invalid html anyway. An `hgroup` can only contain a heading and one or more paragraphs. If the paragraph is rendered using the `tagline` prop, you can only really add a heading.

I've added a `tagline` prop!

storybook/storybook-react/src/Card/Card.docs.mdx
alimpens: Ah, you're right! I thought a screen reader is just a voice reader, but braille readers are part of that group as well. Shorter is better.

VincentSmedinga (approved):

PR #595, opened 2023-10-04
Wrap calculations in clamp css values in calc()

dlnr:
Projects using  postcss-safe-parser  (like CRA 4) are getting a build error on the :root variables that use a calc function without `calc()`.


VincentSmedinga (approved):
The `val` value for `clamp` works without `calc` for direct addition, subtraction, multiplication and division ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp#parameters)) but with `calc` in every case and this adds compatibility.

PR #623, opened 2023-10-11
Add 'Icon' postfix to all React icons

alimpens:


proprietary/react-icons/indexTemplate.js
```diff
@@ -0,0 +1,12 @@
+const path = require('path')
+
+function indexTemplate(filePaths) {
+  const exportEntries = filePaths.map(({ path: filePath }) => {
+    const basename = path.basename(filePath, path.extname(filePath))
+    const exportName = /^\d/.test(basename) ? `Svg${basename}` : basename
+    return `export { default as ${exportName}Icon } from './${basename}'`
```
VincentSmedinga, line 7: This transformation may be hard to find like this – let’s add a comment to highlight it and express our intent. Also I’d be a fan of requiring an empty line before each return statement.

```suggestion

    // The name of the React component for each icon ends with `Icon`
    return `export { default as ${exportName}Icon } from './${basename}'`
```

VincentSmedinga (changes requested):
Thanks for all the renames 🏭

VincentSmedinga (approved):

PR #674, opened 2023-10-25
Add pagination component

alimpens:
Based on [ASC Pagination](https://amsterdam.github.io/amsterdam-styled-components/?path=/docs/ui-pagination--default-story).

Couple of questions:

- This component currently consists of a list of buttons. This makes sense if the pagination and filtering and stuff all happens client-side, ie you're not really navigating to a different page when you interact with the component. For apps that do navigate to a different page, maybe we should add an option to render links instead of buttons? Maybe we should wait until we have users that need this though.
- Currently, what the Pagination shows is calculated in the component based on the `page`, `pageSize` and `collectionSize` props. I can imagine some apps get this calculation already from their backend, so maybe we should make this calculation optional? Again, maybe we should wait until we have users that need this.
- Do we want to statically set the `aria-label`s and previous- and next-button text content? Or should that be configurable?


VincentSmedinga:
> * This component currently consists of a list of buttons. This makes sense if the pagination and filtering and stuff all happens client-side, ie you're not really navigating to a different page when you interact with the component. For apps that do navigate to a different page, maybe we should add an option to render links instead of buttons? Maybe we should wait until we have users that need this though.

These should totally be links from the start. I don’t think something happening either client-side or server-side is a decisive factor to make an action count as navigating or not. If it is, would a user of a single-page app ever be navigating?

An SPA is free to hijack the url and replace it with client-side routing. But they should push the url onto the navigation history to make the end result look like a navigation happened.

https://twitter.com/housecor/status/1717152910713278544

We want users to be able to share the link to a specific page.

https://twitter.com/housecor/status/1717153836823892190

VincentSmedinga:
> Currently, what the Pagination shows is calculated in the component based on the `page`, `pageSize` and `collectionSize` props. I can imagine some apps get this calculation already from their backend, so maybe we should make this calculation optional? Again, maybe we should wait until we have users that need this.

Yes. I had to get used a bit to the idea of the component doing these calculations, but seeing it in action I think it’s the right way ahead.

VincentSmedinga:
> Do we want to statically set the `aria-label`s and previous- and next-button text content? Or should that be configurable?

Static is fine for now.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+    // If page prop changes, set currentPage so component can be managed externally
+    useEffect(() => {
+      setCurrentPage(page)
+    }, [page])
+
+    const totalPages = Math.ceil(collectionSize / pageSize)
+
+    // Get array of page numbers and / or spacers
+    const range = getRange(currentPage, totalPages, maxVisiblePages)
```
hcvdhaar, line 110: For a small performance improvement would it be better to memo this? Every state change outside of the Pagination component would run this function even nothing happens here.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { ForwardedRef, forwardRef, HTMLAttributes, useEffect, useState } from 'react'
+import { Icon } from '../Icon/Icon'
+
+export interface PaginationProps extends HTMLAttributes<HTMLElement> {
+  /**
+   * The number of items in your paginated collection.
+   *
+   * Note that this is not the same as the number of pages, but will be used to calculate it.
+   */
+  collectionSize: number
```
hcvdhaar, line 17: why not offer a totalpages prop? This usually what is provided in a pagination object.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+  // eslint-disable-next-line no-unused-vars
+  onPageChange?: (page: number) => void
+  /**
+   * The current page number.
+   */
+  page?: number
+  /**
+   * The number of items per page.
+   */
+  pageSize?: number
```
hcvdhaar, line 30: If this needed to calculate the number of pages, shouldn't it be required? If it is not provided it defaults to 10. Then it could be you request the wrong page. Would strong opt for a totalPages prop

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const min = 1
+  // the center part of the range starts with the current page minus half of the visible pages
+  let centerStartPage = currentPage - Math.floor(visiblePages / 2)
+  // centerStartPage has a lower limit of 1
+  centerStartPage = Math.max(centerStartPage, min)
+  // centerStartPage has an upper limit of 1 plus total pages minus visible pages
+  centerStartPage = Math.min(centerStartPage, min + totalPages - visiblePages)
+
+  const pages = Array.from({ length: visiblePages }, (_el, i) => centerStartPage + i).reduce<Array<string | number>>(
```
hcvdhaar, line 66: This is now a collection of strings and number. Would it be nicer to have a Page object? Something like:

```
{
  type: 'page' || 'space',
  page: number,
  currentPage: boolean 
  etc...
}
```
I guess it more scalable and at some parts more readable and better to understand the flows/logic

hcvdhaar: but maybe a bit much for now?

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+      pageSize = 10,
+      maxVisiblePages = 7,
+      ...restProps
+    }: PaginationProps,
+    ref: ForwardedRef<HTMLElement>,
+  ) => {
+    const [currentPage, setCurrentPage] = useState(page)
+
+    // If page prop changes, set currentPage so component can be managed externally
+    useEffect(() => {
```
hcvdhaar, line 103: Not best practice to use `useEffect` for these cases. It works, but yeah ;)

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

hcvdhaar (commented):
Looks very nice. It is a complex component! good job

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,192 @@
...
+    // Don't show pagination if you only have one page
+    if (collectionSize <= pageSize) {
+      return null
+    }
+
+    return (
+      <nav {...restProps} className={clsx('amsterdam-pagination', className)} ref={ref} aria-label="Paginering">
+        <ol className="amsterdam-pagination__list">
+          <li>
+            <button
```
alimpens, line 136: Responding to the comment by @VincentSmedinga about links vs buttons here:

> These should totally be links from the start.

Hm, I think it's a bit more complicated than that. I did a quick inventory how other design systems and MDN does this, 5x buttons vs 7x links. I think it very much depends on the structure of your app. 

Most Amsterdam apps I checked use url params to indicate pagination, which makes it more of a filter I think? In those cases, buttons make more sense imo. These are amsterdam.nl itself (which does use links), data.amsterdam.nl and onderzoek.amsterdam.nl. Mijn.amsterdam.nl creates a url with `react-router-dom` I think, so something like /page/2 (which should be a link, maybe? But they use buttons).

Erfpacht has multiple Pagination components on one page, so links wouldn't make a lot of sense there. 

I think for use within Amsterdam a button makes more sense. Maybe we should support both, if it's needed.

packages/react/src/Pagination/Pagination.tsx
alimpens: Good call! I actually don't think `useEffect` is necessary here at all, `useState(page)` already handles the controlled component functionality

packages/react/src/Pagination/Pagination.tsx
alimpens: I did a quick inventory, ASC, Ant, Carbon and Kolibri design systems all use something like `collectionSize` and `pageSize`. Atlassian passes an array with all pages(?) and Lion passes the total number of pages. Do you have other examples of Pagination components that only pass a `totalPages` prop?

hcvdhaar: I checked MUI and Carbon React as well, and they basically ask for `count` or `totalItems`. It seems there are 2 variants of paginations. Like the navigation variant or the table variant. Were for the latter the props `collectionSize` and `pageSize` make more sense. 

- https://react.carbondesignsystem.com/?path=/docs/components-paginationnav--overview
- https://react.carbondesignsystem.com/?path=/docs/components-pagination--overview
- https://mui.com/material-ui/react-pagination/#basic-pagination
- https://mui.com/material-ui/react-pagination/#table-pagination

packages/react/src/Pagination/Pagination.tsx
alimpens: Just discussed this, for now we'll use buttons, because changing it to links makes the component a bit more complex. I imagine we need a variant with links as well, but we should test it thoroughly in an application setting before we add that.

packages/react/src/Pagination/Pagination.tsx
alimpens: I tried this, but I think it makes the `getRange` function a bit harder to read

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+  // eslint-disable-next-line no-unused-vars
+  onPageChange?: (page: number) => void
+  /**
+   * The current page number.
+   */
+  page?: number
+  /**
+   * The maximum amount of pages shown. This has a lower limit of 5
+   */
+  maxVisiblePages?: number
```
VincentSmedinga, line 24: Please sort the properties.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * The total number of pages.
+   */
+  totalPages: number
+}
+
+/**
+ * This returns an array of the range, including spacers
+ *
+ * @example
```
VincentSmedinga, line 34: Nice!

VincentSmedinga (changes requested):
Just submitted my comments – only now I notice they have been ‘pending’ for a week…

packages/css/src/pagination/README.md
```diff
@@ -0,0 +1,18 @@
+# Pagination
+
+Pagination (in het Nederlands: paginering) is een navigatie-element onder een zoekresultatenlijst. Bij grote hoeveelheden zoekresultaten kan het duidelijker of functioneler zijn om de inhoud over meerdere pagina´s te verdelen. Paginering toont op welke zoekresultatenlijst de gebruiker zich bevindt en kan hiermee naar een andere zoekresultatenlijst navigeren.
```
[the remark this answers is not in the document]
alimpens: Most of the docs are from [the 'old' design system page](https://www.amsterdam.nl/digitalservices/design-system/atomen-moleculen/pagination/), they explicitly state it should only be used for a search results page. I can imagine it's useful for other pages / patterns as well, I'll ask Soesanto where this restriction comes from.

packages/css/src/pagination/pagination.scss
```diff
@@ -0,0 +1,73 @@
...
+  list-style-type: none;
+  margin-block: 0;
+  padding-inline-start: 0;
+}
+
+.amsterdam-pagination__list {
+  color: var(--amsterdam-pagination-color);
+  display: flex;
+  flex-wrap: wrap;
+  font-family: var(--amsterdam-pagination-font-family);
```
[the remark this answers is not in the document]
alimpens: > Or is that an unwanted dependency between components?

Yeah, I thought we opted for less dependencies between components.

> Or a paragraph or span between the list items and the links?

Adding HTML markup for style reasons doesn't sound like a good plan to me. Imo it's not really a problem to use the cascade here, seeing as it's within a component that's indivisible and doesn't accept children.

packages/css/src/pagination/pagination.scss
```diff
@@ -0,0 +1,73 @@
...
+
+@mixin button-reset {
+  all: unset;
+  box-sizing: border-box;
+  outline: revert;
+  -webkit-text-size-adjust: 100%;
+}
+
+.amsterdam-pagination__button {
+  @include button-reset;
```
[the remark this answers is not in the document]
alimpens: Yeah, `all: unset` removes the `gap` property if you set it after. I'll add a comment here

packages/css/src/pagination/pagination.scss
```diff
@@ -0,0 +1,73 @@
...
+  text-underline-offset: 3px;
+  touch-action: manipulation;
+
+  &:hover {
+    color: var(--amsterdam-pagination-button-hover-color);
+    text-decoration: underline;
+  }
+
+  &:disabled {
+    visibility: hidden;
```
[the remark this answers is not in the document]
alimpens: I agree. I originally set `visibility: hidden` here so you wouldn't have layout shifts on wide screens, but seeing as we have this guideline, the layout shift doesn't really matter:

> Start een zoekresultatenpagina bovenaan de pagina na het veranderen van pagina

packages/css/src/pagination/pagination.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // Override for icon size
+  span.amsterdam-icon svg {
+    height: 1rem;
+    width: 1rem;
+  }
```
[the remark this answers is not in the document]
alimpens: Maybe we should wait for [this ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-447), so we'll have a coherent system in place?

packages/css/src/pagination/pagination.scss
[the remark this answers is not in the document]
VincentSmedinga: I’ll try and live with it.

packages/css/src/pagination/pagination.scss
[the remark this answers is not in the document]
VincentSmedinga: Sure, added a link to here to it.

VincentSmedinga (approved):

PR #680, opened 2023-10-26
Show correct colors in Highlight story

alimpens:


storybook/storybook-react/src/Highlight/Highlight.stories.tsx
```diff
@@ -32,151 +32,70 @@ const meta = {
       },
     },
   },
+  render: ({ color }) => (
+    <Highlight color={color}>
+      <PageGrid>
+        <GridCell gridColumns={{ start: 3, span: 8 }}>
+          <Blockquote inverseColor={!(color === 'green' || color === 'yellow')}>
```
VincentSmedinga, line 39: More background colours than these two require non-inverted text.

Let’s use `Array.prototype.includes` to remove even more repetition ;)

According to [Stijlweb](https://amsterdam.nl/stijlweb/basiselementen/kleuren/#PagCls_15671872), some of them accept both inverted and non-inverted. Out of scope to add a reference to this doc?

VincentSmedinga (changes requested):
Nice little improvement, crossed my mind as well. Will introduce small merge conflicts as the grid prop has changed, but we can handle that.

VincentSmedinga (approved):

VincentSmedinga (approved):

PR #681, opened 2023-10-27
Add Dialog component

dlnr:


packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+  gap: 1.5rem;
+  justify-items: flex-start;
+  max-block-size: 100%; /* safari */
+  overflow-y: auto;
+  overscroll-behavior-y: contain;
+  padding-inline-end: var(--amsterdam-dialog-article-padding-inline-end);
+
+  &::-webkit-scrollbar-track {
+    background: var(--amsterdam-dialog-background-color);
+  }
```
alimpens, line 52: What does this do? I generally wouldn't try to style scrollbars, it can have a11y implications. Also, this doesn't work on Chrome / Windows

packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+}
+
+.amsterdam-dialog__footer {
+  display: flex;
+  flex-direction: column-reverse;
+  grid-gap: var(--amsterdam-dialog-footer-gap);
+  padding-block: var(--amsterdam-dialog-footer-padding-block);
+
+  @media screen and (min-width: $amsterdam-breakpoint-medium) {
+    align-items: end;
```
alimpens, line 87: Is this necessary?

packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
+    font-size: var(--amsterdam-dialog-title-wide-font-size);
+  }
+
+  @include reset;
+}
+
+.amsterdam-dialog__footer {
+  display: flex;
+  flex-direction: column-reverse;
```
alimpens, line 82: Something we should discuss with design: this changes the tab order, which isn't great for a11y

storybook/storybook-react/config/preview.tsx
```diff
@@ -3,6 +3,8 @@ import '@amsterdam/design-system-assets/font/index.css'
 import '@amsterdam/design-system-css/dist/index.css'
 import { viewports } from './viewports'
 
+import './story.css'
```
alimpens, line 6: We currently use `storybook-react/config/preview-body.html` for story-only styling, this can go there. (@VincentSmedinga: why did we move this from a css file to a html file with style tags?)

storybook/storybook-react/src/Dialog/Dialog.docs.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+<Controls />
+
+## Scroll Content
+
+<Canvas of={DialogStories.ScrollContent} className="amsterdam-dialog-story" />
+
+## No Actions
+
+<Canvas of={DialogStories.NoActions} className="amsterdam-dialog-story" />
```
alimpens, line 19: `DialogStories.ShowDialog` is missing here

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import { Button, Dialog, Heading, Paragraph } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Containers/Dialog',
+  component: Dialog,
+  args: {
+    title: 'Later verder gaan',
+    children: [
```
alimpens, line 14: This doesn't have to be an array, it's a single child

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Paragraph key={0}>
+        U kunt de ingevulde antwoorden opslaan in onze beveiligde database, deze kunt u later openen om verder te gaan
+        met invullen. Nadat u op opslaan heeft geklikt ontvangt u een mail. Met de link in deze mail kunt verder gaan
+        met het formulier.
+      </Paragraph>,
+    ],
+    actions: [
```
alimpens, line 21: Same here I think

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <>
+        <Button
+          key={1}
+          type="button"
```
alimpens, line 25: This isn't necessary, `type="button"` is default for the Utrecht button

storybook/storybook-react/config/story.css
```diff
@@ -0,0 +1,5 @@
+/* Workaround to mimic dark background for dialog Canvas blocks */
+.amsterdam-dialog-story {
+  background-color: #333 !important;
```
alimpens, line 3: Is this used? It seems like it's either `::backdrop` or `parameters.backgrounds.default: 'dark'` that set the background color.

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+    open: true,
+  },
+  render: (args) => (
+    <div style={{ minHeight: '500px' }}>
+      <Dialog {...args} />
+    </div>
+  ),
+}
+
+export const NoActions: Story = {
```
alimpens, line 60: Do we want to keep this story? I thought we decided that actions are required for Dialogs?

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Dialog {...args} />
+    </div>
+  ),
+}
+
+export const ScrollContent: Story = {
+  args: {
+    open: true,
+    title: 'Privacyverklaring gemeente Amsterdam',
+    children: [
```
alimpens, line 76: This array and the keys aren't necessary, it's a single child because you wrap it in a `React.Fragment`

packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../utils/breakpoint";
+
+.amsterdam-dialog {
+  align-content: start;
+  background-color: var(--amsterdam-dialog-background-color);
+  border: var(--amsterdam-dialog-border);
+  display: grid;
```
alimpens, line 12: Overriding the default display property here makes a closed `dialog` accessible for screenreader, which we don't want. `pointer-events: none` isn't enough to fix that.

storybook/storybook-react/src/Dialog/Dialog.docs.mdx
dlnr: There was an issue with the `showModal()` function causing the button not to open the dialog, I've now set an `id` to the modal but typescript interferes with the queryselector forcing me to cast the element to any:
```
document.querySelector('#showdialog' as any)?.showModal()
```

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
dlnr: True but what can we do about the console error?
```Warning: Each child in a list should have a unique "key" prop.```

File: packages/react/src/Dialog/Dialog.test.tsx [hunk 13]
```diff
@@ -0,0 +1,84 @@
...
+
+  it('is not visible when open attribute is not used', () => {
+    const { container } = render(<Dialog />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toBeInTheDocument()
+    expect(component).not.toBeVisible()
+  })
+
```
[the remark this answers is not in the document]
dlnr: I've restored the default `display: none` property so it should not be the case.

packages/react/src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  it('renders DialogClose button', () => {
+    const { container } = render(<Dialog />)
+
+    const closeButton = container.querySelector('.amsterdam-dialog__close')
+
+    expect(closeButton).toBeInTheDocument()
+  })
+})
```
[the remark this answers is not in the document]
dlnr: I've tried but can not seem to get `expect(component).not.toBeVisible()` to work. Can you help out?

packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+
+.amsterdam-dialog__title {
+  color: var(--amsterdam-dialog-title-color);
+  flex: auto;
+  font-family: var(--amsterdam-dialog-title-font-family);
+  font-size: var(--amsterdam-dialog-title-narrow-font-size);
+  font-weight: var(--amsterdam-dialog-title-font-weight);
+  line-height: var(--amsterdam-dialog-title-line-height);
+
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
```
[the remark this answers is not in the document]
dlnr: I thought this was just for typography media queries? When should or are we allowed to use the -medium and -wide breakpoints then?

packages/css/src/dialog/dialog.scss
```diff
@@ -0,0 +1,107 @@
...
+
+  /* no token because dialog does not inherit from any element */
+  &::backdrop {
+    background: #0006;
+  }
+
+  &:not([open]) {
+    opacity: 0%;
+    pointer-events: none;
+  }
```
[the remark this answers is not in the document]
dlnr: Although you can now animate display (in modern browsers) https://developer.chrome.com/blog/entry-exit-animations/

storybook/storybook-react/config/story.css
alimpens: Ah, I missed that one

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
alimpens: Just removing the square brackets and the keys works, no?

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
alimpens: That doesn't cause a console error for me?

packages/react/src/Dialog/Dialog.test.tsx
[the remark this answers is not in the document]
alimpens: Ah, [it seems `dialog` isn't properly implemented in `jsdom` yet](https://github.com/jsdom/jsdom/issues/3294), we probably can't test this currently.

packages/css/src/dialog/README.md
```diff
@@ -0,0 +1,22 @@
+# Dialog
+
+The dialog provides feedback or reveals additional information or functionality while maintaining the context of an underlying page.
```
alimpens, line 3:
```suggestion
A Dialog allows the user to focus on one task or a piece of information, by popping-up and blocking the page content until the modal task is completed, or until the user dismisses the action.
```

packages/css/src/dialog/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Guidelines
+
+Use modals sparingly because they interrupt the user's workflow.
```
alimpens, line 7:
```suggestion
- Use Dialog sparingly, because it interrupts the user's workflow.
- Use Dialog for short and non-frequent tasks. For common tasks consider using the main flow.
```

packages/css/src/dialog/README.md
dlnr: I'd rather not use the term pop-up and I was happy with the phrasing of keeping the context of the page. That's why we use a semi transparent backdrop. Also it's not really blocking the page if you're able to dismiss the dialog.

alimpens: Why do you rather not use the term pop-up? We can get rid of it I think, it doesn't really add much. 

To me it doesn't really feel like the core function of the Dialog that it (only visually) maintains the context of an underlying page. The core function is taking you out of the underlying page and forcing you to take action before you can return. 

> Also it's not really blocking the page if you're able to dismiss the dialog.

Which is taking action, right? So it's blocked until the user does something.

packages/css/src/dialog/dialog.scss
dlnr: It insures the items are aligned to the bottom of the Dialog, since the actions are "open" to what the developer puts in there, this is just extra insurance.

storybook/storybook-react/config/preview.tsx
dlnr: I follow [storybook documentation](https://storybook.js.org/docs/react/configure/styling-and-css), I could move the preview-body css to this file.

packages/react/src/Dialog/Dialog.test.tsx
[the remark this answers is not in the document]
dlnr: I did get it to work in storybook interaction tests. Perhaps we can use the test-storybook method.

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+export const ShowDialog: Story = {
+  args: {
+    id: 'showdialog',
+    actions: [
+      <>
+        <Button
+          key={1}
+          variant="tertiary"
+          autoFocus
+          onClick={() => document.querySelector('#showdialog' as any)?.close()}
```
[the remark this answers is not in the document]
dlnr: This results in `Argument of type 'HTMLDialogElement' is not assignable to parameter of type 'string'.` I solved it in a different way for the close button, but the button outside the Dialog targeting an ID (because there a multiple Dialog elements in de SB Docs) needs the `any`. Typescript is "really" great ;)

packages/react/src/Dialog/Dialog.test.tsx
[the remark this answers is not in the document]
dlnr: How would you test the non-interaction with an event handler??

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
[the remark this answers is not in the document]
alimpens: Are you sure? You might've missed the parenthesis. This works for me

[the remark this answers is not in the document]
dlnr: Ah right, I changed it but for the showModal click.

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { DialogHTMLAttributes, ForwardedRef, forwardRef, PropsWithChildren, ReactNode } from 'react'
+import { Icon } from '../Icon'
+import { VisuallyHidden } from '../VisuallyHidden'
+
+export interface DialogProps extends PropsWithChildren<DialogHTMLAttributes<HTMLDialogElement>> {
+  actions?: ReactNode
+}
+
+const DialogClose = forwardRef(({ ...restProps }, ref: ForwardedRef<HTMLButtonElement>) => (
+  <button role="close" aria-label="close" ref={ref} className="amsterdam-dialog__close" formNoValidate {...restProps}>
```
alimpens, line 17: I don't think making up your own `role` is a good idea, screenreaders ignore it. 

Having both `aria-label` and `VisuallyHidden` seems a bit much, don't we just want one or the other? `VisuallyHidden` has better support, I think

storybook/storybook-react/config/preview.tsx
alimpens: I would do that for now, we'll figure out how we want to implement story-only styling later.

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Button onClick={() => (document.querySelector('#showdialog') as HTMLDialogElement)?.showModal()}>
+          Open Dialog
+        </Button>
+        <Story />
+      </article>
+    ),
+  ],
+  parameters: {
+    backgrounds: {
+      default: 'white',
```
alimpens, line 119: This throws a warning

```suggestion
      default: 'light',
```

dlnr: Light gives you a grey background which I did not want because it makes the backdrop darker then it should be. I've opted to disable the background to combat the dark background I've set for the other stories.

storybook/storybook-react/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+  parameters: {
+    backgrounds: {
+      disable: true,
+    },
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+    await userEvent.click(canvas.getByRole('button'), { delay: 500 })
+    await userEvent.click(canvas.getByText('Sluiten'), { delay: 1000 })
+  },
```
alimpens, line 126: This currently doesn't really test anything, right? It just goes through the interaction? Maybe we can get rid of this for now, but it would be nice to figure out how we can use this for testing in a seperate ticket

dlnr: It tests if you can successfully open and close the dialog, something the unit test fails to do.

alimpens: But it doesn't really test that currently, right? Say the close button doesn't close the Dialog, would we get a notification now?

dlnr: Well just in storybook until I integrate jest or vitest, they recelty release a @storybook/test package that combines a lot of things. I'll comment it out until I have something to show for.

dlnr: Ok done, i'll add the ticket later after researching new development a bit.

alimpens (approved):

PR #684, opened 2023-10-27
Rename Page Grid to Grid and make Grid Cell a subcomponent

VincentSmedinga:


packages/react/src/Grid/GridCell.tsx
```diff
@@ -5,7 +5,7 @@
 
 import clsx from 'clsx'
 import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
-import type { PageGridColumnNumber, PageGridColumnNumbers } from './PageGrid'
+import { GridColumnNumber, GridColumnNumbers } from './Grid'
```
alimpens, line 8:
```suggestion
import type { GridColumnNumber, GridColumnNumbers } from './Grid'
```

packages/css/src/grid-cell/grid-cell.scss
```diff
@@ -5,41 +5,41 @@
 
 @import "../../utils/breakpoint";
```
alimpens, line 7: If we're turning it into a single component, maybe we want to merge the css files as well? Or at least have them both live in `/css/src/grid`?

packages/react/src/Grid/index.ts
```diff
@@ -1,5 +1,3 @@
 export type { GridCellProps } from './GridCell'
```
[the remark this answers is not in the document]
VincentSmedinga: `Grid` doesn’t currently have props. I’m getting used to how subcomponents are defined, our other components use an interface for it. Should I have done something differently? Do we need to export the interface?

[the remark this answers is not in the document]
alimpens: Ah, sorry, I assumed Grid had it's own props, but it doesn't. This is fine

packages/css/src/grid-cell/grid-cell.scss
VincentSmedinga: Ah yes I had planned that but forgot. A slight problem is the two README documents in the CSS package. Is a convention to separate their file names better than directories?

For the React components, having everything in one directory is fine. If we would get tokens for grid cell, they would live in two files. The stories are now also separate for Page Grid and Grid Cell. We can also keep the CSS components separate?

VincentSmedinga: Merged grid and grid cell into one directory in our packages, but not in Storybook. Therefore, chose to have `README.grid.md` and `README.grid-cell.md` in the CSS package as both stories reference this separately.

packages/react/src/Alert/Alert.tsx
```diff
@@ -5,7 +5,7 @@
 
 import { AlertIcon, CheckmarkIcon, CloseIcon } from '@amsterdam/design-system-react-icons'
 import clsx from 'clsx'
-import React, { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren, useMemo } from 'react'
```
dlnr, line 8: Legit change but I see a trend where you make corrections in many other components then what your PR is about. I have the tendency to do the same but I usually get the order to stop it.

packages/react/src/Grid/Grid.test.tsx
dlnr, line 1: The test is just the scaffold of the component, I'm missing tests for cell children and other scenarios you might think of.

packages/react/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ) => (
+    <div {...restProps} ref={ref} className={clsx('amsterdam-grid', className)}>
+      {children}
+    </div>
+  ),
+) as GridComponent
+
+Grid.Cell = GridCell
+Grid.displayName = 'Grid'
+GridCell.displayName = 'Grid.Cell'
```
dlnr, line 43: Funny that you would have to declare this twice, also in GridCell, I'm not sure what the correct method should be it's hard to find "official" documentation on compound components. Perhaps the direction you're going is too far and it should look like:
```
export default Grid
export { GridCell }
```
Feel free to resolve this brainfart.

packages/react/src/Grid/GridCell.test.tsx
dlnr, line 1: Ah I see all the tests I expected for Grid are in here because you renamed `GridCell` to `Grid.Cell`. This is cool but I think we need to apply the same rules to all compound components and I have some thoughts on this. Perhaps we could ask around to find out what developers prefer.

storybook/storybook-react/config/preview.tsx
```diff
@@ -13,12 +13,6 @@ export const parameters = {
       date: /Date$/,
     },
   },
-  docs: {
-    // Show code by default.
-    canvas: {
-      sourceState: 'shown',
```
dlnr, line 19: Also feels unrelated to grid changes.

storybook/storybook-docs/src/grid.stories.mdx
```diff
@@ -105,3 +109,8 @@ Dat zorgt ervoor dat content op zo veel mogelijk vensterbreedtes en met diverse
 Op dit moment reageert de typografie nog op één breakpoint bij een vensterbreedte van 854 pixels.
 Nu het grid doorontwikkeld is, stemmen we de typografie daar weer op af.
```
dlnr, line 110: This is very ambitious since we have yet to do a full layout using the grid. Also, how do we remind ourselves to change this line when the typography has been changed?

dlnr (commented):
The rename to the compounded component feels like it's not consistent with other components that we've grouped like this, because it's not really a component like the others. So I have doubts about the structure in general.

storybook/storybook-docs/src/grid.stories.mdx
VincentSmedinga: Bettina is working on this and I also made a prototype while making the narrow Grid variations, albeit with not too much text inside. I expect we’re going to change the current typography breakpoint to one of the two grid breakpoints, tweak the font sizes a bit and that’s it.

Understand the point about tracing back this text. Made a note of it in the appropriate task on our board.

storybook/storybook-react/config/preview.tsx
VincentSmedinga: True.

packages/css/src/grid/README.grid-cell.md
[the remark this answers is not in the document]
VincentSmedinga: Both correct, we’ll translate all current docs around next week, then continue in English.

packages/css/src/grid/grid-cell.scss
```diff
@@ -5,41 +5,41 @@
 
 @import "../../utils/breakpoint";
 
-.amsterdam-page-grid > .amsterdam-grid-cell--full-width {
+.amsterdam-grid-cell--full-width {
```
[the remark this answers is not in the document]
dlnr: It still seems like a hybrid solution if you know what I mean, but i'll take the refactor.

packages/css/src/grid/grid.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.amsterdam-grid {
+  display: grid;
+  gap: var(--amsterdam-grid-gap);
+  grid-template-columns: repeat(var(--amsterdam-grid-column-count), 1fr);
+  margin-inline: auto;
+  padding-inline: var(--amsterdam-grid-padding-inline);
+}
+
+@media screen and (min-width: $amsterdam-breakpoint-medium) {
```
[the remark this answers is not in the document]
dlnr: What I meant was that in other file we nest the media queries like so:

```
.amsterdam-grid {
  display: grid;
  gap: var(--amsterdam-grid-gap);
  grid-template-columns: repeat(var(--amsterdam-grid-column-count), 1fr);
  margin-inline: auto;
  padding-inline: var(--amsterdam-grid-padding-inline);

  @media screen and (min-width: $amsterdam-breakpoint-medium) {
    .amsterdam-grid {
      grid-template-columns: repeat(var(--amsterdam-grid-medium-column-count), 1fr);
    }
  }

  @media screen and (min-width: $amsterdam-breakpoint-wide) {
    .amsterdam-grid {
      grid-template-columns: repeat(var(--amsterdam-grid-wide-column-count), 1fr);
    }
  }
}
```

Avoiding nesting because it is traditional makes no sense to me, sass and even vanilla css support it and it adds to the readability and scoping of variables (if you need them changed for some reason like a user preference query).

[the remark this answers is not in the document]
VincentSmedinga: Thanks for clarifying, I interpreted your comment totally differently.

With ‘traditional’ I meant the `min-width: x` syntax instead of the `width > x` syntax of Media Queries Level 4. Not the nesting :)

Updated.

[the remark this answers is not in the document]
dlnr: yeah sorry I commented on two things.

PR #730, opened 2023-11-08
Set language on document instead of screen

VincentSmedinga:
On second thought, `<html lang="nl">` is better than `<Screen lang="nl">`.

Even though this explicitly sets the language for the website, the developer still has to remember to do this on the HTML element, for example because the content of some <meta> elements can also be in that language.

In addition, on a page in another language e.g. `<html lang="ar">` you may forget to also set the (planned) `lang` prop on `Screen`, so that the default value of `nl` kicks in. The fact that developers forget something in itself is not our problem, but we can limit the chance of errors occurring.

So this also documents that implementers should consider this and how.

Sources:
* https://www.w3.org/International/techniques/authoring-html.en?open=language&open=textprocessing#textprocessing
* https://www.w3.org/International/questions/qa-html-language-declarations#basics

Regarding the language tag of `nl` instead of `nl-NL`:

> The golden rule when creating language tags is to keep the tag as short as possible. Avoid region, script or other subtags except where they add useful distinguishing information. For instance, use `ja` for Japanese and not `ja-JP`, unless there is a particular reason that you need to say that this is Japanese as spoken in Japan, rather than elsewhere.
(…)
Once again, you should only use region subtags if they are necessary to make a distinction you need. Unless you specifically need to highlight that you are talking about Italian as spoken in Italy you should use `it` for Italian, and not `it-IT`. The same goes for any other possible combination. – https://www.w3.org/International/articles/language-tags/


dlnr (approved):
Maybe also something to communicate with that standards team and the people who are creating the starter kit.

alimpens (approved):

PR #741, opened 2023-11-10
Add initial image component

VincentSmedinga:


packages/react/src/Image/Image.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, ImgHTMLAttributes } from 'react'
+
+export interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {}
+
+export const Image = forwardRef(({ className, ...restProps }: ImageProps, ref: ForwardedRef<HTMLImageElement>) => (
+  <img {...restProps} ref={ref} className={clsx('amsterdam-image', className)} />
```
dlnr, line 12: Are we going to have additional components for figure/figcaption or responsive images?

VincentSmedinga: Yes, I can see a approach like `<Figure caption="…"><Image /></Figure>` in a later change. This is just the MVP Image component.

As `Image` extends the HTML attributes of `img`, implementers can provide multiple resolutions of the image using the `srcset` and `sizes` attributes to let the browser download the appropriate source.

If various the sizes need different cutouts of the image (the ‘[art direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#art_direction)’ aspect), we may offer a React component to wrap the `picture` element later.

packages/css/src/image/README.md
```diff
@@ -0,0 +1,16 @@
...
+- Vergeet niet om een beschrijving van de afbeelding op te nemen in het `alt`-attribuut.
+  Dit zorgt ervoor dat gebruikers van schermlezers deze informatie ook tot zich kunnen nemen.
+  Daarnaast kan het helpen bij zoekmachineoptimalisatie.
+- Alleen voor decoratieve afbeeldingen is zo’n beschrijving niet nodig.
+  Denk aan afbeeldingen die weinig toevoegen aan de nabije tekst of afbeeldingen die louter bijdragen aan het ontwerp of de sfeer van de pagina (bron: [W3C Web Accessibility Initiative](https://www.w3.org/WAI/tutorials/images/decorative/)).
+- Zorg ervoor dat de afbeelding een beeldverhouding heeft die ondersteund wordt door het [Aspect Ratio](?path=/docs/layout-aspect-ratio--docs) component.
+
+## Relevante WCAG-eisen
+
+- [WCAG 1.1.1](https://www.w3.org/TR/WCAG22/#non-text-content): niet-tekstuele content heeft een tekstueel alternatief
```
alimpens, line 16: Do we want to add [1.4.5 and 1.4.9](https://www.w3.org/WAI/tutorials/images/#related) here as well? We should at least mention that you shouldn't use images of text if at all avoidable.

packages/css/src/image/README.md
```diff
@@ -0,0 +1,16 @@
+# Image
+
+Toont een afbeelding.
```
alimpens, line 3: Weren't we writing docs in English from now on?

packages/css/src/image/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Richtlijnen
+
+- Vergeet niet om een beschrijving van de afbeelding op te nemen in het `alt`-attribuut.
+  Dit zorgt ervoor dat gebruikers van schermlezers deze informatie ook tot zich kunnen nemen.
+  Daarnaast kan het helpen bij zoekmachineoptimalisatie.
+- Alleen voor decoratieve afbeeldingen is zo’n beschrijving niet nodig.
```
alimpens, line 10:
```suggestion
- Alleen voor decoratieve afbeeldingen is zo’n beschrijving niet nodig. Gebruik in dit geval `alt=""`.
```

packages/css/src/image/image.scss
```diff
@@ -0,0 +1,21 @@
...
+  height: auto; /* [1] */
+  max-width: 100%; /* [1] */
+  shape-margin: 1rem; /* [5] */
+  vertical-align: middle; /* [2] */
+}
+
+// [1] Allow for fluid image sizing while maintaining aspect ratio governed by width/height attributes
+// [2] Remove ‘phantom’ whitespace
+// [3] Italicise alt text to visually offset it from surrounding copy
+// [4] Set up backgrounds for optional LQI
```
alimpens, line 19: What is LQI?

packages/css/src/image/image.scss
```diff
@@ -0,0 +1,21 @@
...
+// [5] Set up margin for optional `shape-outside`
```
alimpens, line 20: Shouldn't you set this together with `shape-outside` when you use it? Feels a bit unnecessary to set it as default.

proprietary/tokens/src/components/amsterdam/image.tokens.json
```diff
@@ -0,0 +1,5 @@
+{
```
alimpens, line 1: We can get rid of this file, no?

packages/css/src/image/README.md
VincentSmedinga: Well, yes, but maybe not before we translated the rest all at once to prevent having two languages live at the same time?

packages/css/src/image/README.md
VincentSmedinga: Yes, fine, a was already wondering how we decided on the right combination of guidelines – check all of them individually whether they apply or find guidance on grouping like this. I’ve added those.

We might want to create some handy construct to reuse the content of each guidelines.

packages/css/src/image/image.scss
VincentSmedinga: A preloaded Low Quality Image I guess?

packages/css/src/image/image.scss
VincentSmedinga: Yes, maybe. I’ll remove it for now, together with the LQI one.

proprietary/tokens/src/components/amsterdam/image.tokens.json
VincentSmedinga: Sure.

packages/css/src/image/README.md
alimpens: I thought we wanted to do it incrementally? We could do it at once I guess, but that's quite a big job

packages/css/src/image/README.md
alimpens: Which guidelines we add is currently a bit ad hoc I think, but that's not necessarily a problem imo. As long as we add new guidelines in the right place(s) when we run into them.

packages/css/src/image/README.md
VincentSmedinga: I’m happy to block some time for a big bang.

alimpens (approved):

PR #751, opened 2023-11-13
Shorten ‘extra’ prefix to ‘x’

VincentSmedinga:
This prepares for a `2x-wide` aspect ratio rather than an `extra-extra-wide` one and is generally easier to read.

### DIFF
--- a/packages/react/src/Screen/Screen.tsx
+++ b/packages/react/src/Screen/Screen.tsx
@@ -6,7 +6,7 @@
 import clsx from 'clsx'
 import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
 
-type ScreenMaxWidth = 'wide' | 'extra-wide'
+type ScreenMaxWidth = 'wide' | 'x-wide'
 
 export interface ScreenProps extends PropsWithChildren<HTMLAttributes<HTMLDivElement>> {
   maxWidth?: ScreenMaxWidth
--- a/packages/css/src/aspect-ratio/README.md
+++ b/packages/css/src/aspect-ratio/README.md
@@ -4,13 +4,13 @@ Aspect Ratio (in het Nederlands: beeldverhouding) koppelt de hoogte van een elem
 
 Om de vormgeving consistent te houden gebruiken we een vaste set verhoudingen (breedte:hoogte):
 
-| _keyword_    | verhouding |
-| :----------- | ---------: |
-| `extra-wide` |       16:9 |
-| `wide`       |        5:4 |
-| `square`     |        1:1 |
-| `tall`       |        4:5 |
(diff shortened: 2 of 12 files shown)


packages/react/src/Screen/Screen.tsx
```diff
@@ -6,7 +6,7 @@
 import clsx from 'clsx'
 import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
 
-type ScreenMaxWidth = 'wide' | 'extra-wide'
+type ScreenMaxWidth = 'wide' | 'x-wide'
```
alimpens, line 9: Do we want to change the interface for Screen as well? We'll probably never need a `2x-wide` Screen, right?

VincentSmedinga: Maybe, maybe not – but either way, why maintain two approaches?

alimpens: True, consistency might be useful here.

Just had another brain fart: shouldn't this be `medium` and `wide`? 100rem is our default / main width, right? In most cases we use `medium` for the default size.

VincentSmedinga: Yes we do try to use `medium` for the default. Our set of breakpoints is why we haven’t succeeded for `Screen` (just yet):

- 0px – 575px = narrow
- 576px – 1087px = medium
- 1088px – 1599px = wide
- 1600px – 2112px = extra wide

So here, `100rem` is the maximum width for ‘wide’ screens and `132rem` the max for ‘extra wide’.

When reasoning about ‘desktop designs’, you’d expect a viewport width in the region of 1280, 1440, 1536 pixels. A good reason to call them ‘medium’. But we have named them ‘wide’ screens (based on the spreadsheet we designed the grid values with).

Naming them the ‘default’ sounds attractive, but then ‘tablet’ windows must become ‘narrow’ and ‘phone’ windows ‘extra narrow’. Would you prefer that?

It will add a bit of complexity to everyday speech – I think we chose the current approach so that we can refer to phones as having ‘narrow screens’ rather than ‘extra narrow’ ones – but I do agree this would be more precise and very much defensible.

alimpens: Hm, I'm not sure the props for Screen should be related to the grid and our breakpoints. Screen is just a container which sets a maximum width, and that maximum width is the default, or a wider variant.

Having said that: maybe this isn't the PR to discuss this, if you feel the prop has the right options, we can leave it for now and merge this PR

VincentSmedinga: Fair point. At the same time I think the pixel widths of the breakpoints and max widths are related. I’ve made a note to discuss.

alimpens (approved):

PR #755, opened 2023-11-14
Add Logo Component

dlnr:


packages/react/src/Logo/Logo.tsx
```diff
@@ -0,0 +1,41 @@
...
+import LogoStadsbankLening from '@amsterdam/design-system-assets/logo/stadsbank-lening.svg'
+import LogoVGA from '@amsterdam/design-system-assets/logo/vga-verzekeringen.svg'
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, HTMLAttributes } from 'react'
+
+export interface LogoProps extends HTMLAttributes<HTMLElement> {
+  variant?: 'amsterdam' | 'ggd-amsterdam' | 'stadsarchief' | 'stadsbank-lening' | 'vga-verzekeringen'
+}
+
+function getLogoVariant(variant: LogoProps['variant']) {
```
VincentSmedinga, line 18: I think we can simplify this. The switch in a function can be replaced with a TypeScript `Record` like so:

```ts
type LogoBrand = 'amsterdam' | 'ggd-amsterdam' | 'stadsarchief' | 'stadsbank-van-lening' | 'vga-verzekeringen'

export interface LogoProps extends HTMLAttributes<HTMLElement> {
  variant?: LogoBrand
}

const logoConfig: Record<LogoBrand, ReactNode> = {
  amsterdam: LogoGGD,
  'ggd-amsterdam': LogoStadsarchief,
  stadsarchief: LogoStadsbankLening,
  'stadsbank-van-lening': LogoVGA,
  'vga-verzekeringen': LogoAmsterdam,
}
```

(If we need more values later we can extend to `{ component: ReactNode, key: value }` objects.)

This allows for a smaller render function:

```ts
const LogoComponent = logoConfig[brand]

return (
  <span {...restProps} ref={ref} className={clsx('amsterdam-logo', className)}>
    {LogoComponent}
  </span>
)
```

This renders the `svg` inline rather than in an image, which I think we should do. We can store the description in the image itself through the `title` element.

[This Smashing article](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/) looks like a good resource on this. [MDN here](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_In_HTML_Introduction).

proprietary/tokens/src/components/amsterdam/logo.tokens.json
VincentSmedinga, line 1: We don’t need to keep this empty file.

VincentSmedinga (changes requested):

VincentSmedinga:
How could product teams not using React still get our logo svg in their application? Should that be a separate task?

proprietary/assets/logo/ggd-amsterdam.svg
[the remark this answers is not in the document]
VincentSmedinga: Oh, we use the `svgo` package already. Maybe @alimpens has a clever idea how to automate this.

packages/react/src/Logo/Logo.tsx
dlnr: This became kind of a struggle with typescript/rollup/react not importing svg as a ReactNode (or SVGElement) and required a lot of workarounds and extra plugins for the linters. So I changed it to the way Icon works and built components for the svg's (with svgomg optimized svgs this time) and it works without jumping hoops. 

I've used the Smashing article to add a11y to the svg, thanks, good article.

packages/css/src/logo/logo.scss
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-logo {
+  display: inline-block;
+
+  img {
```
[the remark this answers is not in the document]
dlnr: So add a classname to the svg (it changed from img to svg)? And why would we remove nesting?

VincentSmedinga (approved):

packages/css/src/logo/README.md
```diff
@@ -0,0 +1,35 @@
+# Logo
+
+Het logo van de gemeente bestaat uit een beeldmerk (3 Andreaskruisen) en een woordmerk (Gemeente Amsterdam).
+Het logo heeft een vaste kleur, grootte en positie en staat op al onze websites.
+
+## Specificaties
+
+Gebruik altijd het online logo.
```
alimpens, line 8: Aren't all the logo's we offer 'online'? Maybe this should be something like:

```suggestion
Gebruik altijd het algemene 'Amsterdam'-logo.
```

packages/css/src/logo/README.md
```diff
@@ -0,0 +1,35 @@
...
+Het Nederlandse logo voor Nederlandstalige websites en anderstalige websites behalve Engels.
+Het Engelstalige logo gebruiken we voor Engelstalige websites.
```
alimpens, line 10: Do we have the English logo? If so, we should add it to the list. If not, maybe we should mention that here?

packages/css/src/logo/README.md
```diff
@@ -0,0 +1,35 @@
...
+- GGD Amsterdam
+- Stadsarchief
+- Stadsbank van Lening
+- VGA Verzekeringen
+
+## Eisen voor alle logo´s
+
+- Het logo wordt altijd getoond.
+- Het logo is linksboven uitgelijnd in de header.
+- Het logo verwijst naar de homepage.
```
alimpens, line 28: I know this causes confusion sometimes. The logo should be part of a link to the homepage of the relevant website, not always amsterdam.nl, right?

packages/css/src/logo/README.md
```diff
@@ -0,0 +1,35 @@
...
+- Als de applicatie een formulier, applicatie of tool is zonder homepage, dan linkt het logo naar de pagina op Amsterdam.nl van waar naar het formulier, applicatie of tool wordt verwezen.
```
alimpens, line 29: This isn't necessarily amsterdam.nl, right?

```suggestion
- Als de applicatie een formulier, applicatie of tool is zonder homepage, dan linkt het logo naar de pagina van waar naar het formulier, applicatie of tool wordt verwezen.
```

packages/react/src/Logo/LogoAmsterdam.tsx
```diff
@@ -0,0 +1,13 @@
+import type { SVGProps } from 'react'
+
+const LogoAmsterdam = (props: SVGProps<SVGSVGElement>) => (
+  <svg className="amsterdam-logo__image" role="img" viewBox="0 0 115 40" xmlns="http://www.w3.org/2000/svg" {...props}>
+    <title>Gemeente Amsterdam</title>
```
alimpens, line 5: Same here, the logo shouldn't be read by screen readers

packages/react/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "es2020"
+    "target": "es2020",
+    "typeRoots": ["node_modules/@types", "src/**/*.d.ts"]
```
alimpens, line 23: Is this necessary in the current setup?

proprietary/assets/logo/amsterdam.svg
```diff
@@ -0,0 +1,6 @@
+<svg class="amsterdam-logo__image" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 115 40">
+  <title>Gemeente Amsterdam</title>
```
alimpens, line 2: Same comment as earlier here, the logo shouldn't be accessible for screen readers. Same for the other svgs

```suggestion
<svg class="amsterdam-logo__image" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 115 40">
```

packages/css/src/logo/logo.scss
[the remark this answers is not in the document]
VincentSmedinga: To loosen the coupling between our CSS and HTML and to make the intent more explicit. We don’t want all images within the logo selector to apply these styles, only the one that we appoint to be the logo image – even if this is a theoretical discussion.

packages/css/src/logo/README.md
VincentSmedinga: [We do](https://amsterdam.nl/stijlweb/basiselementen/logo-gemeente-amsterdam/#hba498e5c-2388-464f-b769-7622a5d32bd1), but as far as I can see it’s only available in a non-web format. I suggest we create a separate task for this.

packages/css/src/logo/README.md
VincentSmedinga: Not sure what you’re suggesting. To specify ‘the homepage’ further?

packages/css/src/logo/README.md
VincentSmedinga: Sounds fair.

packages/react/src/Logo/LogoAmsterdam.tsx
```diff
@@ -0,0 +1,13 @@
+import type { SVGProps } from 'react'
+
+const LogoAmsterdam = (props: SVGProps<SVGSVGElement>) => (
+  <svg className="amsterdam-logo__image" role="img" viewBox="0 0 115 40" xmlns="http://www.w3.org/2000/svg" {...props}>
```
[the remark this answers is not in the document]
VincentSmedinga: I haven’t researched this.

The [W3C WAI](https://www.w3.org/WAI/search/?q=logo) does add a role and a label:

```
<a lang="en" class="home w3c" href="https://www.w3.org/">
      <svg role="img" aria-label="W3C" viewBox="0 0 68 34" xmlns="http://www.w3.org/2000/svg">
```

packages/css/src/logo/README.md
VincentSmedinga: We can add that link for now?

packages/css/src/logo/README.md
alimpens: These docs are only relevant to the Design System, right? The non-'online' logo's are irrelevant to these docs imo.

packages/css/src/logo/README.md
alimpens: Something like this?

```suggestion
Het Engelstalige logo gebruiken we voor Engelstalige websites. Dit logo moet nog worden toegevoegd, [je kunt hier een plaatje van het logo vinden](https://amsterdam.nl/stijlweb/basiselementen/logo-gemeente-amsterdam/#hba498e5c-2388-464f-b769-7622a5d32bd1). 
```

packages/css/src/logo/README.md
alimpens: Yeah, something like this:

```suggestion
- Het logo verwijst naar de homepage van je website of applicatie.
```

packages/css/src/logo/logo.scss
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+.amsterdam-logo {
+  display: inline-block;
+}
+
+.amsterdam-logo__image {
+  height: 40px;
+  vertical-align: middle;
+  width: auto;
```
alimpens, line 13: Is this necessary? `width: auto` is the default, no?

packages/react/src/Logo/LogoAmsterdam.tsx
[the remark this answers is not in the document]
alimpens: NVDA reads 'clickable link graphic w3c' there, which is not great imo. [gov.uk](https://www.gov.uk/) hides their logo for screenreaders, NVDA reads 'clickable link Go to the GOV.UK homepage' there. We could also get rid of all these 3 attributes and allow the user to set them themselves, for extra flexibility. This is currently not possible though.

packages/react/src/Logo/Logo.tsx
```diff
@@ -0,0 +1,36 @@
...
+  'stadsbank-van-lening': LogoStadsbankVanLening,
+  'vga-verzekeringen': LogoVgaVerzekeringen,
+}
+
+export const Logo = forwardRef(
+  ({ brand = 'amsterdam', className, ...restProps }: LogoProps, ref: ForwardedRef<HTMLElement>) => {
+    const LogoComponent = logoConfig[brand]
+
+    return (
+      <span {...restProps} ref={ref} className={clsx('amsterdam-logo', className)}>
```
alimpens, line 29: Do we need the span here? We could just set all classes and props directly on the svg, no? This has the added benefit that the user can access the svgs attributes (`alt`, etc.), which they can't now

packages/react/src/Logo/Logo.tsx
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, HTMLAttributes } from 'react'
+import { LogoAmsterdam, LogoGgdAmsterdam, LogoStadsarchief, LogoStadsbankVanLening, LogoVgaVerzekeringen } from '.'
```
alimpens, line 8: This causes a circulair dependency, you're also importing from `index.ts` which exports this file.

packages/css/src/logo/README.md
VincentSmedinga: I agree. Let’s add this context then and remove the ambiguity of what the ‘online’ or ‘general Amsterdam’ logo actually is.

```suggestion
Het design system biedt het correcte logo voor online aan.
Er bestaan andere varianten van het logo, maar die mogen op websites en applicaties niet gebruikt worden.
```

packages/css/src/logo/README.md
VincentSmedinga: I suggest slightly different wording, and a sentence per line in Markdown.

```suggestion
Het Engelstalige logo gebruiken we voor Engelstalige websites.
Deze afbeelding wordt later toegevoegd aan het design system.
In de tussentijd kun je het downloaden vanaf [Stijlweb](https://amsterdam.nl/stijlweb/basiselementen/logo-gemeente-amsterdam/#hba498e5c-2388-464f-b769-7622a5d32bd1).
```

packages/react/src/Logo/LogoAmsterdam.tsx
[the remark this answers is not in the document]
dlnr: This is not really helping guys.. Vincents option: screen reader reads 'Image: "Gemeente Amsterdam"', Arams option doesn't read anything. I would argue that the accessible option is that if you have a link around the image (which should always be the case) you do want to have it read something. [The article](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/) explains that you can also use `aria-labelledby` to explain the image is labelled by a visually hidden site title in the link for example. A `<title>` tag in the svg bubbles up to the svg (with role image) but this would make the link less accessible and you do not want the site title twice.

packages/css/src/logo/README.md
alimpens: Small tweak:

```suggestion
Het design system biedt het correcte logo voor online gebruik aan.
Er bestaan andere varianten van het logo, maar die mogen op websites en in applicaties niet gebruikt worden.
```

packages/css/src/logo/README.md
alimpens: Small tweak:

```suggestion
Het Engelstalige logo gebruiken we voor Engelstalige websites.
Dit logo wordt later toegevoegd aan het design system.
In de tussentijd kun je een afbeelding van dit logo downloaden vanaf [Stijlweb](https://amsterdam.nl/stijlweb/basiselementen/logo-gemeente-amsterdam/#hba498e5c-2388-464f-b769-7622a5d32bd1).
```

packages/react/src/Logo/LogoAmsterdam.tsx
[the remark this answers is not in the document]
VincentSmedinga: Hah I understand. I should have been clearer

1. The code snippet doesn’t reflect my opinion, just an example of a site that probably knows what they’re doing on this subject.
2. My implicit question was ‘Can you please research this and implement your best option’.

I would expect a screen reader user to want to hear ‘Homepage, link’ or ‘Link, homepage’ when focusing the link to the logo. Can we make that happen?

packages/react/tsconfig.json
dlnr: No, it was forgotten after testing with loading svg files.

packages/react/src/Logo/LogoAmsterdam.tsx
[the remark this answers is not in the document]
dlnr: Yes, it's up to the developer to wrap the logo in a link. We've decided the logo should just be the image.

packages/css/src/logo/logo.scss
dlnr: Yes it's the default, I left it there because it was previously a pixel value which prevented layout shift. We should probably go for a width and height attribute in the svg instead of viewBox.

packages/react/src/Logo/Logo.tsx
dlnr: Moved to subfolder brand

packages/react/src/Logo/Logo.tsx
dlnr: Do you mean having the svg tag here and putting the paths in `LogoBrand`? It could be done but I have a bunch of complaints from typescript concerning ref and brand as props. Can you make a suggestion? 

It was originaly a span to hold an `` tag and have it scale with the page menu and match the line-height. Something that's also easier keeping the span?

dlnr: Ok, it has been changed. Only thing I wasn't able to get going was the ForwardRef unit test. So I skipped it.

proprietary/tokens/src/components/amsterdam/logo.tokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "amsterdam": {
+    "logo": {
+      "height": { "value": "40px" },
```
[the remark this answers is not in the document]
dlnr: That's correct, the header will look better when people have enlarged fonts and such.

pnpm-lock.yaml
```diff
@@ -510,7 +510,6 @@ packages:
 
   /@aw-web-design/x-default-browser@1.4.126:
```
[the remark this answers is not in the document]
dlnr: I honestly have no idea, pnpm keeps suprising me. This I propably the result of a merge with another branch?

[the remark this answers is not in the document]
dlnr: I ran fixlockfile and did not see any changes to this line, branch is up to date with develop and I have not installed anything in this branch..

packages/react/src/Logo/Logo.tsx
alimpens: I added a commit to pass the refs to the LogoComponents

pnpm-lock.yaml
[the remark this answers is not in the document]
alimpens: I added a commit to fix the lockfile

packages/css/src/logo/logo.scss
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-logo {
+  // display: inline-block;
```
alimpens, line 7: Let's get rid of the unused css

packages/css/src/logo/logo.scss
```diff
@@ -0,0 +1,25 @@
...
+.amsterdam-logo {
+  display: inline-block;
+}
+
+.amsterdam-logo__image {
+  height: var(--amsterdam-logo-height);
+  vertical-align: middle;
+}
+
+.amsterdam-logo__emblem path {
```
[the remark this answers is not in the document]
dlnr: Gave the paths a class and remove groups.

alimpens (approved):

PR #758, opened 2023-11-16
Add compact grid

VincentSmedinga:
This adds:
1. a compact version of the grid through a `density` property
2. options to add vertical whitespace above and below a grid through three ~`margin*`~ `padding-*` properties
3. ~a property for `Screen` to left-align it.~

These changes were made in the context of a prototype. Although we prefer pull requests to focus on one feature, in this case I think it would not be efficient to split them up after the fact.

The sizing of the compact grid has been presented to teams currently designing applications. They are evaluating but I don’t expect any changes.

Implementing generic utility classes for spacing seemed a bridge too far. They’re only available to the grid. Consequently, the class names are many and verbose for now.

A next step is to implement a concept of ‘theming’, as we want all grid instances to have the same density and to have typography react to the same setting. After that, we can extend the `margin` props to generic utility classes.

I’ve removed the temporary inner vertical whitespace in `Highlight` and `Footer`


packages/css/src/grid/grid.scss
```diff
@@ -20,3 +19,80 @@
...
   }
 }
+
+.amsterdam-grid--density-high {
+  gap: var(--amsterdam-grid-density-high-gap);
+  padding-inline: var(--amsterdam-grid-density-high-padding-inline);
+}
+
+.amsterdam-grid--density-low--margin-bottom--small {
+  padding-block-end: calc(var(--amsterdam-grid-density-low-gap) / 2);
```
alimpens: We're setting the `padding`, but the class is called `*-margin-*`, that's a bit odd

packages/css/src/screen/screen.scss
```diff
@@ -9,7 +9,9 @@
 
 .amsterdam-screen {
   @include reset;
+}
 
+.amsterdam-screen--center {
```
alimpens, line 14: Screen should always be center aligned, no? Even for full width applications

packages/react/src/Grid/Grid.tsx
```diff
@@ -22,17 +22,77 @@ export type GridColumnNumbers = {
   wide: GridColumnNumber
 }
 
-interface GridComponent
-  extends ForwardRefExoticComponent<PropsWithChildren<HTMLAttributes<HTMLDivElement>> & RefAttributes<HTMLDivElement>> {
+type GridDensity = 'low' | 'high'
+
+type GridMarginSize = 'small' | 'medium' | 'large'
```
alimpens, line 27: We're setting padding, shouldn't the name reflect this?

packages/react/src/Grid/Grid.tsx
```diff
@@ -22,17 +22,77 @@ export type GridColumnNumbers = {
...
+
+type GridMarginVerticalProp = {
+  marginBottom?: never
+  marginTop?: never
+  /** The amount of vertical whitespace above and below the grid. */
```
alimpens, line 32: I'm not sure, but I think Intellisense only picks up JSDoc type comments in this format:

```suggestion
  /** 
  * The amount of vertical whitespace above and below the grid. 
  */
```

storybook/storybook-docs/src/grid.stories.mdx
```diff
@@ -21,25 +21,45 @@ Op dat punt worden en blijven het er 12.
...
-Het grid staat horizontaal gecentreerd op de pagina.
+In websites staat het grid horizontaal gecentreerd op de pagina.
 Aan beide zijden reserveert het grid marges om afstand te houden tot de randen van het venster.
 Deze ruimte is 1½ keer zo breed als die tussen de kolommen.
 
 Sommige elementen mogen over deze marges gepositioneerd worden, zoals de Page Footer en een schermvullende afbeelding.
 Die zijn dan dus breder dan de rest van de content, maar worden wel beperkt door een maximale breedte.
 
+Voor applicaties is de marge naast het grid gelijk aan de witruimtes erbinnen.
+Ook is het grid in dit geval links uitgelijnd op het scherm.
```
alimpens, line 60: Is this different alignment necessary?

storybook/storybook-react/src/Grid/Grid.docs.mdx
```diff
@@ -29,3 +43,7 @@ Elke cel biedt ruimte voor een afbeelding.
 <Canvas of={GridStories.Cells} />
 
 Voor meer voorbeelden raadpleeg je [Grid Cell](?path=/docs/react_layout-grid-cell--docs) zelf.
+
+## Richtlijnen
+
+- Gebruik de `marginVertical` prop niet tegelijk met de `marginTop` of `marginBottom` props.
```
alimpens, line 60: Is this necessary to document? Typescript already prevents you from doing this, no?

storybook/storybook-react/src/shared/argTypes.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+export const marginArgType = {
+  control: {
+    type: 'inline-radio',
```
alimpens, line 8: These horizontal radio buttons are a bit fiddly imo, any reason to choose them over vertical radio's?

storybook/storybook-react/src/shared/argTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+    direction: 'horizontal',
+    labels: { undefined: 'n/a', small: 'small', medium: 'medium', large: 'large' },
```
alimpens, line 10:
```suggestion
    labels: { undefined: 'none', small: 'small', medium: 'medium', large: 'large' },
```

packages/css/src/screen/screen.scss
dlnr: The internal apps are aligned left if I'm not mistaken.

packages/css/src/grid/grid.scss
```diff
@@ -6,10 +6,12 @@
 @import "../../utils/breakpoint";
 
 .amsterdam-grid {
+  --amsterdam-grid-gap: var(--amsterdam-grid-density-low-gap);
```
dlnr, line 9: Very cool that you're scoping variables but why would you when it doesn't change in the scope of this class? Since you do not nest the density high modifier? And also, is it not the _NLDS way_ to set this via tokens?

packages/css/src/grid/grid.scss
```diff
@@ -20,3 +22,44 @@
...
   }
 }
+
+.amsterdam-grid--density-high {
+  --amsterdam-grid-gap: var(--amsterdam-grid-density-high-gap);
+  --amsterdam-grid-padding-inline: var(--amsterdam-grid-density-high-padding-inline);
+}
+
+.amsterdam-grid--margin-bottom--small {
+  padding-block-end: calc(var(--amsterdam-grid-gap) / 2);
```
dlnr, line 32: I believe @alimpens already said this but the class is "margin" and the rule is "padding"?

packages/react/src/Grid/Grid.tsx
```diff
@@ -22,17 +22,76 @@ export type GridColumnNumbers = {
   wide: GridColumnNumber
 }
 
-interface GridComponent
-  extends ForwardRefExoticComponent<PropsWithChildren<HTMLAttributes<HTMLDivElement>> & RefAttributes<HTMLDivElement>> {
+type GridDensity = 'low' | 'high'
```
dlnr, line 25: I'm not convinced the "internal app theme" should be a property on all components used. It should be a theme (provider) or different set of tokens. Perhaps we can investigate further?

storybook/storybook-docs/src/grid.stories.mdx
dlnr: Yes :)

packages/react/src/Grid/Grid.tsx
```diff
@@ -22,17 +22,76 @@ export type GridColumnNumbers = {
...
+}
+
+interface GridComponent extends ForwardRefExoticComponent<GridProps & RefAttributes<HTMLDivElement>> {
   Cell: typeof GridCell
 }
 
 export const Grid = forwardRef(
   (
-    { children, className, ...restProps }: PropsWithChildren<HTMLAttributes<HTMLDivElement>>,
+    { children, className, density = 'low', marginBottom, marginTop, marginVertical, ...restProps }: GridProps,
```
dlnr, line 82: If the default density is `low` shouldn't you rather use a boolean called `condesed` so you only have one modifier `--condesed` instead of having to define `density-low` and `density-high` everywhere?

storybook/storybook-react/src/shared/argTypes.ts
VincentSmedinga: I like that their compactness allows more props in view, and that props with the same options appear related directly.

As we current only use vertical radios everywhere I’ll change these to match. Let’s revisit later.

storybook/storybook-react/src/shared/argTypes.ts
VincentSmedinga: Yes, better. Done.

packages/react/src/Grid/Grid.tsx
alimpens: Hm, JSDoc annotations don't show very reliably on my machine anyway... Let's keep it single line, if it causes issues we can always change it later

storybook/storybook-react/src/Grid/Grid.docs.mdx
alimpens: But this is React specific documentation? Feels like it's not necessary to add it here. Of course, it doesn't really hurt either, so we can keep it if you feel it adds something.

packages/react/src/Grid/Grid.tsx
alimpens: Hm, I'm not really convinced. `margin` means something in CSS-land, and it's not what we're setting here. It's confusing for developers I think to have a prop called `margin*` which doesn't set a CSS `margin`.

If `padding` is what we're setting, why not just call it 'padding' (for now at least)?

`paddingBottom`, `paddingVertical` etc. is a pretty good description of what it does, imo

VincentSmedinga: Okay.

storybook/storybook-react/src/Grid/Grid.docs.mdx
VincentSmedinga: Allright, removed both.

packages/css/src/grid/grid.scss
VincentSmedinga: Changed it to padding now.

alimpens (approved):

alimpens (approved):

PR #807, opened 2023-11-24
feat: Add Header Component

dlnr:
The header has a title that appears in the center on desktop sizes, this presents some challenges. The solution I've used depends on the developer not using too much page menu items or a long subsite title. So basically if a subsite has a long title they should not use that many menu items and if you have a lot of menu items you should consider not adding a site title.

When mega menu is developed we should consider separating the menu button so we can remove the hack to hide menu items on mobile.


packages/react/src/Header/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, HTMLAttributes, ReactNode } from 'react'
+import { Heading } from '../Heading'
+import { Logo, LogoBrand } from '../Logo'
+import { VisuallyHidden } from '../VisuallyHidden'
+
+export interface HeaderProps extends HTMLAttributes<HTMLElement> {
+  logoBrand?: LogoBrand
+  logoLink?: string
```
alimpens, line 14: `logoLink` is required, but we could set the default to `"/"`

packages/react/src/Header/Header.tsx
```diff
@@ -0,0 +1,57 @@
...
+  logoLinkTitle?: string
+  title?: string
+  menu?: ReactNode
+}
+
+export const Header = forwardRef(
+  (
+    { className, logoBrand, logoLink, logoLinkTitle, title, menu, ...restProps }: HeaderProps,
```
alimpens, line 22: Let's set defaults here, `logoBrand="amsterdam"`, `logoLinkTitle="Ga naar homepage"`. That way we could get rid of a bunch of the checks below

VincentSmedinga: ‘Ga naar *de* homepage’?

dlnr: Why would you disallow people from having a header without a link? There can be cases like a landingpage (without any navigation) where the header is a static thing right?

alimpens: It's described [here](https://www.amsterdam.nl/digitalservices/design-system/atomen-moleculen/logo-and-favicon/#hab-7cac11-e64e-44cb-acb7-23dbfc88b972), a logo in a header links to the apps homepage. If the app doesn't have a homepage, it links to the page of origin on amsterdam.nl

dlnr: Ok

packages/react/src/Header/Header.tsx
dlnr: If you set it to a default there is no longer a need to require it right? Or the default wil never "kick in"?

storybook/storybook-react/src/Header/Header.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  args: {
+    title: 'Aan de Amsterdamse Grachten',
+  },
+}
+
+export const WithLogoVariant: Story = {
+  args: {
+    logoBrand: 'ggd-amsterdam',
+    logoLink: 'https://www.ggd.amsterdam.nl/',
+    logoLinkTitle: 'Naar de website van de GGD Amsterdam',
```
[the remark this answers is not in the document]
dlnr: I think we've discussed that we do need to add the site title here for screen readers because the logo is "silent" and they do not know on what site they are.

packages/react/src/Header/Header.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+describe('Header', () => {
+  const defaultProps: HeaderProps = {
+    logoLink: '/',
+  }
+
+  it('renders', () => {
+    const { container } = render(<Header {...defaultProps} />)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 14: [Testing library recommends getting elements by role if that's possible](https://testing-library.com/docs/queries/about/#priority). So this should be something like `screen.getByRole('banner')`. Same for the link and heading

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+    min-width: 0;
+    order: 2;
+
+    .amsterdam-heading {
+      display: block;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+      width: 100%;
+    }
```
VincentSmedinga, line 50: Components are responsible for their own appearance. Other components must not influence them. If we want to keep this appearance and behaviour, it must either become a prop of Heading, or changed to local HTML and CSS.

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+.amsterdam-header__column--menu {
+  flex: 2 1 calc(100% - 160px);
```
VincentSmedinga, line 55: What’s the 160px for?

packages/react/src/Header/Header.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export interface HeaderProps extends HTMLAttributes<HTMLElement> {
+  logoBrand?: LogoBrand
+  logoLink?: string
+  logoLinkTitle?: string
+  title?: string
+  menu?: ReactNode
+}
+
+const SubsiteTitle = ({ children }: { children: ReactNode }) => (
```
VincentSmedinga, line 21: It doesn’t seem necessary to have a separate component for this.

packages/react/src/PageMenu/PageMenu.tsx
```diff
@@ -45,16 +45,18 @@ export const PageMenu = forwardRef(
 
 export interface PageMenuLinkProps extends PropsWithChildren<AnchorHTMLAttributes<HTMLAnchorElement>> {
   icon?: Function
+  showOnMobile?: boolean
```
VincentSmedinga, line 48: We mustn’t use the word ‘mobile’ and think mobile-first. Maybe something like `hideInNarrowGrid`?

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+.amsterdam-header {
+  align-items: center;
+  background-color: var(--amsterdam-header-background-color);
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  padding-block: 0.88rem;
+  row-gap: 1.5rem;
+
+  @media screen and (max-width: $amsterdam-breakpoint-medium) {
```
VincentSmedinga, line 17: We could make it easy for ourselves and display the page menu links from on the wide grid only?

VincentSmedinga (changes requested):
Yeah, some difficult bits in this design – the struggle shows. Let’s try to simplify. Use flex shorthands, hide the page menu links earlier, just let things wrap if they don’t fit, remove some of the CSS that has no or little effect. And revisit later.

packages/react/src/Header/Header.test.tsx
dlnr: Ah right, didn't change it from the plop template.

packages/css/src/components/page-menu/page-menu.scss
```diff
@@ -22,15 +22,25 @@
...
   row-gap: var(--amsterdam-page-menu-row-gap);
 
   @include reset-list;
+
+  &__item {
+    &--hide {
+      display: none;
+      visibility: hidden;
+    }
+  }
```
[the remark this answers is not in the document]
dlnr: Yes

packages/react/src/PageMenu/PageMenu.tsx
dlnr: Since we've decided the menu button should be placed in the megamenu, and this solution is only to hide everything but the menu button i'll go ahead and remove this and the menu button from page menu.

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+
+@import "../../common/breakpoint";
+
+.amsterdam-header {
+  align-items: center;
+  background-color: var(--amsterdam-header-background-color);
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  padding-block: 0.88rem;
```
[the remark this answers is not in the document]
dlnr: The desktop padding in Figma is 14px, must be a mistake. I'll change it and also put it in the tokens.

packages/css/src/components/header/header.scss
dlnr: Maximum width of the logo, the column would try to occupy that space instead of `auto` where it would divide the available space with the title.

dlnr: Mistake, it was the width the logo would like to have but this is the menu..

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
+    column-gap: var(--amsterdam-header-column-gap);
+    flex-wrap: nowrap;
+  }
+}
+
+.amsterdam-header__column--logo {
+  flex: 0 0 auto;
```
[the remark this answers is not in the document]
VincentSmedinga: I haven’t tested it, but the logo is in a flex row so I can imagine it will shrink horizontally and distort due to the fixed height.

packages/css/src/components/header/header.scss
dlnr: Or extend the className of heading?

packages/css/src/components/header/header.scss
dlnr: Only if there is a subsite title?

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+}
+
+.amsterdam-header__column--title {
+  flex: 1 1 100%;
+
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
+    flex: 1 1 auto;
+    min-width: 0;
+    order: 2;
```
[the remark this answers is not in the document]
dlnr: Yes the title comes second at this breakpoint. I've opted to use order instead of some absolutely positioned solution.

packages/react/src/Header/Header.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    const component = screen.getByRole('banner')
+
+    expect(ref.current).toBe(component)
+  })
+
+  it('renders with a logo link', () => {
+    const { container } = render(<Header {...defaultProps} logoLink="/home" />)
+
+    const logoLink = container.querySelector('.amsterdam-header__logo')
```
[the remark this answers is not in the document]
dlnr: Fixed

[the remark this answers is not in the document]
dlnr: Fixed

packages/css/src/components/page-menu/page-menu.scss
```diff
@@ -40,6 +50,7 @@
 @mixin page-menu-item {
   color: var(--amsterdam-page-menu-item-color);
   display: inline-flex;
+  flex: 1;
```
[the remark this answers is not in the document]
dlnr: Nope, don't remember how this got here, probably testing things to get the page menu to work in the header.

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
+    column-gap: var(--amsterdam-header-column-gap);
+    flex-wrap: nowrap;
+  }
+}
+
+.amsterdam-header__column--logo {
+  flex: 0 0 auto;
+}
```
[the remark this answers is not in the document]
dlnr: The whitespace is actualy caused by the line-height of the subsite title.

packages/react/src/Header/Header.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    expect(logoLink).toHaveAttribute('href', '/home')
+  })
+
+  it('renders with a logo link title', () => {
+    const { container } = render(<Header {...defaultProps} logoLinkTitle="Go to homepage" />)
+
+    const logoLinkTitle = container.querySelector('.amsterdam-visually-hidden')
```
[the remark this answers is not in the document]
dlnr: Fixed

packages/css/src/components/header/header.scss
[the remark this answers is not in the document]
dlnr: This was all done to get the header to behave the way I want to, if you feel your assumptions about shrinking or growing are correct then you're welcome to commit.

packages/css/src/components/header/header.scss
```diff
@@ -0,0 +1,62 @@
...
+
```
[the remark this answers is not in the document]
dlnr: Done

VincentSmedinga (approved):

PR #811, opened 2023-11-24
This breaks the (old) CRA builds

dlnr:
Hee Chan reported that this grid CSS breaks the CRA build. This was previously the case in typography.

Since the old CRA (v4) uses a buggy safe css parser, it does not accept "unsafe" calculations. It needs to be used with `calc()`. Same issue with the Typography.

### DIFF
--- a/proprietary/tokens/src/components/amsterdam/grid.tokens.json
+++ b/proprietary/tokens/src/components/amsterdam/grid.tokens.json
@@ -6,21 +6,21 @@
       },
       "density-low": {
         "gap": {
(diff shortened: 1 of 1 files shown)


hcvdhaar:
For me it is not possible yet to update to a newer version of CRA or replace it. Hope you can take this in consideration for the upcoming releases.

VincentSmedinga (approved):

alimpens (approved):

VincentSmedinga:
Sure. Silly I forgot. Maybe there’s a linter rule that we can add for it.

PR #812, opened 2023-11-25
fix: Add vertical whitespace to grids in Footer and Highlight examples

VincentSmedinga:
Whitespace was missing in the examples. Plus one improved grid cell configuration.


storybook/storybook-react/src/Footer/Footer.stories.tsx
```diff
@@ -81,7 +81,7 @@ export const Default: Story = {
               </UnorderedList>
             </div>
           </Grid.Cell>
-          <Grid.Cell span={3} start={{ narrow: 1, medium: 1, wide: 9 }}>
+          <Grid.Cell span={{ narrow: 4, medium: 8, wide: 4 }} start={{ narrow: 1, medium: 1, wide: 9 }}>
```
alimpens, line 84: I don't think this should change, the third column should have the same dimensions as the other 2

VincentSmedinga: Alright, reverted.

alimpens (approved):

PR #813, opened 2023-11-25
refactor: Use highlight in footer

VincentSmedinga:
The background colour of `Footer.Top` can be applied through a blue `Highlight`.


alimpens (approved):

VincentSmedinga:
> Maybe not a decision we have to make now: can't we get rid of the Footer component entirely? `Footer` just renders a HTML `footer`, `Footer.Bottom` renders a `div`, and `Footer.Top` is a `Highlight`. Maybe we can just turn this into a recipe or something, instead of a component?

While I agree that the footer for our websites can technically be implemented through composition and configuration, I also think there’s value in having a semantic component that people don’t have to think twice about. Even if it’s not doing that much, having a thing called footer can be reassuring that they use the right thing in the right way. 

Additionally, the footer has specific usage guidelines (prescribing what goes in which column) and its documentation is easier found if it’s an independent component.

PR #814, opened 2023-11-27
feat: Add common link tokens and use them in all link-like components

alimpens:
I've done 2 things in this PR:

- I've moved our css utils to `src/common`, and all components are moved to `src/components`. This caused a lot of changes, sorry!
- I've added `link-types` to our common tokens layer, which is now used in all our components that look like links. If anyone has a better name than `link-types`, let me know!


proprietary/tokens/src/common/amsterdam/link-type.tokens.json
```diff
@@ -0,0 +1,47 @@
...
+        "hover": {
+          "text-decoration-line": { "value": "underline" }
+        }
+      },
+      "in-list": {
+        "text-decoration-line": { "value": "none" },
+        "text-decoration-thickness": { "value": "2px" },
+        "text-underline-offset": { "value": "6px" },
+        "hover": {
+          "text-decoration-line": { "value": "underline" }
```
VincentSmedinga, line 42: The underline should be 3px on hover here as well.

```suggestion
          "text-decoration-line": { "value": "underline" },
          "text-decoration-thickness": { "value": "3px" }
```

We could even lift that token up a level as it’s the same for all link types?

proprietary/tokens/src/common/amsterdam/link-type.tokens.json
```diff
@@ -0,0 +1,47 @@
...
+      "regular": {
+        "text-decoration-line": { "value": "underline" },
+        "text-decoration-thickness": { "value": "2px" },
+        "text-underline-offset": { "value": "6px" },
+        "hover": {
+          "text-decoration-thickness": { "value": "3px" },
+          "text-underline-offset": { "value": "5px" }
+        }
+      },
+      "apparent": {
```
VincentSmedinga, line 29: I understand the quest for a better term than ‘obvious’. That one didn’t stick with me either. We should make time for it now.

The problem with both of these is that they go like ‘that this is a link is so obvious / apparent that it doesn't need an initial underline’. But the visual effect of this link type is the opposite: making it look a bit _less_ obvious than a link. And I think the characteristic of the link visualisation is what we should describe, not the context it’s in.

I think I would propose ‘muted’. It’s a term I come across regularly in design systems – albeit with colours, but we can use it there as well to prevent having two similar terms. It’s short, pretty clear, easy to pronounce, not too particular.

Plenty of [alternatives](https://www.thesaurus.com/browse/muted) though: soft, subtle, plain, dimmed, light, slight, discreet, subdued, silent, quiet, mellow, toned-down.

What would you say?

proprietary/tokens/src/common/amsterdam/link-type.tokens.json
```diff
@@ -0,0 +1,47 @@
+{
+  "amsterdam": {
+    "link-type": {
```
VincentSmedinga, line 3: A bit long maybe, but `link-appearance` might be perfect.

If not, [synonyms here](https://www.thesaurus.com/browse/appearance). Presentation?

proprietary/tokens/src/components/amsterdam/page-menu.tokens.json
```diff
@@ -4,21 +4,24 @@
...
       "item": {
-        "color": { "value": "{amsterdam.color.primary-black}" },
+        "color": { "value": "{amsterdam.link-type.color}" },
         "font-family": { "value": "{amsterdam.typography.font-family}" },
         "font-weight": { "value": "{amsterdam.typography.font-weight.normal}" },
         "gap": { "value": "0.5rem" },
         "line-height": { "value": "{amsterdam.typography.text-level.7.line-height}" },
-        "text-decoration": { "value": "none" },
+        "outline-offset": { "value": "{amsterdam.focus.outline-offset}" },
+        "text-decoration-line": { "value": "{amsterdam.link-type.in-list.text-decoration-line}" },
```
VincentSmedinga, line 13: Page menu uses the ‘apparent’ link type I think?

storybook/storybook-react/src/Link/Link.stories.tsx
```diff
@@ -68,23 +67,20 @@ const meta = {
   args: {
     children: 'Linktekst',
```
VincentSmedinga, line 68: This text has no descenders, unfortunately – they would help indicate the underline behaviour.

```suggestion
    children: 'Klik op mij',
```

We’ll need to think of something English in a few days or so.

alimpens:
> Which focus playlist did you use to get this done? :)

I think it was mostly some Aphex Twin, which I wouldn't recommend for a focus playlist 😅

proprietary/tokens/src/common/amsterdam/link-type.tokens.json
alimpens: See one of my previous comments, in Figma there's a difference between 'in-list' and 'apparent'. I think we could merge them, I'd opt for 2px thickness everywhere though. 3px is a bit heavy.

Edit: I just checked how it looks with a 6px offset, 3px thickness looks fine then. So I don't really have a preference

storybook/storybook-react/src/Link/Link.stories.tsx
alimpens: Is this still necessary with the 6px offset?

proprietary/tokens/src/common/amsterdam/link-type.tokens.json
VincentSmedinga: It’s 2px initial and 3px on hover everywhere, right?

storybook/storybook-react/src/Link/Link.stories.tsx
VincentSmedinga: I’m still not fond of ‘Linktekst’, let’s revisit when translating.

proprietary/tokens/src/components/amsterdam/pagination.tokens.json
```diff
@@ -16,15 +12,17 @@
         "font-size": { "value": "{amsterdam.typography.text-level.6.wide.font-size}" }
       },
       "button": {
+        "outline-offset": { "value": "{amsterdam.focus.outline-offset}" },
+        "text-decoration-line": { "value": "{amsterdam.link-type.apparent.text-decoration-line}" },
```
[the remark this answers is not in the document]
VincentSmedinga: I thought we ended up with only regular and (now) subtle link appearances.

What difference has remained between subtle and in-list? If it is font-size, that’s not part of list-appearance?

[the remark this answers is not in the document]
VincentSmedinga: Great!

VincentSmedinga (approved):

PR #815, opened 2023-11-27
fix: Merge grid cell files into grid files and directories

VincentSmedinga:
As `Grid.Cell` is a subcomponent of `Grid`, it makes sense to implement it in context rather than distinctly. We follow the same approach with things like `Breadcrumb.Item` and `Footer.Top`, even though these subcomponents have significantly less responsibilities than grid cell.


storybook/storybook-react/src/Grid/Grid.Cell.stories.tsx
VincentSmedinga, line 1: @alimpens or @dlnr Can either of you figure out how to best display the grid cell stories on the grid page?

Doesn't seem [too complicated](https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components) but I’m not getting my head around it.

An important requirement is that people can still experiment with grid cell args [like we currently allow](https://amsterdam.github.io/design-system/?path=/story/react_layout-grid-cell--start-position&args=span:4).

We should remove the grid cell entry from the side bar if we keep the stories file.

Bonus points for making the `{ narrow, medium, wide } object editable in the args 
table as well!

dlnr: I've changed Grid docs and argTypes to show controls in Grid Cell. And there are now grid cell stories in the grid doc. However, for controls I think you'll need to go to the grid cell stories where I've added argTypes to play with cols and span.

VincentSmedinga: If the only way we can use Storybook controls for the Grid Cell stories is through their own entry in the docs, we should maybe keep the stories and docs separated while merging the rest.

Or is there another way to have Grid Cell stories in the Grid docs?

storybook/storybook-react/src/Grid/Grid.Cell.stories.tsx
```diff
@@ -10,6 +10,20 @@ import { Meta, StoryObj } from '@storybook/react'
 const meta = {
   title: 'Layout/Grid Cell',
```
dlnr, line 11: If you remove this line it will be hidden from the navigation tree and the stories described in grid docs will have to suffice.

storybook/storybook-react/src/Grid/Grid.Cell.stories.tsx
dlnr: This would only work if both do not use `children`, if you would rewrite grid so that it uses something like `cells` it could work.

storybook/storybook-react/src/Grid/Grid.Cell.stories.tsx
VincentSmedinga: Have you seen that working? Doesn’t change anything for me.

As indicated here, I’ve tried the following in both the docs and the react instance, with `tags: ['hidden']` in the grid cell `meta`, unfortunately to no effect either:

```js
addons.setConfig({
  theme: create(themeConfig),
  sidebar: {
    filters: {
      hidden: (item) => {
        console.log(item.title, item.tags)
        return !item.tags.includes('hidden')
      },
    },
  },
})
```

It‘s only logging the docs entries to the console – so might have something to do with the composition. @alimpens does that ring a bell for you?

I’ve just pushed it as a WIP commit.

VincentSmedinga: That is the goal yes. We discussed we wanted to remove the separation between `Grid` and `Grid.Cell` as much as possible because they are inseparable.

The problem I am trying to overcome is that both have different props, so they don’t share the same `meta` for example, and Storybook doesn’t seem to work well with one `*.stories.tsx` file defining two sets of meta’s, stories etc.

Removing `.stories` from the file name doesn’t work either. It results in the error

> Invalid value passed to the 'of' prop. The value was resolved to a 'component or unknown' type but the only types for this block are: story. (…) Did you pass a story, CSF file or meta to the 'of' prop that is not indexed, ie. is not targeted by the 'stories' globs in the main configuration?”

alimpens: Ah, you're right, that doesn't work. I think you're trying to do the same thing I did for [Search Field](https://github.com/Amsterdam/design-system/pull/892/commits/afdfbfb08f5d4e53b15d21cd2d89b95a6dc9f16a), did you look at that?

VincentSmedinga: I did, and in the end the render function helped – but I think the context is a bit different so my approach is as well. Still, it worked.

I’d like to limit the controls to the ones that the specific story pertains to. Is that worthwhile?

dlnr (approved):

packages/react/src/Grid/Grid.tsx
```diff
@@ -99,6 +99,41 @@ export const Grid = forwardRef(
   ),
 ) as GridComponent
 
+type GridCellFullWidthProp = {
```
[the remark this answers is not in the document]
dlnr: I agree, should be a seperate story.

dlnr (approved):
Very cool solution getting the grid cell stories in!

packages/react/src/Grid/Grid.tsx
```diff
@@ -99,6 +99,41 @@ export const Grid = forwardRef(
...
+  /** The amount of grid columns the cell spans. */
+  span?: GridColumnNumber | GridColumnNumbers
+  /** The index of the grid column the cell starts at. */
+  start?: GridColumnNumber | GridColumnNumbers
+}
+
+export type GridCellProps = (GridCellFullWidthProp | GridCellColumnProps) &
+  PropsWithChildren<HTMLAttributes<HTMLDivElement>>
+
+const GridCell = forwardRef(
```
alimpens, line 120: I thought we wanted to keep subcomponents in seperate files? Similar to `Accordion`

alimpens, line 120: Can you add tests for the `Grid` and `GridCell` components? If that's out of scope for this PR, we could do it in a seperate ticket

storybook/storybook-docs/config/manager.js
```diff
@@ -2,4 +2,14 @@ import { addons } from '@storybook/manager-api'
...
+  theme: create(themeConfig),
+  sidebar: {
+    filters: {
+      hidden: (item) => {
+        console.log(item.title, item.tags)
+        return !item.tags.includes('hidden')
+      },
+    },
+  },
+})
```
alimpens, line 15: We don't need this anymore, right?

storybook/storybook-react/config/manager.js
```diff
@@ -2,4 +2,14 @@ import { addons } from '@storybook/manager-api'
...
+  theme: create(themeConfig),
+  sidebar: {
+    filters: {
+      hidden: (item) => {
+        console.log(item.title, item.tags)
+        return !item.tags.includes('hidden')
+      },
+    },
+  },
+})
```
alimpens, line 15: Same here

storybook/storybook-react/src/Grid/Grid.docs.mdx
```diff
@@ -26,31 +32,17 @@ Daarom bestaat er een compacte variant van het grid.
 
 ### Verticale marge
 
-In tegenstelling tot de horizontale marges zijn de verticale instelbaar.
-De opties `paddingVertical`, `paddingTop` en `paddingBottom` voegen witruimte boven en/of onder het grid toe.
-Dit is handig in een vlak met een achtergrondkleur zoals [Footer](?path=/docs/react_containers-footer--docs) of
+In tegenstelling tot de horizontale marges tussen de kolommen zijn de verticale erboven en eronder instelbaar.
+De opties `paddingVertical`, `paddingTop` en `paddingBottom` voegen daar witruimte toe.
+Dit is handig in een gekleurd vlak zoals [Footer](?path=/docs/react_containers-footer--docs) of [Highlight](?path=/docs/react_containers-highlight--docs) of tussen twee opvolgende grids.
 [Spotlight](?path=/docs/react_containers-spotlight--docs) of tussen twee opvolgende grids.
```
alimpens, line 38: Copy paste error? It says 'of tussen twee opvolgende grids' twice.

storybook/storybook-react/src/Grid/Grid.docs.mdx
```diff
@@ -61,4 +53,47 @@ In sommige gevallen kan daar beter meer of minder witruimte staan.
...
+<Canvas of={GridStories.FullWidth} />
+
+### Startpositie
+
+Elke cel start automatisch in de eerstvolgende beschikbare positie in het grid.
+Een waarde van 1 hoef je dus meestal niet expliciet op te geven.
+Je kunt de startpositie van een cel aanpassen met de `start` prop.
+Zo kun je bijvoorbeeld cellen in meerdere rijen met elkaar uitlijnen.
+Of een cel in het midden van een breed grid plaatsen.
+Ook komt dit voor als een kolom als extra witruimte wilt gebruiken.
```
alimpens, line 85: Do we want to encourage this? I'd rather have people using `grid-row`

alimpens, line 85:
```suggestion
Ook komt dit voor als je een kolom als extra witruimte wilt gebruiken.
```

storybook/storybook-react/src/Grid/Grid.docs.mdx
```diff
@@ -61,4 +53,47 @@ In sommige gevallen kan daar beter meer of minder witruimte staan.
...
+
+Een voorbeeld met `start={2}`:
+
+<Canvas of={GridStories.StartPosition} />
+
+## Richtlijnen
+
+Zorg ervoor dat het aantal kolommen dat je aan een cel toekent past bij het aantal kolommen van het grid voor de betreffende vensterbreedte.
+Datzelfde geldt voor het starten van een cel in een latere kolom.
+Als het totaal van beide waarden is te groot is voegt de browser kolommen toe aan het grid.
```
alimpens, line 98:
```suggestion
Als het totaal van beide waarden te groot is voegt de browser kolommen toe aan het grid.
```

alimpens, line 98: Do we want to implement a check for this?

storybook/storybook-react/src/Grid/Grid.stories.tsx
```diff
@@ -3,25 +3,62 @@
...
+
+const gridCellArgTypes = {
+  span: {
+    control: { type: 'inline-radio' },
+    options: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12],
+  },
+  start: {
+    control: { type: 'inline-radio' },
+    options: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12],
+  },
```
alimpens, line 43: Wouldn't number inputs make more sense here?

storybook/storybook-react/src/Grid/Grid.stories.tsx
```diff
@@ -31,66 +68,96 @@ const StoryTemplate: Story = {
...
+        <Grid>
+          <Story />
+        </Grid>
+      </Screen>
+    ),
+  ],
+  render: ({ children, ...args }) => <Grid.Cell {...args}>{children}</Grid.Cell>,
+}
+
+export const Default: GridStory = {
```
alimpens, line 85: As a developer, the first thing I'd want to see in the docs is the grid with one cell spanning the entire grid, together with the controls for the grid and that one cell. That would get you directly to the code you need. The code in this example will usually not be useful for a dev.

storybook/storybook-react/src/Grid/Grid.stories.tsx
```diff
@@ -31,66 +68,96 @@ const StoryTemplate: Story = {
...
+  ...GridStoryTemplate,
   args: {
     children: Array.from(Array(12).keys()).map((i) => <Grid.Cell className="amsterdam-docs-pink-box" key={i} />),
```
alimpens, line 88: We're using this same array thrice, can't we add it to the `GridStoryTemplate`?

packages/css/src/components/grid/grid.scss
```diff
@@ -71,3 +71,43 @@
 .amsterdam-grid--padding-vertical--large {
   padding-block: calc(var(--amsterdam-grid-gap) * 2);
 }
+
+.amsterdam-grid-cell--full-width {
```
alimpens, line 75: Shouldn't all grid cell classes start with `amsterdam-grid__cell`?

storybook/storybook-react/src/Grid/Grid.docs.mdx
VincentSmedinga: Yes, merge error after renaming Highlight to Spotlight. Fixed.

storybook/storybook-react/src/Grid/Grid.docs.mdx
VincentSmedinga: Can you be more specific? Encourage what? Use `grid-row` how?

We offer no API for rows (yet?). Our current approach relies on automatic rows. They require less configuration to place items on the 4, 8, and 12-column grids.

VincentSmedinga: Yes, updated.

storybook/storybook-react/src/Grid/Grid.docs.mdx
VincentSmedinga: I don’t think so.

storybook/storybook-react/src/Grid/Grid.stories.tsx
VincentSmedinga: Not sure. A number input becomes full width and requires typing or clicking very small buttons. I think radio options are easier to use, even here.

storybook/storybook-react/src/Grid/Grid.stories.tsx
VincentSmedinga: The current examples form a walkthrough of the features of grid and grid cell. Yes, that’s a different approach than what you describe.

Not sure how an example with one cell would get you directly to ‘the code you need’. Can you write the story as you see it?

storybook/storybook-react/src/Grid/Grid.stories.tsx
VincentSmedinga: Seems logical, but results in the second and third canvas being empty and showing incorrect code. I tried extracting a variable and a function as well. Can you give it a shot?

storybook/storybook-react/src/Grid/Grid.stories.tsx
```diff
@@ -31,66 +68,96 @@ const StoryTemplate: Story = {
...
   },
-  name: 'Basis',
+  name: 'Ruimtelijk',
```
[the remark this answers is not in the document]
VincentSmedinga: They set the text for the link in the sidebar. If not specified, Storybook uses the Story component name, changing pascal case to separate words.

This makes me notice I used Dutch story names while all our others are English. Updated.

packages/react/src/Grid/Grid.tsx
alimpens: Yeah, we discussed this earlier. I wrote down:

> 1 story, 1 css file, seperate files for child components, with their own test files. We do not directly expose child components to consumers.

If we agree on that convention, we should document it somewhere.

It's not a problem for the exports. If you try to import `Accordion*` now, you only get `Accordion`, `AccordionProps` and `AccordionSectionProps`, ie only the things specified in `index.ts` of the component.

alimpens: Ah, sorry, missed that somehow

storybook/storybook-react/src/Grid/Grid.docs.mdx
alimpens: Why not? Would be doable, right? We're allowing users to break the 4, 8, 12 column grid now, which isn't ideal imo

storybook/storybook-react/src/Grid/Grid.stories.tsx
alimpens: A number input allows you to use the arrows on your keyboard though, which makes them easier to use I think

storybook/storybook-react/src/Grid/Grid.docs.mdx
alimpens: I'd rather not encourage users to use empty elements for layout purposes. We don't have an API for rows yet, but you can extend the existing css to make use of them.

storybook/storybook-react/src/Grid/Grid.stories.tsx
alimpens: Discussed this in person, we'll leave it as is for now.

storybook/storybook-react/src/Grid/Grid.docs.mdx
alimpens: Ah, I misunderstood what this said.

storybook/storybook-react/src/Grid/Grid.docs.mdx
alimpens: We'll make a seperate story for this

packages/react/src/Grid/Grid.tsx
VincentSmedinga: Yes, pushed.

storybook/storybook-react/src/Grid/Grid.stories.tsx
VincentSmedinga: Updated.

alimpens (approved):

PR #821, opened 2023-11-28
feat: Allow small and large vertical whitespace on grids

VincentSmedinga:
This implements a `gapVertical` property on `Grid` and applies it to the `Footer` example.

### DIFF
--- a/packages/css/src/grid/grid.scss
+++ b/packages/css/src/grid/grid.scss
@@ -28,6 +28,14 @@
   --amsterdam-grid-padding-inline: var(--amsterdam-grid-density-high-padding-inline);
 }
 
+.amsterdam-grid--gap-vertical--small {
+  row-gap: calc(var(--amsterdam-grid-gap) / 2);
+}
+
+.amsterdam-grid--gap-vertical--large {
+  row-gap: calc(var(--amsterdam-grid-gap) * 2);
+}
+
 .amsterdam-grid--padding-bottom--small {
   padding-block-end: calc(var(--amsterdam-grid-gap) / 2);
 }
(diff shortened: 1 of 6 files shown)


packages/css/src/grid/grid.scss
```diff
@@ -28,6 +28,14 @@
   --amsterdam-grid-padding-inline: var(--amsterdam-grid-compact-padding-inline);
 }
 
+.amsterdam-grid--gap-vertical--small {
+  row-gap: calc(var(--amsterdam-grid-gap) / 2);
```
dlnr, line 32: So you'll only allow a natural flow of grid cells into rows and not `grid-template-columns`? Or are developers free to add their own CSS to force cells into rows like I'm doing to the Header? Fidgeting with just empty cells, start and span is not my idea of a good time.

dlnr (approved):

packages/react/src/Grid/Grid.tsx
```diff
@@ -47,6 +47,8 @@ export type GridProps = {
    * @todo Implement more generally – it will be moved into a theme soon.
    */
   compact?: boolean
+  /** The amount of vertical whitespace between rows of the grid. */
+  gapVertical?: 'small' | 'large'
```
alimpens, line 51: Shouldn't you be able to set different values for different breakpoints? Same for the `padding*` props

VincentSmedinga: I’m not sure, I hope we can avoid that extra complexity. Basing padding and gap on our responsive horizontal grid gap already helps it behave on narrow and wide screens.

I can imagine we’ll end up with other sizes than the grid gutter. Should we research that first or can we merge this initial approach?

alimpens (approved):

packages/react/src/Grid/Grid.tsx
VincentSmedinga: I understand. Yes, the spacing approach is the main concern popping up in the [prototypes](https://github.com/Amsterdam/design-system-prototypes) currently. It’s the very next thing we should tackle.

PR #836, opened 2023-11-28
feat: Add initial overview of available icons

VincentSmedinga:
This uses the [IconGallery doc block](https://storybook.js.org/docs/api/doc-block-icongallery#page-top) of Storybook to quickly render an overview of all our current icons.

There’s probably a smarter way to import them all? If so, we’ll probably lose the spaces in the labels, I added them manually.


storybook/storybook-docs/src/icon-gallery.stories.mdx
```diff
@@ -0,0 +1,446 @@
+import { Icon } from "@amsterdam/design-system-react";
+import {
```
alimpens, line 2: You can do `import * as Icons from '@amsterdam/design-system-react-icons'` to import them all, see [the Icon component](https://github.com/Amsterdam/design-system/blob/develop/storybook/storybook-react/src/Icon/Icon.stories.tsx#L7) for an example

alimpens: Can't you do the javascript stuff in a `tsx` file, and import it in the `mdx`? That would get rid of the issues I think?

VincentSmedinga: Smart. I still need to ignore a TypeScript warning as `Icons` is not typed.

alimpens: I think you can set the type like this: `(Object.keys(Icons) as Array<keyof typeof Icons>).map(...)`

VincentSmedinga: Ah yes, updated.

storybook/storybook-docs/src/icon-gallery.stories.mdx
```diff
@@ -0,0 +1,14 @@
+import { Meta } from "@storybook/blocks";
+import { AmsterdamIconGallery } from "./components/AmsterdamIconGallery";
+
+<Meta title="Docs/Designrichtlijnen/Iconen" />
```
alimpens, line 4: An overview of the icons isn't really a 'designrichtlijn', right? Shouldn't we show them directly under docs?

VincentSmedinga: I never really understood the title ‘Design rightlijnen’. If we rename it to ‘Ontwerpkeuzes’ or ‘Vormgeving’ or ‘Digitale huisstijl’, everything we currently have fits again.

alimpens: The idea behind this section is to define the rationale behind design choices, which flows into the actual design. So something like: our interactive color is blue, which results in our links being blue. 

Something like 'Ontwerpkeuzes' would work as well, but 'Vormgeving' means something else imo. Our _vormgeving_ is the result of our _ontwerpkeuzes_.

In the same way, our icons are the result of our _ontwerpkeuzes_, not the _ontwerpkeuze_ itself.

VincentSmedinga: Alright, I moved icons into a new ‘assets’ directory and the parts of the typography dealing with the font itself as well.

alimpens (approved):

PR #880, opened 2023-12-04
chore: Enhance examples of text components through random content

VincentSmedinga:
This makes our examples of headings, paragraphs, and lists more representative and interesting.

The sentence containing all letters of the alphabet is still the best candidate for the typography docs.


storybook/storybook-react/src/Accordion/Accordion.stories.tsx
```diff
@@ -36,13 +37,13 @@ export const Default: Story = {
   args: {
     children: [
       <Accordion.Section key={1} label="Eerste sectie">
-        <Paragraph>Jouw typograaf biedt mij zulke exquise schreven!</Paragraph>
+        <Paragraph>{exampleParagraph()}</Paragraph>
```
alimpens, line 40: Hm, I'm not really a fan of randomizing text on a single page. It makes it harder to compare different variants of the same component imo (for example a small, default or large paragraph).

I'm also wondering if randomizing for every page visit is a good idea. As a developer, I go to Storybook to see what variants there are, to read the documentation and to copy the code I need. Changing dummy content could be an unnecessary cognitive load? And it adds complexity to our Storybook.

VincentSmedinga: > Hm, I'm not really a fan of randomizing text on a single page. It makes it harder to compare different variants of the same component imo (for example a small, default or large paragraph).

Comparing the actual sizes of paragraph variations isn’t necessary. Discovering that there are three and recognising that they are different in general is enough. Users don’t need to know what the exact design decisions between variations are, just that the variations exist. And they shouldn’t decide which variation to use based on their size or another visual characteristic, but from their names and usage instructions.

> I'm also wondering if randomizing for every page visit is a good idea. As a developer, I go to Storybook to see what variants there are, to read the documentation and to copy the code I need. Changing dummy content could be an unnecessary cognitive load? And it adds complexity to our Storybook.

The tasks you mention can all be completed without parsing the actual text content. If they visit the page for the first time, any of the candidate texts they end up seeing will be equally new to them. On consecutive visits, either they will notice that the text is different or they won’t. The first case may just as well be a good reminder that components do their thing irrespective of their precise content, and helps recognise that the content for the component they are implementing on a page will also be different every time.

The extra complexity in Storybook is limited – just one separate file with lists of content and two functions to retrieve them.

Additionally, having these texts in place makes it easier for us to generate representative examples for new components or variations.

So I still propose merging this in.

VincentSmedinga: I discussed this with Inge to get a democratic majority. She agrees the advantages outweigh the downsides.

PR #894, opened 2023-12-06
fix: Clean up redundant subcomponent types

alimpens:


VincentSmedinga:
Is something like ‘Remove redundant `children` props from components` a better PR title to go into the change log? Child components are not the same as subcomponents, right?

VincentSmedinga (approved):

PR #898, opened 2023-12-06
feat!: Update line heights and text levels

VincentSmedinga:
1. Merges text levels 4 and 5
2. Applies newly decided line heights
3. Allows visual (not semantic) heading levels 5 and 6

Typography documentation already reflected this plan.

I don’t think there’s a need to document line heights explicitly.

### DIFF
--- a/packages/react/src/Pagination/Pagination.tsx
+++ b/packages/react/src/Pagination/Pagination.tsx
@@ -122,7 +122,7 @@ export const Pagination = forwardRef(
               onClick={onPrevious}
               type="button"
             >
-              <Icon svg={ChevronLeftIcon} size="level-6" />
+              <Icon svg={ChevronLeftIcon} size="level-5" />
               vorige
             </button>
           </li>
@@ -162,7 +162,7 @@ export const Pagination = forwardRef(
               type="button"
             >
               volgende
-              <Icon svg={ChevronRightIcon} size="level-6" />
+              <Icon svg={ChevronRightIcon} size="level-5" />
             </button>
           </li>
         </ol>


proprietary/tokens/src/components/amsterdam/form-label.tokens.json
```diff
@@ -5,12 +5,12 @@
       "font-family": { "value": "{amsterdam.typography.font-family}" },
       "font-weight": { "value": "{amsterdam.typography.font-weight.bold}" },
       "spacious": {
-        "font-size": { "value": "{amsterdam.typography.spacious.text-level.7.font-size}" },
-        "line-height": { "value": "{amsterdam.typography.spacious.text-level.7.line-height}" }
+        "font-size": { "value": "{amsterdam.typography.spacious.text-level.6.font-size}" },
```
alimpens, line 8: This looks a little small, shouldn't it be level 4?

storybook/storybook-react/src/Icon/Icon.stories.tsx
```diff
@@ -13,7 +13,7 @@ const meta = {
   argTypes: {
     size: {
       control: { type: 'select' },
-      options: ['level-3', 'level-4', 'level-5', 'level-6', 'level-7'],
+      options: ['level-3', 'level-4', 'level-5', 'level-6'],
```
alimpens, line 16: Icon doesn't align to the text anymore in the [Link component](https://amsterdam.github.io/design-system/demo-DES-534-update-line-heights-and-text-levels/?path=/docs/react_navigation-link--docs) amd the [Pagination component](https://amsterdam.github.io/design-system/demo-DES-534-update-line-heights-and-text-levels/?path=/docs/react_navigation-pagination--docs)

proprietary/tokens/src/components/amsterdam/form-label.tokens.json
VincentSmedinga: Ah yes 6 is too small. I would say 4 is too big, a form field is such a small section that a label or heading with a font size equal to the control’s content would be fine. But most (not all) designs suggest 4 as well. Updated.

storybook/storybook-react/src/Icon/Icon.stories.tsx
VincentSmedinga: Ah, that bug must have been introduced with the list items going to the regular level. Fixed now.

alimpens: It's fixed for Link, Pagination is still off I think

VincentSmedinga: Ah. Done as well.

PR #912, opened 2023-12-08
feat: Add text input component

alimpens:
Out of scope:

- Passing the following types as a prop:

  - email
  - password
  - search
  - tel
  - text
  - url


VincentSmedinga (approved):
Great! I think we need more docs about how to use form fields exactly, but that’s a larger operation so let’s get going with this start.

PR #913, opened 2023-12-10
fix: Adjust size of breadcrumb separator and allow natural text wrapping

VincentSmedinga:
* Adjusts the size of the separator from `1rem` to `1ex`
* Makes the separator black instead of blue
* Allows text within an item to wrap
* Improves the Storybook example


packages/css/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -25,27 +25,22 @@
...
 .amsterdam-breadcrumb__item:not(:last-child)::after {
   background-color: var(--amsterdam-breadcrumb-separator-background-color);
-  clip-path: path("M 4.725,16 3.275,14.545 9.815,8 3.275,1.455 4.725,0 l 8,8 z");
+  clip-path: path("M2.725 8 2 7.272 5.27 4 2 .728 2.725 0l4 4-4 4Z");
   content: "";
   display: inline-block;
-  height: 1rem;
-  margin-inline-start: 1rem;
-  width: 1rem;
+  height: 0.5rem;
```
alimpens, line 41: Are we sure about this size? It looks really tiny imo

packages/css/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -55,10 +50,6 @@
   text-decoration-thickness: var(--amsterdam-breadcrumb-item-link-text-decoration-thickness);
   text-underline-offset: var(--amsterdam-breadcrumb-item-link-text-underline-offset);
 
-  &::first-letter {
-    text-transform: capitalize;
```
alimpens, line 59: Don't we want to enforce a capital first letter?

packages/css/src/components/breadcrumb/breadcrumb.scss
VincentSmedinga: It matches the current design. Doesn’t bother me very much, but I can imagine we make the chevrons as large as the font’s x-height. I’ll check with design.

(Although making it respond to font size requires a different approach, the current path is in static pixels.)

packages/css/src/components/breadcrumb/breadcrumb.scss
VincentSmedinga: No, I think we shouldn’t tamper with text people write.

packages/css/src/components/breadcrumb/breadcrumb.scss
VincentSmedinga: Chevrons will become as tall as the x-height of the test. I’ll update this PR.

VincentSmedinga:
Renaming the branch closed the PR, that wasn’t intended.

packages/css/src/components/breadcrumb/breadcrumb.scss
alimpens: Nice!

alimpens (approved):

storybook/storybook-react/src/Breadcrumb/Breadcrumb.stories.tsx
```diff
@@ -15,14 +15,12 @@ export default meta
...
   render: () => (
     <Breadcrumb>
-      <Breadcrumb.Item href="#">home</Breadcrumb.Item>
-      <Breadcrumb.Item href="#">path 1</Breadcrumb.Item>
-      <Breadcrumb.Item href="#">path 2</Breadcrumb.Item>
-      <Breadcrumb.Item href="#">path 3</Breadcrumb.Item>
-      <Breadcrumb.Item href="#">path 5</Breadcrumb.Item>
+      <Breadcrumb.Item href="#">Home</Breadcrumb.Item>
+      <Breadcrumb.Item href="#">Nieuws</Breadcrumb.Item>
+      <Breadcrumb.Item href="#">Kennisgevingen en bekendmakingen</Breadcrumb.Item>
```
dlnr, line 23: A page depth of 5 like the Figma examples will help development for small screens, I think breaking the lines instead of a horizontal scroll will cause a lot of vertical space.

packages/css/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -25,27 +25,22 @@
...
-  background-color: var(--amsterdam-breadcrumb-separator-background-color);
-  clip-path: path("M 4.725,16 3.275,14.545 9.815,8 3.275,1.455 4.725,0 l 8,8 z");
+  background-image: var(--amsterdam-breadcrumb-separator-background-image);
+  background-repeat: no-repeat;
   content: "";
   display: inline-block;
-  height: 1rem;
-  margin-inline-start: 1rem;
-  width: 1rem;
+  height: 1ex;
```
dlnr, line 41: This results in a much smaller caret then the design. And why not use an icon here that matches the font level of the item?

storybook/storybook-react/src/Breadcrumb/Breadcrumb.stories.tsx
VincentSmedinga: This was the longest example I could find on our current website, so maybe a depth of 5 isn’t all that realistic.

I don’t think breadcrumb items should overflow the screen. That’s a useful pattern for a menu, but applying it to breadcrumbs would hide the most important items from sight.

packages/css/src/components/breadcrumb/breadcrumb.scss
VincentSmedinga: This makes the chevron icon equally large as in Figma, unless I’m missing something?

Inge and I agreed it must be as tall as the x-height of the font. That’s exactly what the `ex` unit does, with the added bonus that the icon sizes with its text.

Not sure what you mean with ‘an icon that matches the font level of the item’. If you mean inserting an icon component, that would make it more difficult to implement in plain HTML.

storybook/storybook-react/src/Breadcrumb/Breadcrumb.stories.tsx
dlnr: Agreed, I did find one [example](https://www.gov.uk/plan-shared-parental-leave-pay) where there are 3 long breadcrumbs and when viewing on mobile they hide the second item.

packages/css/src/components/breadcrumb/breadcrumb.scss
dlnr: Ok, I must have been comparing to the wrong chevron in Figma, I can see the "OLD breadcrumb" now has a scaled down icon.

dlnr (approved):

PR #932, opened 2023-12-12
feat: Introduce ‘Hero Image’ component

VincentSmedinga:


packages/react/src/HeroImage/HeroImage.test.tsx
dlnr, line 1: I'm missing tests for `HeroImageProps`.

packages/react/src/HeroImage/HeroImage.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const HeroImage = forwardRef<HTMLDivElement, HeroImageProps>(
+  ({ as = 'div', children, className, src, ...restProps }: HeroImageProps, ref) => {
+    const Component = as
+
+    return (
+      <Component
+        {...restProps}
+        ref={ref}
+        className={clsx('amsterdam-hero-image', className)}
+        style={{ backgroundImage: `url(${src ?? 'https://picsum.photos/1280/360'})` }}
```
dlnr, line 24: Why did you choose a css background image instead of an image (with aspect ratio element) to allow for some a11y. You could use stacking to put the form over an image element:
```
grid-area: 1 / 1;
align-items: center;
```

storybook/storybook-react/src/HeroImage/HeroImage.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  args: {
+    src: 'https://picsum.photos/1280/360?random=1',
+  },
+}
+
+export const WithSearchField: Story = {
+  args: {
+    src: 'https://picsum.photos/1280/360?random=2',
+    children: (
+      <Grid paddingVertical="medium">
```
dlnr, line 33: This is not needed if you use the grid-area trick, and you'll avoid misalignment with the grid that it could already be in.

storybook/storybook-react/src/HeroImage/HeroImage.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  },
+} satisfies Meta<typeof HeroImage>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    src: 'https://picsum.photos/1280/360?random=1',
```
dlnr, line 25: All of these are suitable for work right? Maybe we could add some demo images in the storybook theme config to align with Figma.

packages/react/src/HeroImage/HeroImage.tsx
dlnr: https://codepen.io/dlnr/pen/eYxoaWN

storybook/storybook-react/src/HeroImage/HeroImage.stories.tsx
VincentSmedinga: I expect them to be, yes. Having a couple of our own images instead is a good idea.

packages/react/src/HeroImage/HeroImage.tsx
dlnr: A css background will not be described to a screenreader. I think it would be a fair to say something like "image of two people in front of city hall" or something like that.

packages/react/src/HeroImage/HeroImage.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface HeroImageProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {
+  /** The source of the image. */
+  src: ImgHTMLAttributes<HTMLImageElement>['src']
+}
+
+export const HeroImage = forwardRef<HTMLDivElement, HeroImageProps>(
+  ({ children, className, src, ...restProps }: HeroImageProps, ref) => (
+    <div {...restProps} ref={ref} className={clsx('amsterdam-hero-image', className)}>
+      <img alt="" className="amsterdam-hero-image__image" src={src} />
```
dlnr, line 17: Perhaps using the Image component is better here and I think the alt text should describe something to screenreaders. Even if it's decorative. But perhaps @alimpens has different thoughts.

This large image could also benefit from being responsive.

packages/css/src/components/hero-image/hero-image.scss
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+.amsterdam-hero-image {
+  aspect-ratio: var(--amsterdam-hero-image-aspect-ratio);
+  display: grid;
+}
+
+.amsterdam-hero-image > * {
+  align-self: center;
+  grid-area: 1 / -1;
```
dlnr, line 13: A negative integer doesn't make a difference here since it's one cell.

packages/css/src/components/hero-image/hero-image.scss
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.amsterdam-hero-image__image {
+  object-fit: cover;
+  width: 100%;
```
dlnr, line 18: This is a theory but if you do not specify `height: 100%` or `overflow: hidden` for the parent and somebody uses a higher image it will not honor the aspect-ratio right?

dlnr: Ok, turns out grid does not handle aspect-ratio, the image will need a wrapper.

dlnr:
```suggestion
.amsterdam-hero-image__wrapper {
  aspect-ratio: var(--amsterdam-hero-image-aspect-ratio);
}

.amsterdam-hero-image__image {
  height: 100%;
  object-fit: cover;
  width: 100%;
}
```

packages/react/src/HeroImage/HeroImage.tsx
dlnr:
```suggestion
      <div className="amsterdam-hero-image__wrapper">
        <img alt="" className="amsterdam-hero-image__image" src={src} />
      </div>
```

packages/react/src/HeroImage/HeroImage.test.tsx
dlnr: A section with the search role comes to mind, perhaps we can discuss this with @alimpens ?

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#technical_summary
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role#prefer_html

packages/css/src/components/hero-image/hero-image.scss
VincentSmedinga: Ah, good find. I updated for this, slightly differently though.

packages/react/src/HeroImage/HeroImage.tsx
dlnr: It doesn't use a `srcset`, in stories you're loading a 1280px wide image in a 320px breakpoint. This becomes a performance penalty.

packages/css/src/components/hero-image/hero-image.scss
dlnr: Oh nice, you don't need a wrapper.

dlnr (approved):

dlnr (approved):

packages/react/src/HeroImage/HeroImage.test.tsx
dlnr:
```
<section role="search">
  <form>
</section>
```

dlnr: Wrong thread, my bad!

VincentSmedinga: The Hero Image component is for general use, anything can go in.

In this case, I’d expect the search role on the form although the `input type="search"` might be enough as well?

I wouldn’t expect this to be a section as it doesn’t have content or a title.

packages/react/src/HeroImage/HeroImage.tsx
VincentSmedinga: People can use `srcset` if they want. Responsive images wasn’t really in scope in my mind, but we could include it here as this is the widest an image will get on our sites.

dlnr (approved):

dlnr (approved):

packages/react/src/HeroImage/index.ts
```diff
@@ -0,0 +1,2 @@
+export { HeroImage } from './HeroImage'
+export type { HeroImageProps } from './HeroImage'
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, updated separately.

storybook/storybook-react/src/HeroImage/HeroImage.docs.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Canvas of={HeroImageStories.WithSearchField} />
+
+## Richtlijnen
+
+- Gebruik een voldoende grote afbeelding.
+  Als het nodig is, wordt de afbeelding uitgerekt om de volledige oppervlakte te bedekken.
+- Specificeer meerdere kandidaten voor de afbeelding via de `srcSet` prop.
+  Bijvoorbeeld een kleine, middelgrote en grote variant voor de diverse schermformaten.
+  Zo voorkom je het onnodige downloaden van grote bestanden.
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, moved and linked to this guideline.

packages/react/src/Image/Image.tsx
```diff
@@ -4,12 +4,16 @@
...
+export interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
+  objectFit?: 'cover' | 'contain'
+}
 
-export const Image = forwardRef(({ className, ...restProps }: ImageProps, ref: ForwardedRef<HTMLImageElement>) => (
-  <img {...restProps} ref={ref} className={clsx('amsterdam-image', className)} />
-))
+export const Image = forwardRef(
+  ({ className, objectFit, ...restProps }: ImageProps, ref: ForwardedRef<HTMLImageElement>) => (
+    <img {...restProps} ref={ref} className={clsx('amsterdam-image', className)} style={{ objectFit }} />
```
alimpens, line 15: Using inline styles deviates from the pattern we use in the rest of the codebase. Any reason to not use a BEM class here?

packages/react/src/HeroImage/HeroImage.tsx
```diff
@@ -0,0 +1,30 @@
...
+   * Note that the image must fill the entire content box, e.g. through `object-fit: cover`.
+   */
+  ImageComponent?: ReactNode
+}
+
+export const HeroImage = forwardRef<HTMLImageElement, HeroImageProps>(
+  ({ children, className, ImageComponent, ...restProps }: HeroImageProps, ref) => (
+    <div className={clsx('amsterdam-hero-image', className)}>
+      <div className="amsterdam-hero-image__image">
+        {ImageComponent ?? <Image {...restProps} alt="" objectFit="cover" ref={ref} />}
```
alimpens, line 22: Do we want to render `<Image />` here at all? We could just say that HeroImage is a container with slots where other components are rendered. We might not even need `amsterdam-hero-image__image`, if you can just pass `<AspectRatio><Image /></AspectRatio>` in the first slot.

Maybe slots aren't even necessary, if HeroImage just renders its children :D All HeroImage would do is just render its children over one another.

packages/react/src/HeroImage/HeroImage.tsx
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+      {children}
+      <HeroImage src="abc.jpg" />
```
alimpens, line 25: This is probably not intentional, right? This makes it a recursive function :D

packages/react/src/HeroImage/HeroImage.tsx
```diff
@@ -0,0 +1,29 @@
...
+  src: ImgHTMLAttributes<HTMLImageElement>['src']
+  /** A list of candidate sources of the image. */
+  srcSet?: ImgHTMLAttributes<HTMLImageElement>['srcSet']
+}
+
+export const HeroImage = forwardRef<HTMLDivElement, HeroImageProps>(
+  ({ children, className, sizes, src, srcSet, ...restProps }: HeroImageProps, ref) => (
+    <div {...restProps} ref={ref} className={clsx('amsterdam-hero-image', className)}>
+      <div className="amsterdam-hero-image__image">
+        <img alt="" sizes={sizes} src={src} srcSet={srcSet} />
```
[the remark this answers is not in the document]
alimpens: > Do we want to take this road already?

Don't we already do this with other container components? A Footer for example renders its children, a user has to configure those children themselves.

> If so, do we follow the convention to start component props with a capital letter?

Hm, I'm not sure why we would want to? It's just a prop, right?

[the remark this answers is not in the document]
VincentSmedinga: > A Footer for example renders its children, a user has to configure those children themselves.

Well, the `children` prop is a special case – `<Parent><Child /></Parent>` vs. `<Parent prop={Child} />`. Composition through child components vs. entire components as props. The latter is for ‘multiple children’ – in this case the content of the hero image container, and the component for the background image. Like Vue employs slots for this.

> Hm, I'm not sure why we would want to? It's just a prop, right?

We followed that approach for the Amsterdam App. I forgot the exact rationale. Couldn’t find sources online as well. Renamed the prop from `ImageComponent` to `image`.

packages/react/src/Image/Image.tsx
VincentSmedinga: I was actually thinking of what you sometimes say about not wanting to replicate CSS :)

Using classes here would introduce a couple of `.amsterdam-image-object-fit--cover { object-fit: cover; }` – would that be better?

packages/react/src/HeroImage/HeroImage.tsx
VincentSmedinga: Oh wow, I must have been typing in the wrong window there 😂 How did this not fail a test? Removed.

packages/react/src/Image/Image.tsx
alimpens: Yeah, it feels a bit unnecessary, but it makes sense in the system we have imo. Inline styles have some downsides, they aren't cached for example.

Do we ever need `contain`? I'd expect that to mess up the perceived vertical spacing, it would visually add transparent bars to an image.

VincentSmedinga: I’ve changed this to a Boolean `cover` prop.

packages/react/src/HeroImage/HeroImage.tsx
VincentSmedinga: It is with a heavy heart that I have stripped this component completely to a General Thing that assumes nothing and stacks its children 🥲

We can use the name `Stack`, which some other libraries use for one-dimensional layout components, as we will separate our future `Row` and `Column` components.

packages/react/src/HeroImage/HeroImage.tsx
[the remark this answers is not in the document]
VincentSmedinga: And obliterated now.

packages/react/src/HeroImage/HeroImage.tsx
VincentSmedinga: Discussed and decided to go with ‘Overlap’.

storybook/storybook-react/src/Overlap/Overlap.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+} satisfies Meta<typeof Overlap>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    children: (
+      <>
```
alimpens, line 28: This fragment isn't necessary if we make `children` an array

VincentSmedinga: Yes, updated.

alimpens (approved):

PR #933, opened 2023-12-12
feat: Introduce ‘Mark’ component

VincentSmedinga:


packages/css/src/components/mark/mark.scss
```diff
@@ -0,0 +1,8 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-mark {
```
dlnr, line 6: There apears to be some inline-padding to the mark if you compare it to Figma, but you'll never get the same padding on a new line if the mark breaks, so probably better to leave it default.

packages/css/src/components/mark/README.md
```diff
@@ -0,0 +1,18 @@
...
+Markeert een fragment tekst met een gele achtergrond.
+Het zorgt voor sterke nadruk op de tekst en leidt de aandacht ernaartoe.
+
+## Richtlijnen
+
+### Zo gebruiken
+
+- Maximaal 1 markering per pagina.
+  De voorkeur is geen markering.
+- Maximaal 1 zin.
```
dlnr, line 12: 1 Sentence per paragraph or document? The latter would seem strange.

packages/css/src/components/mark/mark.scss
VincentSmedinga: I agree.

dlnr (approved):

dlnr (approved):

alimpens (approved):

PR #965, opened 2023-12-18
chore: Remove unused date-fns package

VincentSmedinga:
We’re not using this.

@hcvdhaar Did I forget why we added this in the first place?


alimpens:
I think it's there because it's in [the template](https://github.com/nl-design-system/example/tree/main), but we don't seem to be using it. Let's remove it

alimpens (approved):

PR #967, opened 2023-12-18
feat: Use node 20 and update dependencies

VincentSmedinga:
Node 20 has been LTS for a while now, and I have started to see packages requiring this. How will it affect our own users?

The major update affected Utrecht DS and Stylelint; I haven’t tested whether we can upgrade the Markdown packages now.


VincentSmedinga:
[Upgrading Stylelint](https://stylelint.io/migration-guide/to-16/) could use more attention; I’ll create a new PR to upgrade the node only.

PR #636, opened 2023-10-13
Add Highlight component

dlnr:
Simple highlight version with no offset rows, just a colored background.

PR #968, opened 2023-12-18
feat: Allow ‘Grid Cell’ to render a custom HTML element

VincentSmedinga:
And update the approach on ‘Spotlight’.

### DIFF
--- a/packages/react/src/Spotlight/Spotlight.test.tsx
+++ b/packages/react/src/Spotlight/Spotlight.test.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import { createRef } from 'react'
 import { Spotlight } from './Spotlight'
 import '@testing-library/jest-dom'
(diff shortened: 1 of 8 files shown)


packages/react/src/Spotlight/Spotlight.test.tsx
```diff
@@ -48,4 +48,10 @@ describe('Spotlight', () => {
 
     expect(component).toHaveClass('amsterdam-spotlight--green')
   })
+
+  it('renders a custom tag', () => {
+    render(<Spotlight as="article" />)
```
dlnr, line 53: Spotlight should be named in the PR title for future reference.

dlnr (approved):

packages/react/src/Spotlight/Spotlight.test.tsx
VincentSmedinga: Spotlight already supported this and its API hasn’t changed. What would you like to mention in the PR title?

dlnr: Allow ‘Grid Cell’ and 'Spotlight' to render a custom HTML element

VincentSmedinga: No, Spotlight didn’t change here. It had an `as` prop from the start in #636, you created it yourself.

PR #969, opened 2023-12-19
chore: Require explicit and inline type imports

VincentSmedinga:
We want to export and import types explicitly.

We can use [inline type modifiers](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names) to combine type and component imports in one statement.

Let’s use ESLint to do this for us.

### DIFF
--- a/packages/react/.eslintrc.json
+++ b/packages/react/.eslintrc.json
@@ -3,6 +3,9 @@
     "jest/globals": true
   },
   "rules": {
+    "@typescript-eslint/consistent-type-exports": "error",
+    "@typescript-eslint/consistent-type-imports": "error",
+    "import/consistent-type-specifier-style": ["error", "prefer-inline"],
     "react/react-in-jsx-scope": "off"
   }
 }
--- a/storybook/storybook-react/.eslintrc.json
+++ b/storybook/storybook-react/.eslintrc.json
@@ -1,5 +1,8 @@
 {
   "rules": {
+    "@typescript-eslint/consistent-type-exports": "error",
+    "@typescript-eslint/consistent-type-imports": "error",
+    "import/consistent-type-specifier-style": ["error", "prefer-inline"],
     "react/prop-types": "off",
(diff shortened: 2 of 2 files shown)


VincentSmedinga:
This results in the following error and my attempts to resolve it are fruitless. Can either of you get this to work?

> Error: Error while loading rule '@typescript-eslint/consistent-type-exports': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Occurred while linting /Volumes/Code/amsterdam/design-system/design-system/packages/react/.eslintrc.json

alimpens:
> This results in the following error and my attempts to resolve it are fruitless. Can either of you get this to work?
> 
> > Error: Error while loading rule '@typescript-eslint/consistent-type-exports': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
> > Occurred while linting /Volumes/Code/amsterdam/design-system/design-system/packages/react/.eslintrc.json

I can't quite get it to work either. Do we need both `import/consistent-type-specifier-style` and `@typescript-eslint/consistent-type-imports`? Everything seems to work when just using the first rule

VincentSmedinga:
> I can't quite get it to work either. Do we need both `import/consistent-type-specifier-style` and `@typescript-eslint/consistent-type-imports`? Everything seems to work when just using the first rule

For me it only inlines existing type imports with the former, doesn’t fix the currently untyped ones.

They should be usable at the same time: https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how

The `packages/react/.eslintrc.json` doesn’t explicitly extend the root `.eslintrc.json` where `parserOptions.project` does get defined. Can that be a problem or are they merged in the background?

If you can’t get it to work in a reasonable amount of time, just close this.

PR #988, opened 2023-12-20
feat: Add Skip link component

alimpens:
I've used the `visually-hidden` class as a css util here. The `:not(:active, :focus)` part doesn't work if you use the `VisuallyHidden` component. Maybe we should think about properly turning this into a css util?

### DIFF
--- a/packages/css/src/components/skip-link/README.md
+++ b/packages/css/src/components/skip-link/README.md
@@ -1,1 +1,30 @@
 # Skip Link
+
+Gebruik een Skip link om makkelijk met het toetsenbord naar de belangrijkste inhoud te navigeren.
+Met een Skip link kun je terugkerende navigatieblokken (zoals het hoofdmenu of het kruimelpad) overslaan.
+Een Skip link is verborgen totdat deze met het toetsenbord geactiveerd wordt.
+
+## Richtlijnen
+
+### Zo gebruiken
+
+- Plaats de Skip link als eerste element in `<body`, tenzij je een cookie-banner hebt.
+  Plaats de Skip link dan direct na de cookie-banner.
+- Gebruik de Skip link om naar de belangrijkste inhoud te navigeren.
+  Op een artikelpagina is dat bijvoorbeeld de titel van het artikel, op een zoekpagina is dat het zoekveld.
+- Voor complexe pagina's met meerdere secties kun je meer dan 1 Skip link gebruiken.
+  In de meeste gevallen is dit niet nodig.
+
+### Dit vermijden
+
+- Skip links zijn niet nodig op een simpele pagina waar alleen tekst staat en weinig navigatie.
+  Het doel van een Skip link is om terugkerende navigatieblokken over te slaan.
+  Als die blokken er niet zijn, is een Skip link niet nodig.
+- Plaats de Skip link niet in een `nav` regio, of in de Header.
+
+## Relevante WCAG eisen
+
+- Voor dit component gelden dezelfde WCAG eisen als voor [het link component](https://amsterdam.github.io/design-system/?path=/docs/react_navigation-link--docs).
+- [WCAG 2.4.1](https://www.w3.org/TR/WCAG22/#bypass-blocks): gebruik een Skip link op elke pagina die begint met een terugkerend navigatieblok.
+- [WCAG 3.2.3](https://www.w3.org/TR/WCAG22/#consistent-navigation): een Skip link staat op elke pagina op dezelfde plek.
+- [WCAG 3.2.4](https://www.w3.org/TR/WCAG22/#consistent-identification): een Skip link heeft dezelfde labels op alle pagina's. Bijvoorbeeld niet: "Navigatie overslaan" op een gedeelte van de site, en "Naar de inhoud" op andere pagina's.
--- a/packages/css/src/components/skip-link/README.md
+++ b/packages/css/src/components/skip-link/README.md
@@ -1,0 +1,1 @@
+# Skip Link
--- a/packages/css/src/components/skip-link/skip-link.scss
+++ b/packages/css/src/components/skip-link/skip-link.scss
@@ -15,6 +15,7 @@
   outline-offset: var(--amsterdam-skip-link-outline-offset);
   padding-block: 0.5rem;
   padding-inline: 1rem;
+  text-align: center;
   text-decoration: none;
   width: 100%;
 
--- a/packages/css/src/components/skip-link/skip-link.scss
+++ b/packages/css/src/components/skip-link/skip-link.scss
@@ -1,0 +1,28 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-skip-link {
+  background-color: var(--amsterdam-skip-link-background-color);
+  color: var(--amsterdam-skip-link-color);
+  display: flex;
+  font-family: var(--amsterdam-skip-link-font-family);
+  font-size: var(--amsterdam-skip-link-font-size);
+  font-weight: var(--amsterdam-skip-link-font-weight);
+  justify-content: center;
+  line-height: var(--amsterdam-skip-link-line-height);
+  outline-offset: var(--amsterdam-skip-link-outline-offset);
+  padding-block: 0.5rem;
+  text-decoration: none;
+  width: 100%;
+
+  &:hover {
+    background-color: var(--amsterdam-skip-link-hover-background-color);
+  }
+
+  .amsterdam-theme--compact & {
+    font-size: var(--amsterdam-skip-link-compact-font-size);
+    line-height: var(--amsterdam-skip-link-compact-line-height);
+  }
+}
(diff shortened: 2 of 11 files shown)


packages/css/src/components/skip-link/skip-link.scss
```diff
@@ -0,0 +1,30 @@
...
+  font-size: var(--amsterdam-skip-link-font-size);
+  font-weight: var(--amsterdam-skip-link-font-weight);
+  justify-content: center;
+  line-height: var(--amsterdam-skip-link-line-height);
+  outline-offset: var(--amsterdam-skip-link-outline-offset);
+  padding-block: 0.5rem;
+  padding-inline: 1rem;
+  text-align: center;
+  text-decoration: none;
+  width: 100%;
```
VincentSmedinga, line 20: Is there a good reason to use flexbox here? I’ve had some trouble with flexbox on text elements in the past. Using block display seems to allow shaving off (`justify-content` and `width`) as well.

```suggestion
  background-color: var(--amsterdam-skip-link-background-color);
  color: var(--amsterdam-skip-link-color);
  display: block;
  font-family: var(--amsterdam-skip-link-font-family);
  font-size: var(--amsterdam-skip-link-font-size);
  font-weight: var(--amsterdam-skip-link-font-weight);
  line-height: var(--amsterdam-skip-link-line-height);
  outline-offset: var(--amsterdam-skip-link-outline-offset);
  padding-block: 0.5rem;
  padding-inline: 1rem;
  text-align: center;
  text-decoration: none;
```

storybook/storybook-react/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+export const Default: Story = {
+  args: {
+    style: {
+      clip: 'auto',
+      clipPath: 'none',
+      height: 'auto',
+      overflow: 'auto',
+      position: 'static',
+      whiteSpace: 'normal',
+      width: 'auto',
```
VincentSmedinga, line 49: Is this clearer?

```suggestion
      clip: 'initial',
      clipPath: 'initial',
      height: 'initial',
      overflow: 'initial',
      position: 'initial',
      whiteSpace: 'initial',
      width: 'initial',
```

storybook/storybook-react/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ],
+} satisfies Meta<typeof SkipLink>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    style: {
```
VincentSmedinga, line 42: Do we need to explain this in a code comment or when introducing the story?

packages/css/src/components/skip-link/README.md
```diff
@@ -0,0 +1,30 @@
+# Skip Link
+
+Gebruik een Skip link om makkelijk met het toetsenbord naar de belangrijkste inhoud te navigeren.
```
VincentSmedinga, line 3: We capitalize all words in a multi-word component name to make it easier to recognize – Skip Link.

packages/react/src/SkipLink/SkipLink.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { render, screen } from '@testing-library/react'
+import { createRef } from 'react'
+import { SkipLink } from './SkipLink'
+import '@testing-library/jest-dom'
+
+describe('Skip link', () => {
+  it('renders', () => {
+    render(<SkipLink href="/" />)
+
```
VincentSmedinga, line 9: How about removing all empty lines from these simple and repetitive test functions?

packages/react/src/SkipLink/SkipLink.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(component).toHaveClass('amsterdam-skip-link')
+  })
+
+  it('renders an additional class name', () => {
+    render(<SkipLink href="/" className="extra" />)
+
+    const component = screen.getByRole('link')
+
+    expect(component).toHaveClass('extra')
+    expect(component).toHaveClass('amsterdam-skip-link')
```
VincentSmedinga, line 30: I think this is equivalent.

```suggestion
    expect(component).toHaveClass('amsterdam-skip-link extra')
```

packages/react/src/SkipLink/index.ts
```diff
@@ -0,0 +1,2 @@
+export { SkipLink } from './SkipLink'
+export type { SkipLinkProps } from './SkipLink'
```
VincentSmedinga, line 2:
```suggestion
export { SkipLink, type SkipLinkProps } from './SkipLink'
```

storybook/storybook-react/src/SkipLink/SkipLink.docs.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
+
+## Toon bij focus
+
+Een Skip link wordt pas getoond als deze focus krijgt.
```
VincentSmedinga, line 15: This file too:

```suggestion
Een Skip Link wordt pas getoond als deze focus krijgt.
```

storybook/storybook-react/src/SkipLink/SkipLink.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+      width: 'auto',
+    },
+  },
+}
+
+export const OnFocus: Story = {
+  decorators: [
+    (Story) => (
+      <>
+        <Paragraph size="small" style={{ marginBottom: '2rem' }}>
```
VincentSmedinga, line 58: I have been using `.amsterdam-docs-paragraph` for this. Which of the two would you pick to use everywhere?

VincentSmedinga (changes requested):
> Maybe we should think about properly turning [visually-hidden] into a css util?

What more would we need to make it proper? Or why not use it like this in its current state?

packages/css/src/components/skip-link/skip-link.scss
alimpens: I think I originally centered the text with `display: flex` and `justify-content`, but `display: block` and `text-align: center` does the same job

packages/react/src/SkipLink/SkipLink.test.tsx
alimpens: Hm, I don't know. I think the empty lines between the render, selection and test parts make it a bit more readable?

packages/react/src/SkipLink/SkipLink.test.tsx
alimpens: Yeah, good idea. I changed it in the plop template as well

proprietary/tokens/src/components/amsterdam/skip-link.tokens.json
```diff
@@ -0,0 +1,20 @@
...
+  "amsterdam": {
+    "skip-link": {
+      "background-color": { "value": "{amsterdam.color.primary-blue}" },
+      "color": { "value": "{amsterdam.color.primary-white}" },
+      "font-family": { "value": "{amsterdam.typography.font-family}" },
+      "font-weight": { "value": "{amsterdam.typography.font-weight.normal}" },
+      "font-size": { "value": "{amsterdam.typography.spacious.text-level.6.font-size}" },
+      "line-height": { "value": "{amsterdam.typography.spacious.text-level.6.line-height}" },
+      "outline-offset": { "value": "{amsterdam.focus.outline-offset}" },
+      "compact": {
```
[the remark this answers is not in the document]
alimpens: Hm, if we treat `spacious` as the default and `compact` as an exception, it's easier to refactor when we move them to a theme, I think? Same goes for the css

storybook/storybook-react/src/SkipLink/SkipLink.stories.tsx
alimpens: A code comment is a good idea I think. In Storybook the difference is already explained in the Story itself, I think? Or isn't that obvious enough?

storybook/storybook-react/src/SkipLink/SkipLink.stories.tsx
alimpens: I think using a `Paragraph` makes more sense. The styling is exactly the same for all stories, and we only have to maintain that styling in one place.

alimpens:
> > Maybe we should think about properly turning [visually-hidden] into a css util?
> 
> What more would we need to make it proper? Or why not use it like this in its current state?

If we want to make it into an exposed css util, we would have to document it, and maybe think about our namespacing (`amsterdam-visually-hidden` is a bit long).

Also, having both a `VisuallyHidden` component and a css util is a bit confusing. The css util would be the cleaner solution (less HTML bloat) in most cases, also for internal use. If we do want to use it this way, we should implement that everywhere we're using `VisuallyHidden` now

packages/react/src/SkipLink/index.ts
alimpens: Discussed this, we're going with importing and exporting types on separate lines for now

proprietary/tokens/src/components/amsterdam/skip-link.tokens.json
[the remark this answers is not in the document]
VincentSmedinga: Oh yeah, this is not long-lived anyway.

VincentSmedinga (approved):

PR #992, opened 2023-12-21
feat: Add MegaMenu component

dlnr:
Initial grid setup of the Mega Menu component, which will be developed further when added to header.


packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+.amsterdam-mega-menu .amsterdam-grid__cell {
+  .amsterdam-heading:nth-child(n + 1) {
+    padding-block-start: 3.5rem;
+  }
+}
+
+.amsterdam-mega-menu .amsterdam-unordered-list {
```
alimpens, line 24: This creates a dependency between Megamenu and UnorderedList. We could create a MegaMenu subcomponent (MegaMenu.ListWrapper or something) which sets the `columns` styling, so we could get rid of that dependency.

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,34 @@
...
+  --amsterdam-grid-gap: var(--amsterdam-grid-spacious-gap);
```
alimpens, line 25: Is this var necessary? This creates a dependency between MegaMenu and Grid. Couldn't we just set the correct column gap in our tokens?

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  column-gap: var(--amsterdam-grid-gap);
+  columns: 20rem auto;
+  display: block;
+
+  &__item {
+    padding-block-start: 1rem;
```
alimpens, line 32: Maybe we should do this by making the `gap` value for lists configurable?

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,34 @@
...
+  .amsterdam-heading:nth-child(n + 1) {
+    padding-block-start: 3.5rem;
+  }
+}
+
+.amsterdam-mega-menu .amsterdam-unordered-list {
+  --amsterdam-grid-gap: var(--amsterdam-grid-spacious-gap);
+
+  column-gap: var(--amsterdam-grid-gap);
+  columns: 20rem auto;
```
alimpens, line 28: This feels a bit magic number-y. Couldn't we create a component with a configurable column count per breakpoint?

packages/react/src/MegaMenu/MegaMenu.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { render } from '@testing-library/react'
+import { createRef } from 'react'
+import { MegaMenu } from './MegaMenu'
+import '@testing-library/jest-dom'
+
+describe('Mega menu', () => {
+  it('renders', () => {
+    const { container } = render(<MegaMenu />)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 10: Try to use the ARIA roles for component selection, so `getByRole('navigation')` in this case (if we choose to keep this a `nav`, see other comment)

packages/react/src/MegaMenu/MegaMenu.tsx
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import clsx from 'clsx'
+import { ForwardedRef, forwardRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export interface MegaMenuProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {}
+
+export const MegaMenu = forwardRef(
+  ({ children, className, ...restProps }: MegaMenuProps, ref: ForwardedRef<HTMLElement>) => (
+    <nav {...restProps} ref={ref} className={clsx('amsterdam-mega-menu', className)}>
```
alimpens, line 13: I'd keep this a div for now, if MegaMenu is going to be a child of PageMenu, the entire thing should be wrapped in a `nav`

proprietary/tokens/src/components/amsterdam/mega-menu.tokens.json
```diff
@@ -0,0 +1,5 @@
+{
+  "amsterdam": {
+    "mega-menu": {}
```
alimpens, line 3: We can remove this if we're not using it

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+import { Grid, Heading, Link, MegaMenu, UnorderedList } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Navigation/Mega Menu',
+  component: MegaMenu,
+  args: {
+    children: 'Nieuw component',
+  },
```
alimpens, line 14: This can be removed I think

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+  parameters: {
+    layout: 'fullscreen',
+  },
```
alimpens, line 17: Does this do anything?

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+} satisfies Meta<typeof MegaMenu>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    children: [
```
alimpens, line 26: This doesn't have to be an array, there's only one child

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </UnorderedList>
+        </Grid.Cell>
+      </Grid>,
+    ],
+  },
+}
+
+export const MultipleCategories: Story = {
+  args: {
+    children: [
```
alimpens, line 144: Same here

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
dlnr: This removes the extra 1rem padding of the story, especially useful when you are working with a grid to avoid double padding.

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  @media screen and (min-width: $amsterdam-breakpoint-wide) {
+    padding-block: 5rem 7.5rem;
+    padding-inline: 5.24rem;
+  }
+}
+
+.amsterdam-mega-menu .amsterdam-grid__cell {
+  .amsterdam-heading:nth-child(n + 1) {
+    padding-block-start: 3.5rem;
```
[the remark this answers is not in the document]
dlnr: Sure, i'll have to compensate the difference in the padding (changes with breakpoints) of the menu but good idea.

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: Added a token but it is no longer a gap, to allow for columns you have to make it a block.

packages/react/src/MegaMenu/MegaMenu.tsx
dlnr: Probably a sibling of header but i'll make it a div.

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: No, columns do not work on flex, an alternative would be to size the items as flex: 0 0 50%; and wrap them, but I would advise against that.

packages/react/src/MegaMenu/MegaMenu.test.tsx
dlnr: Sure, I did have to revert to using a `nav` element, which I would prefer.

packages/css/src/components/mega-menu/mega-menu.scss
alimpens: The `columns` property would be inherited by the list, when we set it on something like `MegaMenu.ListWrapper`

dlnr: I would rather target the tag then introduce a new wrapper.
```suggestion
.amsterdam-mega-menu ul, .amsterdam-mega-menu ol {
```

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-mega-menu {
+  padding-block: var(--amsterdam-mega-menu-padding-block);
```
alimpens, line 7: Do we need this padding? Can't we do all padding with the Grids `paddingVertical` prop?

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,18 @@
...
+}
+
+.amsterdam-mega-menu__list-wrapper {
+  column-gap: var(--amsterdam-mega-menu-list-wrapper-column-gap);
+  columns: var(--amsterdam-mega-menu-list-wrapper-column-width, 20rem) auto;
+  padding-block-start: var(--amsterdam-mega-menu-list-wrapper-padding-block-start);
+
+  &:has(+ .amsterdam-heading) {
+    padding-block-end: var(--amsterdam-mega-menu-list-wrapper-padding-block-end);
+  }
```
alimpens, line 17: Does this work? You can't check previous siblings with `has`, right?

packages/react/src/MegaMenu/MegaMenu.tsx
alimpens: I think the general pattern is to wrap everything (PageMenu + MegaMenu) in a `nav`, see also [Accessible Mega Menu](https://adobe-accessibility.github.io/Accessible-Mega-Menu/)

packages/css/src/components/mega-menu/mega-menu.scss
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-mega-menu {
+  padding-block: var(--amsterdam-mega-menu-padding-block);
+}
+
+.amsterdam-mega-menu__list-wrapper {
+  column-gap: var(--amsterdam-mega-menu-list-wrapper-column-gap);
+  columns: var(--amsterdam-mega-menu-list-wrapper-column-width, 20rem) auto;
```
alimpens, line 12: We generally don't define fallbacks

packages/css/src/components/mega-menu/mega-menu.scss
alimpens: Let's discuss which approach we want to take here tomorrow

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+} satisfies Meta<typeof MegaMenu>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    children: (
+      <Grid key="nav" paddingVertical="medium">
```
alimpens, line 24: We don't need this `key`, same for the other story

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: That property only allows small, medium and large, and they are not the same as the padding for the menu.

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: This works, it checks if the list is followed by a heading and then applies some padding to the bottom.

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
dlnr: Removed

dlnr: Also added a ListWrapper like discussed

packages/css/src/components/mega-menu/mega-menu.scss
VincentSmedinga: We can also use `column-width` if the longhand sets `column-count` to its default of `auto` anyway.

packages/css/src/components/mega-menu/mega-menu.scss
VincentSmedinga: Pretty nifty, but a bit indirect.

This should be a grid or flex gap in the future, either by wrapping the second and third sections in a `Column` or applying a row span to the first.

For now, we can set the value of the `--amsterdam-mega-menu-list-wrapper-padding-block-end` token to the grid gap to reach the same visual effect. Probably separate for spacious and compact?

packages/react/src/MegaMenu/MegaMenu.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const MegaMenu = forwardRef(
+  ({ children, className, ...restProps }: MegaMenuProps, ref: ForwardedRef<HTMLDivElement>) => (
+    <nav {...restProps} ref={ref} className={clsx('amsterdam-mega-menu', className)}>
+      {children}
+    </nav>
+  ),
+) as MegaMenuComponent
+
+const MegaMenuListWrapper = forwardRef(
```
VincentSmedinga, line 30: I’m not fond of the ‘wrapper’ concept; we should find a more semantic approach in cases like these.

We can call this a ‘mega menu section’ if we move the title into it, replace the top padding with a flex gap, and set the bottom padding through `:not(:last-child)`.

proprietary/tokens/src/components/amsterdam/mega-menu.tokens.json
```diff
@@ -0,0 +1,15 @@
+{
+  "amsterdam": {
+    "mega-menu": {
+      "padding-block": {
+        "value": "clamp(2.5rem, calc(2.2917rem + 1.0417vw), 3rem) clamp(3.5rem, calc(3.2917rem + 1.0417vw), 4rem)"
+      },
+      "list-wrapper": {
+        "column-gap": { "value": "{amsterdam.grid.spacious.gap}" },
+        "column-width": { "value": "20rem" },
```
VincentSmedinga, line 9: This value introduces a narrow window between 1088px and 1174px, where we have twelve columns, but the first list is still in one column. It’s a bit magic number-y, as I saw Aram comment somewhere.

How about ‘4 * grid column width + 3 * grid column gap’? That’s precisely what we want to express, right? Splitting or merging of lists should happen at our grid breakpoints.

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+import { Grid, Heading, Link, MegaMenu, UnorderedList } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Navigation/Mega Menu',
+  component: MegaMenu,
+  parameters: {
+    layout: 'fullscreen',
+  },
```
VincentSmedinga, line 14: Let’s hide `children` from the controls table.

```suggestion
  },
  argTypes: {
    children: {
      table: {
        disable: true,
      },
    },
  },
```

VincentSmedinga (changes requested):

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: This is for when two lists live in one grid cell, see the multiple categories variant in stories.

packages/css/src/components/mega-menu/mega-menu.scss
dlnr: The fallback is defined to prevent `columns:  auto` if the slim chance exists the variable is not defined.

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
dlnr: Should we always do this when children are used?

packages/react/src/MegaMenu/MegaMenu.tsx
dlnr: The `columns` are not going to behave nice if you put the title in there. I've added a wrapper after Aram requested me to do so and he explained you can define `columns` to the wrapper.

proprietary/tokens/src/components/amsterdam/mega-menu.tokens.json
dlnr: Also, the issue of the narrow window appeared in the other PR (#1020) when MegaMenu was placed in the `<Screen>` component and the Header. So perhaps we should adress this in the other PR? To tweak this value?

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
VincentSmedinga: Yes, I think so.

proprietary/tokens/src/components/amsterdam/mega-menu.tokens.json
dlnr: As discussed, this approach will stay and value has changed to 22rem.

packages/react/src/MegaMenu/MegaMenu.tsx
dlnr: Discussed with Aram, section is misleading because it is not a section tag. I think "group" is not quite right because you are not grouping lists, you're wrapping one list. So I've changed it to "ListCategory", I'm not happy about the change but I hope you can resolve it like this. Also note that the wrapper is entirely unnecessary, you can just use a `.amsterdam-mega-menu ul` selector.

dlnr (commented):
> I think the general pattern is to wrap everything (PageMenu + MegaMenu) in a `nav`, see also [Accessible Mega Menu](https://adobe-accessibility.github.io/Accessible-Mega-Menu/) 

As discussed we will resolve this in the header-megamenu PR.

packages/react/src/MegaMenu/MegaMenu.tsx
VincentSmedinga: Alright, let’s have a chat about our naming approach and cooperation strategy later.

A type selector would couple CSS more tightly to the HTML. I don’t think that’s a good idea. It requires changing the CSS if the HTML gets changed, doesn’t allow e.g. an ordered list, and unintendedly targets other possible elements in the tree. We should always express our intent through classes.

Category is acceptable (or maybe Part or Region are a bit more structural?). Can you update the class name accordingly?

VincentSmedinga: Or just `MegaMenu.List`.

VincentSmedinga (approved):

PR #993, opened 2023-12-22
fix!: Update alert appearance, make ‘warning’ the default and icon non-optional

VincentSmedinga:
- Updates to latest design: background now white, corrected text levels
- Adds a border token; not sure if the current set is exactly correct
- Uses `Heading` component to display title
- Adds ‘warning’ severity and makes it default
- Removes `icon` prop – distinguishing colour alone is not enough
- Hmm, this could / should have been 3 separate PRs
- Extends and organizes documentation, although not complete yet

PR #1021, opened 2024-01-05
feat!: Add icon button

alimpens:
Some questions:

- Do we want to offer a disabled variant? An Icon Button usually doesn't have any context / text, which makes a disabled button even worse than usual for a11y. -> Discussed, a disabled Icon Button could be useful for application menu bar type things.
- Is onBackground dark / light the right API? We also use onBackground dark for the Search Bar, which isn't necessarily on a dark background. -> We'll discuss this in a [separate ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-568)
- ~~The Alert component should use the Icon Button as well. I didn't do that in this PR to prevent conflict with [this PR](https://github.com/Amsterdam/design-system/pull/993)~~
- The component currently differs from the design, because having a default padding is a bit cumbersome in use. This padding is necessary for the SearchField component though, so we don't use IconButton there (yet). Discuss this with design.


proprietary/tokens/src/components/amsterdam/icon-button.tokens.json
```diff
@@ -0,0 +1,28 @@
...
+  "amsterdam": {
+    "icon-button": {
+      "background-color": { "value": "transparent" },
+      "color": { "value": "{amsterdam.color.primary-blue}" },
+      "outline-offset": { "value": "{amsterdam.focus.outline-offset}" },
+      "hover": {
+        "background-color": { "value": "rgba(0, 70, 153, 0.13)" },
+        "color": { "value": "{amsterdam.color.dark-blue}" }
+      },
+      "on-background-light": {
```
dlnr, line 11: Light is the default, I think we will have a hell of a time when we add a general dark theme because you will have 4 variations in some places and 2 in most.

packages/css/src/components/icon-button/icon-button.scss
```diff
@@ -0,0 +1,39 @@
...
+  padding-inline: 0.5rem;
+  touch-action: manipulation;
+
+  &:hover {
+    background-color: var(--amsterdam-icon-button-hover-background-color);
+    color: var(--amsterdam-icon-button-hover-color);
+  }
+}
+
+.amsterdam-icon-button--on-background-light {
```
VincentSmedinga, line 22: Unsure where this would be used. I’d be inclined to leave it out for now.

proprietary/tokens/src/components/amsterdam/icon-button.tokens.json
```diff
@@ -0,0 +1,28 @@
+{
+  "amsterdam": {
+    "icon-button": {
+      "background-color": { "value": "transparent" },
+      "color": { "value": "{amsterdam.color.primary-blue}" },
+      "outline-offset": { "value": "{amsterdam.focus.outline-offset}" },
+      "hover": {
+        "background-color": { "value": "rgba(0, 70, 153, 0.13)" },
```
VincentSmedinga, line 8:
```suggestion
        "background-color": { "value": "rgba(0, 70, 153, 0.125)" },
```

proprietary/tokens/src/components/amsterdam/icon-button.tokens.json
```diff
@@ -0,0 +1,28 @@
...
+        "color": { "value": "{amsterdam.color.dark-blue}" }
+      },
+      "on-background-light": {
+        "color": { "value": "{amsterdam.color.primary-black}" },
+        "hover": {
+          "background-color": { "value": "rgba(0, 0, 0, 0.13)" },
```
VincentSmedinga, line 14:
```suggestion
          "background-color": { "value": "rgba(0, 0, 0, 0.125)" },
```

storybook/storybook-react/src/Icon/Icon.docs.mdx
```diff
@@ -33,6 +33,12 @@ Iconen kunnen naast tekst gebruikt worden. Door dezelfde tekstniveaus te gebruik
 
 <Canvas of={IconStories.WithText} />
 
+### Square
+
+Soms is het handig als het icoon vierkante ruimte inneemt, bijvoorbeeld bij de Icon Button.
```
VincentSmedinga, line 38:
```suggestion
Soms is het handig als het icoon vierkante ruimte inneemt, bijvoorbeeld bij de [Icon Button](?path=/docs/react_buttons-icon-button--docs).
```

VincentSmedinga (commented):
Nice cleanup in the other components!

VincentSmedinga:
> Do we want to offer a disabled variant? An Icon Button usually doesn't have any context / text, which makes a disabled button even worse than usual for a11y.

We should strongly advise against disabling buttons in general, but I don’t think we can leave them out. I expect some disabled icon buttons in applications, e.g. to add or remove something an item from a set.

> Is onBackground dark / light the right API? We also use onBackground dark for the Search Bar, which isn't necessarily on a dark background.

Let‘s explore this separately. If we keep using it, we should make clear which backgrounds are light and which are dark. I’ve left some comments on this with the code as well. We use it in other places; no need to hold back this PR.

> The Alert component should use the Icon Button as well. I didn't do that in this PR to prevent conflict with [this PR](https://github.com/Amsterdam/design-system/pull/993)

Yes, I’ll make a task to remember that.

proprietary/tokens/src/components/amsterdam/icon-button.tokens.json
VincentSmedinga: No, see Figma. The default is the white background, ‘light’ means yellow, magenta etc. On white, the icon button background is light blue. On light, it is 12.5% black.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -55,6 +55,7 @@
   font-size: var(--amsterdam-dialog-title-spacious-font-size);
   font-weight: var(--amsterdam-dialog-title-font-weight);
   line-height: var(--amsterdam-dialog-title-spacious-line-height);
+  padding-block: 0.5rem;
```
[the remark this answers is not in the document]
alimpens: That would align the close button in the center of multiline text, which isn't what we want I think

packages/css/src/components/icon-button/icon-button.scss
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+.amsterdam-icon-button {
```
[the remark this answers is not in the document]
alimpens: Which ones? Only `border` can be considered a reset, I think?

packages/css/src/components/icon-button/icon-button.scss
alimpens: Maybe we should discuss this with design? It was added quite recently.

packages/css/src/components/icon-button/icon-button.scss
```diff
@@ -0,0 +1,39 @@
...
+    color: var(--amsterdam-icon-button-on-background-light-hover-color);
+  }
+}
+
+.amsterdam-icon-button--on-background-dark {
+  background-color: var(--amsterdam-icon-button-on-background-dark-background-color);
+  color: var(--amsterdam-icon-button-on-background-dark-color);
+
+  &:hover {
+    background-color: var(--amsterdam-icon-button-on-background-dark-hover-background-color);
```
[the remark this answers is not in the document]
alimpens: Good point, it's not very flexible this way. [I've made a ticket to discuss this!](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378/backlog?selectedIssue=DES-568)

packages/react/src/Icon/Icon.tsx
```diff
@@ -10,11 +10,12 @@ import { ForwardedRef, forwardRef, HTMLAttributes } from 'react'
 
 export interface IconProps extends HTMLAttributes<HTMLSpanElement> {
   size?: 'level-3' | 'level-4' | 'level-5' | 'level-6'
+  square?: boolean
```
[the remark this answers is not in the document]
alimpens: The viewbox is square, but the Icon component it taller than just the viewbox, in order to be able to align the icon to text.

packages/css/src/components/dialog/dialog.scss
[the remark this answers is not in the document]
VincentSmedinga: True. Removing padding as discussed this morning, in combination with a text size that matches the heading, will resolve this just as well.

packages/css/src/components/icon-button/icon-button.scss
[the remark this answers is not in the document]
VincentSmedinga: I was thinking about the background colour, too, as we reset it from the browser's default grey to transparent.

packages/css/src/components/icon-button/icon-button.scss
VincentSmedinga: I’ve asked some questions in Figma. Let’s leave it in.

VincentSmedinga (approved):

PR #1040, opened 2024-01-10
feat!: Allow zero row gap for Grid

alimpens:


packages/css/src/components/grid/grid.scss
```diff
@@ -9,8 +9,8 @@
   --amsterdam-grid-gap: var(--amsterdam-grid-spacious-gap);
   --amsterdam-grid-padding-inline: var(--amsterdam-grid-spacious-padding-inline);
 
+  column-gap: var(--amsterdam-grid-gap);
```
VincentSmedinga, line 12: This removes the row gap by default, which is not what we want. We rely on the vertical gap being the same as the horizontal one by default, e.g. when moving cells around respponsively and in grids of cards or top task buttons.

packages/react/src/Grid/Grid.tsx
```diff
@@ -48,7 +48,7 @@ export type GridProps = {
    */
   compact?: boolean
   /** The amount of vertical whitespace between rows of the grid. */
-  gapVertical?: 'small' | 'large'
+  gapVertical?: 'small' | 'medium' | 'large'
```
VincentSmedinga, line 51: So I think `medium` should be the default, and this prop should accept a `none` value.

VincentSmedinga (approved):

PR #1045, opened 2024-01-16
feat: Add Badge component

dlnr:


packages/react/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,22 @@
...
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export interface BadgeProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {
+  color?: 'blue' | 'dark-blue' | 'green' | 'dark-green' | 'orange' | 'magenta' | 'purple' | 'yellow'
+}
+
+export const Badge = forwardRef(
+  ({ children, className, color = 'dark-green', ...restProps }: BadgeProps, ref: ForwardedRef<HTMLElement>) => (
+    <span {...restProps} ref={ref} className={clsx('amsterdam-badge', `amsterdam-badge--${color}`, className)}>
+      {children}
```
VincentSmedinga, line 17: I’m leaning towards using a `label` prop for the text. We want one or two words in here, maybe an icon someday. Rendering any children may open up too many options for misuse.

dlnr: Your suggestion of using `label: string | number` would also allow them to enter any string they want right? It would prevent abuse for something like a react node in a coloured block but you can do that with `SpotLight`.

VincentSmedinga: Yes, obviously, but a `label` prop does prevent people from putting paragraphs, lists, buttons, tables, images etc. in a `Badge` – like `children` would allow to.

storybook/storybook-react/src/Badge/Badge.docs.mdx
```diff
@@ -0,0 +1,19 @@
...
+<Meta of={BadgeStories} />
+
+<Markdown>{README}</Markdown>
+
+## Voorbeelden
+
+### Default
+
+<Primary />
+
```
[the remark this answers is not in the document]
VincentSmedinga: I always delete them 🥸 But I agree it’s current practice until we decide otherwise.

storybook/storybook-react/src/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+      selected: 'dark-green',
+    },
+  },
+} satisfies Meta<typeof Badge>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
```
VincentSmedinga, line 30: Why delete the example of a Tip in a list? It’s the only extended example we had, and it’s present in Figma as well.

VincentSmedinga: Oh, wait – Badge vs Tip.

I’m starting to wonder whether we need Tip at all. But let’s finish Badge first.

packages/react/src/Badge/Badge.tsx
dlnr: Changed it.

storybook/storybook-react/src/Badge/Badge.stories.tsx
dlnr: I think you do need Tip if you look at it's CSS but perhaps if we move some of that here and add some padding.

packages/css/src/components/badge/README.md
```diff
@@ -0,0 +1,14 @@
+# Badge
+
+A prominently colored box containing 1 or 2 words to guide the user in taking a specific action.
```
VincentSmedinga, line 3:
```suggestion
A prominently coloured box containing 1 or 2 words.
Guides the user in taking a specific action or describes its surrounding content.
```

packages/css/src/components/badge/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Design
+
+The badge can contain a string of text or a number.
+
+## Guidelines
+
+The default background colour is green.
+Suggestions when to use the other colours will follow soon.
```
VincentSmedinga, line 12: These are not guidelines, so let’s merge the sentences into the ‘Design’ section.

```suggestion
The badge can contain a short text or a number.
The default background colour is dark green.
Suggestions on when to use the other colours will follow soon.
```

packages/css/src/components/badge/badge.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.amsterdam-badge--blue {
+  background-color: var(--amsterdam-badge-blue-background-color);
+  color: var(--amsterdam-badge-blue-color);
+}
+
+.amsterdam-badge--dark-blue {
+  background-color: var(--amsterdam-badge-dark-blue-background-color);
+  color: var(--amsterdam-badge-dark-blue-color);
```
VincentSmedinga, line 27: This should use the ‘primary blue’ colour.

It highlights an inconsistency in our colour names:
- Figma: ‘Light Blue’ (#009DE6) and ‘Blue’ (#004699)
- React: ‘Blue’ (#009DE6), ‘Primary Blue’ (#004699) and ‘Dark Blue’ (#102E62)

It appears we work around this in the `Spotlight` tokens with a slightly mischievous mapping from blue to dark blue etc.

We shouldn’t change too much in the names of our base colours in the scope of this PR, but we are defining a new API here. To prevent a breaking change next week already, let’s use ‘blue’ and ‘light blue’, and ‘green’ and ‘light green’ as prop values (Figma is also inconsistent between blue and green) and do the mapping in the tokens here as well.

I’d be happy to implement this if you like.

packages/react/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2024 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes } from 'react'
+
+export interface BadgeProps extends HTMLAttributes<HTMLElement> {
+  label: string | number
+  color?: 'blue' | 'dark-blue' | 'green' | 'dark-green' | 'orange' | 'magenta' | 'purple' | 'yellow' | 'black' | 'white'
```
VincentSmedinga, line 12: Now might also be a good time to create types for these.

VincentSmedinga (changes requested):

packages/css/src/components/badge/badge.scss
dlnr: Sure this confused me I must say, they use blue and light-blue, and we change it to blue = light-blue, dark-blue = blue. 

Perhaps we should not have (brand color) primary blue as a supplemental color to begin with and change our blue to light blue.

I also think that the background colours we make available should only consist of the supplemental colours and not add black and white just because we can.

packages/react/src/Badge/Badge.tsx
dlnr: Perhaps when we start using primitives and semantic layers for colour tokens? Might be easier to get all supplemental colours that way.

packages/css/src/components/badge/badge.scss
dlnr: > I’d be happy to implement this if you like.

Please, be my guest! I've commented on colours a couple of times in Figma and do not fully follow what change is wrong or right.

packages/react/src/Badge/Badge.tsx
VincentSmedinga: Yes, that will be great.

Created just a local variable for now, and derived the type and the tests from it.

VincentSmedinga (dismissed):

VincentSmedinga (approved):

VincentSmedinga (dismissed):

packages/react/src/Badge/Badge.tsx
```diff
@@ -0,0 +1,38 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2024 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes } from 'react'
+
+export const badgeColors = [
```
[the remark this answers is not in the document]
VincentSmedinga: https://gemeente-amsterdam.atlassian.net/browse/DES-594

storybook/storybook-react/src/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+const meta = {
+  title: 'Feedback/Badge',
+  component: Badge,
+  args: {
+    label: 'Tip',
+  },
+  argTypes: {
+    color: {
+      control: {
+        type: 'inline-radio',
```
alimpens, line 18: I think a select is more user-friendly here

VincentSmedinga: Done! And checked all components using a select, radio, or inline radio.

VincentSmedinga (approved):

PR #1048, opened 2024-01-18
feat: Add Tip Component

dlnr:
Adding the Badge to the tip raises a couple of questions. Do we need to inherit the font-size of the other child? And do we need to add a required "badge" prop to enforce the badge (and badge color)?


VincentSmedinga:
We decided against including this component now. It’s unclear if and how it will be used, and the implementation is not much more than a flex row with a gap and some padding. Until we offer something specific, people can wrap a Badge and a Paragraph in a small custom container.

PR #1050, opened 2024-01-19
fix: Refactor subcomponents

alimpens:


packages/react/src/Breadcrumb/index.ts
```diff
@@ -1,2 +1,3 @@
 export { Breadcrumb } from './Breadcrumb'
-export type { BreadcrumbProps, BreadcrumbItemProps } from './Breadcrumb'
+export type { BreadcrumbProps } from './Breadcrumb'
+export type { BreadcrumbItemProps } from './BreadcrumbItem'
```
dlnr, line 3: Your IDE will still recommend importing BreadcrumbItem from '@amsterdam/design-system-react/dist/Breadcrumb/BreadcrumbItem'. I would say it is even more confusing to the (new) developer since using `<Breadcrum.Item>` also autocompletes with the `<BreadcrumbItem>`. This goes for all the separated child components.

packages/react/src/OrderedList/index.ts
```diff
@@ -1,2 +1,3 @@
 export { OrderedList } from './OrderedList'
 export type { OrderedListProps } from './OrderedList'
+export type { OrderedListItemProps } from './OrderedListItem'
```
dlnr, line 3: What is the reason for exporting the ItemProps but not the component? The component is recognized by your IDE anyway so now there are different paths. This may look weird when a developer is extending the item for some reason.

packages/react/src/PageMenu/PageMenu.tsx
dlnr, line 1: This change will probably not be needed since these links will probably by passed to header as an object. Only Header and Footer use this component and both look very different, maybe it's obsolete and we can have `Header.Link` and `Footer.Link`?

dlnr (commented):
Autocompletion is an issue for me, the benefit of the child components was that they were exclusively to be used inside the parent and now they are an import autocomplete everywhere.

packages/react/src/Breadcrumb/index.ts
alimpens: Good point, there are way too many type declarations exported currently. I've changed the rollup config to bundle typescript files as well, this should fix that problem. Can you check?

packages/react/src/PageMenu/PageMenu.tsx
alimpens: Maybe, but let's change one thing at a time ;)

packages/react/src/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+  ({ children, className, href, ...restProps }: BreadcrumbItemProps, ref: ForwardedRef<HTMLLIElement>) => (
+    <li {...restProps} className={clsx('amsterdam-breadcrumb__item', className)} ref={ref}>
+      <a className="amsterdam-breadcrumb__link" href={href}>
+        {children}
+      </a>
+    </li>
+  ),
+)
+
+BreadcrumbItem.displayName = 'BreadcrumbItem'
```
VincentSmedinga, line 24: We’re not consistent with including the dot here or not. I say we do. (More elsewhere.)

```suggestion
BreadcrumbItem.displayName = 'Breadcrumb.Item'
```

packages/react/src/Card/CardHeadingGroup.test.tsx
```diff
@@ -0,0 +1,48 @@
+import { render, screen } from '@testing-library/react'
+import { createRef } from 'react'
+import { CardHeadingGroup } from './CardHeadingGroup'
+import '@testing-library/jest-dom'
+
+describe('CardHeadingGroup', () => {
+  it('renders', () => {
+    const { container } = render(<CardHeadingGroup tagline="test" />)
+
+    const component = container.querySelector(':only-child')
```
VincentSmedinga, line 10: Can we query by the role of ‘group’? https://www.w3.org/TR/html-aria/#el-hgroup

packages/react/src/Footer/Footer.test.tsx
```diff
@@ -7,59 +7,35 @@ describe('Footer', () => {
   it('renders an HTML footer element', () => {
     render(<Footer />)
 
-    const footer = screen.getByRole('contentinfo')
+    const component = screen.getByRole('contentinfo')
```
VincentSmedinga, line 10: Should we rename all instances instead of just one?

VincentSmedinga (commented):
Ahh so pretty.

packages/react/src/Breadcrumb/BreadcrumbItem.tsx
alimpens: We currently define the display name twice, here and in the main component file. That's probably not necessary. 

I made [a new ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-591) to fix some issues with the subcomponent typings, let's resolve that there.

packages/react/src/Card/CardHeadingGroup.test.tsx
alimpens: Hm, that's odd. It doesn't show up as role `group` in Chrome's a11y tree. Does it for you?

packages/react/src/OrderedList/index.ts
alimpens: Yeah, good catch, I'm not sure we're exporting the right types for the top component. I've made [a new ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-591) where we can figure that out

packages/react/src/Card/CardHeadingGroup.test.tsx
VincentSmedinga: You’re right, it doesn’t. Does our test runner use Chromium as well?

alimpens: No, it uses Jest with JSDOM, which gives you a subset of the DOM I think? I'm not sure, it's a little black box-y. 

But it's easy to test. When I use `screen.getByRole('group')`, Testing Library is unable to find the element, so the spec doesn't seem to be implemented in Jest / JSDOM either.

VincentSmedinga (approved):

packages/react/src/Breadcrumb/index.ts
dlnr: Yes, that did the trick!

dlnr (approved):

PR #1051, opened 2024-01-22
feat: Introduce separate Link List component

VincentSmedinga:
As discussed, we prefer implementing the current variants of our Link component as separate components. One reason is that the icon for the ‘in list’ variant is an essential difference from the others. A second one is that combining a (un)ordered list, a list item, and a link was cumbersome.

I tried the `small` and `onBackground` props on the list first because that is the more natural API and its value should be the same for all items anyway. But because the list items define both their text size and colors and we want to uphold components being responsible for their own appearance, this is not an option (yet). We might find a good compromise, e.g. with CSS custom properties, but that’s outside the scope of this PR.

The ‘in list’ stories of `Link` has been removed and that value for the `variant` prop has been marked deprecated.

When setting up separate stories for links (without the list) I improved some findings in the Grid stories that served as inspiration. I haven’t added individual link stories yet.


packages/react/src/Link/Link.tsx
```diff
@@ -10,7 +10,9 @@ import type { AnchorHTMLAttributes, ForwardedRef, PropsWithChildren } from 'reac
 import { Icon } from '../Icon/Icon'
 
 type LinkOnBackground = 'default' | 'light' | 'dark'
-type LinkVariant = 'standalone' | 'inline' | 'inList'
+/** @deprecated Use `LinkList` instead. */
```
dlnr, line 13: Do we need deprecation messages in the beta stage?

storybook/storybook-react/src/LinkList/LinkList.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+        Werken bij de gemeente Amsterdam
+      </LinkList.Link>,
+    ],
+  },
+}
+
+export const CustomIcons: Story = {
+  args: {
+    children: [
+      <LinkList.Link key="form" href="#" icon={ChattingIcon}>
```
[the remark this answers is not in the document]
VincentSmedinga: I don’t understand that sentence.

The standalone variant of `Link` isn’t affected by `LinkList`.

Links in a `LinkList` have either a chevron icon or a semantic one.

Icons at the end of the text can still be added because links render their children.

packages/react/src/LinkList/LinkListLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ForwardRefExoticComponent,
+  PropsWithChildren,
+  RefAttributes,
+} from 'react'
+import { Icon } from '../Icon'
+
+type BackgroundName = 'default' | 'light' | 'dark'
+
+export interface LinkListLinkProps extends PropsWithChildren<AnchorHTMLAttributes<HTMLAnchorElement>> {
+  /** The target url for the link. */
```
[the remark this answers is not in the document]
dlnr: I think you should use 'autodocs' or use the [Description](https://storybook.js.org/docs/api/doc-block-description#of) block.

packages/css/src/components/link/README.md
```diff
@@ -24,7 +24,7 @@ Geef een externe link altijd `rel="external"` mee. Vermijd `target="_blank"`, ma
 
 ### Bezochte stijl
 
-Bezochte koppelingen geven aan dat een gebruiker de koppeling al heeft geopend. Gebruikte stijlen raden we af bij koppelingen, omdat ze de pagina vaak onoverzichtelijk maken. Het maakt navigeren door een pagina moeilijker. De inline koppeling heeft wel een bezochte stijl. Het vormt geen onderdeel van navigatie elementen die veel gescand worden.
+Bezochte koppelingen geven aan dat een gebruiker de koppeling al heeft geopend. Gebruikte stijlen raden we af bij koppelingen, omdat ze de pagina vaak onoverzichtelijk maken. Het maakt navigeren door een pagina moeilijker. De inline koppeling heeft wel een bezochte stijl. Het vormt geen onderdeel van navigatie-elementen die veel gescand worden.
```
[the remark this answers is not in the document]
dlnr: If we want to prevent mixing languages we should probably have some sort of ~~boy~~ scout rule. I confess I should have already changed some docs in that regard.

packages/react/src/LinkList/LinkListLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * Don’t mix custom icons with chevrons in one list.
+   */
+  icon?: Function
+  /** Whether the link sits on a dark background. */
+  onBackground?: BackgroundName
+  /**
+   * The text size for the link.
+   * Use the same size for all items in the list.
+   */
+  size?: 'small' | 'large'
```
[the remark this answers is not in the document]
dlnr: Paragraph also has 'medium' but not the links list that will be used in combination with paragraphs? It just confuses me that you would scope the font sizes like this in the tokens. Not a fan of 3 different font sizes in one document I guess.

proprietary/tokens/src/components/amsterdam/link-list.tokens.json
```diff
@@ -0,0 +1,59 @@
...
+      "gap": { "value": "0.5em" },
+      "link": {
+        "color": { "value": "{amsterdam.link-appearance.color}" },
+        "font-family": { "value": "{amsterdam.typography.font-family}" },
+        "gap": { "value": "0.5em" },
+        "text-decoration-line": { "value": "{amsterdam.link-appearance.subtle.text-decoration-line}" },
+        "text-decoration-thickness": { "value": "{amsterdam.link-appearance.text-decoration-thickness}" },
+        "text-underline-offset": { "value": "{amsterdam.link-appearance.text-underline-offset}" },
+        "spacious": {
+          "small": {
```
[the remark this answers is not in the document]
dlnr: Ah the confusing is my fault, we used to have a typography breakpoint..

packages/css/src/components/link/README.md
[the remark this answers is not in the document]
dlnr: Perhaps a translation PR is more effective at this point.

packages/react/src/LinkList/LinkListLink.tsx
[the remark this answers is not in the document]
VincentSmedinga: They both have ‘medium’ as the implicit default – the `size` prop is optional. So all headings still use level numbers and all body text components use small, medium, and large.

packages/react/src/LinkList/LinkListLink.tsx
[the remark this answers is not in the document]
dlnr: I've tested it and you've used render in all the stories, and it does not seem to render any controls that way. Normally storybook would put the jsdoc in the description field of the control.

packages/css/src/components/link/README.md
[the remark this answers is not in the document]
VincentSmedinga: Alright, I’ve translated the link docs now.

packages/react/src/LinkList/LinkListLink.tsx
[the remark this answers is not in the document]
VincentSmedinga: I hadn’t seen Autodocs yet, seems interesting! The Description block parses JSDoc in the stories; props should be documented in the component itself.

The complexity from `LinkList` and `LinkListLink` doesn’t help indeed – but simpler components like `Alert` don’t show the JSDoc descriptions either.

I created a bug on our board.

packages/react/src/LinkList/LinkListLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+            'amsterdam-link-list__link',
+            onBackground && `amsterdam-link-list__link--on-background-${onBackground}`,
+            size && `amsterdam-link-list__link--${size}`,
+            className,
+          )}
+          href={href}
+          ref={ref}
+          {...restProps}
+        >
+          <Icon svg={icon ?? ChevronRightIcon} size={size === 'small' ? 'level-6' : 'level-5'} />
```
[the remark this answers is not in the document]
VincentSmedinga: No wait, `large` is text level 4. This already supports the primary list in the mega menu.

dlnr (approved):

packages/css/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,90 @@
...
+  color: var(--amsterdam-link-list-link-color);
+  display: inline-flex;
+  font-family: var(--amsterdam-link-list-link-font-family);
+  font-size: var(--amsterdam-link-list-link-spacious-medium-font-size);
+  gap: var(--amsterdam-link-list-link-gap);
+  line-height: var(--amsterdam-link-list-link-spacious-medium-line-height);
+  text-decoration-line: var(--amsterdam-link-list-link-text-decoration-line);
+  text-decoration-thickness: var(--amsterdam-link-list-link-text-decoration-thickness);
+  text-underline-offset: var(--amsterdam-link-list-link-text-underline-offset);
+  vertical-align: middle;
```
alimpens, line 32: `font-weight` and `outline-offset` are missing

alimpens, line 32: What does this do? It's doesn't seem to change the alignment

packages/css/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,90 @@
...
+    line-height: var(--amsterdam-link-list-link-compact-medium-line-height);
+  }
+
+  &:hover {
+    text-decoration-line: var(--amsterdam-link-list-link-hover-text-decoration-line);
+  }
+
+  &:visited {
+    color: var(--amsterdam-link-list-link-visited-color);
+  }
```
alimpens, line 45: I don't think this is necessary. This was only necessary because we set the visited color in the Link component, but we don't do that here.

packages/css/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,90 @@
...
+  text-underline-offset: var(--amsterdam-link-list-link-text-underline-offset);
+  vertical-align: middle;
+
+  .amsterdam-theme--compact & {
+    font-size: var(--amsterdam-link-list-link-compact-medium-font-size);
+    line-height: var(--amsterdam-link-list-link-compact-medium-line-height);
+  }
+
+  &:hover {
+    text-decoration-line: var(--amsterdam-link-list-link-hover-text-decoration-line);
```
alimpens, line 40: We should set the color to dark blue here as well, I think

packages/css/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,90 @@
...
+.amsterdam-link-list__link--on-background-dark {
+  color: var(--amsterdam-link-list-link-on-background-dark-color);
+
+  &:hover {
+    color: var(--amsterdam-link-list-link-on-background-dark-hover-color);
+  }
+
+  &:visited {
+    color: var(--amsterdam-link-list-link-on-background-dark-visited-color);
+  }
```
alimpens, line 77: Same here, I don't think the visited states are necessary

packages/css/src/components/link-list/link-list.scss
```diff
@@ -0,0 +1,90 @@
...
+.amsterdam-link-list__link--on-background-light {
+  color: var(--amsterdam-link-list-link-on-background-light-color);
+
+  &:hover {
+    color: var(--amsterdam-link-list-link-on-background-light-hover-color);
+  }
+
+  &:visited {
+    color: var(--amsterdam-link-list-link-on-background-light-visited-color);
+  }
```
alimpens, line 89: Same here, I don't think the visited states are necessary

packages/react/src/LinkList/LinkListLink.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * The text size for the link.
+   * Use the same size for all items in the list.
+   */
+  size?: 'small' | 'large'
+}
+
+interface LinkListLinkComponent
+  extends ForwardRefExoticComponent<LinkListLinkProps & RefAttributes<HTMLAnchorElement>> {}
+
+/** One link with a Link List. */
```
alimpens, line 35: Is this shown anywhere?

packages/react/src/LinkList/LinkListLink.tsx
alimpens, line 54: The default chevrons are too large compared to the designs. The designs have a level 6 chevron next to level 5 text.

storybook/storybook-react/src/LinkList/LinkList.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+<Meta of={LinkListStories} />
+
+<Markdown>{README}</Markdown>
+
+## Examples
+
+### Default
+
+<Primary />
+
```
alimpens, line 14: The Controls are missing here

storybook/storybook-react/src/LinkList/LinkList.docs.mdx
```diff
@@ -0,0 +1,35 @@
...
+Avoid using both meaningful and list icons in the same list.
+
+<Canvas of={LinkListStories.CustomIcons} />
+
+### On Dark Background
+
+For a list on a dark background color, specify this for each link in the list.
+The link remains responsible for using appropriate text colors.
+
+<Canvas of={LinkListStories.OnDarkBackground} />
```
alimpens, line 35: The On Light Background stories are missing

storybook/storybook-react/src/LinkList/LinkList.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { Meta, StoryObj } from '@storybook/react'
+import { exampleLinkList } from '../shared/exampleContent'
+
+const links = exampleLinkList()
+
+const meta = {
+  title: 'Navigation/Link List',
+  component: LinkList,
+  args: {
+    children: 'Default children',
```
alimpens, line 18: I think we can remove this

storybook/storybook-react/src/LinkList/LinkList.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+  },
```
alimpens, line 19: Don't we want the small, custom icons, on dark / light controls here?

storybook/storybook-react/src/LinkList/LinkList.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    return (
+      <LinkList {...args}>
+        {links.map((text, index) => (
+          <LinkList.Link key={index} href="#" size={args['size']}>
+            {text}
+          </LinkList.Link>
+        ))}
+      </LinkList>
+    )
+  },
```
alimpens, line 44: This seems to break the Show code window

VincentSmedinga: The stories have gotten a big update.

storybook/storybook-react/src/LinkList/LinkList.docs.mdx
VincentSmedinga: I’ve added them a bit lower now, to the first story for Link.

packages/react/src/LinkList/LinkListLink.tsx
VincentSmedinga: I assume that’s a bug in the design.

packages/css/src/components/link-list/link-list.scss
alimpens: We explicitly set font weight for all typographic components now, even if they're not bold. We could decide not to do that, but we should do that for all typographic components in one go.

We change the font weight from `normal` to `400`, which probably doesn't matter in most browsers.

packages/react/src/LinkList/LinkListLink.tsx
alimpens: I'm not sure, the smaller size makes sense for chevrons imo. We should discuss icon sizing separately though, I'll make a ticket for it

storybook/storybook-react/src/LinkList/LinkList.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+  ...LinkStoryTemplate,
+  args: {
+    children: links[0],
+    href: '#',
+    onBackground: 'dark',
+  },
+  decorators: [
+    ...LinkStoryTemplate.decorators,
+    (Story) => (
+      <div style={{ background: '#004699', padding: '1rem' }}>
```
alimpens, line 138: This should be dynamic, see [PageHeading](https://github.com/Amsterdam/design-system/blob/d346cdfd2b3c7fab08e6bae10314420f1fde8a5f/storybook/storybook-react/src/PageHeading/PageHeading.stories.tsx#L30). We should implement this for the Link story as well, because it has controls for onBackground

VincentSmedinga: We agree to address this in a separate PR for all components with a `background` prop.

alimpens (approved):

alimpens (approved):

PR #1056, opened 2024-01-26
feat: Hyphenate headings and other large text components

dlnr:
For testing use the word "Carnavalsoptochtvoorbereidingswerkzaamheden" in Heading and Card stories. Make sure the html lang is set to "nl" and it only seems to work if you open the frame of the story.


packages/css/src/common/hyphenation.scss
```diff
@@ -0,0 +1,6 @@
+@mixin hyphenation {
+  hyphenate-limit-chars: 12 8 4;
+  hyphens: auto;
+  overflow-wrap: break-word;
+  word-break: break-word;
```
alimpens, line 5: We should use the tokens here

proprietary/tokens/src/common/amsterdam/hyphenation.tokens.json
```diff
@@ -0,0 +1,11 @@
+{
+  "amsterdam": {
+    "hyphenation": {
+      "hyphenate-limit-chars": { "value": "12 8 4" },
+      "hyphenate-limit-lines": { "value": "2" },
```
alimpens, line 5: This is unused

packages/css/src/common/hyphenation.scss
alimpens, line 5: `break-word` is deprecated

packages/css/src/components/top-task-link/top-task-link.scss
```diff
@@ -32,6 +34,7 @@
     line-height: var(--amsterdam-top-task-link-label-compact-line-height);
   }
 
+  @include hyphenation;
```
alimpens, line 37: Top task Link has text level 4. Shouldn't we hyphenate all text level 4 elements if Top task needs to be hyphenated?

proprietary/tokens/src/common/amsterdam/hyphenation.tokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "amsterdam": {
+    "hyphenation": {
+      "hyphenate-limit-chars": { "value": "12 8 4" },
+      "hyphens": { "value": "auto" },
+      "overflow-wrap": { "value": "break-word" },
+      "word-break": { "value": "auto-phrase" }
```
alimpens, line 7: This only works for Japanese? Doesn't seem very necessary imo

VincentSmedinga: No indeed, removed.

storybook/storybook-react/config/preview-head.html
```diff
@@ -1,3 +1,7 @@
 <script>
   window.global = window;
+
+  // Set language to Dutch for Canvas and Stories
+  document.documentElement.setAttribute("lang", "nl");
+  document.querySelectorAll(".docs-story").forEach((e) => e.setAttribute("lang", "nl"));
```
VincentSmedinga, line 6: This should work, right? Any idea why it doesn’t?

proprietary/tokens/src/common/amsterdam/hyphenation.tokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "amsterdam": {
+    "hyphenation": {
+      "hyphenate-limit-chars": { "value": "12 8 4" },
+      "hyphens": { "value": "auto" },
+      "overflow-wrap": { "value": "break-word" },
```
VincentSmedinga: Should we use `overflow-wrap` or `word-break: break-word`? I think I’d prefer the latter as it fills up the first line as much as possible. See the second example [here on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text/Wrapping_breaking_text#breaking_long_words). @dlnr Did you investigate this?

proprietary/tokens/src/common/amsterdam/hyphenation.tokens.json
VincentSmedinga: I’ve removed it.

storybook/storybook-react/config/preview-head.html
alimpens: I think stories are loaded after the document is loaded, so elements with the `docs-story` class don't exist yet on document load. I've added this script to the story `decorators`, that seems to work

proprietary/tokens/src/common/amsterdam/hyphenation.tokens.json
alimpens: `word-break: break-word` is deprecated, we shouldn't use that. The example uses `word-break: break-all`. `word-break: break-all` beats `hyphens: auto` though, so we can't really use that either 😅

dlnr: I've tried "auto-phrase" but it did not seem to do anything.
> Has the same effect as word-break: normal except that language-specific analysis is performed to improve word breaks by not placing them in the middle of natural phrases.

storybook/storybook-react/config/preview-head.html
dlnr: Does this also work when you open the frame directly?

alimpens: > although it fires too often – it queries the entire document for each Story on it

Derp, missed that. I've gone with another approach, simply wrapping the story in a div with the correct language. Should be more robust.

> Does this also work when you open the frame directly?

Yeah, the new approach does. The decorators also cascade properly, so the stories are wrapped in `<div lang="nl">` first, and any other decorators defined on the story level after.

VincentSmedinga (approved):

PR #1060, opened 2024-01-31
feat: Add initial documentation for colour palette

VincentSmedinga:
This shows a simple list of the colours we currently have. No semantic grouping or explanation or anything; these are to be discussed.


storybook/storybook-docs/src/color.stories.mdx
```diff
@@ -0,0 +1,35 @@
+import { Meta } from "@storybook/blocks";
+import { ColourPalette } from "./components/ColourPalette";
```
alimpens, line 2: Did you choose not to use [Storybooks ColorPalette](https://storybook.js.org/docs/api/doc-block-colorpalette) on purpose?

storybook/storybook-docs/src/color.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+<Meta title="Docs/Assets/Colour" />
+
+# Colour
+
+The basic colours of Amsterdam are white, red, and black.
+We have 8 additional colours and 3 shades of grey.
+
+<ColourPalette>
+  <ColourPalette.Section>
+    <ColourPalette.Tile name="white" rgb="#ffffff" />
```
alimpens, line 13: Can we get these values from the design tokens? So we don't have to maintain them in 2 places.

storybook/storybook-docs/src/color.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+    <ColourPalette.Tile name="red" rgb="#ec0000" />
+    <ColourPalette.Tile name="black" rgb="#000000" />
```
alimpens, line 15: If we're keeping the custom component: maybe change the prop name from `rgb` to `color`?

storybook/storybook-docs/src/color.stories.mdx
VincentSmedinga: Yes, my first attempt used it. Its design doesn’t match ours, we don’t have swatches of shades and tints of one hue, and no names to display on the left.

I’ll be replacing their IconGallery with one of our own as well.

storybook/storybook-docs/src/color.stories.mdx
VincentSmedinga: I’ll be making a case to move to [HSL or even fancier models](https://www.smashingmagazine.com/2021/11/guide-modern-css-colors/) when the time is right. But if we do, we should probably still display the RGB notation as well. I wanted to keep the options open. Should we change to `color` until then?

storybook/storybook-docs/src/color.stories.mdx
VincentSmedinga: Certainly!

The page even reloads automatically when updating a colour token, pretty amazing.

storybook/storybook-docs/src/color.stories.mdx
alimpens: Yeah, I think `color` is a more robust api. We could even allow all types of color notation in the `color` prop, and do any conversion necessary in the component itself... But that's something for later.

VincentSmedinga: Alright, done.

alimpens (approved):

PR #1062, opened 2024-01-31
chore: Simplify token scripts

alimpens:


package.json
```diff
@@ -74,14 +74,14 @@
...
+    "prewatch:storybook": "lerna run build",
     "update-patch": "npm-check-updates --configFileName .ncurc.patch.js --deep --dep dev,prod --target patch --upgrade && pnpm install",
     "update-minor": "npm-check-updates --configFileName .ncurc.minor.js --deep --dep dev,prod --target minor --upgrade && pnpm install",
     "update-major": "npm-check-updates --configFileName .ncurc.major.js --deep --dep dev,prod --target latest --upgrade && pnpm install",
     "watch:css": "npm run --workspace packages/css build:watch",
     "watch:react": "npm run --workspace packages/react build:watch",
     "watch:storybook-docs": "npx wait-on http://localhost:6007 && npm run --workspace storybook/storybook-docs start",
     "watch:storybook-react": "npm run --workspace storybook/storybook-react start",
-    "watch:style-dictionary": "npm run --workspace proprietary/tokens watch:style-dictionary"
+    "watch:style-dictionary": "npm run --workspace proprietary/tokens watch"
```
dlnr, line 85:
```suggestion
    "watch": "npm run --workspace proprietary/tokens watch"
```

VincentSmedinga (commented):
This change is not clear to me, please explain.

package.json
alimpens: I'm not sure I understand this suggestion. We're in the root package.json here, where there are `watch` scripts for several packages and both storybooks. All these `watch` scripts together are run with the `"start": "npm-run-all --parallel watch:**"` command.

The script to watch the tokens shouldn't just be called `watch`, I think. We could call it `watch:tokens` or something to make it more clear, though?

alimpens:
> This change is not clear to me, please explain.

The `tokens` package currently has 5 scripts: `clean`, `build`, `build:style-dictionary`, `watch` and `watch:style-dictionary`. The `build` and `watch` scripts simply run 'all' commands starting with `build:*` or `watch:*`, essentially making them aliases for the other 2 scripts. 

Simply put: we only need 3 scripts here, which this change implements.

package.json
VincentSmedinga: Yes, `watch:tokens` seems consistent with `:css` and `:react`.

package.json
```diff
@@ -74,14 +74,14 @@
     "test": "npm run test-workspaces",
     "test-update": "npm-run-all --sequential clean lint build test",
     "test-workspaces": "lerna run --no-bail test",
-    "prewatch:storybook": "lerna run build:style-dictionary",
+    "prewatch:storybook": "lerna run build",
```
VincentSmedinga, line 77: Let’s sort these alphabetically.

VincentSmedinga (dismissed):

VincentSmedinga (approved):

PR #1064, opened 2024-01-31
feat!: Remove deprecated ‘in list’ variant of Link

VincentSmedinga:
We deprecated this value for the `variant` prop of List in yesterday’s release. This PR removes it for the next release.

It also updates all our examples – something we should actually have done with the deprecation. This includes making the links in the Footer small, and the primary ones in the Mega Menu large. Sorry for not separating commits.


packages/react/src/Link/Link.tsx
dlnr, line 1: You've opted to use jsdocs everywhere. Should we add them if we touch a file without jsdocs?

dlnr (approved):

packages/react/src/Link/Link.tsx
VincentSmedinga: Yes, but let’s first discuss how we want to write this aspect of documentation.

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -156,147 +76,79 @@ export const MultipleCategories: Story = {
...
-                </Link>
-              </UnorderedList.Item>
-              <UnorderedList.Item>
-                <Link variant="inList" href="#">
-                  Subsidies
-                </Link>
-              </UnorderedList.Item>
-            </UnorderedList>
+            <LinkList>
+              <LinkList.Link href="#" size="large">
```
alimpens, line 83: Out of scope for this PR, but I missed it in the LinkList PR: do we really want a `large` link variant? It looks a little off imo

alimpens (approved):

storybook/storybook-react/src/MegaMenu/MegaMenu.stories.tsx
VincentSmedinga: Noted for discussion.

PR #1065, opened 2024-02-01
feat: Refactor button component

dlnr:
To slim down our bundle and allow for more of our customizations we opted to use our own button.


packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,96 @@
...
  */
 
-@import "../../../node_modules/@utrecht/components/button/css";
-
-@mixin reset {
-  -webkit-text-size-adjust: 100%;
-}
-
-.amsterdam-button {
+@mixin amsterdam-button {
```
VincentSmedinga, line 6: What’s the added value of using mixins? A button already has both `.amsterdam-button` and `.amsterdam-button-${variant}`, so we could use classes as well. No other components use the mixins approach.

packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,96 @@
...
     font-size: var(--amsterdam-breadcrumb-compact-font-size);
     line-height: var(--amsterdam-breadcrumb-compact-line-height);
   }
+
+  &:disabled,
+  &[aria-disabled="true"] {
+    cursor: var(--amsterdam-action-disabled-cursor);
+  }
+
+  &:hover:not(:disabled, [aria-disabled="true"]) {
```
VincentSmedinga, line 29: Can we lose the `:not` selector by moving this block to before the `disabled` block?

packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,96 @@
...
+  }
+}
+
+@mixin amsterdam-button-forced-color-mode {
+  @media screen and (-ms-high-contrast: active), screen and (forced-colors: active) {
+    border: 2px solid ButtonBorder; // add border because forced colors changes box-shadow to none
+  }
+}
+
+.amsterdam-button--busy {
```
VincentSmedinga, line 40: I’m not seeing this in action. Does it need a `:hover` selector added? Can we add the arg to the stories?

packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,96 @@
...
+  cursor: var(--amsterdam-action-busy-cursor);
+}
+
+.amsterdam-button--primary {
+  background-color: var(--amsterdam-button-primary-background-color);
+  color: var(--amsterdam-button-primary-color);
```
VincentSmedinga, line 46: Should we set `box-shadow` (with the same colour as the background) for consistency and reusability with the other variants?

packages/css/src/components/button/button.scss
dlnr: It's usually a rendered state, for when you're waiting for other tasks to complete. I'll add some stories.

VincentSmedinga: Yes I mean I tested it locally, but didn’t see the cursor change.

packages/css/src/components/button/button.scss
dlnr: Good idea

VincentSmedinga (dismissed):

proprietary/tokens/src/components/amsterdam/button.tokens.json
```diff
@@ -1,6 +1,13 @@
 {
   "amsterdam": {
     "button": {
+      "font-family": { "value": "{amsterdam.typography.font-family}" },
+      "gap": { "value": "{amsterdam.spacing.inset.md}" },
+      "padding-inline-start": { "value": "{amsterdam.spacing.inset.md}" },
+      "padding-inline-end": { "value": "{amsterdam.spacing.inset.md}" },
+      "padding-block-start": { "value": "{amsterdam.spacing.inset.sm}" },
+      "padding-block-end": { "value": "{amsterdam.spacing.inset.sm}" },
```
alimpens, line 9: The spacing tokens aren't really well thought out yet, I wouldn't use them here.

storybook/storybook-react/src/Button/Button.docs.mdx
```diff
@@ -1,15 +1,11 @@
...
 import README from "../../../../packages/css/src/components/button/README.md?raw";
 
 <Meta of={ButtonStories} />
 
 <Markdown>{README}</Markdown>
 
-<Primary />
-
-<Controls />
-
```
alimpens, line 12: Why did you remove these?

storybook/storybook-react/src/Button/Button.stories.tsx
```diff
@@ -25,16 +41,93 @@ export const Primary: Story = {
...
+export const SecondaryWithIconAndBadge: Story = {
+  args: {
+    variant: 'secondary',
+    children: [
+      'Secondary',
+      <Badge key="badge" label={99} color="dark-blue" />,
+      <Icon key="icon" svg={CloseIcon} size="level-5" />,
+    ],
+  },
+}
```
alimpens, line 73: I don't think you have to show you can add other components to Button 4 times, PrimaryWithIcon is enough imo

packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,95 @@
...
+  font-family: var(--amsterdam-button-font-family);
   font-size: var(--amsterdam-button-spacious-font-size);
+  gap: var(--amsterdam-button-gap);
   line-height: var(--amsterdam-button-spacious-line-height);
-
-  @include reset;
+  outline-offset: var(--amsterdam-button-outline-offset);
+  padding-block: var(--amsterdam-button-padding-block-start) var(--amsterdam-button-padding-block-end);
+  padding-inline: var(--amsterdam-button-padding-inline-start) var(--amsterdam-button-padding-inline-end);
+  -webkit-text-size-adjust: 100%;
```
alimpens, line 17: I think we want to wrap this in a reset mixin, like we did before

packages/css/src/components/button/button.scss
```diff
@@ -3,47 +3,95 @@
...
```
alimpens, line 18: We're missing `touch-action: manipulation` here

proprietary/tokens/src/components/amsterdam/button.tokens.json
dlnr: They were also used in the previous button setup and they're just `1rem` and `.5rem` values.

storybook/storybook-react/src/Button/Button.docs.mdx
dlnr: They interupted the docs without added value.

storybook/storybook-react/src/Button/Button.stories.tsx
dlnr: This was left in here on purpose to test the theory that we don't need a filter button

dlnr: Changed it and moved busy/disabled to controls. I kept secondary with a badge as an example for the filter button discussion. Is that ok?

alimpens: I'm not sure we need to show a filter button with a Badge, that design doesn't seem to be very fleshed out. In most examples, a filter button is a Button with a Close Icon. That's all I would show here. Maybe we should call the story 'Button with an icon', so we don't imply that only the Primary Button can have an Icon. I'd move it to after the Tertiary Button story, so that distinction is more clear

packages/css/src/components/button/button.scss
```diff
@@ -3,42 +3,96 @@
...
 
-@import "../../../node_modules/@utrecht/components/button/css";
-
 @mixin reset {
   -webkit-text-size-adjust: 100%;
 }
 
 .amsterdam-button {
+  border: none;
+  cursor: var(--amsterdam-action-activate-cursor);
```
alimpens, line 12: We shouldn't import common level tokens directly in the css. The component level tokens should import the brand/common level ones.

That way, external users only have to override the component level tokens to get their styling.

packages/css/src/components/button/button.scss
```diff
@@ -3,42 +3,96 @@
...
+  display: inline-flex;
+  font-family: var(--amsterdam-button-font-family);
   font-size: var(--amsterdam-button-font-size);
+  gap: var(--amsterdam-button-gap);
   line-height: var(--amsterdam-button-line-height);
+  outline-offset: var(--amsterdam-button-outline-offset);
+  padding-block: var(--amsterdam-button-padding-block-start) var(--amsterdam-button-padding-block-end);
```
alimpens, line 19: Why do we use a shorthand here if we're going to define both values anyway? As far as I'm concerned, we only have to define 1 value here, but if we want to define both, it makes more sense to not use the shorthand here imo.

alimpens (approved):

PR #1066, opened 2024-02-02
feat: Image srcset documentation

dlnr:


VincentSmedinga (approved):
I’ve shortened it a bit. We do want to suggest best practices for components, but explaining how the web APIs work is a bit much; we can refer to MDN instead.

PR #1071, opened 2024-02-12
feat: Create unique token classes

alimpens:


proprietary/tokens/build.js
```diff
@@ -3,7 +3,7 @@ const StyleDictionary = require('style-dictionary')
 const modes = ['compact']
 
 function generateSharedConfig(mode) {
-  const name = mode ? `${mode}` : 'index'
+  const name = mode || 'index'
```
VincentSmedinga, line 6: I think we can use the null coalescing operator more often – this is one example where our intent is to check for a value that is not defined, not a falsy one ([SO](https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or)).

```suggestion
  const name = mode ?? 'index'
```

E.g. an unfortunate mode of `''` would result in a class of `.amsterdam-theme--`.

proprietary/tokens/build.js
```diff
@@ -28,7 +28,7 @@ function generateSharedConfig(mode) {
           destination: `${name}.theme.css`,
           format: 'css/variables',
           options: {
-            selector: '.amsterdam-theme',
+            selector: `.amsterdam-theme${name === 'index' ? '' : `--${name}`}`,
```
VincentSmedinga, line 31: This is a stylistic choice: should we prefer a logical conjuncation over the ternary operator in compact cases like this?

```suggestion
            selector: `.amsterdam-theme${name !== 'index' && `--${name}`}`,
```

It’s a bit more compact, not difficult to read, and we already use it often.

proprietary/tokens/build.js
alimpens: > E.g. an unfortunate mode of '' would result in a class of .amsterdam-theme--.

But this is something we don't want, right? I think you would get this using the nullish coalescing operator, but not with the OR operator. I don't think it's necessarily a bad idea to not allow falsy values as mode names...

Tangentially related: maybe we should convert this to Typescript, to prevent possible issues like these?

proprietary/tokens/build.js
alimpens: > This is a stylistic choice

Not quite. AND returns a boolean value, so this would result in `.amsterdam-themefalse`

proprietary/tokens/build.js
VincentSmedinga: The current approach will work as well; let’s discuss later.

proprietary/tokens/build.js
VincentSmedinga: Hah, I already doubted whether I wasn’t making a point that opposed my other comment

PR #1074, opened 2024-02-12
feat: Allow lists to appear in white on a dark background

dlnr:
Just like paragraphs the lists are now able to handle inverseColor for dark backgrounds.


packages/css/src/components/ordered-list/ordered-list.scss
```diff
@@ -36,6 +36,12 @@
   }
 }
 
+.amsterdam-ordered-list--inverse-color,
+.amsterdam-ordered-list--inverse-color:not(.amsterdam-ordered-list--no-markers),
```
VincentSmedinga, line 40: Only a list without markers has a text colour.

```suggestion
.amsterdam-ordered-list--inverse-color:not(.amsterdam-ordered-list--no-markers),
```

packages/css/src/components/ordered-list/ordered-list.scss
```diff
@@ -36,6 +36,12 @@
...
+.amsterdam-ordered-list--inverse-color .amsterdam-ordered-list {
```
VincentSmedinga, line 41: Each instance of a component is responsible for its own appearance. Nested lists should have the `inverseColor` prop explicitly as well.

```suggestion
.amsterdam-ordered-list--inverse-color:not(.amsterdam-ordered-list--no-markers) {
```

packages/css/src/components/unordered-list/unordered-list.scss
```diff
@@ -36,6 +36,12 @@
   }
 }
 
+.amsterdam-unordered-list--inverse-color,
+.amsterdam-unordered-list--inverse-color:not(.amsterdam-unordered-list--no-markers),
+.amsterdam-unordered-list--inverse-color .amsterdam-unordered-list {
```
VincentSmedinga, line 41:
```suggestion
.amsterdam-unordered-list--inverse-color:not(.amsterdam-unordered-list--no-markers) {
```

packages/react/src/OrderedList/OrderedList.tsx
```diff
@@ -10,17 +10,26 @@ import { OrderedListItem } from './OrderedListItem'
...
 } & PropsWithChildren<OlHTMLAttributes<HTMLOListElement>>
 
 type OrderedListComponent = {
   Item: typeof OrderedListItem
 } & ForwardRefExoticComponent<OrderedListProps & RefAttributes<HTMLOListElement>>
 
 export const OrderedList = forwardRef(
-  ({ children, markers = true, className, ...restProps }: OrderedListProps, ref: ForwardedRef<HTMLOListElement>) => (
+  (
+    { children, markers = true, inverseColor, className, ...restProps }: OrderedListProps,
```
VincentSmedinga, line 22: Let’s sort lists of props alphabetically. This can make Git diffs less confusing (although dangling commas help there as well) and helps humans find specific props easier.

```suggestion
    { children, className, inverseColor, markers = true, ...restProps }: OrderedListProps,
```
```suggestion
    { children, markers = true, inverseColor, className, ...restProps }: OrderedListProps,
```

packages/react/src/OrderedList/OrderedList.tsx
```diff
@@ -10,17 +10,26 @@ import { OrderedListItem } from './OrderedListItem'
...
+    ref: ForwardedRef<HTMLOListElement>,
+  ) => (
     <ol
       ref={ref}
-      className={clsx('amsterdam-ordered-list', !markers && 'amsterdam-ordered-list--no-markers', className)}
+      className={clsx(
+        'amsterdam-ordered-list',
+        !markers && 'amsterdam-ordered-list--no-markers',
+        inverseColor && 'amsterdam-ordered-list--inverse-color',
```
VincentSmedinga, line 30: This is only a reorder.

```suggestion
        inverseColor && 'amsterdam-ordered-list--inverse-color',
        !markers && 'amsterdam-ordered-list--no-markers',
```

packages/react/src/UnorderedList/UnorderedList.tsx
```diff
@@ -10,18 +10,27 @@ import { UnorderedListItem } from './UnorderedListItem'
 
 export type UnorderedListProps = {
   markers?: boolean
+  inverseColor?: boolean
```
VincentSmedinga, line 13: This is only a reorder.

```suggestion
  inverseColor?: boolean
  markers?: boolean
```

packages/react/src/UnorderedList/UnorderedList.tsx
```diff
@@ -10,18 +10,27 @@ import { UnorderedListItem } from './UnorderedListItem'
...
 } & PropsWithChildren<HTMLAttributes<HTMLUListElement>>
 
 type UnorderedListComponent = {
   Item: typeof UnorderedListItem
 } & ForwardRefExoticComponent<UnorderedListProps & RefAttributes<HTMLUListElement>>
 
 export const UnorderedList = forwardRef(
-  ({ children, markers = true, className, ...restProps }: UnorderedListProps, ref: ForwardedRef<HTMLUListElement>) => {
+  (
+    { children, markers = true, inverseColor, className, ...restProps }: UnorderedListProps,
```
VincentSmedinga, line 22: This is only a reorder.

```suggestion
    { children, className, inverseColor, markers = true, ...restProps }: UnorderedListProps,
```

packages/react/src/UnorderedList/UnorderedList.tsx
```diff
@@ -10,18 +10,27 @@ import { UnorderedListItem } from './UnorderedListItem'
...
+    ref: ForwardedRef<HTMLUListElement>,
+  ) => {
     return (
       <ul
         ref={ref}
-        className={clsx('amsterdam-unordered-list', !markers && 'amsterdam-unordered-list--no-markers', className)}
+        className={clsx(
+          'amsterdam-unordered-list',
+          !markers && 'amsterdam-unordered-list--no-markers',
+          inverseColor && 'amsterdam-unordered-list--inverse-color',
```
VincentSmedinga, line 31: This is only a reorder.

```suggestion
          inverseColor && 'amsterdam-unordered-list--inverse-color',
          !markers && 'amsterdam-unordered-list--no-markers',
```

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
```diff
@@ -14,10 +14,18 @@ const meta = {
...
   args: {
     children: orderedList,
+    inverseColor: false,
   },
   argTypes: {
     markers: { control: 'boolean' },
   },
+  decorators: [
+    (Story, context) => (
+      <div className={context.args.inverseColor ? 'amsterdam-docs-dark-background' : undefined}>
```
VincentSmedinga, line 24: Nice. Can you still add an initially visible example as well? For both ordered and unordered.

VincentSmedinga (commented):
Thanks for picking this one up! I think the selector should be more specific, we can sort lists of restructured props and prop objects, and we could use explicit examples.

packages/css/src/components/unordered-list/unordered-list.scss
dlnr: This will break the stories with more levels. I did not find a solution to pass `args` to `children` and it would seem like something that will go wrong during normal usage.

How likely will it be that there is a nested list of an `inverseColor` list that does not appear in white?

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
VincentSmedinga: Thanks, I forgot to mention that we use the same text for this prop on all components, so I’ve copied that.

VincentSmedinga (approved):

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
```diff
@@ -104,3 +104,16 @@ export const WithoutMarkers: Story = {
     markers: false,
   },
 }
+
+export const InverseColor: Story = {
+  args: {
+    inverseColor: true,
```
alimpens, line 110: We should add the `inverseColor` boolean to the `argTypes` as well, the control currently isn't visible for all stories. Same for the Unordered list stories.

VincentSmedinga: Uh-oh, don’t let @dlnr read this 🫣. He and I decided earlier today to move the arg and decorator over from the other stories, because a nested list isn’t affected by the arg and would appear incorrectly.

dlnr: So do I add the `argType` that break the stories or leave it to the story to illustrate how the prop works?

alimpens: Preferably neither ;) You can pass the args to the children of a story with the render function, right? So you can pass the value of `inverseColor` to the nested list?

VincentSmedinga: Done!

VincentSmedinga (approved):

PR #1079, opened 2024-02-13
fix: Refactor Mega menu, use correct list category class

alimpens:


packages/react/src/MegaMenu/MegaMenu.tsx
```diff
@@ -26,7 +26,7 @@ const MegaMenuListCategory = forwardRef(
     { children, className, ...restProps }: PropsWithChildren<HTMLAttributes<HTMLDivElement>>,
     ref: ForwardedRef<HTMLDivElement>,
   ) => (
-    <div {...restProps} ref={ref} className={clsx('amsterdam-mega-menu__list-wrapper', className)}>
+    <div {...restProps} ref={ref} className={clsx('amsterdam-mega-menu__list-category', className)}>
```
VincentSmedinga, line 29: I guess we can move this subcomponent into its own file and add our default unit tests, as we did in #1050?

VincentSmedinga: Nice catch though!

VincentSmedinga (approved):

PR #1081, opened 2024-02-13
feat: Add Fieldset component

alimpens:


packages/css/src/components/fieldset/fieldset.scss
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+@import "../../common/hyphenation";
+
+@mixin reset {
+  border: none;
+  margin-inline: 0;
+  padding-block: 0;
+  padding-inline: 0;
+}
```
dlnr, line 13: This is unused.

packages/css/src/components/fieldset/fieldset.scss
```diff
@@ -0,0 +1,37 @@
...
+
+@mixin legend-reset {
+  padding-inline: 0;
+  -webkit-text-size-adjust: 100%;
+}
```
dlnr, line 25: This is only used once. Why make it reusable?

packages/react/src/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export type FieldsetProps = PropsWithChildren<HTMLAttributes<HTMLFieldSetElement>> & {
+  legend: string
```
dlnr, line 11: Legend is not required

packages/css/src/components/fieldset/README.md
```diff
@@ -0,0 +1,12 @@
...
+A component to group related form inputs.
+
+## Guidelines
+
+- Use Fieldset when you need to show a relationship between multiple form inputs. For example, you may need to group a set of text inputs into a single fieldset when asking for an address.
+- When grouping radio inputs, use `role="radiogroup"` on Fieldset to have this grouping explicitly announced as a radio group. Fieldset has a default role of `group`.
+
+## Relevant WCAG Requirements
+
+- [WCAG 1.3.5](https://www.w3.org/WAI/WCAG22/Understanding/identify-input-purpose.html): Fieldset labels the purpose of a group of inputs.
```
dlnr, line 12: THis might help:
- https://www.w3.org/TR/WCAG-TECHS/H71.html

packages/css/src/components/fieldset/fieldset.scss
```diff
@@ -0,0 +1,37 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright (c) 2023 Gemeente Amsterdam
+ */
+
+@import "../../common/hyphenation";
+
+@mixin reset {
```
dlnr, line 8: This is unused

dlnr (commented):
Looks good, weird how the legend has always been a thing that wants to live in the middle of the fieldset line. And also that the legend is not actually required, just for a11y purposed it is recommended.

packages/css/src/components/fieldset/fieldset.scss
alimpens: We have an undocumented convention to wrap reset rules in a mixin, to show the difference between reset rules and true styling rules. We should really document this, I'll make a note of it

alimpens:
> Looks good, weird how the legend has always been a thing that wants to live in the middle of the fieldset line. And also that the legend is not actually required, just for a11y purposed it is recommended.

Yeah, the default styling is a bit odd. I think it makes sense to require the legend in ADS, a fieldset without a legend is a bit pointless imo.

dlnr:
> > Looks good, weird how the legend has always been a thing that wants to live in the middle of the fieldset line. And also that the legend is not actually required, just for a11y purposed it is recommended.
> 
> Yeah, the default styling is a bit odd. I think it makes sense to require the legend in ADS, a fieldset without a legend is a bit pointless imo.

You're probably right, but.. you can pass a `form` attribute to the fieldset to connect it to a form outside of the `form` tag. Perhaps developer might need this in some odd sidecases. And they might not need a heading then.

VincentSmedinga (approved):

PR #1086, opened 2024-02-20
chore: Implement all props on all List stories

VincentSmedinga:
- All list stories show all controls, e.g. for the `inverseColor` prop
- Exception: the ‘Without Markers’ story, where it doesn’t apply
- Restored the `Primary` and `Controls` doc blocks
- Adds a bit of consistency to docs: ‘markers’ and title case for headings
- Gave Boolean args of other stories an initially `false` value as well
- Only `markers` now shows a ‘Set boolean’ button to prevent it from appearing in code to be copied


storybook/storybook-react/src/IconButton/IconButton.stories.tsx
```diff
@@ -12,11 +12,12 @@ const meta = {
   component: IconButton,
   args: {
     label: 'Sluiten',
+    onBackground: undefined,
+    disabled: false,
+    size: undefined,
```
alimpens, line 17: This doesn't do anything, I think. Maybe we should make `level-5` undefined, similar to what we do with `onBackground`?

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
```diff
@@ -7,19 +7,29 @@ import { Heading, OrderedList, Paragraph } from '@amsterdam/design-system-react'
...
   args: {
-    children: orderedList,
-    inverseColor: undefined,
+    children: orderedListItems,
+    inverseColor: false,
     markers: undefined,
   },
   argTypes: {
+    inverseColor: { control: 'boolean' },
     markers: { control: 'boolean' },
```
alimpens, line 24: We generally hide `children` from the controls

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
```diff
@@ -105,19 +115,17 @@ export const WithoutMarkers: Story = {
     ],
     markers: false,
   },
+  argTypes: {
+    inverseColor: {
+      table: { disable: true },
+    },
+  },
```
alimpens, line 122: I'd keep the controls for all stories the same, it's good to show that certain combinations of props don't work imo

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
alimpens, line 24: Do we want to add `start` and `reversed` here as well? That would unify all controls for all stories

VincentSmedinga: Yes, I’ve configured that globally now. It seems that the arg still appears in the table if it has an initial value, e.g. for `Button` or `Heading`.

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
VincentSmedinga: Hmm… alright.

storybook/storybook-react/src/OrderedList/OrderedList.stories.tsx
alimpens: That works. I think we sometimes do want to show the `children` control though, it's useful to be able to change the text of a Button or Heading imo. Is that possible?

VincentSmedinga: Whoah! I had checked that and have seen `children` args with this change in effect, but that must have been cache or something – they’re now gone indeed. I’ve specified them explicitly where we want to see them.

storybook/storybook-react/src/Button/Button.stories.tsx
```diff
@@ -11,20 +11,16 @@ const meta = {
   title: 'Buttons/Button',
   component: Button,
   args: {
-    children: 'Default',
+    children: 'Tekst',
```
alimpens, line 14: I'd use the text of the primary story here, no need to set text we'll never show imo

storybook/storybook-react/src/PageHeading/PageHeading.docs.mdx
```diff
@@ -6,12 +6,14 @@ import README from "../../../../packages/css/src/components/page-heading/README.
 
 <Markdown>{README}</Markdown>
 
+## Examples
```
alimpens, line 9: We generally use this heading after the Primary and Controls blocks, no?

storybook/storybook-react/src/Button/Button.stories.tsx
```diff
@@ -11,20 +11,16 @@ const meta = {
...
   },
   argTypes: {
     children: {
-      table: {
-        disable: true,
-      },
-    },
-    disabled: {
-      control: 'boolean',
+      table: { disable: false },
```
alimpens, line 21: We should hide `children` in the ButtonWithAnIcon story again, I think

storybook/storybook-react/src/PageHeading/PageHeading.docs.mdx
VincentSmedinga: No, wasn’t my idea. It’s the primary example, right?

storybook/storybook-react/src/Button/Button.stories.tsx
VincentSmedinga: I thought ‘Tekst’ would be the best global default for the `children` prop. I had been typing ‘Button label’, ‘Checkbox label’, ‘Form label’ here but 1 that’s English en 2. too much work.

alimpens: But we don't really need a global default, right? We always have a primary story, so the `children` prop can just have the primary story text, 'Primary' in this case.

storybook/storybook-react/src/PageHeading/PageHeading.docs.mdx
alimpens: Hm, I see we don't really follow a convention here. Sometimes ## Examples is above the Primary story, sometimes it's below. And sometimes we don't use it at all 😅 Let leave it for now and discuss this in [this ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-598)

storybook/storybook-react/src/Button/Button.stories.tsx
VincentSmedinga: Alright, [updated](https://github.com/Amsterdam/design-system/pull/1086/commits/aaa891a90e4d8a7e1e9e37363c06f8649e7c3ce4) for 4 components.

alimpens: I think you misunderstood me, the primary button label should be 'Primary'

VincentSmedinga: Sorry, I disagree. I don’t want to mix up examples with documentation. Examples should display realistic content or an essential phrase; the variation of a component is to be inferred from the title and the props.

We could go for ‘Opslaan’, ‘Bewerken’ etc. if you prefer, although we should think about the best phrase for a link, a checkbox and a form label as well.

alimpens (dismissed):

PR #1088, opened 2024-02-21
feat: Logo Museum Weesp 

dlnr:
Added Logo for Museum Weesp with Figma markup of the SVG. If you like the solution I can setup the other logo in Figma just like this.


dlnr (commented):
Added proposition to use var in fill. This way you wouldn't need a css for logo and you can set dark mode with an override.

packages/react/src/Logo/brand/LogoMuseumWeesp.tsx
```diff
@@ -0,0 +1,45 @@
...
+      fill="var(--amsterdam-logo-title-color, #EC0000)"
+    />
+    <path
+      className="amsterdam-logo__emblem"
+      d="M0 38.3642L3.6 34.9356L0 31.507L2.4 29.2213L6 32.6499L9.6 29.2213L12 31.507L8.4 34.9356L12 38.3642L9.6 40.6499L6 37.2213L2.4 40.6499L0 38.3642Z"
+      fill="var(--amsterdam-logo-emblem-color, #EC0000)"
```
VincentSmedinga, line 28: What’s the added value? The CSS class already applies the same custom property.

This also makes the image less portable – it requires the device rendering it to support CSS in the `fill` attribute.

packages/react/src/Logo/brand/LogoMuseumWeesp.tsx
```diff
@@ -0,0 +1,45 @@
+import { forwardRef } from 'react'
+import type { ForwardedRef, SVGProps } from 'react'
+
+const LogoMuseumWeesp = forwardRef((props: SVGProps<SVGSVGElement>, ref: ForwardedRef<SVGSVGElement>) => (
```
VincentSmedinga, line 4: Don’t we generate the logo components from the SVG sources yet, like we do with the icons?

The plain SVG is missing from `proprietary/assets/logo`.

dlnr: We're talking about the same ticket right? DES-589?

VincentSmedinga: It wasn’t, just a question, I could have looked up the answer myself. So please add the SVG for consistency and don’t worry about generating components.

packages/react/src/Logo/brand/LogoMuseumWeesp.tsx
dlnr: If a device support rendering an svg with `fill="#EC0000"` I think it supports CSS. Using variables has the added benefit that you can easily scope those custom properties for example on a dark background with messing with stylesheets. Basically I'm trying to convince you guys you can have a `<style>` tag in your svg so you can support dark mode.

VincentSmedinga (dismissed):

packages/react/src/Logo/brand/LogoMuseumWeesp.tsx
dlnr: This is removed

alimpens (dismissed):

VincentSmedinga (approved):

PR #1089, opened 2024-02-21
feat!: Use spacing tokens in Grid component

VincentSmedinga:
This makes the Grid component use the new Space tokens, which were, in their turn, based on the Grid gap token.

- ~~Uses Sass to generate classes for gap and padding. I intended to use shorter, local variable names, but prefixing with `amsterdam` is required.~~
- ~~Introduces a `.amsterdam-grid--gap--medium` class that wasn’t previously there as the base component includes the medium gap class.~~
- The compact grid gap token is no longer needed; it now uses the token from compact mode if it’s loaded. The padding one is because its value is different.
- Replaces ‘white space’ with ‘gap’ in the docs; it is more specific and matches nicely with the prop name.

The change is breaking because:
1. the `amsterdam-grid-gap` token is now `amsterdam-grid-gap-md`
2. the `large` vertical padding of the grid is now 1.5 (instead of 2) times as wide as the regular gap width to match the space token naming


packages/css/src/components/grid/grid.scss
```diff
@@ -3,11 +3,37 @@
  * Copyright (c) 2023 Gemeente Amsterdam
  */
 
+@use "sass:list";
+@use "sass:map";
```
alimpens, line 7: Using these functions creates kind of hard to follow loops, and it doesn't really reduce the amount of code. It's not really an improvement imo.

proprietary/tokens/src/components/amsterdam/grid.tokens.json
```diff
@@ -3,13 +3,20 @@
     "grid": {
       "column-count": { "value": "4" },
       "gap": {
```
alimpens, line 5: Maybe we should separate `column-gap` and `row-gap`? `column-gap` only needs one value, makes it a bit easier to follow maybe.

proprietary/tokens/src/components/amsterdam/grid.compact.tokens.json
```diff
@@ -1,14 +1,7 @@
...
     "grid": {
-      "gap": {
-        "value": "clamp(1rem, calc(1.5625vw - 0.0625rem), 2.5rem)",
-        "comment": "Grows from 16px at 1088px wide to 40px at 2624px wide."
-      },
-      "padding-inline": {
-        "value": "clamp(1rem, calc(1.5625vw - 0.0625rem), 2.5rem)",
-        "comment": "Equals the gap."
-      }
+      "padding-inline": { "value": "amsterdam.space.md" }
```
alimpens, line 4:
```suggestion
      "padding-inline": { "value": "{amsterdam.space.md}" }
```

Same for `grid.compact.tokens.json`

packages/css/src/components/grid/grid.scss
```diff
@@ -24,48 +50,20 @@
...
-  padding-block-start: calc(var(--amsterdam-grid-gap) * 2);
-}
-
-.amsterdam-grid--padding-vertical--small {
-  padding-block: calc(var(--amsterdam-grid-gap) / 2);
-}
-
-.amsterdam-grid--padding-vertical--medium {
-  padding-block: var(--amsterdam-grid-gap);
+@each $class, $property in $amsterdam-grid-gap-classes {
```
dlnr, line 53: This has only one value, a loop should not be needed. Never in CSS IMO.

packages/css/src/components/grid/grid.scss
```diff
@@ -3,11 +3,37 @@
...
+  "gap-vertical": "row-gap",
+);
+$amsterdam-grid-padding-classes: (
+  "padding-bottom": "padding-block-end",
+  "padding-top": "padding-block-start",
+  "padding-vertical": "padding-block",
+);
+
+// Filter sizes to sm, md, lg
+@each $size, $label in $amsterdam-sizes {
```
dlnr, line 23: Not a big fan of sass loops. It is not very readable and since there are only a couple of options I would prefer to have them all here.

proprietary/tokens/src/brand/amsterdam/space.compact.tokens.json
```diff
@@ -3,7 +3,10 @@
     "space": {
       "xs": { "value": "clamp(0.25rem, calc(0.390625vw - 0.015625rem), 0.625rem)" },
       "sm": { "value": "clamp(0.5rem, calc(0.78125vw - 0.03125rem), 1.25rem)" },
-      "md": { "value": "clamp(1rem, calc(1.5625vw - 0.0625rem), 2.5rem)" },
+      "md": {
+        "value": "clamp(1rem, calc(1.5625vw - 0.0625rem), 2.5rem)",
+        "comment": "Grows from 16px at 1088px wide to 40px at 2624px wide."
```
dlnr, line 8: Why does only 'md' have a description?

proprietary/tokens/src/components/amsterdam/grid.tokens.json
```diff
@@ -3,13 +3,20 @@
...
-      "padding-inline": {
-        "value": "clamp(1.5rem, calc(4.6875vw + 0.5625rem), 5.25rem)",
-        "comment": "Equals 1.5 times the gap."
+      "padding-block": {
+        "xs": { "value": "amsterdam.space.xs" },
+        "sm": { "value": "amsterdam.space.sm" },
+        "md": { "value": "amsterdam.space.md" },
+        "lg": { "value": "amsterdam.space.lg" },
+        "xl": { "value": "amsterdam.space.xl" }
       },
```
dlnr, line 18: I'm not getting this, all the grid.gap xs - xl values are stored in the mobile (4 cols) section of the grid and medium/wide do not hold any of those values. I think the hierarchy does not make sense. I would expect grid>gap to hold 1 value, and the grid>medium>gap etc.

packages/css/src/components/grid/grid.scss
VincentSmedinga: It thought it was consistent and forward-looking, but it’s been removed due to the other comments.

proprietary/tokens/src/brand/amsterdam/space.compact.tokens.json
VincentSmedinga: Because `md` is the base value from which the other ones are derived. The comment has been copied over from the grid tokens, where ‘medium’ was the only one.

We could document the multiplication factors of ¼, ½, 1½, and 2, but I guess the minimum values already do so?

proprietary/tokens/src/components/amsterdam/grid.compact.tokens.json
VincentSmedinga: Thanks

proprietary/tokens/src/components/amsterdam/grid.tokens.json
VincentSmedinga: Now it’s my turn to think this would be overcomplicating things.

The CSS `gap` property is a shorthand that sets the same value for both `column-gap` and `row-gap`. I think that’s what we want: the same options for both horizontal and vertical spacing. I wouldn’t expect the community to need two different scales, either.

Is it worthwhile to have both `amsterdam-grid-column-gap` and `amsterdam-grid-row-gap` with the same options?

proprietary/tokens/src/components/amsterdam/grid.tokens.json
VincentSmedinga: I understand how it looks like that, but it’s really not different from how we define variations for other components.

In this context of a grid, it reflects our thinking ‘mobile first’. All default values are in the `amsterdam.grid` namespace, including the 4 columns it initially has. In the `medium` and `wide` variations, only the value for `column-count` needs to change; the options for `gap` and `padding-*` remain the same.

We could move `"column-count": { "value": "4" },` into a `"narrow"` object to clarify this, but I think it would deviate from our regular approach.

proprietary/tokens/src/components/amsterdam/grid.tokens.json
```diff
@@ -3,13 +3,20 @@
     "grid": {
       "column-count": { "value": "4" },
       "gap": {
-        "value": "clamp(1rem, calc(3.125vw + 0.375rem), 3.5rem)",
-        "comment": "Grows from 16px at 320px wide to 56px at 1600px wide."
+        "xs": { "value": "{amsterdam.space.xs}" },
+        "sm": { "value": "{amsterdam.space.sm}" },
+        "md": { "value": "{amsterdam.space.md}" },
+        "lg": { "value": "{amsterdam.space.lg}" },
+        "xl": { "value": "{amsterdam.space.xl}" }
```
alimpens, line 10: `--amsterdam-grid-gap-xs` and `xl` are never used, why are they defined?

proprietary/tokens/src/components/amsterdam/grid.tokens.json
```diff
@@ -3,13 +3,20 @@
...
       },
-      "padding-inline": {
-        "value": "clamp(1.5rem, calc(4.6875vw + 0.5625rem), 5.25rem)",
-        "comment": "Equals 1.5 times the gap."
+      "padding-block": {
+        "xs": { "value": "{amsterdam.space.xs}" },
```
alimpens, line 13: Same here, `xs` and `xl` are never used

packages/css/src/components/grid/grid.scss
```diff
@@ -25,47 +25,41 @@
...
 .amsterdam-grid--padding-vertical--medium {
-  padding-block: var(--amsterdam-grid-gap);
+  padding-block-start: var(--amsterdam-grid-padding-block-md);
 }
 
+.amsterdam-grid--padding-top--large,
 .amsterdam-grid--padding-vertical--large {
-  padding-block: calc(var(--amsterdam-grid-gap) * 2);
+  padding-block-start: var(--amsterdam-grid-padding-block-lg);
 }
```
alimpens, line 63: The previous way of organising the classes results in only a couple of characters more, while improving the readability a lot imo.

(Sorry, adding a suggestion is not supported)

```
.amsterdam-grid--padding-bottom--small {
	padding-block-end: var(--amsterdam-grid-padding-block-sm);
}

.amsterdam-grid--padding-bottom--medium {
	padding-block-end: var(--amsterdam-grid-padding-block-md);
}

.amsterdam-grid--padding-bottom--large {
	padding-block-end: var(--amsterdam-grid-padding-block-lg);
}

.amsterdam-grid--padding-top--small {
	padding-block-start: var(--amsterdam-grid-padding-block-sm);
}

.amsterdam-grid--padding-top--medium {
	padding-block-start: var(--amsterdam-grid-padding-block-md);
}

.amsterdam-grid--padding-top--large {
	padding-block-start: var(--amsterdam-grid-padding-block-lg);
}

.amsterdam-grid--padding-vertical--small {
	padding-block: var(--amsterdam-grid-padding-block-sm);
}

.amsterdam-grid--padding-vertical--medium {
	padding-block: var(--amsterdam-grid-padding-block-md);
}

.amsterdam-grid--padding-vertical--large {
	padding-block: var(--amsterdam-grid-padding-block-lg);
}
```

proprietary/tokens/src/components/amsterdam/grid.tokens.json
VincentSmedinga: Yes, `Grid` only allows for a `small` or `large` gap and vertical padding. We never really tested these quick choices. Let’s keep them unchanged (although `large` is now 1.5x as large as `medium`, not 2x) and add the `extra-` options for Grid gap and padding if we need them. Removed these two.

proprietary/tokens/src/components/amsterdam/grid.tokens.json
VincentSmedinga: Removed.

packages/react/src/Grid/Grid.test.tsx
```diff
@@ -1,63 +1,117 @@
...
+  })
+
+  it(`renders the correct class name for a small bottom padding`, () => {
+    const { container } = render(<Grid paddingBottom="small" />)
+
     const component = container.querySelector(':only-child')
-    expect(component).toHaveClass('amsterdam-grid--gap-vertical--small amsterdam-grid--padding-vertical--large')
+
+    expect(component).toHaveClass('amsterdam-grid--padding-bottom--small')
+  })
```
alimpens, line 76: These are the same tests, and they are already handled by the next loop, no?

alimpens (approved):

dlnr (approved):

PR #1091, opened 2024-02-22
fix!: Remove whitespace at the right end of dialog content

VincentSmedinga:
I can’t explain why we need whitespace at the right of the Dialog body.


packages/css/src/components/dialog/dialog.scss
```diff
@@ -36,7 +36,6 @@
   max-block-size: 100%; /* safari */
   overflow-y: auto;
   overscroll-behavior-y: contain;
-  padding-inline-end: var(--amsterdam-dialog-article-padding-inline-end);
```
VincentSmedinga, line 39: The intention was to reserve space for a scrollbar. I don’t think this is the way to do that – an arbitrary width of white space in a single component. It’s the OS's responsibility to fix this. Please confirm things look good on macOS, Windows, iOS and Android, then merge this. If necessary, create a new ticket to research how we should deal with scrollbars in general.

alimpens: On Windows and Android the text sometimes touches the scrollbar, we might want to reserve space to prevent that?

scrollbar

dlnr: On iOS and macos the scrollbar overlaps the text without the padding.
Screenshot 2024-02-23 at 14 12 55 (2)

VincentSmedinga: I'm still not convinced we must add this space explicitly.

On iOS and macOS, the scrollbar is transparent and only shows when scrolling (unless the user configures otherwise) – probably exactly to keep the right margin of the overflowing content aligned with its surroundings.

On Windows, the text does come close to the scrollbar in some of the lines, but they never overlap, and it doesn’t appear broken.

A scrollbar will not display if the content doesn’t overflow, making the alignment to the right look bad.

The current width of the whitespace is an arbitrary number, doesn’t have a token, and isn’t applied to other containers that may scroll.

So again I propose we merge this.

alimpens: > A scrollbar will not display if the content doesn’t overflow, making the alignment to the right look bad.

> The current width of the whitespace is an arbitrary number, doesn’t have a token, and isn’t applied to other containers that may scroll.

Good points imo, we can merge this as far as I'm concerned

alimpens (approved):

PR #1093, opened 2024-02-22
feat: Add and use tokens for inside spacing

VincentSmedinga:
This introduces tokens for inside space. No breaking changes, only additions.

They have fixed sizes; making them fluid doesn’t seem necessary. A large difference between the min and max values makes for too much space on wide screens and too little on narrow ones.

Five options again. The default in the spacious theme is 1.5rem – 1rem seemed logical but appeared to be too small in practice.

Applied them to all existing components, which changes some sizing which has to be fed back to the designers for review and application in Figma. Most notably, the button has become a bit wider and slightly less tall, making it look much more button-y, in my opinion.

Note: #1089 and #1091 should be merged first.

### DIFF
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -26,6 +26,8 @@
   gap: var(--amsterdam-dialog-form-gap);
   grid-template-rows: auto 1fr auto;
   max-block-size: var(--amsterdam-dialog-form-max-block-size);
+
+  // TODO Decide on these widths
   padding-block: var(--amsterdam-dialog-form-padding-block);
   padding-inline: var(--amsterdam-dialog-form-padding-inline);
 }
--- a/packages/css/src/components/header/header.scss
+++ b/packages/css/src/components/header/header.scss
@@ -9,7 +9,7 @@
   align-items: center;
   display: flex;
   flex-wrap: wrap;
-  padding-block: 1rem;
+  padding-block: var(--amsterdam-header-padding-block);
   row-gap: 1.5rem;
 
   @media screen and (min-width: $amsterdam-breakpoint-wide) {
--- a/proprietary/tokens/src/brand/amsterdam/space.tokens.json
+++ b/proprietary/tokens/src/brand/amsterdam/space.tokens.json
@@ -5,7 +5,14 @@
       "sm": { "value": "clamp(0.5rem, calc(1.5625vw + 0.1875rem), 1.75rem)" },
       "md": { "value": "clamp(1rem, calc(3.125vw + 0.375rem), 3.5rem)" },
       "lg": { "value": "clamp(1.5rem, calc(4.6875vw + 0.5625rem), 5.25rem)" },
-      "xl": { "value": "clamp(2rem, calc(6.25vw + 0.75rem), 7rem)" }
+      "xl": { "value": "clamp(2rem, calc(6.25vw + 0.75rem), 7rem)" },
+      "inside": {
+        "xs": { "value": ".375rem" },
+        "sm": { "value": ".75rem" },
+        "md": { "value": "1.5rem" },
+        "lg": { "value": "2.25rem" },
+        "xl": { "value": "3rem" }
+      }
     }
   }
 }
--- a/packages/css/src/components/accordion/accordion.scss
+++ b/packages/css/src/components/accordion/accordion.scss
@@ -24,8 +24,8 @@
   font-weight: var(--amsterdam-accordion-button-font-weight);
   justify-content: space-between;
   line-height: var(--amsterdam-accordion-button-line-height);
-  padding-block: 0.75rem;
-  padding-inline: 1rem;
+  padding-block: var(--amsterdam-accordion-button-padding-block);
+  padding-inline: var(--amsterdam-accordion-button-padding-inline);
   width: 100%;
 
   &:focus {
@@ -54,8 +54,8 @@
 
 .amsterdam-accordion__panel {
   display: none;
-  padding-block: 1rem;
-  padding-inline: 1rem;
+  padding-block: var(--amsterdam-accordion-panel-padding-block);
+  padding-inline: var(--amsterdam-accordion-panel-padding-inline);
 }
 
 .amsterdam-accordion__panel--expanded {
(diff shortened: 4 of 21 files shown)


proprietary/tokens/src/brand/amsterdam/space.tokens.json
```diff
@@ -8,7 +8,14 @@
         "comment": "Grows from 16px at 320px wide to 56px at 1600px wide."
       },
       "lg": { "value": "clamp(1.5rem, calc(4.6875vw + 0.5625rem), 5.25rem)" },
-      "xl": { "value": "clamp(2rem, calc(6.25vw + 0.75rem), 7rem)" }
+      "xl": { "value": "clamp(2rem, calc(6.25vw + 0.75rem), 7rem)" },
+      "inside": {
```
dlnr, line 12: After reading the docs I would expect a different structure here:

* space
	* grid
		* xs
		* sm
		* md
		* lg
		* xl
	* stack
		* xs
		* sm
		* md
		* lg
		* xl
	* inside
		* xs
		* sm
		* md
		* lg
		* xl
	* type
		* xs
		* sm
		* md
		* lg
		* xl

packages/css/src/components/header/header.scss
```diff
@@ -38,7 +38,7 @@
 
 .amsterdam-header__menu {
   flex: 1;
-  padding-inline-start: var(--amsterdam-page-menu-column-gap);
+  padding-inline-start: var(--amsterdam-page-menu-column-gap); // TODO Don’t use tokens of another component
```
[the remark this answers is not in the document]
dlnr: If i remember correctly I had to align this to the grid, which would imply that is uses the same tokens as the grid.

proprietary/tokens/src/brand/amsterdam/space.tokens.json
VincentSmedinga: Yes, that’s how it will be. I mentioned having to group the grid tokens in the PR description.

packages/css/src/components/header/header.scss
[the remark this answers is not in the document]
VincentSmedinga: Yeah it seems to be a gap between the logo and the menu button on narrow screens, we’ll check that out later.

proprietary/tokens/src/brand/amsterdam/space.tokens.json
VincentSmedinga: I’ve prepared that separate PR: #1098

alimpens:
> Applied them to all existing components, which changes some sizing which has to be fed back to the designers for review and application in Figma. Most notably, the button has become a bit wider and slightly less tall, making it look much more button-y, in my opinion.

Hm, I don't think it's a good idea to change styling so it fits the system at this point. Seeing as we're just setting up the system, it should change to fit the styling imo. 

To be more specific: I think the new button design looks off. Maybe we should discuss this with design first, before we introduce these changes?

VincentSmedinga:
> Hm, I don't think it's a good idea to change styling so it fits the system at this point. Seeing as we're just setting up the system, it should change to fit the styling imo.

I would present it as bring System into the Design. Most current choices for whitespace appear consistent, but not all, e.g. a horizontal padding of 16 for an Accordion Panel but 24 for an Alert. Introducing a solid set of design tokens while having a good set of designed components creates the oppurtunity to consider white space more holistically, which may indeed change the value we choose for some components and hence their visual appearence.

The root of the changes is in the fact that my research indicates 1.5rem to be the best value for the medium amount of whitespace, where the designs are based on 1rem. Transforming the current to the proposed values requires making some choices when rounding up and down. As everyting in a PR, this is a proposal and a request for feedback.

What I don’t think we can do is 100% implement the current values and have a consistent scale at the same time.

> To be more specific: I think the new button design looks off. Maybe we should discuss this with design first, before we introduce these changes?

Whether the visual changes are improvements is indeed to be discussed, as I mentioned in the PR description. I had already brought Inge up to speed to my effort. Most probably we can look into this together this Tuesday or Wednesday. I think we should evaluate the assignment of `xs` – `xl` values to all components, then visually find the best numbers of pixels for the sizing scale.

alimpens:
> The root of the changes is in the fact that my research indicates 1.5rem to be the best value for the medium amount of whitespace, where the designs are based on 1rem. Transforming the current to the proposed values requires making some choices when rounding up and down. As everyting in a PR, this is a proposal and a request for feedback.

I guess I don't really understand why 1.5rem would be the best medium value. 1rem is used for padding in something like 9 components, whereas 1.5rem is used in 5, I think? Why is 1.5 better, in your opinion?

VincentSmedinga:
> Why is 1.5 better, in your opinion?

Happy to explore this, but rather not through GitHub comments! 😊

alimpens:
> Also, again, it’s not about the numbers but about the approach.

I think the numbers are kind of important :D But you're right, let's discuss this in realtime!

dlnr (approved):

proprietary/tokens/src/components/amsterdam/alert.tokens.json
```diff
@@ -4,10 +4,10 @@
       "border-width": { "value": "4px" },
       "border-style": { "value": "solid" },
       "gap": { "value": "1rem" },
-      "padding-block-start": { "value": "1rem" },
-      "padding-block-end": { "value": "1rem" },
-      "padding-inline-start": { "value": "1.5rem" },
-      "padding-inline-end": { "value": "1.5rem" },
+      "padding-block-start": { "value": "{amsterdam.space.inside.md}" },
```
alimpens, line 7: Didn't we want to merge `padding-block-start` and `*-end` into one style rule? Or will we do that in a different PR?

VincentSmedinga: Yes, I’d prefer a separate change for that because it is a breaking change.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -26,6 +26,8 @@
   gap: var(--amsterdam-dialog-form-gap);
   grid-template-rows: auto 1fr auto;
   max-block-size: var(--amsterdam-dialog-form-max-block-size);
+
+  // TODO Decide on these widths
   padding-block: var(--amsterdam-dialog-form-padding-block);
   padding-inline: var(--amsterdam-dialog-form-padding-inline);
```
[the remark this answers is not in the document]
VincentSmedinga: [DES-638](https://gemeente-amsterdam.atlassian.net/browse/DES-638).

alimpens (approved):

PR #1094, opened 2024-02-23
chore: Add and update copyright notices

alimpens:


documentation/code-reviews.md
```diff
@@ -1,3 +1,5 @@
+<!-- @license CC0-1.0 -->
```
dlnr, line 1: Are we mixing two licenses?

dlnr (commented):
Removing the year is fine but why the Creative Commons license in components we are supposed to share with NLDS?

VincentSmedinga (approved):

PR #1095, opened 2024-02-26
feat: Add TextArea Component

dlnr:
This pull request adds a new TextArea component and its corresponding documentation. The component allows users to input text over multiple lines. It includes features such as vertical and horizontal resizing, growing as the user types, and support for invalid and disabled states. The documentation provides guidelines on when to use the TextArea component and how to ensure accessibility.


packages/css/src/components/text-area/README.md
```diff
@@ -0,0 +1,21 @@
+# TextArea
+
+A form field that allows the user to input text over multiple lines.
+
+## Guidelines
+
+- Use a TextArea when users need to enter more than 1 sentence of text, such as a comment or description.
+- The height of the TextArea should be appropriate for the information to be entered.
+  - Use cols and rows attributes to set the width and height of the TextArea.
```
alimpens, line 9: The `cols` attribute doesn't really work with our current implementation. We give `textarea` `width: 100%`, which overrides `cols`. What do we want to use, `cols` or `width: 100%`?

`cols` are easier to customise, `width: 100%` works better with `Grid.Cell`

packages/css/src/components/text-area/README.md
```diff
@@ -0,0 +1,21 @@
...
+  - Use the experimental grow attribute to allow the TextArea to grow in height as the user types, this only works in modern browsers.
```
alimpens, line 10: The current implementation breaks the layout for browsers that do not support this property, again because of the `width: 100%` thing. We should decide what we want to do there

packages/css/src/components/text-area/README.md
```diff
@@ -0,0 +1,21 @@
...
+- A TextArea must have a label, and in most cases, this label should be visible.
+- Use `spellcheck="false"` for fields that may contain sensitive information, such as passwords and personal data.
+  Some browser extensions for spell-checking send this information to external servers.
+
+## Accessibility
+
+- [WCAG 2.1](https://www.w3.org/WAI/WCAG21/Techniques/html/H91.html) requires that the TextArea has a label ot title attribute.
```
alimpens, line 17:
```suggestion
- [WCAG 2.1](https://www.w3.org/WAI/WCAG21/Techniques/html/H91.html) requires that the TextArea has a label or title attribute.
```

packages/css/src/components/text-area/text-area.scss
```diff
@@ -0,0 +1,71 @@
...
+
+.amsterdam-text-area::placeholder {
+  color: var(--amsterdam-text-area-placeholder-color);
+}
+
+.amsterdam-text-area:disabled {
+  background-color: var(--amsterdam-text-area-disabled-background-color);
+  box-shadow: var(--amsterdam-text-area-disabled-box-shadow);
+  color: var(--amsterdam-text-area-disabled-color);
+  cursor: not-allowed;
```
alimpens, line 42: You can use the action tokens for this in the component level tokens.

packages/css/src/components/text-area/text-area.scss
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+.amsterdam-text-area--resize-none {
+  resize: none;
+}
+
+.amsterdam-text-area--resize-horizontal {
+  resize: horizontal;
+}
```
alimpens, line 61: Do we want to allow horizontal resizing? This breaks the container it's in (eg a `Grid.Cell`)

packages/react/src/TextArea/TextArea.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import userEvent from '@testing-library/user-event'
+import { createRef, useState } from 'react'
+import { TextArea } from './TextArea'
+import '@testing-library/jest-dom'
+
+describe('Text area', () => {
+  it('renders', () => {
+    const { container } = render(<TextArea />)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 11: Use the correct role here

packages/react/src/TextArea/TextArea.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  it('renders an additional class name', () => {
+    const { container } = render(<TextArea className="extra" />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('amsterdam-text-area extra')
+  })
+
+  it('should be working in a controlled state', async () => {
```
alimpens, line 33:
```suggestion
  it('can be controlled', async () => {
```

packages/react/src/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,32 @@
...
+      {...restProps}
+      ref={ref}
+      className={clsx(
+        'amsterdam-text-area',
+        resize && `amsterdam-text-area--resize-${resize}`,
+        grow && 'amsterdam-text-area--grow',
+        className,
+      )}
+    >
+      {children}
```
alimpens, line 27: [`textarea` doesn't accept children in React](https://react.dev/reference/react-dom/components/textarea#props)

packages/css/src/components/text-area/text-area.scss
```diff
@@ -0,0 +1,71 @@
...
+
+.amsterdam-text-area {
+  border: none;
+  box-shadow: var(--amsterdam-text-area-box-shadow);
+  color: var(--amsterdam-text-area-color);
+  font-family: var(--amsterdam-text-area-font-family);
+  font-size: var(--amsterdam-text-area-font-size);
+  font-weight: var(--amsterdam-text-area-font-weight);
+  line-height: var(--amsterdam-text-area-line-height);
+  min-height: var(--amsterdam-text-area-min-height);
```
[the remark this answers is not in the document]
dlnr: min-height is removed but I do not think this will conflict with the rows. Either way cols and rows are attributes you can add to the textarea.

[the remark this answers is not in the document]
dlnr: We can add a modifier to set width and/or height to auto if cols/rows is set.

packages/css/src/components/text-area/README.md
dlnr: Width is set to auto for textareas that grow. Browser support is > 97%.

packages/css/src/components/text-area/text-area.scss
dlnr: The ticket described this for me but if you and @VincentSmedinga can come to a solution I'll change it.

storybook/src/components/TextInput/TextInput.stories.tsx
dlnr, line 1: Don't know how this got in here or why but it is removed.

packages/css/src/components/text-area/README.md
```diff
@@ -0,0 +1,20 @@
+# TextArea
```
alimpens, line 1:
```suggestion
<!-- @license CC0-1.0 -->

# TextArea
```

packages/react/src/TextArea/README.md
```diff
@@ -0,0 +1,3 @@
+# React TextArea component
```
alimpens, line 1:
```suggestion
<!-- @license CC0-1.0 -->

# React TextArea component
```

packages/css/src/components/text-area/README.md
dlnr: https://caniuse.com/?search=form-sizing%3A%20content

packages/css/src/components/text-area/README.md
dlnr: I've set the width to auto whenever cols are used.

alimpens: Nice, we can use both then. I would add `max-width: 100%` as well (same for resizing), so you can never break out of the container

packages/css/src/components/text-area/README.md
alimpens: You're not using `form-sizing`, but `field-sizing`.

packages/css/src/components/text-area/text-area.scss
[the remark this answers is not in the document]
alimpens: I've thought about it a bit, I think `min-height` is a good idea. We should make it the height of the text + padding though, so it looks exactly the same as `rows="1"`.

packages/css/src/components/text-area/README.md
dlnr: Don't remember how I got that reference but I still think we should leave it in :)

packages/css/src/components/text-area/text-area.scss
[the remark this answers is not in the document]
dlnr: calculating it is kind of awkward, because the token for line-height returns "1.6" and you can't do `calc(1.6 + 1rem)`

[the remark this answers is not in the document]
dlnr: Setting it to `2.6rem` will make the min-height smaller then the regular text input. Any ideas how to calculate the min-height?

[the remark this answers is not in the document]
dlnr: I can do a `calc(1.6em + 1rem)` that works

packages/css/src/components/text-area/README.md
alimpens: Well, the current implementation reverts TextArea to its native sizing for browsers that do not support `field-sizing` (which is pretty much all of them, I think), which is kind of unusable. So in its current implementation, I would not leave it in.

packages/react/src/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,31 @@
...
+      {...restProps}
+      ref={ref}
+      className={clsx(
+        'amsterdam-text-area',
+        resize && `amsterdam-text-area--resize-${resize}`,
+        grow && 'amsterdam-text-area--grow',
+        restProps.cols && 'amsterdam-text-area--cols',
+        className,
+      )}
+    ></textarea>
```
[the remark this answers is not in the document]
dlnr: Fixed this

packages/css/src/components/text-area/README.md
dlnr: Added max-width

packages/css/src/components/text-area/README.md
dlnr: removed

packages/react/src/TextArea/TextArea.test.tsx
dlnr: Applied suggestion

storybook/storybook-react/src/TextInput/TextInput.stories.tsx
```diff
@@ -42,6 +42,12 @@ type Story = StoryObj<typeof meta>
 
 export const Default: Story = {}
 
+export const DefaultValue: Story = {
+  args: {
+    defaultValue: 'Default value',
+  },
+}
+
```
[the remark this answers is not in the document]
dlnr: removed

storybook/src/components/TextArea/TextArea.docs.mdx
```diff
@@ -0,0 +1,27 @@
...
+<Controls />
+
+## Vertical Resize
+
+<Canvas of={TextAreaStories.VerticalResize} />
+
+## Horizontal Resize
+
+<Canvas of={TextAreaStories.HorizontalResize} />
+
```
alimpens, line 20: We're missing a story for no resize here

storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const paragraph = exampleParagraph()
+
+type StoryProps = TextAreaProps & { invalid?: boolean }
+
+const meta = {
+  title: 'Components/Forms/Text Area',
+  component: TextArea,
+  args: {
+    children: paragraph,
```
alimpens, line 19: [textarea doesn't accept children in React](https://react.dev/reference/react-dom/components/textarea#props), you should set the contents with `value` or `defaultValue`

storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+    disabled: false,
+  },
+  argTypes: {
+    resize: {
+      control: {
+        type: 'select',
+      },
+      options: ['default', 'none', 'horizontal', 'vertical'],
```
alimpens, line 27: `'default'` causes a non-existant `amsterdam-text-area--resize-default` to be added

packages/react/src/TextArea/TextArea.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  it('renders an additional class name', () => {
+    render(<TextArea className="extra" />)
+
+    const component = screen.getByRole('textbox')
+
+    expect(component).toHaveClass('amsterdam-text-area extra')
+  })
+
```
alimpens, line 32: Can you add tests for the classes that are added by the `resize` and `cols` props?

packages/css/src/components/text-area/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+# TextArea
+
+A form field that allows the user to input text over multiple lines.
+
+## Guidelines
+
+- Use a TextArea when users need to enter more than 1 sentence of text, such as a comment or description.
+- The height of the TextArea should be appropriate for the information to be entered.
+  - Use the experimental grow attribute to allow the TextArea to grow in height as the user types.
```
alimpens, line 11: This is removed now, no?

storybook/src/components/TextArea/TextArea.stories.tsx
dlnr: Feels wrong to change a component to make options in storybook work but here you go.

alimpens: I agree, you kind of solved it the wrong way around 😅 You can give an option the value of undefined, which doesn't set the prop. 

[Similar to what we do here](https://github.com/Amsterdam/design-system/blob/develop/storybook/src/components/IconButton/IconButton.stories.tsx#L22-26)

packages/react/src/TextArea/TextArea.test.tsx
alimpens: One of the classes added by `resize` is now tested, could you add the rest?

storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+const paragraph = exampleParagraph()
+
+type StoryProps = TextAreaProps & { invalid?: boolean }
+
+const meta = {
+  title: 'Components/Forms/Text Area',
+  component: TextArea,
+  args: {
+    defaultValue: paragraph,
```
alimpens, line 19: Using `defaultValue` makes the interaction a bit weird, using `value` is probably a better idea

storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+    disabled: false,
+    name: 'text-area',
```
alimpens, line 21: What is this for?

dlnr: Otherwise it generates an error in the browser console.

alimpens: You mean an issue? A `name` is not strictly necessary, and it's a bit confusing here imo.

dlnr: image

Yes an issue, I've removed it.

storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+  args: {
+    resize: 'horizontal',
+  },
+}
+
+export const NoResize: Story = {
+  args: {
+    resize: 'none',
+  },
+}
```
alimpens, line 97: Can you order these stories to be the same as they are in the Docs? Otherwise the order in the Docs and in the TOC is different

alimpens (approved):

PR #1096, opened 2024-02-27
chore: Remove code of conduct

alimpens:
The code of conduct is [already added at the Amsterdam GitHub group level](https://github.com/Amsterdam/.github/blob/main/CODE_OF_CONDUCT.md), no need to add it ourselves.


README.md
```diff
@@ -28,7 +28,7 @@ See the [documentation for contributors](./CONTRIBUTING.md).
 ## Code of Conduct
 
 We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.
-Read [our Code of Conduct](CODE_OF_CONDUCT.md) if you haven’t already.
+Read [our Code of Conduct](./?tab=coc-ov-file) if you haven’t already.
```
VincentSmedinga, line 31: This link doesn’t work for https://github.com/Amsterdam?tab=coc-ov-file

VincentSmedinga:
Will it still appear at its current position on https://github.com/Amsterdam/design-system if we remove it here?

README.md
alimpens: It should link to <https://github.com/Amsterdam/design-system/?tab=coc-ov-file>, which should select the correct tab. I'm not 100% sure though.

We could also link to <https://github.com/Amsterdam/.github/blob/main/CODE_OF_CONDUCT.md>, but that takes you out of our repo.

alimpens:
> Will [the tab on our home page](https://github.com/Amsterdam/design-system?tab=coc-ov-file) display the Amsterdam Code of Conduct if we remove our copy?

Yeah, it should. All repos in the Amsterdam group have a Code of Conduct tab now, which uses the generic one.

README.md
VincentSmedinga: Discussed that we need a fully qualified URL anyway in case people use it from their IDE or elsewhere. Let’s change that, merge and see what happens in the GitHub UI.

VincentSmedinga (approved):

PR #1105, opened 2024-03-04
chore: Improve name and signature of focus hook

VincentSmedinga:
This renames the hook to ‘use focus with arrows’ to ‘use keyboard focus’ because we support more keys than only the arrows to manage focus.

It changes the export type from default to named.

And I’ve moved the `ref` parameter out of the options object. That object was added because of a request of mine in a pull request, but I had intended only the Boolean configuration options to be in it, not the ref object.

We don’t export this function, so I don’t consider it a breaking change, correct?

### DIFF
--- a/packages/react/src/common/useFocusByKeyboard.tsx
+++ b/packages/react/src/common/useFocusByKeyboard.tsx
@@ -31,7 +31,7 @@ type FocusByKeyboardOptions = {
   horizontally?: boolean
 }
 
-const useFocusByKeyboard = ({
+export const useFocusByKeyboard = ({
   ref,
   rotating = false,
   directChildrenOnly = false,
@@ -110,5 +110,3 @@ const useFocusByKeyboard = ({
 
   return { keyDown }
 }
-
-export default useFocusByKeyboard
--- a/packages/react/src/common/useFocusByKeyboard.tsx
+++ b/packages/react/src/common/useFocusByKeyboard.tsx
@@ -25,20 +25,17 @@ const FOCUSABLE_ELEMENTS = [
 ]
 
 type FocusByKeyboardOptions = {
-  ref: RefObject<HTMLDivElement>
   rotating?: boolean
   directChildrenOnly?: boolean
   horizontally?: boolean
 }
 
-export const useFocusByKeyboard = ({
-  ref,
-  rotating = false,
-  directChildrenOnly = false,
-  horizontally = false,
-}: FocusByKeyboardOptions) => {
+export const useFocusByKeyboard = (ref: RefObject<HTMLDivElement>, options: FocusByKeyboardOptions) => {
+  const { rotating = false, directChildrenOnly = false, horizontally = false } = options
+
   const next = horizontally ? KeyboardKeys.ArrowRight : KeyboardKeys.ArrowDown
   const previous = horizontally ? KeyboardKeys.ArrowLeft : KeyboardKeys.ArrowUp
+
   const keyDown = (e: KeyboardEvent) => {
     if (ref.current) {
       const element = ref.current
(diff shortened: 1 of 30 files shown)


packages/react/src/common/useFocusByKeyboard.tsx
```diff
@@ -24,21 +24,18 @@ const FOCUSABLE_ELEMENTS = [
...
   horizontally?: boolean
 }
 
-const useFocusWithArrows = ({
-  ref,
-  rotating = false,
-  directChildrenOnly = false,
-  horizontally = false,
-}: FocusWithArrowsOptions) => {
+export const useFocusByKeyboard = (ref: RefObject<HTMLDivElement>, options: FocusByKeyboardOptions) => {
```
VincentSmedinga, line 33: Wait… is it correct that we use `HTMLDivElement` here? Why div?

alimpens, line 33: I like `useKeyboardFocus`!

alimpens: No, I don't think so, should be `HTMLElement` I think

packages/react/src/common/useFocusByKeyboard.tsx
```diff
@@ -24,21 +24,18 @@ const FOCUSABLE_ELEMENTS = [
   'select:not([disabled])',
 ]
 
-type FocusWithArrowsOptions = {
-  ref: RefObject<HTMLDivElement>
```
alimpens, line 28: Why move ref from the params object? Adding it to the object documents it nicely, imo

packages/react/src/common/useFocusByKeyboard.tsx
```diff
@@ -24,21 +24,18 @@ const FOCUSABLE_ELEMENTS = [
   'select:not([disabled])',
```
alimpens, line 24: Maybe not for now, but: do we want to document this hook? We expose it to our users already

packages/react/src/common/useFocusByKeyboard.tsx
alimpens: Hm, does changing it to `HTMLElement` cause TypeScript errors? I think maybe the 'correct' way to do this is to [use a generic and pass the type to the function when calling it](https://stackoverflow.com/questions/57201223/typescript-how-to-declare-a-type-that-includes-all-types-extending-a-common-typ#answer-57201395), but that might be a bit much for now.

packages/react/src/common/useFocusByKeyboard.tsx
alimpens: Ah, sorry, I was wrong. It is a function in `index.esm.js` and `index.cjs.js`, but it's not exported, so you probably can't reach it?

VincentSmedinga: Yes, nothing seems to indicate it’s public.

packages/react/src/common/useFocusByKeyboard.tsx
VincentSmedinga: I think I’ve applied that correctly; can you verify?

alimpens: Hm, I don't this this is any different from what we had? `ref: RefObject<HTMLElement>` seems to work for all types of html elements, and `<T extends HTMLElement>(ref: RefObject<T>)` seems to do the same thing.

The first option is good enough imo, and a bit simpler than the second one.

VincentSmedinga: I did get the TS error I quoted, but not anymore now. Maybe some other change. Reverted the generic.

alimpens (approved):

PR #1106, opened 2024-03-04
chore: Clean up stories

alimpens:
This PR removes unused keys, uses `args.children` instead of `render` when possible, removes unnecessary `React.Fragment`s, etc.

### DIFF
--- a/storybook/src/components/PageMenu/PageMenu.stories.tsx
+++ b/storybook/src/components/PageMenu/PageMenu.stories.tsx
@@ -17,54 +17,49 @@ export default meta
 type Story = StoryObj<typeof meta>
 
 export const Default: Story = {
-  render: () => (
-    <PageMenu>
-      <PageMenu.Link href="#">English</PageMenu.Link>
-      <PageMenu.Link href="#" icon={LoginIcon}>
+  args: {
+    children: [
+      <PageMenu.Link href="#" key={1}>
+        English
+      </PageMenu.Link>,
+      <PageMenu.Link href="#" icon={LoginIcon} key={2}>
         Inloggen Mijn Amsterdam
-      </PageMenu.Link>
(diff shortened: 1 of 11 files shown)


storybook/src/components/PageMenu/PageMenu.stories.tsx
```diff
@@ -17,54 +17,49 @@ export default meta
...
         Onderzoeken
-      </PageMenu.Link>
-      <PageMenu.Link href="#" key="bezoeken">
+      </PageMenu.Link>,
+      <PageMenu.Link href="#" key={2}>
         Bezoeken
-      </PageMenu.Link>
-      <PageMenu.Link href="#" key="archiveren">
+      </PageMenu.Link>,
+      <PageMenu.Link href="#" key={3}>
```
VincentSmedinga, line 42: I understand how numbers seem simple and nice, but keys prevent having to change other values when inserting items into a list.

VincentSmedinga (commented):
Thanks! Please add a little update to `documentation/storybook.md` with the guidelines you followed here.

storybook/src/components/PageMenu/PageMenu.stories.tsx
alimpens: I'm fine either way, as long as we do it consistently. Sometimes using a string for a key and mostly using numbers implies that the string has meaning imo, which it hasn't. Always using a string would fix that, but it'd be a lot more code.

I think changing everything to strings is more work then we'd save when we have to insert an item into a list once or twice, so I'm leaning towards numbers.

VincentSmedinga: Fair enough, I can go along with numbers.

VincentSmedinga (approved):

PR #1109, opened 2024-03-06
chore: Add example of wrapping button label

VincentSmedinga:
I added this story a couple of days ago while working on padding. It shows the effect of vertical padding and line height in action simultaneously (it seems like line height could be a bit smaller here), and it prepares a case against the potential claim that “button labels don’t wrap.”

### DIFF
--- a/storybook/src/components/Button/Button.docs.mdx
+++ b/storybook/src/components/Button/Button.docs.mdx
@@ -41,9 +41,9 @@ Use tertiary buttons for unimportant calls to action – as many as necessary.
 
 ### Text wrapping
 
-The label can occasionally wrap over multiple lines, for various reasons.
-A shorter text might be unavailable.
-Two buttons could sit next to each other on a narrow screen.
-Or the user has configured a large font on their device.
+Keep in mind that the label may occasionally wrap over multiple lines:
+a shorter text might be unavailable,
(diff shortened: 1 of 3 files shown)


storybook/src/components/Button/Button.docs.mdx
```diff
@@ -37,4 +37,13 @@ Use tertiary buttons for unimportant calls to action – as many as necessary.
...
 
-<Canvas of={ButtonStories.ButtonWithAnIcon} />
+<Canvas of={ButtonStories.WithIcon} />
+
+### Text wrapping
+
+The label can occasionally wrap over multiple lines, for various reasons.
+A shorter text might be unavailable.
+Two buttons could sit next to each other on a narrow screen.
+Or the user has configured a large font on their device.
```
alimpens, line 47: Do we want to add that this isn't recommended?

VincentSmedinga: I don’t think we can – there’s no way around a wrapping button label if it’s on a phone with a text size zoomed to 300%.

We can update to ‘Keep in mind that the label can occasionally wrap over multiple lines’. We don’t suggest using short labels yet; I can add that to the top of the docs in support?

alimpens (approved):

PR #1113, opened 2024-03-07
feat: Introduce tokens for stack spacing and Row component

VincentSmedinga:
This introduces a `Row` component, for now exactly the same as `Column`, but a separate component as we foresee different alignment properties between the two.

Setting this up made me lean towards the idea that vertical space in `Grid` should be responsive, but in `Column` it should be fixed-width just like in `Row`. If so, we can keep the name ‘stack space’ for the tokens and set them as values for the column gap tokens as well.


packages/css/src/components/row/row.scss
```diff
@@ -0,0 +1,17 @@
...
+
+@import "../../common/size";
+
+.amsterdam-row {
+  display: flex;
+  flex-direction: row;
+}
+
+@each $size, $label in $amsterdam-sizes {
+  .amsterdam-row--#{$label} {
```
VincentSmedinga, line 14: Should we include ‘gap’ in this class name: `.amsterdam-row--gap-#{$label}`? If yes, same for `Column`.

storybook/src/components/Row/Row.docs.mdx
```diff
@@ -0,0 +1,30 @@
...
+import * as RowStories from "./Row.stories.tsx";
+import README from "../../../../packages/css/src/components/row/README.md?raw";
+
+<Meta of={RowStories} />
+
+<Markdown>{README}</Markdown>
+
+## Design
+
+The five [spacing](/docs/docs-design-guidelines-space--docs) sizes are available for the size of the gap.
```
VincentSmedinga, line 11:
```suggestion
The five [spacing](/docs/docs-design-guidelines-space--docs) sizes are available for the width of the gap.
```

packages/css/src/components/row/row.scss
alimpens: We're going to offer these classes as utility classes as well, right? If so, shorter is better. OTOH, if we're going to add other props to these components, we have to make a distinction between the rules we're setting. 

Maybe keep it as short as possible for now, and change it when we need to?

storybook/src/components/Row/Row.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  args: {
+    children: ThreeButtons,
+  },
+  argTypes: {
+    as: {
+      control: { type: 'radio' },
+      options: ['article', 'div', 'section'],
+    },
+    children: {
+      table: { disable: true },
```
alimpens, line 23: This is default, no?

PR #1114, opened 2024-03-07
feat: Add Form Field Character Counter

alimpens:
In the designs the color changes to red on exceeding the max count. Do we want to implement that? That would mean we need a `count` and `max` prop. -> This is now implemented

If we do, maybe we want to generate the shown string with these props as well? With string replacement or something? We have to keep in mind that the string should be able to be translated. -> The string is static for now, we'll look at translations later


dlnr:
Yes I would have expected a function to count the number of characters in a assigned form field, and a max prop.

Also, I expected the text color to match the design? Grey like the FormFieldDescription.

proprietary/tokens/src/components/ams/form-field-character-counter.tokens.json
```diff
@@ -0,0 +1,11 @@
+{
+  "ams": {
+    "form-field-character-counter": {
+      "color": { "value": "{ams.color.neutral-grey3}" },
```
VincentSmedinga, line 4:
```suggestion
      "color": { "value": "{ams.color.primary-black}" },
```

packages/react/src/FormFieldCharacterCounter/FormFieldCharacterCounter.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes } from 'react'
+
+export type FormFieldCharacterCounterProps = HTMLAttributes<HTMLDivElement> & {
+  label: string
```
VincentSmedinga, line 11: Please add `length` and `maxLength` props.

The component can render `${length} van ${maxLength} tekens` – we’ll worry about i18n later.

dlnr (approved):

VincentSmedinga (approved):

PR #1116, opened 2024-03-08
feat: Remove border from accordion section and move chevron to the left

dlnr:
Moving the border to panel has some consequences, the focus state thickens the border but only the button in the top is clickable. I've suggested another solution in Figma so I'll await the response before marking this as ready.

Update @VincentSmedinga: the border has been removed in a subsequent design iteration, and the chevron moved to the start of the line.


packages/css/src/components/accordion/accordion.scss
```diff
@@ -34,20 +40,21 @@
...
-    transition: transform 0.3s ease;
+.ams-accordion__icon svg {
+  transform: rotate(0deg);
+  transition: transform 0.3s ease;
 
-    @media (prefers-reduced-motion) {
-      transition: none;
-    }
+  [aria-expanded="true"] & {
+    transform: rotate(180deg);
```
VincentSmedinga, line 52: We can start using the `rotate` property ([94.54% usage in NL](https://caniuse.com/?search=rotate)) rather than the transform function.

VincentSmedinga, line 52: Now that the chevron is at the start of the line, rotating clockwise feels more natural.

packages/css/src/components/accordion/accordion.scss
```diff
@@ -34,20 +40,21 @@
...
   }
 
-  @include text-rendering;
-  @include reset-button;
+  @media (prefers-reduced-motion) {
+    transition: none;
+  }
```
VincentSmedinga, line 57: A simpler pattern could be to set a transition only once – if people do not prefer reduced motion.

```suggestion
  @media not (prefers-reduced-motion) {
    transition: transform 0.3s ease;
  }
```

packages/css/src/components/accordion/accordion.scss
```diff
@@ -5,25 +5,31 @@
...
-.ams-accordion__button {
   color: var(--ams-accordion-button-color);
   cursor: pointer;
   display: flex;
   font-family: var(--ams-accordion-button-font-family);
   font-size: var(--ams-accordion-button-font-size);
   font-weight: var(--ams-accordion-button-font-weight);
-  justify-content: space-between;
+  gap: var(--ams-accordion-button-gap);
+  justify-content: flex-start;
```
VincentSmedinga, line 32: We can remove this: the initial value is `normal` and “[a]s `stretch` behaves as `start` in flex containers, `normal` also behaves as `start`” ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#normal)).

```suggestion
```

packages/css/src/components/accordion/accordion.scss
```diff
@@ -34,20 +39,21 @@
...
-    @media (prefers-reduced-motion) {
-      transition: none;
-    }
+  [aria-expanded="true"] & {
+    transform: rotate(180deg);
   }
 
-  @include text-rendering;
-  @include reset-button;
+  @media not (prefers-reduced-motion) {
```
dlnr, line 54: @VincentSmedinga I do "not" agree on using `not` here.

packages/css/src/components/accordion/accordion.scss
dlnr: Good point

VincentSmedinga (approved):

PR #1123, opened 2024-03-08
fix: Use semantic Heading for the title of a Dialog

dlnr:
Styling is supposed to be level 4 but what about the level? H2?


alimpens:
> Styling is supposed to be level 4 but what about the level? H2?

This is actually a fairly complex topic 😅 [Here's an informative thread in the Material UI repo on it](https://github.com/mui/material-ui/issues/34250).

An argument can be made that a `dialog` is a document that lives outside the main document, and as such needs an `h1`. I think that makes sense, but there are different views on this subject.

dlnr:
Changed the heading level to 1, I think it makes sense.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -44,6 +44,7 @@
   align-items: flex-start;
   display: flex;
   gap: var(--ams-dialog-header-gap);
+  justify-content: space-between;
 }
 
 @mixin reset {
```
alimpens, line 50: Something went wrong with the merge, `ams-dialog__title` and its tokens should be deleted

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -17,7 +18,9 @@ export const Dialog = forwardRef(
     <dialog {...restProps} ref={ref} className={clsx('ams-dialog', className)}>
       <form method="dialog" className="ams-dialog__form">
         <header className="ams-dialog__header">
-          <span className="ams-dialog__title">{title}</span>
+          <Heading level={1} size="level-4">
+            {title}
+          </Heading>
           <IconButton label="Sluiten" formNoValidate />
```
alimpens, line 24: If the heading is now level 4, shouldn't the close button be as well?

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -17,7 +18,9 @@ export const Dialog = forwardRef(
     <dialog {...restProps} ref={ref} className={clsx('ams-dialog', className)}>
       <form method="dialog" className="ams-dialog__form">
         <header className="ams-dialog__header">
-          <span className="ams-dialog__title">{title}</span>
+          <Heading level={1} size="level-4">
```
alimpens, line 21: `level={1}` is the default

packages/react/src/Dialog/Dialog.tsx
VincentSmedinga: Yes, adjusted.

VincentSmedinga (approved):

PR #1125, opened 2024-03-10
feat!: Shorten namespaces in theme and tokens

VincentSmedinga:
This shortens various namespaces in our theme, classes, and tokens:

1. `amsterdam` to `ams`
2. `typography` to `text`

Token directories have been renamed in accordance.

Additionally, we could simplify `text.text-level` to `text.level`.

It decreases our bundle sizes by around 10% (CSS) and 20% (tokens), although gzip will override most of that effect.


proprietary/tokens/src/components/ams/icon.tokens.json
```diff
@@ -0,0 +1,24 @@
...
+        },
+        "5": {
+          "font-size": { "value": "{ams.text.level.5.font-size}" },
+          "line-height": { "value": "{ams.text.level.5.line-height}" }
+        },
+        "6": {
+          "font-size": { "value": "{ams.text.level.6.font-size}" },
+          "line-height": { "value": "{ams.text.level.6.line-height}" }
+        }
+      }
```
alimpens, line 21: I'm not sure this is an improvement. The icon variants are called `size-3`, `size-4`, etc [1]. Maybe you wanted to align it to the Heading props, but those variants are called `1`, `2`, etc. 

[1] The Icon `size` prop actually has `level-3`, `level-4`, etc. Maybe we should align those 2 as well? Maybe not in this PR though

VincentSmedinga: Whoah, did you actually read everything? 😮 

I think we can still split these names, but I agree that a separate task is better for it – even more so if we consider changing icon props anyway. Reverted.

alimpens: No, just checked the separate commits 😁 Sounds like a plan to me!

alimpens (approved):

PR #1127, opened 2024-03-11
chore: Fix Radio Group story code view

alimpens:


storybook/src/components/Radio/Radio.stories.tsx
```diff
@@ -71,4 +71,9 @@ export const RadioGroup: Story = {
       </Radio>
     </Fieldset>
   ),
+  parameters: {
+    docs: {
+      source: { type: 'dynamic' },
```
VincentSmedinga, line 76: Renders the story source with dynamically updated arg values – https://storybook.js.org/docs/api/doc-block-source#type

VincentSmedinga (approved):

PR #1131, opened 2024-03-14
chore: Storybook 8 upgrade

dlnr:
Also tried to resolve the vite cjs warning but ran into jest issues.


storybook/config/main.ts
```diff
@@ -1,5 +1,4 @@
 import type { StorybookConfig } from '@storybook/react-vite'
-import remarkGfm from 'remark-gfm'
```
VincentSmedinga, line 2: I now understand why we need this – it allows us to use GitHub Flavoured Markdown in Storybook.

For example, our tables no longer work without: https://amsterdam.github.io/design-system/demo-DES-618-storybook-8/?path=/docs/docs-design-guidelines-grid--docs#spacious-2

> This is a plugin to enable the extensions to markdown that GitHub adds with GFM: autolink literals, footnotes, strikethrough, tables, and tasklists. These extensions by GitHub to CommonMark are called [GFM](https://github.github.com/gfm/) (GitHub Flavored Markdown).

> Users frequently believe that some of these extensions, specifically autolink literals and tables, are part of normal markdown, so using remark-gfm will help match your implementation to their understanding of markdown.

– [source](https://github.com/remarkjs/remark-gfm)

dlnr: Thanks for the headsup, it works with the latest version.

VincentSmedinga (approved):

storybook/config/manager-head.html
```diff
@@ -1,20 +1,9 @@
 <script id="storybook-mod-expand-all-folders" type="text/javascript">
-  // Somewhat hacky way to expand all folders. Storybook does not support that natively.
```
alimpens, line 2: This is still a hacky way to do this, right? I'd keep the comment

storybook/package.json
```diff
@@ -26,34 +26,32 @@
...
     "react-dom": "18.2.0",
     "require-from-string": "2.0.2",
-    "storybook": "7.6.17",
+    "storybook": "8.0.0",
     "vite": "5.1.4"
   },
   "dependencies": {
-    "remark-gfm": "3.0.1"
+    "@storybook/test": "8.0.0",
+    "remark-gfm": "4.0.0"
```
alimpens, line 55: These should be devDependencies as well, I think?

storybook/package.json
```diff
@@ -26,34 +26,32 @@
...
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "require-from-string": "2.0.2",
-    "storybook": "7.6.17",
+    "storybook": "8.0.0",
     "vite": "5.1.4"
   },
   "dependencies": {
-    "remark-gfm": "3.0.1"
+    "@storybook/test": "8.0.0",
```
alimpens, line 54: We don't seem to be using this (yet)

storybook/package.json
dlnr: It was in the deps previously, and is used in the build. However it's probably baked in during the static build so I'll move it.

storybook/package.json
dlnr: Sure, I've added to replace @storybook/testing-library, which was also unused I guess? After removing my interaction tests for Dialog?

alimpens: Yeah, I think so

alimpens (approved):

PR #1134, opened 2024-03-14
feat: Add Avatar component

RubenSibon:
Please pay extra attention to:

- Avatar's README
- Colours have been copied from the Badge component; the colour red has been added
- With an empty `label` property the rendered label will be a user/account icon
- Labels longer than two characters will be cut off
- Labels are always capitalized.

Considerations:

- Should the Avatar be an anchor or a button (and thus have interaction states)? Currently, there is no interaction at all => **The avatar will be static; an anchor can be wrapped around it; interaction states should be added at a later time.**
- Should an empty `label` give back "n.b." or something else? **Empty label will give back the user icon.**

### DIFF
--- a/packages/css/src/components/avatar/avatar.scss
+++ b/packages/css/src/components/avatar/avatar.scss
@@ -24,7 +24,7 @@
   overflow: hidden;
   padding-block: 0;
   padding-inline: 0;
-  vertical-align: middle;
+  vertical-align: middle; /* Remove ‘phantom’ white space when image doesn’t load */
   width: calc(var(--ams-avatar-line-height) * var(--ams-avatar-font-size) + 2 * var(--ams-avatar-padding-inline));
 }
 


proprietary/tokens/src/components/ams/avatar.tokens.json
```diff
@@ -0,0 +1,48 @@
+{
+  "ams": {
+    "avatar": {
+      "border-width": { "value": "{ams.border-width.md}" },
+      "border-style": { "value": "solid" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-size": { "value": "{ams.text.level.6.font-size}" },
```
dlnr, line 7: Since the line height determines the size of the Avatar, does this level align with menu items in the Header? Maybe we need an example story?

proprietary/tokens/src/components/ams/avatar.tokens.json
```diff
@@ -0,0 +1,48 @@
+{
+  "ams": {
+    "avatar": {
+      "border-width": { "value": "{ams.border-width.md}" },
```
dlnr, line 4: Do we want to use a border? If you make this work with an image you'll lose the with and height.

packages/css/src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,70 @@
...
+
+.ams-avatar {
+  border-radius: 50%;
+  border-style: var(--ams-avatar-border-style);
+  border-width: var(--ams-avatar-border-width);
+  display: inline-block;
+  font-family: var(--ams-avatar-font-family);
+  font-size: var(--ams-avatar-font-size);
+  line-height: var(--ams-avatar-line-height);
+  min-width: 1lh;
```
dlnr, line 14:
```suggestion
  min-width: var(--ams-avatar-line-height);
```

packages/css/src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,70 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-avatar {
+  border-radius: 50%;
```
dlnr, line 7:
```suggestion
  aspect-ratio: 1 / 1;
  border-radius: 50%;
```

proprietary/tokens/src/components/ams/avatar.tokens.json
RubenSibon: I'll add an example story in the next commit.

proprietary/tokens/src/components/ams/avatar.tokens.json
RubenSibon: Borders have been removed in latest commit due to the addition of an optional background image.

packages/css/src/components/avatar/avatar.scss
RubenSibon: Went for:

```
width: calc(var(--ams-avatar-line-height) * var(--ams-avatar-font-size));
```

packages/css/src/components/avatar/avatar.scss
RubenSibon: Implemented.

RubenSibon:
- [ ] Add example story with Avatar in a Header.

packages/css/src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,74 @@
...
+  justify-content: center;
+  line-height: var(--ams-avatar-line-height);
+  padding-block: var(--ams-avatar-padding);
+  padding-inline: var(--ams-avatar-padding);
+  width: calc(var(--ams-avatar-line-height) * var(--ams-avatar-font-size));
+}
+
+.ams-avatar--blue {
+  background-color: var(--ams-avatar-blue-background-color);
+  background-image: var(--ams-avatar-blue-background-image);
```
alimpens, line 24: If we render the default Icon as an SVG in the React component, we can set the color with CSS I think. This would get rid of the need for the encoded SVGs in the tokens.

packages/css/src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,74 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-avatar {
+  aspect-ratio: 1/1;
+  background-position: center;
```
alimpens, line 8: Can we separate the rules you need for an Avatar with an image from the rules for an Avatar without?

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export type AvatarProps = {
+  color?: AvatarColor
+  imageUrl?: string
+  label: string
+} & HTMLAttributes<HTMLElement>
+
+export const Avatar = forwardRef(
+  ({ label, imageUrl, className, color = 'dark-blue', ...restProps }: AvatarProps, ref: ForwardedRef<HTMLElement>) => {
+    const initials: string | ReactElement = useMemo(() => {
```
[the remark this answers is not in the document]
RubenSibon: Correct, should've been removed.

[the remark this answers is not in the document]
RubenSibon: I'll remove it

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }, [imageUrl, label])
+
+    return (
+      <span
+        {...restProps}
+        ref={ref}
+        className={clsx('ams-avatar', `ams-avatar--${color}`, className)}
+        style={{ backgroundImage: backgroundImageValue }}
+        title={title}
```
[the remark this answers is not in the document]
RubenSibon: I used `title` to show a tooltip to users as well.

Previously we decided to use `aria-label`, but that is also recommended against I see.

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    const backgroundImageValue: string | undefined = useMemo(() => {
+      if (imageUrl) {
+        return `url(${imageUrl})`
+      } else if (label.length) {
+        return 'none'
+      } else {
+        return undefined
+      }
+    }, [imageUrl, label])
```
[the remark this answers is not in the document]
RubenSibon: Let's discuss

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+} & HTMLAttributes<HTMLElement>
+
+export const Avatar = forwardRef(
+  ({ label, imageUrl, className, color = 'dark-blue', ...restProps }: AvatarProps, ref: ForwardedRef<HTMLElement>) => {
+    const initials: string | ReactElement = useMemo(() => {
+      return (label.length > 2 ? label.slice(0, 2) : label).toUpperCase()
+    }, [label])
+
+    const title = useMemo(() => {
+      return !initials.length ? 'Niet-ingelogde gebruiker' : `Initialen gebruiker: ${initials.split('').join('.')}.`
```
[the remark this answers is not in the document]
RubenSibon: What would be a better way to insert a period?

Something like this?

```
`${initials[0]}.${initials[1]}.`
```

packages/react/src/Avatar/Avatar.tsx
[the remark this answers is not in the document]
RubenSibon: Addressed

packages/react/src/Avatar/Avatar.tsx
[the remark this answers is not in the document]
RubenSibon: Removed

packages/react/src/Avatar/Avatar.tsx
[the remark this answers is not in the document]
RubenSibon: Adjusted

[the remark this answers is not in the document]
RubenSibon: Adjusted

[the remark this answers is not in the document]
RubenSibon: Using the initials without any periods.

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+  'purple',
+  'red',
+  'yellow',
+] as const
+
+type AvatarColor = (typeof avatarColors)[number]
+
+export type AvatarProps = {
+  color?: AvatarColor
+  imageUrl?: string
```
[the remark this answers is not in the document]
RubenSibon: Adjusted

packages/css/src/components/avatar/avatar.scss
RubenSibon: Background Images have been removed for Icon and Image components.

packages/css/src/components/avatar/avatar.scss
RubenSibon: I think this is no longer relevant.

proprietary/tokens/src/components/ams/avatar.tokens.json
RubenSibon: Example story added.

packages/css/src/components/avatar/avatar.scss
```diff
@@ -0,0 +1,74 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-avatar {
+  aspect-ratio: var(--ams-aspect-ratio-square);
+  border-radius: 50%;
+  box-sizing: content-box;
```
alimpens, line 9: Do we need this?

VincentSmedinga: No, thanks, it’s a leftover. Removed.

VincentSmedinga (dismissed):

VincentSmedinga (approved):

PR #1136, opened 2024-03-20
feat: Add document-level classes and guidelines

VincentSmedinga:
Our prototyping environment taught us that we need some base styling at the document level to ensure a correct context for our CSS and layout.

This adds a class to remove the default browser margin from the `body` element ~~and sets 100% height and a white background colour (which can later depend on light and dark mode)~~.

~~It also uses our extra bold font weight for strongly emphasised text, scoped to our theme class.~~

Our typography system expects `1rem` to be `16px`; we explain that the root font size must not be overridden.


packages/css/src/components/document/document.scss
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
```
VincentSmedinga, line 5: I found this in the create-vite-app scaffold, should we add it?

```css
:root {
  font-synthesis: none;
  text-rendering: optimizeLegibility;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}
```

packages/css/README.md
```diff
@@ -49,3 +49,15 @@ To use the compact mode, import the compact css **after** theme css, like so:
...
+
+Some elements of your document are out of reach for the design system components.
+However, their default styles or common boilerplate approaches can influence on the appearance on the CSS components.
+Apply the following to the document elements:
+
+The font-size must not be changed – e.g. through `html { font-size: 62.5% }` – because the typography system expects a rem to be 16 pixels.
+
+The `ams-body` class must be added to the `<body>` element – it prepares the base layout and sets a background colour.
+
+After the `theme-ams` class has been set, e.g. on the `<body>` as well, bold text will use our extra bold font weight.
```
VincentSmedinga, line 63: We don’t instruct to add this theme class anywhere. I couldn’t find it on the NLDS website yet, either.

alimpens, line 63: Can't we merge these two classes?

alimpens: Do you mean the `ams-theme` class rendered by our tokens packages? There are two ways to use our tokens in an app:

1. You can import `index.css`, which sets the tokens on `:root`
2. You can import `index.theme.css`, which contains an `ams-theme` class. You can set that class on the body tag for example. This is useful when you switch themes in your app (spacious and compact, or light and dark in the future for example)

We should definitely write docs about that. I don't quite understand why you reuse that `ams-theme` class here though?

packages/css/README.md
```diff
@@ -49,3 +49,15 @@ To use the compact mode, import the compact css **after** theme css, like so:
...
 import "@amsterdam/design-system-tokens/dist/compact.css";
 ```
+
+## Document styles
+
+Some elements of your document are out of reach for the design system components.
+However, their default styles or common boilerplate approaches can influence on the appearance on the CSS components.
+Apply the following to the document elements:
+
+The font-size must not be changed – e.g. through `html { font-size: 62.5% }` – because the typography system expects a rem to be 16 pixels.
```
alimpens, line 59: Can we make this part of `ams-body`? `html { font-size: 100% !important; }`

packages/css/README.md
```diff
@@ -49,3 +49,15 @@ To use the compact mode, import the compact css **after** theme css, like so:
 import "@amsterdam/design-system-tokens/dist/index.css";
 import "@amsterdam/design-system-tokens/dist/compact.css";
 ```
+
+## Document styles
+
+Some elements of your document are out of reach for the design system components.
+However, their default styles or common boilerplate approaches can influence on the appearance on the CSS components.
```
alimpens, line 56:
```suggestion
Some elements of your document are out of reach for the Design System components.
However, their styles can influence the appearance of the components.
```

packages/css/src/components/document/document.scss
alimpens: - `font-synthesis: none`: Is our font still missing subscript and superscript? And small caps is probably missing as well. I don't really see the need to disable font-synthesis. 
- `text-rendering: optimizeLegibility`: I think Chrome uses this by default, but I think in general it makes sense for us to optimize legibility over speed.
- `-webkit-font-smoothing: antialiased` and `-moz-osx-font-smoothing: grayscale`: This is somewhat of a contested topic I think. I'd let the browser decide for now, and add it if that really causes issues. 
- `-webkit-text-size-adjust`: We add this on a component basis, but I think it's useful to add this on the document level as well, for text that doesn't use our DS components.

packages/css/src/components/document/document.scss
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Add this class to your HTML `<body>` element.
+ * It overrides margins from the user agent styles,
+ * ans sets a white background and full height.
+ */
+.ams-body {
+  @include reset-body;
+
+  background-color: var(--ams-color-primary-white);
+  min-height: 100vh;
```
alimpens, line 20: Are these necessary? The background color is always white for me, and it isn't an issue if the document is shorter than the screen.

packages/css/src/components/document/document.scss
VincentSmedinga: Alright, so we want `text-rendering: optimizeLegibility` and `-webkit-text-size-adjust: 100%;`. On the latter, [this article](https://kilianvalkhof.com/2022/css-html/your-css-reset-needs-text-size-adjust-probably/) suggests adding two other prefix variations and using a value of `none` instead of `100%`.

We can follow one of three approaches:
1. Add both to `:root` or `.ams-html` or `.ams-body` or `.ams-screen` and use inheritance.
2. Declare for each component separately as we do now.
3. Same, but through a common mixin.

I would go for the latter; it is more in line with our component architecture but reuses the actual declarations as we do for other concepts already. Agree?

alimpens: Yeah, three is definitely a good idea. We could also add it to `:root` or `.ams-html` or `.ams-body` or `.ams-screen`, so components outside of the Design System still benefit from it? But that might be overkill

packages/css/README.md
alimpens: Yeah, applying it to `:root` makes more sense. 

> We could, but isn’t asking people to add a class that overrides an explicit declaration more difficult than asking them to remove that declaration in the first place?

Hm, not sure. Asking them to remove something requires them to read and understand text and figure out if it's relevant for them. Adding an override to CSS you need to import anyway to get it to work doesn't.

packages/css/src/components/document/document.scss
alimpens: I'd only add both if they're necessary, maybe we can make it a prop on Screen?

packages/css/README.md
alimpens: On second thought, maybe it's not such a good idea to enforce `font-size: 100%` with a css rule. If we add it to `:root`, `body` or `html`, projects that currently use `font-size: 62.5%` would have to refactor their entire project before being able to use our packages. 

If we just describe it, users can incrementally adopt our packages, having `font-size: 100%` on the parts that use our components, and `font-size: 62.5%` on the legacy parts. So maybe leave it as is?

packages/css/README.md
alimpens: I agree that it's a good idea to scope the `font-weight` declaration block, but maybe we should use a different class name for it? `ams-theme` is a bit confusing, it's already used for something else.

packages/css/README.md
```diff
@@ -49,3 +49,23 @@ To use the compact mode, import the compact css **after** theme css, like so:
...
+
+After the `theme-ams` class has been set, e.g. on the `<body>` as well, bold text will use our extra bold font weight.
+
+## Common patterns
+
+We use Sass mixins to extract common patterns, especially if they need more than 1 declaration.
+We collect reset styles in mixins as well.
+Both the name of the mixins and their documentation help explain the approach.
+Mixins sit at the end of the declation block to ensure their effect.
+The general ‘reset’ mixin is the very last.
```
alimpens, line 71: Isn't this more of a code convention thing? It's not really relevant for our consumers, right?

VincentSmedinga: Correct, I’ve moved the section.

packages/css/README.md
VincentSmedinga: Very good point. Updated.

packages/css/README.md
VincentSmedinga: I can’t think of anything better now, and it might be a bit out of scope for this PR as well. I’ve removed it.

packages/css/documentation/coding-conventions.md
```diff
@@ -35,3 +35,11 @@ Still, we define its thickness and offset for the initial state.
...
 }
 ```
+
+## Use mixins for patterns and resets
+
+We use Sass mixins to extract common patterns, especially if they need more than 1 declaration.
+We collect reset styles in mixins as well.
+Both the name of the mixins and their documentation help explain the approach.
+Mixins sit at the end of the declaration block to ensure their effect.
+The general ‘reset’ mixin is the very last.
```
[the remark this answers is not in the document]
VincentSmedinga: I’m also used to putting the reset first, but somehow they ended up being last in our code. Made me think about it and concluded that last may be better to prevent styles overriding the reset. The reset should always win.

packages/css/README.md
```diff
@@ -49,3 +49,25 @@ To use the compact mode, import the compact css **after** theme css, like so:
...
+
+## Global styles
+
+Some elements of your document are out of reach for the Design System components.
+However, their styles can influence the appearance of the components.
+
+### Base font size
+
+Do not change the base font-size – e.g. through `html { font-size: 62.5% }`.
+This interferes with users choosing a different default font size in their browser.
```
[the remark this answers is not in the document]
VincentSmedinga: Ah yes, my assumption was incorrect, thanks.

packages/css/documentation/coding-conventions.md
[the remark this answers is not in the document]
alimpens: > The reset should always win.

Should it? It kind of prevents us from making generic resets. Say we make a generic button reset, with `background-color: transparent` to get rid of the grey background color. We would be unable to override that color in a component. 

(Again, not really an issue with our current resets, so we could leave it for now.)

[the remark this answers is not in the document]
VincentSmedinga: I didn’t think hard enough, apparently! Removed these as well. What a crazy PR this has become.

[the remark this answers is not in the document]
alimpens: Touching basic building blocks is always tricky :D I do like the idea of describing how our resets work and where you should define them though, do you want to do that in a separate ticket?

packages/css/README.md
```diff
@@ -49,3 +49,24 @@ To use the compact mode, import the compact css **after** theme css, like so:
...
+### Base font size
+
+Do not change the base font-size – e.g. through `html { font-size: 62.5% }`.
+Out typography system expects `1rem` to be the browser default of 16 pixels.
+
+If your application does use a rule like this and you can’t easily remove it, adopt the Design System components gradually by setting `font-size: 100%` on the parts that use them.
+
+### Remove body margin
+
+If you haven’t removed the margin on the `<body>` element that browsers set by default, you can add our `ams-body` to it to do so.
```
alimpens, line 67:
```suggestion
If you haven’t removed the margin on the `<body>` element that browsers set by default, you can add our `ams-body` class to it to do so.
```

packages/css/documentation/coding-conventions.md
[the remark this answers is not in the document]
VincentSmedinga: Yes! Please! 😝

alimpens (approved):

PR #1139, opened 2024-03-21
Feature/DES-576-description-list

dlnr:


packages/react/src/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+type DescriptionListComponent = {
+  Term: typeof DescriptionListTerm
+  Details: typeof DescriptionListDetails
+  Row: typeof DescriptionListRow
+} & ForwardRefExoticComponent<DescriptionListProps & RefAttributes<HTMLDListElement>>
+
+export const DescriptionList = forwardRef(
+  ({ children, className, inverseColor, ...restProps }: DescriptionListProps, ref: ForwardedRef<HTMLDListElement>) => {
+    const hasDescriptionListRow = Children.toArray(children).some(
```
dlnr, line 25: Tried to avoid using a prop but it seems to not always return true in time for the render. Any react suggestions are welcome.

dlnr: [This](https://react.dev/reference/react/Children#alternatives) suggests using props

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @include reset;
+}
+
+.ams-description-list--inverse-color {
+  color: var(--ams-description-list-inverse-color);
+}
+
+.ams-description-list__row {
```
VincentSmedinga, line 41: Have you tried supporting multiple terms or descriptions without a row? We did need divs while we were floating and clearing things, but here’s an opportunity for named grid areas if I’m not overlooking something.

packages/react/src/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export type DescriptionListProps = PropsWithChildren<HTMLAttributes<HTMLDListElement>> & {
+  inverseColor?: boolean
+}
+
+type DescriptionListComponent = {
+  Term: typeof DescriptionListTerm
+  Details: typeof DescriptionListDetails
+  Row: typeof DescriptionListRow
+} & ForwardRefExoticComponent<DescriptionListProps & RefAttributes<HTMLDListElement>>
```
VincentSmedinga, line 21: This approach got updated in #1130; please apply that here, too.

proprietary/tokens/src/components/ams/description-list.tokens.json
```diff
@@ -0,0 +1,20 @@
...
+    "description-list": {
+      "color": { "value": "{ams.color.primary-black}" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-size": { "value": "{ams.text.level.5.font-size}" },
+      "font-weight": { "value": "{ams.text.font-weight.normal}" },
+      "gap": { "value": "{ams.space.stack.md}" },
+      "inverse-color": { "value": "{ams.color.primary-white}" },
+      "line-height": { "value": "{ams.text.level.5.line-height}" },
+      "row": {
+        "gap": { "value": "{ams.space.inside.md}" }
```
VincentSmedinga, line 12: If we keep the row, this should use the Stack scale as well.
```suggestion
        "gap": { "value": "{ams.space.stack.md}" }
```

packages/css/src/components/description-list/description-list.scss
alimpens: Or just `grid-column-start: 2` on the details?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+  decorators: [
+    (Story, context) => (
+      <div className={context.args.inverseColor ? 'ams-docs-dark-background' : undefined}>
+        <Story />
+      </div>
+    ),
+  ],
+  args: {
+    children: [
+      <DescriptionList.Term key={3}>Gebied</DescriptionList.Term>,
```
alimpens, line 34: Why are the keys out of order?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <DescriptionList.Term key={5}>Stadsdeel</DescriptionList.Term>,
+      <DescriptionList.Details key={6}>West</DescriptionList.Details>,
+      <DescriptionList.Term key={1}>Opmerkingen</DescriptionList.Term>,
+      <DescriptionList.Details key={2}>{paragraph}</DescriptionList.Details>,
+    ],
+  },
+}
+
+export const WrapRows: Story = {
+  render: () => (
```
alimpens, line 45: Please use `children` here as well, `render` creates a messy code view.

packages/css/src/components/description-list/description-list.scss
dlnr: Sure I've also come across documentation where it is allowed to have multiple terms and details per row. 

@alimpens  This is de case? It just happens to conflict if you also have multiple terms.

alimpens: I'm not sure I understand your question.

If we always set `grid-column-start: 1` on the terms and `grid-column-start: 2` on the details, we wouldn't need DescriptionListRow, right?

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,72 @@
...
+    padding-inline-start: 0;
+  }
+
+  .ams-description-list--rows & {
+    @media screen and (min-width: $ams-breakpoint-medium) {
+      grid-column-start: 2;
+    }
+  }
+
+  @include hyphenation;
```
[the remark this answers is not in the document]
dlnr: I was under the impression that we use this wherever we have a title like field.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,72 @@
...
+
+@import "../../common/breakpoint";
+@import "../../common/hyphenation";
+
+@mixin reset {
+  box-sizing: border-box;
+  list-style: none;
+  margin-block: 0;
+  margin-inline: 0;
+  -webkit-text-size-adjust: 100%;
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, that’s merged.

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: That’s also what I was hoping we can do.

dlnr: Sure, I'm trying to come up with a solution for multiple terms though. I've [suggested](https://github.com/Amsterdam/design-system/pull/1139#discussion_r1535390371) to use an object instead of children. What do you guys think?

dlnr: Also see [this link](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element:~:text=The%20following%20example%20illustrates%20the%20use%20of%20the%20dl%20element%20to%20mark%20up%20metadata%20of%20sorts)

VincentSmedinga: Ah, I guess I didn’t quite understand what you meant there.

I’m not a fan of object props; they’re difficult to use and compose. Let’s keep using child components.

Having multiple terms is an edge case. We can happily put that out of scope.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  @include reset;
+}
+
+.ams-description-list--inverse-color {
+  color: var(--ams-description-list-inverse-color);
+}
+
+.ams-description-list__term {
+  @include reset;
```
alimpens, line 37: Does `term` need this reset? The css rules don't seem to do anything.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    padding-inline-start: 0;
+  }
+
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    grid-column-start: 2;
+  }
+
+  @include reset;
```
alimpens, line 53: `details` already gets `box-sizing: border-box` from the parent I think, it only needs `margin-inline: 0;`

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.ams-description-list--inverse-color {
+  color: var(--ams-description-list-inverse-color);
+}
+
+.ams-description-list__term {
+  @include reset;
+  @include text-rendering;
```
alimpens, line 38: It's a little weird imo that the `font-family` etc. is defined on the `list`, and the font rendering is defined on `term` and `details`. Why not define everything on `list`?

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+}
```
alimpens, line 39: Wouldn't adding the following rules here solve the whole multiple term, multiple details thing?

```suggestion
  @media screen and (min-width: $ams-breakpoint-medium) {
    grid-column-start: 1;
  }
}
```

packages/react/src/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <dl
+        {...restProps}
+        ref={ref}
+        className={clsx('ams-description-list', inverseColor && 'ams-description-list--inverse-color', className)}
+      >
+        {children}
+      </dl>
+    )
+  },
```
alimpens, line 27:
```suggestion
  ({ children, className, inverseColor, ...restProps }: DescriptionListProps, ref: ForwardedRef<HTMLDListElement>) => (
    <dl
      {...restProps}
      ref={ref}
      className={clsx('ams-description-list', inverseColor && 'ams-description-list--inverse-color', className)}
    >
      {children}
    </dl>
  ),
```

packages/react/src/DescriptionList/DescriptionListDetails.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render } from '@testing-library/react'
+import { createRef } from 'react'
+import { DescriptionList } from './DescriptionList'
+import '@testing-library/jest-dom'
+
+describe('Description list details', () => {
+  it('renders', () => {
+    const { container } = render(<DescriptionList.Details>Test</DescriptionList.Details>)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 10: Can you select the component by role here? `definition` in this case

packages/react/src/DescriptionList/DescriptionListTerm.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render } from '@testing-library/react'
+import { createRef } from 'react'
+import { DescriptionList } from './DescriptionList'
+import '@testing-library/jest-dom'
+
+describe('Description list term', () => {
+  it('renders', () => {
+    const { container } = render(<DescriptionList.Term>Test</DescriptionList.Term>)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 10: Same here, `term` in this case

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+    inverseColor: false,
+  },
+} satisfies Meta<typeof DescriptionList>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  decorators: [
```
alimpens, line 33: I'd add this decorator to the `meta` object. The text in the MultipleDetails story is now invisible if you set `inverseColor` to true there.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+
+.ams-description-list {
+  color: var(--ams-description-list-color);
+  display: grid;
+  font-family: var(--ams-description-list-font-family);
+  font-size: var(--ams-description-list-font-size);
+  font-weight: var(--ams-description-list-font-weight);
+  gap: var(--ams-description-list-gap);
+  grid-template-columns: 1fr;
```
VincentSmedinga, line 22: Removing this doesn’t change the appearance. Still useful to define it explicitly?

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: Indeed, a reset should be applied to one element only – make multiple resets for multiple elements.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -0,0 +1,55 @@
...
+  font-weight: var(--ams-description-list-details-font-weight);
+  padding-inline-start: var(--ams-description-list-details-padding-inline-start);
+
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    padding-inline-start: 0;
+  }
+
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    grid-column-start: 2;
+  }
```
VincentSmedinga, line 51: We can merge the identical selectors:

```suggestion
  @media screen and (min-width: $ams-breakpoint-medium) {
    grid-column-start: 2;
    padding-inline-start: 0;
  }
```

packages/css/src/components/description-list/README.md
```diff
@@ -0,0 +1,10 @@
+<!-- @license CC0-1.0 -->
+
+# Description List
+
+A collection of terms and their details.
+
+## References
```
VincentSmedinga, line 7:
```suggestion
## Design

On a narrow screen, details appear indented below their term.
From the medium breakpoint, terms and details appear next to each other.
The column for the details is twice as wide as the one for the term.

Details are set in bold text.

## References
```

packages/css/src/components/description-list/description-list.scss
dlnr: No it does not, it will skip a row in the case of two `dt` elements followed by 3 `dd` elements. But I guess it's better then leaving it out. Perhaps should UX should describe how this case should look.

alimpens: Terms in a column followed by definitions in a column is kind of how I would expect it to look. It's pretty close to the native styling as well. I'd go for it

packages/css/src/components/description-list/description-list.scss
alimpens: `margin-inline: 0;` is missing after your change

packages/react/src/DescriptionList/DescriptionListDetails.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    expect(component).toHaveClass('ams-description-list__details extra')
+  })
+
+  it('supports ForwardRef in React', () => {
+    const ref = createRef<HTMLElement>()
+
+    const { container } = render(<DescriptionList.Details ref={ref}>Test</DescriptionList.Details>)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 37: Why not use `getByRole` here?

packages/react/src/DescriptionList/DescriptionListTerm.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    expect(component).toHaveClass('ams-description-list__term extra')
+  })
+
+  it('supports ForwardRef in React', () => {
+    const ref = createRef<HTMLElement>()
+
+    const { container } = render(<DescriptionList.Term ref={ref}>Test</DescriptionList.Term>)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 37: Why not use `getByRole` here?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  args: {
+    inverseColor: true,
+  },
+  decorators: [
+    (Story, context) => (
+      <div className={context.args['inverseColor'] ? 'ams-docs-dark-background' : undefined}>
+        <Story />
+      </div>
+    ),
+  ],
```
alimpens, line 65: This isn't necessary anymore

packages/css/src/components/description-list/description-list.scss
dlnr: It's on line 12

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
dlnr: Done

packages/react/src/DescriptionList/DescriptionListTerm.test.tsx
dlnr: Testing the ref is never done by `getByRole` in our components, look at `Heading` or `Button`.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
dlnr: Should we also change this for paragraph now?

packages/css/src/components/description-list/description-list.scss
dlnr: It's in there now

packages/css/src/components/description-list/description-list.scss
alimpens: It shouldn't be part of the general reset though, it should be part of a `reset-details` mixin. The default inline margin currently isn't reset

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
alimpens: I wouldn't, can you create a separate ticket for that?

packages/react/src/DescriptionList/DescriptionListTerm.test.tsx
dlnr: Perhaps incorporate a role in a plop prompt?

VincentSmedinga (approved):

VincentSmedinga (approved):

PR #1142, opened 2024-03-24
feat!: Allow setting a background colour on Tabs List

VincentSmedinga:
The tabs list may not always have the same background colour as the parent container. Our applications have a grey background, so we need something to make the tabs list's background colour white. [An example here.](https://amsterdam.github.io/design-system-prototypes/amopis/)

Additionally, I think setting the inactive button background to white is more intentional than making it transparent, and it may be easier to animate the colour later.


proprietary/tokens/src/components/ams/tabs.tokens.json
```diff
@@ -2,10 +2,11 @@
   "ams": {
     "tabs": {
       "list": {
+        "background-color": { "value": "{ams.color.primary-white}" },
         "border-bottom": { "value": ".125rem solid {ams.color.primary-blue}" }
       },
       "button": {
-        "background-color": { "value": "transparent" },
+        "background-color": { "value": "{ams.color.primary-white}" },
```
alimpens, line 9: Setting the button `background-color` to `transparent` is a reset of sorts. Setting it to white is unnecessary if we set the list `background-color` imo.

We could move button `background-color` and `border` to a reset mixin to make this more explicit?

VincentSmedinga: Yes, I’ve added a mixin to reset the background and border.

Explicitly stating that both the list and the buttons have a white background is a good thing, in my opinion. They are different layers of this component, so I don’t think using inheritance would be correct. As a bonus, it may also help animate the button's background colour in the future. I prefer to leave `white` in.

alimpens: I disagree. The layer of the component that's styled is the list (because the part of the list that doesn't have buttons is also white). The buttons do not have a different color than their background, so its implied in the designs that they are transparent imo.

If I were to run into this code 'in the wild', I'd probably remove the button background color, because setting the same color in two places is more error-prone, and harder to change.

I also don't quite get the animation argument? Animating the button background color isn't easier or harder if you set the background color to white, is it?

VincentSmedinga: Yes, the list is white in the first place. I still think the buttons should be white as well, but it’s time to draw a conclusion – I’ve made them transparent again.

The animation thing was from the top of my head. I expect a transition to work better from white to another colour than from transparency to colour, but I haven’t checked and am happy to trust browser makers.

proprietary/tokens/src/components/ams/tabs.tokens.json
```diff
@@ -2,11 +2,12 @@
   "ams": {
     "tabs": {
       "list": {
+        "background-color": { "value": "{ams.color.primary-white}" },
         "border-bottom": { "value": ".125rem solid {ams.color.primary-blue}" }
       },
       "button": {
         "background-color": { "value": "transparent" },
-        "border": { "value": "none" },
+        "border": { "value": "" },
```
alimpens, line 10:
```suggestion
        "border": { "value": "0" },
```

packages/css/src/components/tabs/tabs.scss
```diff
@@ -5,7 +5,13 @@
 
 @import "../../common/breakpoint";
 
+@mixin reset-button {
+  background-color: transparent;
+  border: none;
```
alimpens, line 10:
```suggestion
  border: 0;
```

proprietary/tokens/src/components/ams/tabs.tokens.json
```diff
@@ -2,11 +2,12 @@
   "ams": {
     "tabs": {
       "list": {
+        "background-color": { "value": "{ams.color.primary-white}" },
         "border-bottom": { "value": ".125rem solid {ams.color.primary-blue}" }
       },
       "button": {
         "background-color": { "value": "transparent" },
-        "border": { "value": "none" },
+        "border": { "value": "0" },
```
alimpens, line 10: These and the corresponding css rules aren't necessary anymore, right? These values are already set in the reset

VincentSmedinga: They would be helpful for teams theming our application who have borders on their tab buttons. The `0` value in our token then expresses that our Tab Buttons don’t have a border.

VincentSmedinga: We decided to focus on tokens we use ourselves for now.

alimpens (approved):

PR #1143, opened 2024-03-24
feat: Start-align Dialog buttons and place the primary button first

VincentSmedinga:
The conclusion of our research is to put the primary button first in a Dialog while still aligning them to the end of the line. See [our ticket in Jira](https://gemeente-amsterdam.atlassian.net/browse/DES-553) for Bas’ writeup.

I’ve implemented this with CSS that stacks the buttons vertically and stretches them if they don’t fit next to each other – so without a breakpoint 🎉

I documented the guideline as well.


VincentSmedinga:
Previous description:

> We’re still considering this; posting a PR to get a deployment to evaluate.
> 
> Even if we don’t follow through with the interaction design, we should merge the documentation improvements.
> 
> See also:
> 
> - OK-Cancel or Cancel-OK? The Trouble With Buttons – https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/
> - “primary-to-secondary from left-to-right wins!.. but not by much” – https://medium.theoremone.co/button-ambiguity-alignment-order-a42736e25334
> - Gov.uk (old) left-aligned time-out modal – https://github.com/alphagov/govuk-design-system-backlog/issues/30#issuecomment-537013244
> - “Left is my preference too, mainly to maintain consistent button alignment across forms and dialog boxes” – https://twitter.com/AdhamDannaway/status/1415444582720888832
> - “[Buttons to the right] is much more common. However not on Windows” – https://twitter.com/vponamariov/status/1717445421188964659

packages/css/src/components/dialog/README.md
```diff
@@ -10,6 +10,13 @@ A Dialog allows the user to focus on one task or a piece of information by poppi
 - Use a dialog for short and non-frequent tasks.
   Consider using the main flow for regular tasks.
 
+## The order of buttons
+
+If your Dialog needs more than one button, put the one for the primary action first and other buttons behind it.
```
RubenSibon, line 15:
```suggestion
If your Dialog needs more than one button, put the one for the primary action first and the other buttons behind it.
```

packages/css/src/components/dialog/README.md
```diff
@@ -10,6 +10,13 @@ A Dialog allows the user to focus on one task or a piece of information by poppi
...
+Sighted users will read the primary action first, in line with the natural reading order.
+The same goes for users of screen readers, who will hear the primary action first, and users of a keyboard, who will focus the primary action first.
+Additionally, this approach keeps the order of buttons consistent on both narrow and wide screens: if the buttons don’t fit next to each other, they get stacked vertically, where having the primary action on top makes the most sense.
```
RubenSibon, line 18:
```suggestion
Also, this approach keeps the order of buttons consistent on both narrow and wide screens: if the buttons do not fit next to each other, they get stacked vertically with the primary action on top.
```

packages/css/src/components/dialog/dialog.scss
```diff
@@ -48,11 +46,13 @@
...
-  flex-direction: column;
-  grid-gap: var(--ams-dialog-footer-gap);
+  flex-wrap: wrap; // [1]
+  gap: var(--ams-dialog-footer-gap);
+  margin-inline-start: auto; // [1]
 
-  @media screen and (min-width: $ams-breakpoint-medium) {
-    flex-direction: row;
-    justify-content: end;
+  > * {
```
RubenSibon, line 53: Is it possible to use a more precise selector than a wildcard? Aren't all children buttons?

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -5,26 +5,29 @@
 
 import { Button, Dialog, Heading, Paragraph } from '@amsterdam/design-system-react/src'
 import { Meta, StoryObj } from '@storybook/react'
+import { MouseEvent } from 'react'
+
+const closeDialog = (event: MouseEvent<HTMLButtonElement>) => event.currentTarget.closest('dialog')?.close()
```
RubenSibon, line 10: For these long one-liners I'd prefer curly braces and the return keyword, but I don't if our linter/prettier allows that.

RubenSibon (approved):

alimpens:
Hm, this feels a bit off for me because of the 'right is next' convention of LTR languages, but I don't know how strong that convention is.

[NLS recommends aligning the submit button in a form to the left](https://www.nldesignsystem.nl/richtlijnen/formulieren/buttons/#plaatsing-van-een-button-binnen-een-formulier), to create a natural scan line. Isn't that also relevant here? A dialog is sort of form-like.

VincentSmedinga:
> Hm, this feels a bit off for me because of the 'right is next' convention of LTR languages, but I don't know how strong that convention is.
> 
> [NLS recommends aligning the submit button in a form to the left](https://www.nldesignsystem.nl/richtlijnen/formulieren/buttons/#plaatsing-van-een-button-binnen-een-formulier), to create a natural scan line. Isn't that also relevant here? A dialog is sort of form-like.

You’re mixing up forms and dialogs here.

Forms in dialogs are against our design guidelines. We will only use dialogs for simple confirmations, warnings or information. They will either have 1 button ‘I understand’, or 2, coming down to ‘Yes’ and ‘No’. In these contexts, having the primary action as the first one is logical when reading, listening and focusing.

Forms or other flows with multiple steps will be on a regular page and have their own rules for button placement – basically ‘Previous’ aligned to the left and ‘Next’ to the right.

alimpens:
> You’re mixing up forms and dialogs here.

No, not really. I'm just saying the guidelines on forms seems relevant here as well. Having a single scan line and having the button more visible for users that zoom in seem relevant for this case too. 

> Forms or other flows with multiple steps will be on a regular page and have their own rules for button placement – basically ‘Previous’ aligned to the left and ‘Next’ to the right.

This probably won't be the button placement, based on the guidelines I linked. Both 'Previous' and 'Next' will be left-aligned. My point is that that same alignment makes sense here imo, and it also makes our overall button alignment more consistent.

VincentSmedinga:
> > You’re mixing up forms and dialogs here.
> 
> No, not really. I'm just saying the guidelines on forms seems relevant here as well. Having a single scan line and having the button more visible for users that zoom in seem relevant for this case too.

Sorry for my misunderstanding. What change do you suggest for this PR?

> > Forms or other flows with multiple steps will be on a regular page and have their own rules for button placement – basically ‘Previous’ aligned to the left and ‘Next’ to the right.
> 
> This probably won't be the button placement, based on the guidelines I linked. Both 'Previous' and 'Next' will be left-aligned. My point is that that same alignment makes sense here imo, and it also makes our overall button alignment more consistent.

That would be fine as well – what I really meant was that form button alignment is out of scope here; shouldn’t have mentioned the explicit example. Let’s discuss that in that future PR.

alimpens:
> Sorry for my misunderstanding. What change do you suggest for this PR?

I suggest left-aligning the buttons. That way, the primary button is visually more 'anchored', and it has the 2 benefits mentioned in the NLDS guidelines on button placement in forms.

VincentSmedinga:
> I suggest left-aligning the buttons.

Ah! Only now I understand what you meant with my first comment. I will take this up with Bas. And please comment to a line or file the next time :)

storybook/src/components/Dialog/Dialog.docs.mdx
```diff
@@ -21,3 +21,10 @@ Content taller than the dialog itself will scroll.
 Click or tap the button to open a dialog.
 
 <Canvas of={DialogStories.TriggerButton} />
+
+## Vertically Stacked Buttons
+
+If the buttons don’t fit next to each other, they will stack vertically and stretch to the full width.
+View this example in a narrow window, enlarge your browser’s text size, or [use the ‘Phone’ size of the preview](?path=/story/components-containers-dialog--vertical-buttons&globals=viewport:phone).
```
alimpens, line 28: Can't we set the inline story to be less wide? So the user doesn't have to resize their window?

VincentSmedinga: Can’t find a way. There is only a `height` prop on `docs`, not `width`.

Setting a width on the dialog through an inline style strangely makes the button labels wrap, breaking the example.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -48,11 +46,13 @@
...
+    flex: auto; // [1]
   }
 }
+
+// [1] This combination stacks the buttons vertically and stretches them, until they fit next to each other.
```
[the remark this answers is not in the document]
VincentSmedinga: I think we should comment things that aren’t obvious. These three lines of code need each other to achieve the desired result. Hopefully this prevents someone in the future (maybe even ourselves) from changing or deleting one property without understanding the effect.

storybook/src/components/Dialog/Dialog.docs.mdx
VincentSmedinga: Nah, screenshots are static and age quickly.

The current setup immediately shows the vertical stacking, let’s leave it as is.

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -5,26 +5,31 @@
 
 import { Button, Dialog, Heading, Paragraph } from '@amsterdam/design-system-react/src'
 import { Meta, StoryObj } from '@storybook/react'
+import { MouseEvent } from 'react'
+
+const closeDialog = (event: MouseEvent<HTMLButtonElement>) => {
+  return event.currentTarget.closest('dialog')?.close()
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, added.

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -58,10 +63,9 @@ export const Default: Story = {
 
 export const WithScrollbar: Story = {
   args: {
-    open: true,
-    title: 'Privacyverklaring gemeente Amsterdam',
+    actions: <Button onClick={closeDialog}>Sluiten</Button>,
     children: [
-      <Heading size="level-4" key={1}>
+      <Heading size="level-5" key={1}>
```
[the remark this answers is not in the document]
VincentSmedinga: Dialog Headings have Text Level 4 (visually) in our design, so headings for child sections should be Text Level 5. I did update the semantic level to 2, following the `h1` element we use for the Dialog Heading.

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -129,3 +127,31 @@ export const TriggerButton: Story = {
...
+        <Story />
+      </div>
+    ),
+  ],
+  parameters: {
+    layout: 'fullscreen',
+    docs: {
+      story: { height: '32em' },
+    },
+    viewport: { defaultViewport: 'phone' },
```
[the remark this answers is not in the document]
VincentSmedinga: Ah, yes. Not what I expected, but not really a bad thing either. It may even help people discover the feature :)

storybook/src/components/Dialog/Dialog.docs.mdx
alimpens: We could make the button labels longer?

dlnr: I agree, but you'll need some long button text.

storybook/src/components/Dialog/Dialog.stories.tsx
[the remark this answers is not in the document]
dlnr: The long button label could fix this.

PR #1147, opened 2024-03-27
feat: Add border tokens and use rems for widths

alimpens:
- I moved the border tokens from `common` to `brand`, they make more sense there imo
- Negative values don't seem to work with these tokens, so [Tabs button still has a negative hardcoded value](https://github.com/Amsterdam/design-system/compare/feat/DES-578-border-tokens?expand=1#diff-64509d5294e86b2b12c8e2cb3deb74ddeb8654eb63a09bd25743a5d8e307f9edR21)
- In Table, the border is set on the cell, because a border on the row doesn't work with the default `border-collapse: separate`. `border-collapse: collapse` caused other issues I think, but I can't quite remember what they were.

### DIFF
--- a/packages/css/src/components/button/button.scss
+++ b/packages/css/src/components/button/button.scss
@@ -28,7 +28,7 @@
 
 @mixin ams-button-forced-color-mode {
   @media screen and (-ms-high-contrast: active), screen and (forced-colors: active) {
-    border: 2px solid ButtonBorder; // add border because forced colors changes box-shadow to none
+    border: var(--ams-button-forced-color-mode-border); // add border because forced colors changes box-shadow to none
   }
 }
 
--- a/packages/css/src/components/button/button.scss
+++ b/packages/css/src/components/button/button.scss
@@ -6,7 +6,7 @@
 @import "../../common/text-rendering";
 
 .ams-button {
-  border: none;
+  border: 0;
   cursor: var(--ams-button-cursor);
   display: inline-flex;
   font-family: var(--ams-button-font-family);
--- a/packages/css/src/components/button/button.scss
+++ b/packages/css/src/components/button/button.scss
@@ -5,8 +5,11 @@
 
 @import "../../common/text-rendering";
 
-.ams-button {
+@mixin reset-button {
   border: 0;
+}
+
+.ams-button {
   cursor: var(--ams-button-cursor);
   display: inline-flex;
   font-family: var(--ams-button-font-family);
@@ -24,6 +27,7 @@
   }
 
   @include text-rendering;
+  @include reset-button;
 }
 
 @mixin ams-button-forced-color-mode {
--- a/proprietary/tokens/src/brand/ams/border.tokens.json
+++ b/proprietary/tokens/src/brand/ams/border.tokens.json
@@ -1,0 +1,10 @@
+{
+  "ams": {
+    "border-width": {
+      "sm": { "value": "0.0625rem" },
+      "md": { "value": "0.125rem" },
+      "lg": { "value": "0.1875rem" },
+      "xl": { "value": "0.25rem" }
+    }
+  }
+}
--- a/proprietary/tokens/src/brand/ams/border.tokens.json
+++ b/proprietary/tokens/src/brand/ams/border.tokens.json
@@ -1,10 +1,12 @@
 {
   "ams": {
-    "border-width": {
-      "sm": { "value": "0.0625rem" },
-      "md": { "value": "0.125rem" },
-      "lg": { "value": "0.1875rem" },
-      "xl": { "value": "0.25rem" }
+    "border": {
+      "width": {
+        "sm": { "value": "0.0625rem" },
+        "md": { "value": "0.125rem" },
+        "lg": { "value": "0.1875rem" },
+        "xl": { "value": "0.25rem" }
+      }
     }
   }
 }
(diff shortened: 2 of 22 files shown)


packages/css/src/components/button/button.scss
```diff
@@ -28,7 +28,7 @@
 
 @mixin ams-button-forced-color-mode {
   @media screen and (-ms-high-contrast: active), screen and (forced-colors: active) {
-    border: 2px solid ButtonBorder; // add border because forced colors changes box-shadow to none
+    border: var(--ams-button-forced-color-mode-border); // add border because forced colors changes box-shadow to none
```
VincentSmedinga, line 31: Should we add a ticket to check all our box shadows for forced color mode and add some docs for it?

proprietary/tokens/src/brand/ams/border.tokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ams": {
+    "border-width": {
```
VincentSmedinga, line 3: Considering these are the ‘border’ tokens, as the file name implies, it would make sense to split `border` and `width` here. This allows us to add the likes of `ams.border.style` later as siblings below `border`.

The custom properties in CSS will still match the property name. In JSON and JS, ‘width’ and ‘style’ will become children of ‘border’ which is conceptually also correct.

proprietary/tokens/src/components/ams/button.tokens.json
```diff
@@ -12,41 +12,44 @@
       "disabled": {
         "cursor": { "value": "{ams.action.disabled.cursor}" }
       },
+      "forced-color-mode": {
+        "border": { "value": "{ams.border-width.md} solid ButtonBorder" }
```
VincentSmedinga, line 16: Do we keep `ButtonBorder`?

proprietary/tokens/src/brand/ams/border.tokens.json
VincentSmedinga, line 1: Please describe these tokens in a little documentation page as we do for e.g. Space.

proprietary/tokens/src/components/ams/button.tokens.json
alimpens: This uses a [system color](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color) for forced colors mode, which makes sense imo.

storybook/src/docs/borders.docs.mdx
```diff
@@ -0,0 +1,20 @@
...
+## Widths
+
+We have 4 border widths:
+
+| Size        | px  | rem    | Token name          |
+| ----------- | --- | ------ | ------------------- |
+| Small       | 1   | 0.0625 | ams.border.width.sm |
+| Medium      | 2   | 0.125  | ams.border.width.md |
+| Large       | 3   | 0.1875 | ams.border.width.lg |
+| Extra-large | 4   | 0.25   | ams.border.width.xl |
```
VincentSmedinga, line 16: This is a bit more inline with the design of other tables:

```suggestion
| Size            | px  |  rem   | Token name            |
| :-------------- | :-: | :----: | :-------------------- |
| **Small**       |  1  | 0.0625 | `ams.border.width.sm` |
| **Medium**      |  2  | 0.125  | `ams.border.width.md` |
| **Large**       |  3  | 0.1875 | `ams.border.width.lg` |
| **Extra large** |  4  |  0.25  | `ams.border.width.xl` |
```

packages/css/src/components/button/button.scss
```diff
@@ -6,7 +6,7 @@
 @import "../../common/text-rendering";
 
 .ams-button {
-  border: none;
+  border: 0;
```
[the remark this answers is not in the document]
VincentSmedinga: I’ve researched a bit. The actual effect of `border: 0` and `border: none` for all longhand properties are the same because the browser sets all others to their initial values, and `0` and `none`  are the initial values for `-width` and `-style`, respectively.

Theoretically, there would be different results with partial cascading.

If used as a reset, I am happy with `border: 0`.

To set or override the width of a particular border, `border-width: 0` would be more precise.

packages/css/src/components/button/button.scss
VincentSmedinga: Yes, separating feels better because it doesn’t mix up testing with adding features.

proprietary/tokens/src/components/ams/button.tokens.json
VincentSmedinga: Agreed. But should we declare it explicitly? My understanding is that [the browser forces this](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors#usage_notes) already and “In general, web authors should not be using the forced-colors media feature to create a separate design for users with this feature enabled.”

alimpens: Without `ButtonBorder` the browser uses the currentColor I think, which is the same as the button text color. So we do not need to define it explicitly.

packages/css/src/components/button/button.scss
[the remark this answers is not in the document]
VincentSmedinga: Discussed: having `border-width` doesn’t seem to have real advantages after all.

Overriding `border` with only a width doesn’t work because the shorthand had already reset `style` to `none`.

Additionally, providing a generic `border` token in every situation gives people who theme our components the most freedom to set all longhand properties to their liking.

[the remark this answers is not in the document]
alimpens: As discussed: `border: 0` does the same thing as `border-width: 0` in every practical sense, and having 1 way of setting this is a simpler convention than having multiple ways.

VincentSmedinga (approved):

PR #1149, opened 2024-03-29
fix: Fix autodocs for stories

alimpens:


storybook/src/components/Mark/Mark.stories.tsx
```diff
@@ -3,14 +3,14 @@
...
-import { Mark, Paragraph } from '@amsterdam/design-system-react'
+import { Mark, Paragraph } from '@amsterdam/design-system-react/src'
 import { Meta, StoryObj } from '@storybook/react'
 
 const meta = {
   title: 'Components/Text/Mark',
   component: Mark,
   args: {
-    children: 'Nieuw component',
+    children: 'Wat vinden Amsterdammers belangrijk?',
```
VincentSmedinga, line 13: Not sure whether this should be the default value if we only have one story. Chances are that a second story would mark a different phrase and we’d move this text back again. I’d say move it back and remove the default arg.

storybook/src/components/OrderedList/OrderedList.stories.tsx
```diff
@@ -22,8 +22,6 @@ const meta = {
     start: undefined,
   },
   argTypes: {
```
VincentSmedinga, line 24: Can we document that HTML attributes still require configuring `argTypes`?

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -19,11 +19,6 @@ const meta = {
     children: {
       table: { disable: false },
```
VincentSmedinga, line 20: Props for not nuking everything of all of the `argTypes` (pun intended). Can we document this config for people to consider?

storybook/src/components/PageMenu/PageMenu.stories.tsx
```diff
@@ -12,15 +12,7 @@ const meta = {
...
-    wrap: true,
-  },
-  argTypes: {
-    alignEnd: {
-      control: 'boolean',
-    },
-    wrap: {
-      control: 'boolean',
-    },
+    wrap: undefined,
```
VincentSmedinga, line 15: I liked how the Args Table would show the default values if we specify them but I understand and agree we don’t want them to appear in the Code View.

Can we remove `alignEnd` with the same reasoning?

Stronger yet: why would we still define args in `meta` at all?

storybook/src/components/Screen/Screen.docs.mdx
```diff
@@ -8,6 +8,8 @@ import README from "../../../../packages/css/src/components/screen/README.md?raw
 
 <Primary />
 
+<Controls />
+
```
VincentSmedinga, line 12: Grid is missing Controls as well and I’m finding some others if I search for `import { Canvas, M`.

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -12,6 +12,10 @@ import { HeadingLevel } from '../Heading/Heading'
 import { useKeyboardFocus } from '../common/useKeyboardFocus'
 
 export type AccordionProps = {
+  /**
+   * The hierarchical level of the accordion title within the document.
+   * @default 1
```
VincentSmedinga, line 17: Can we document that we need to specify default values for components that don’t define them but just pass the prop on?

VincentSmedinga (commented):
Great improvement for our users! And good for us that we don’t need to replicate prop configs in `argTypes` anymore.

packages/react/src/Accordion/Accordion.tsx
alimpens: Hm, this is actually an error. I copied this from Alert, but the Accordion headingLevel doesn't have a default. 

For the doc gen we don't need to define these defaults, it gets it from the default props. It does show in your IDE, but documenting that may be a bit out of scope for this PR?

storybook/src/components/Button/Button.stories.tsx
alimpens: I'm not sure what you mean. Do you mean documenting that you automatically get your typing as controls, but if you want to show something other than the automatic controls, you have to use `argTypes`?

storybook/src/components/PageMenu/PageMenu.stories.tsx
alimpens: I don't think we need to, `alignEnd` doesn't show in the code view. I generally think it's better when booleans have a default value. I only changed this line because `wrap` has a default value of `true`, which means it shows in the doc view.

storybook/src/components/Screen/Screen.docs.mdx
alimpens: I think the others are all components without props or complicated components which don't have obvious controls (like `Grid`). `Screen` is a simple component with obvious controls, which is why I added them. 

I think we have to think about how we want to document the more complicated components.

storybook/src/components/Button/Button.stories.tsx
VincentSmedinga: No, sorry I meant that hiding `children` from the Args Table is one thing that may still be necessary to configure, next to declaring HTML attributes.

documentation/storybook.md
```diff
@@ -20,6 +20,11 @@ We write our documentation in English, the stories are Dutch.
 
 ## Best practices for controls
 
+Controls are automatically generated based on the component’s typing.
+If you want to document native HTML attributes, you can use [`argTypes`](https://storybook.js.org/docs/api/arg-types).
+You can also use `argTypes` to override the automatically generated controls.
+Be sure to follow these guidelines when you do:
+
 1. For props offering five options or less, use radio buttons rather than a select.
```
VincentSmedinga, line 28: Is this still valid with Auto Docs on?

storybook/src/components/Button/Button.stories.tsx
alimpens: Got it, I've added it to the docs

storybook/src/components/Screen/Screen.docs.mdx
VincentSmedinga: Discussed and added them for `Grid` and `Header`.

VincentSmedinga (approved):

PR #1150, opened 2024-04-04
fix: Use correct text size in Text Input, Search Field, and Text Area

dlnr:


dlnr (commented):
Padding for textarea looks kind of small now, we should also have a look with the ux-ers at the inside spacing stuff.

VincentSmedinga (approved):

PR #1151, opened 2024-04-05
feat: Select component

dlnr:
As discussed the combobox MVP will have a native dropdown. I've created child components to align without our approach in other components and align with NLDS.


storybook/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+const meta = {
+  title: 'Components/Forms/Select',
+  component: Select,
+  args: {
+    invalid: false,
+    required: false,
+    disabled: false,
+    multiple: false,
+    children: [
+      <Select.Option key="0" disabled selected hidden>
```
dlnr, line 18: [This article](https://www.w3docs.com/snippets/css/how-to-create-a-placeholder-for-an-html5-select-box-by-using-only-html-and-css.html) has a cool way to create a placeholder for the select and I decided to adopt it. What do you guys think? Shall I document this?

dlnr: Does not seem to hide on iOs though:
image

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
...
+
+  &:hover {
+    box-shadow: var(--ams-select-hover-box-shadow);
+  }
+
+  @include reset;
+}
+
+.ams-select:disabled,
+.ams-select--disabled {
```
alimpens, line 41: This class isn't used

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
...
+  box-shadow: var(--ams-select-disabled-box-shadow);
+  color: var(--ams-select-disabled-color);
+  cursor: not-allowed;
+
+  &:not([multiple]) {
+    background-image: var(--ams-select-disabled-background-image);
+  }
+}
+
+.ams-select--invalid,
```
alimpens, line 51: I don't think we need a class for this, we can use `[aria-invalid="true"]`

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
...
+.ams-select:required:invalid {
```
alimpens, line 52: Do we need the `:required` part? If the select is required and doesn't have a value, it is invalid with or without that part, right?

packages/react/src/Select/SelectOption.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, OptionHTMLAttributes, PropsWithChildren } from 'react'
+
+export type SelectOptionProps = {
+  value?: string
```
alimpens, line 11: I don't think we really need this prop, `restProps` already does this, no?

packages/react/src/Select/SelectOption.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { render, screen } from '@testing-library/react'
+import { createRef } from 'react'
+import { Select } from './Select'
+import '@testing-library/jest-dom'
+
+describe('Select', () => {
```
alimpens, line 6:
```suggestion
describe('Select option', () => {
```

packages/react/src/Select/SelectOptionGroup.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { render, screen } from '@testing-library/react'
+import { createRef } from 'react'
+import { Select } from './Select'
+import '@testing-library/jest-dom'
+
+describe('Select', () => {
```
alimpens, line 6:
```suggestion
describe('Select option group', () => {
```

storybook/src/components/Select/Select.docs.mdx
```diff
@@ -0,0 +1,36 @@
...
+<Controls />
+
+## Multiple
+
+With `multiple` enabled the attribute `size` is allowed to set the number of rows. Multi select is harder to use
+on desktop as they require the user to hold down the `Ctrl` or `Cmd` key while clicking on the options.
+You might need to describe this behavior to the user. It is recommended to use checkboxes instead.
+On the other hand, it is a good choice for mobile devices.
+
+<Canvas of={SelectStories.Multiple} />
```
alimpens, line 22: I'm not sure we should show an example of `multiple`. A single select is already tricky to use for users, a multi select is downright horrible on desktops. It's nice that it works better on mobile devices, but you can't really know if something is only going to be used on a mobile device (if you're not building apps).

What do you think @VincentSmedinga?

storybook/src/components/Select/Select.docs.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Grouped
+
+Use the `Select.Group` component to group options, the component requires a `label` attribute.
```
alimpens, line 26:
```suggestion
Use the `Select.Group` component to group options.
The component requires a `label` attribute.
```

storybook/src/components/Select/Select.stories.tsx
alimpens: If it causes issues, I wouldn't use it

storybook/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+import { Select } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Components/Forms/Select',
+  component: Select,
+  args: {
+    invalid: false,
+    required: false,
```
alimpens, line 14: `required` is a native attribute, that doesn't change the styling in our implementation. I don't think we really need to have it as a control.

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
...
+  box-shadow: var(--ams-select-invalid-box-shadow);
+
+  &:hover {
+    box-shadow: var(--ams-select-invalid-hover-box-shadow);
+  }
+}
```
alimpens, line 58: Shouldn't we style the disabled option as well? It currently is its native grey.

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.ams-select {
+  background-color: var(--ams-select-background-color);
+  box-shadow: var(--ams-select-box-shadow);
+  color: var(--ams-select-color);
+  font-family: var(--ams-select-font-family);
+  font-size: var(--ams-select-font-size);
+  font-weight: var(--ams-select-font-weight);
+  line-height: var(--ams-select-line-height);
+  max-inline-size: var(--ams-select-max-inline-size);
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, sizing inputs is of later concern.

storybook/src/components/Select/Select.docs.mdx
VincentSmedinga: It does work great on (some, most?) mobile devices, but I agree that usability on desktops is bad. So I’m afraid I must agree to remove this for now and rewrite this paragraph to a ‘Do not use’ guideline.

storybook/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+export const Multiple: Story = {
+  args: {
+    multiple: true,
+    size: 5,
+  },
+}
+
+export const Grouped: Story = {
+  args: {
+    children: [
```
VincentSmedinga, line 61: Nice example content :)

storybook/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <Select.Option>Marnixbuurt Noord</Select.Option>
+        <Select.Option>Marnixbuurt Zuid</Select.Option>
+        <Select.Option>Passeerdersgrachtbuurt</Select.Option>
+        <Select.Option>Zaagpoortbuurt</Select.Option>
+      </Select.Group>,
+    ],
+  },
+}
+
+export const Invalid: Story = {
```
VincentSmedinga, line 113: Should we select ‘Westpoort’ in this example as this is the invalid option?

If yes, you could refactor the `meta.children` options to a function that accepts a `selected` value.

packages/css/src/components/select/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `Home` or `End` key to navigate to the first or last option
+- `PageUp` or `PageDown` key to navigate to the previous or next option that is currently not visible
+- `Escape` key to close the dropdown
+- `Tab` key to close the dropdown and move focus to the next focusable element
+- `Shift + Tab` key to close the dropdown and move focus to the previous focusable element
+- Select the item that starts with the typed character(s) when the dropdown is open
+
+## References
+
+- [Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)
```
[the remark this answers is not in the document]
dlnr: This is the default role of the select.

packages/css/src/components/select/select.scss
dlnr: Like the disabled class this is copied from NLDS to anticipate different usage (form frameworks or handling) or visual regression testing. But it can be removed.

packages/css/src/components/select/select.scss
dlnr: Copied from NLDS, I assumed it was to prevent showing red borders on a select that still needs you to select an option. Using `invalid` on input and select gives mixed results in my experience, and do not bahve the same.

packages/css/src/components/select/select.scss
dlnr: So turns out I left this in to fix the order of things. Stylelint is acting strange because there is a hover on disabled.

proprietary/tokens/src/components/ams/select.tokens.json
```diff
@@ -0,0 +1,36 @@
+{
+  "ams": {
+    "select": {
+      "background-color": { "value": "{ams.color.primary-white}" },
+      "background-image": {
+        "value": "url(\"data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><path fill='%23004699' fill-rule='evenodd' d='m16 25.757-16-16 2.91-2.9L16 19.937l13.09-13.08 2.91 2.9z'/></svg>\")"
+      },
+      "box-shadow": { "value": "inset 0 0 0 {ams.border.width.sm} {ams.color.primary-black}" },
```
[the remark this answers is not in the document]
dlnr: TextInput only has a medium border on hover.

proprietary/tokens/src/components/ams/select.tokens.json
```diff
@@ -0,0 +1,36 @@
...
+      "color": { "value": "{ams.color.primary-black}" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-size": { "value": "{ams.text.level.5.font-size}" },
+      "font-weight": { "value": "{ams.text.font-weight.normal}" },
+      "line-height": { "value": "{ams.text.level.5.line-height}" },
+      "outline-offset": { "value": "{ams.focus.outline-offset}" },
+      "padding-block": { "value": "{ams.space.inside.xs}" },
+      "padding-inline": { "value": "{ams.space.inside.md}" },
+      "max-inline-size": { "value": "28rem" },
```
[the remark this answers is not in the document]
dlnr: The width was copied from NLDS, not improvised.

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
alimpens: I think we still need `max-inline-size: 100%` to prevent the component from breaking out of its container

storybook/src/components/Select/Select.docs.mdx
dlnr: Could you make a suggestion on a 'Do not use' policy without example?

packages/css/src/components/select/select.scss
alimpens: >  I assumed it was to prevent showing red borders on a select that still needs you to select an option

I don't really understand what you mean here. Is this something you've tested and is an issue in our case? If not, I wouldn't add it

storybook/src/components/Select/Select.stories.tsx
dlnr: It was to demonstrate you can set an indiviual option to disabled.

proprietary/tokens/src/components/ams/select.tokens.json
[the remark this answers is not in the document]
VincentSmedinga: Ah yes, my bad.

packages/css/src/components/select/select.scss
```diff
@@ -0,0 +1,58 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@mixin reset {
+  appearance: none;
+  border: none;
+}
```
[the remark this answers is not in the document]
VincentSmedinga: I’ve updated the suggestion. Should we consider it a reset?

packages/react/src/Select/SelectOption.tsx
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, OptionHTMLAttributes, PropsWithChildren } from 'react'
+
+export type SelectOptionProps = {} & OptionHTMLAttributes<HTMLOptionElement>
```
alimpens, line 10:
```suggestion
export type SelectOptionProps = OptionHTMLAttributes<HTMLOptionElement>
```

storybook/src/components/Select/Select.docs.mdx
alimpens: Maybe try it yourself first, and we can add onto it? [You can look at gov.uk for inspiration](https://design-system.service.gov.uk/components/select/#avoid-adding-functionality-to-allow-selecting-multiple-options).

proprietary/tokens/src/components/ams/select.tokens.json
```diff
@@ -0,0 +1,35 @@
...
+      },
+      "box-shadow": { "value": "inset 0 0 0 {ams.border.width.sm} {ams.color.primary-black}" },
+      "color": { "value": "{ams.color.primary-black}" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-size": { "value": "{ams.text.level.5.font-size}" },
+      "font-weight": { "value": "{ams.text.font-weight.normal}" },
+      "line-height": { "value": "{ams.text.level.5.line-height}" },
+      "outline-offset": { "value": "{ams.focus.outline-offset}" },
+      "padding-block": { "value": "{ams.space.inside.xs}" },
+      "padding-inline": { "value": "{ams.space.inside.md} calc({ams.space.inside.md} * 2 + 32px)" },
```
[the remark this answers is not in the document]
alimpens: Not quite, it currently is `ams.space.inside.md` + 32px(?).

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
alimpens: Yeah, I think so. The comment can be a bit more verbose imo, maybe something like `// Reset rounded borders for Safari on MacOS`?

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
dlnr: The reason I've created a token is because NLDS seems to set this and perhaps you're like to be able to override this in your application somehow.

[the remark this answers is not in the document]
alimpens: I don't think we should. The Utrecht Select combines `max-inline-size` with `inline-size: 100%` to make sure the select doesn't break out of its container. Seeing as we don't do that, allowing users to set `max-inline-size` to a static value would break the component.

storybook/src/components/Select/Select.docs.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Controls />
+
+## Multiple
+
+Avoid adding functionality to allow selecting multiple options. Multi select is harder to use
+on desktop as they require the user to hold down the `Ctrl` or `Cmd` key while clicking on the options.
+It is recommended to use checkboxes instead.
+
+- [WCAG 2.2 success criterion](https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html)
```
alimpens, line 21: Is this relevant to multi select interaction?

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
dlnr: I've added `inline-size: 100%`. Preventing users to set a width seems like a bad idea.

storybook/src/components/Select/Select.docs.mdx
dlnr: Yes this is the gov.uk reference to why you should not allow multiple. Since we don't actively prevent the usage (you can still add the attribute and the styling is there) I included this like for the curious developer.

proprietary/tokens/src/components/ams/select.tokens.json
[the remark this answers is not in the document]
dlnr: Changed

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
VincentSmedinga: People can still add a class name and set a width. We haven’t researched how to size inputs yet. On the grid? Through a set of t-shirt sizes? A number of variants? And how to implement variations for varying screen widths. So 100% is a good, predictable start.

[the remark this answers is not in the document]
dlnr: So it's ok as is?

proprietary/tokens/src/components/ams/select.tokens.json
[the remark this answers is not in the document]
alimpens: It doesn't seem to be?

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
dlnr: I don't realy get the argument to not use a token but here you go.

alimpens (dismissed):

packages/css/src/components/index.scss
```diff
@@ -4,6 +4,8 @@
  */
 
 /* Append here */
+
```
alimpens, line 7:
```suggestion
```

PR #1152, opened 2024-04-05
feat: Add date input

alimpens:
The native date input looks different on Chromium browsers and Firefox, and even more so on Android and iOS. 

The icon currently looks as designed on Chromium. On Firefox the icon is difficult to style. Android and iOS do not have a calender icon, but we could add it of course.


dlnr:
Off the bat, the input appears blank on mobile devices (both android and apple)

image

VincentSmedinga (approved):
MVP to the max but let’s roll!

PR #1153, opened 2024-04-05
feat: Plop prompts

dlnr:
To include the right tag, type, attributes and role.


alimpens:
Nice! It adds some extra prompts, but it is easy enough to skip them if you can't / don't want to fill those out

alimpens (approved):

PR #1155, opened 2024-04-10
chore: Update minor dependencies

alimpens:


packages/css/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "devDependencies": {
     "@amsterdam/design-system-tokens": "workspace:*",
-    "sass": "1.72.0"
+    "sass": "1.74.1"
   },
   "dependencies": {
     "@utrecht/components": "3.0.0"
```
VincentSmedinga, line 26: I don’t think we use the Utrecht components anymore.

alimpens: Good point, I'll remove it

VincentSmedinga (approved):

PR #1179, opened 2024-04-11
feat: Refactor inverseColor stories

dlnr:


storybook/src/components/Blockquote/Blockquote.stories.tsx
```diff
@@ -21,31 +21,23 @@ const meta = {
...
     },
   },
-} satisfies Meta<typeof Blockquote>
-
-export default meta
-
-type Story = StoryObj<typeof meta>
-
-export const Default: Story = {
   decorators: [
```
alimpens, line 24: Nice fix! We discussed extracting this decorator into a reusable one, is that possible? Maybe add a `common` folder next to `docs` and `components`

VincentSmedinga: Yes, added!

PR #1213, opened 2024-04-24
chore: Translate prop descriptions

alimpens:


RubenSibon (approved):

packages/react/src/PageHeading/PageHeading.tsx
```diff
@@ -8,10 +8,7 @@ import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
 
 export type PageHeadingProps = {
-  /**
-   * De kleur van de titel
-   * Gebruik deze property om de titel in tegenovergestelde kleur te tonen.
-   */
+  /** Property to show the heading in the opposite color. */
```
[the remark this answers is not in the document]
VincentSmedinga: We could use ‘opposite’ instead of ‘inverse’ later, but I think ‘complementary’ has the connotation of ‘colours that work well with each other’ in everyday speech.

I’ve changed the description for the `inverseColor` prop.

packages/react/src/Blockquote/Blockquote.tsx
```diff
@@ -8,10 +8,7 @@ import { forwardRef } from 'react'
 import type { BlockquoteHTMLAttributes, ForwardedRef, PropsWithChildren } from 'react'
 
 export type BlockquoteProps = {
-  /**
-   * De kleur van het citaat.
-   * Gebruik deze property om het citaat in tegenovergestelde kleur te tonen.
-   */
+  /** Changes the text color for readability on a dark background. */
```
VincentSmedinga, line 11: @alimpens This is the most compact description ChatGPT and I could think of. What do you think?

I preferred not to start with ‘A property to…’ – it’s clear that it’s a prop and then we could start every description with it. I’m looking for an active voice, capturing the essence of what a prop means or does, and sometimes adding a fragment on how or when to use.

I’ll keep notes of guidelines when I pick up the other two tasks :)

alimpens: I like it! It's concise, generic and describes what it does and why it does it. My suggestion only did the former.

VincentSmedinga (approved):

PR #1217, opened 2024-05-03
feat: Allow small text for unordered lists

alimpens:
Do we want the use the same API as for Paragraph? So `size="small"`? Or a `small` boolean?

### DIFF
--- a/packages/react/src/UnorderedList/UnorderedList.tsx
+++ b/packages/react/src/UnorderedList/UnorderedList.tsx
@@ -12,11 +12,13 @@ export type UnorderedListProps = {
   /** Changes the text color for readability on a dark background. */
   inverseColor?: boolean
   markers?: boolean
+  /** The size of the unordered list. */
+  size?: 'small'
 } & PropsWithChildren<HTMLAttributes<HTMLUListElement>>
 
 const UnorderedListRoot = forwardRef(
   (
-    { children, className, inverseColor, markers = true, ...restProps }: UnorderedListProps,
+    { children, className, inverseColor, markers = true, size, ...restProps }: UnorderedListProps,
     ref: ForwardedRef<HTMLUListElement>,
   ) => {
     return (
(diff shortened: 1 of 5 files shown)


RubenSibon:
> Do we want the use the same API as for Paragraph? So `size="small"`? Or a `small` boolean?

Yes, keep it consistent for the developers.

RubenSibon (approved):

packages/react/src/UnorderedList/UnorderedList.tsx
```diff
@@ -12,11 +12,13 @@ export type UnorderedListProps = {
   /** Changes the text color for readability on a dark background. */
   inverseColor?: boolean
   markers?: boolean
+  /** The size of the unordered list. */
```
RubenSibon, line 15: What are our guidelines for documenting comments in code? Not every prop has comment, I see. When do we add one and when not? Why not JSDoc styled comments?

RubenSibon (approved):

packages/react/src/UnorderedList/UnorderedList.tsx
alimpens: We haven't done so in the past, but all props should have documentation. We have [two](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378?selectedIssue=DES-752) [tickets](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378?selectedIssue=DES-753) to fix that.

> Why not JSDoc styled comments?

I'm not quite sure what you mean? This is JSDoc.

RubenSibon: Right, a different style than I'm used to (documenting props/params on the Function/Component level), but this is fine by me. Is it an idea to add comments for all props when we touch a file (i.e. Boy Scout Rule)?

alimpens: The idea now is to add comments for all props in 2 batches (see the tickets I linked earlier). I think that's doable, there's a lot of repetition there.

PR #1218, opened 2024-05-07
feat: File Input

dlnr:
Improvised (no design) implementation of the native file input element.


packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+@import "../../common/text-rendering";
+
+@mixin reset-button {
+  border: 0;
+}
+
+.ams-file-input {
+  align-items: center;
```
dlnr, line 13: Required to align the button text with the filename text on iOS.

packages/css/src/components/file-input/README.md
```diff
@@ -0,0 +1,9 @@
+<!-- @license CC0-1.0 -->
+
+# File Input
+
+The file input component is a styled input element that allows users to select a file from their device.
```
VincentSmedinga, line 5:
```suggestion
Allows the user to upload one or more files from their device.
```

packages/css/src/components/file-input/file-input.scss
alimpens: Are you sure? This doesn't seem to do anything on an iPhone 13 in Safari

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  color: var(--ams-file-input-color);
+  display: flex;
```
alimpens, line 15: This doesn't seem to do anything

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.ams-file-input:disabled {
+  color: var(--ams-file-input-disabled-color);
+  cursor: not-allowed;
+}
+
+.ams-file-input::file-selector-button {
+  background-color: var(--ams-file-input-file-selector-button-background-color);
+  box-shadow: var(--ams-file-input-file-selector-button-box-shadow);
```
alimpens, line 35: This border is missing on iOS

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  color: var(--ams-file-input-file-selector-button-color);
+  cursor: var(--ams-file-input-file-selector-button-cursor);
+  display: inline-block;
```
alimpens, line 38: Does this do anything?

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  font-family: inherit;
+  font-size: inherit;
```
alimpens, line 40: This is the default

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+.ams-file-input:disabled {
+  color: var(--ams-file-input-disabled-color);
+  cursor: not-allowed;
+}
+
+.ams-file-input::file-selector-button {
+  background-color: var(--ams-file-input-file-selector-button-background-color);
+  box-shadow: var(--ams-file-input-file-selector-button-box-shadow);
+  color: var(--ams-file-input-file-selector-button-color);
+  cursor: var(--ams-file-input-file-selector-button-cursor);
```
alimpens, line 37: Shouldn't `cursor: pointer` be defined for the whole input? The entire thing is clickable

packages/react/src/FileInput/FileInput.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  it('supports ForwardRef in React', () => {
+    const ref = createRef<HTMLInputElement>()
+
+    const { container } = render(<FileInput ref={ref} />)
+    const component = container.querySelector(':only-child')
+
+    expect(ref.current).toBe(component)
+  })
+})
```
alimpens, line 37: Maybe we should check that it has `type="file"`? With something like this:

`const component = container.querySelector('input[type="file"]')`

Maybe we could simply do that in the existing tests, instead of:

`const component = container.querySelector(':only-child')`

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,32 @@
+{
+  "ams": {
+    "file-input": {
+      "color": { "value": "{ams.color.primary-black}" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-size": { "value": "{ams.text.level.5.font-size}" },
+      "line-height": { "value": "{ams.text.level.5.line-height}" },
```
alimpens, line 7: `font-weight` is not defined here

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,32 @@
...
+      "outline-offset": { "value": "{ams.focus.outline-offset}" },
+      "placeholder": {
+        "color": { "value": "{ams.color.neutral-grey3}" }
+      },
```
alimpens, line 11: Does input type=file have a placeholder? It's not defined in the CSS

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,32 @@
...
+      "disabled": {
+        "color": { "value": "{ams.color.neutral-grey2}" }
+      },
+      "file-selector-button": {
+        "background-color": { "value": "{ams.color.primary-white}" },
+        "color": { "value": "{ams.color.primary-blue}" },
+        "box-shadow": { "value": "inset 0 0 0 {ams.border.width.md} {ams.color.primary-blue}" },
```
alimpens, line 18: `cursor` is used in the CSS, but it's not defined here. Although it's probably a better idea to define the cursor for the whole input

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  padding-inline: var(--ams-file-input-file-selector-button-padding-inline);
+
+  @include reset-button;
+}
+
+.ams-file-input:disabled::file-selector-button {
+  background-color: var(--ams-file-input-file-selector-button-disabled-background-color);
+  box-shadow: var(--ams-file-input-file-selector-button-disabled-box-shadow);
+  color: var(--ams-file-input-disabled-color);
+  cursor: var(--ams-file-input-file-selector-button-disabled-cursor);
```
alimpens, line 54: This is unnecessary, the disabled cursor is already defined on the input.

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  /* stylelint-disable-next-line property-disallowed-list */
+  margin-right: 2ex; // Add space between the file selector button and the file name
+  padding-block: var(--ams-file-input-file-selector-button-padding-block);
+  padding-inline: var(--ams-file-input-file-selector-button-padding-inline);
+
+  @include reset-button;
+}
+
+.ams-file-input:disabled::file-selector-button {
+  background-color: var(--ams-file-input-file-selector-button-disabled-background-color);
```
alimpens, line 51: I don't think this is necessary, the background color is always white.

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,32 @@
...
+        "padding-block": { "value": "{ams.space.inside.xs}" },
+        "padding-inline": { "value": "{ams.space.inside.md}" },
+        "hover": {
+          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.lg} {ams.color.dark-blue}" },
+          "color": { "value": "{ams.color.dark-blue}" }
```
alimpens, line 23: This is not used in the CSS, but it should be

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  box-shadow: var(--ams-file-input-file-selector-button-disabled-box-shadow);
+  color: var(--ams-file-input-disabled-color);
+  cursor: var(--ams-file-input-file-selector-button-disabled-cursor);
+}
+
+.ams-file-input:not(:disabled):hover::file-selector-button {
+  background-color: var(--ams-file-input-file-selector-button-hover-background-color);
```
alimpens, line 58: This isn't necessary I think, the background color is always white.

storybook/src/components/FileInput/FileInput.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+    accept: '',
+    multiple: false,
+  },
+  argTypes: {
+    accept: {
+      control: {
+        type: 'text',
+      },
+    },
+    multiple: {
```
alimpens, line 22: I think we should add `disabled` here as well

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright Gemeente Amsterdam
+ */
+
+@import "../../common/text-rendering";
+
+@mixin reset-button {
+  border: 0;
+}
+
+.ams-file-input {
```
alimpens, line 12: I think we should add a border around the entire clickable area, to make it more obvious that it's clickable

packages/css/src/components/file-input/file-input.scss
dlnr: No it falls back to sans, you're welcome to try.

packages/css/src/components/file-input/file-input.scss
dlnr: it allows me to align the text on iOS

packages/css/src/components/file-input/file-input.scss
dlnr: It was looking too much like a text input before?

packages/css/src/components/file-input/file-input.scss
dlnr: Added this rule before changing line-height and font, not needed

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,60 @@
...
+  background-color: var(--ams-file-input-file-selector-button-background-color);
+  box-shadow: var(--ams-file-input-file-selector-button-box-shadow);
+  color: var(--ams-file-input-file-selector-button-color);
+  cursor: var(--ams-file-input-file-selector-button-cursor);
+  display: inline-block;
+  font-family: inherit;
+  font-size: inherit;
+  font-weight: inherit;
+  /* stylelint-disable-next-line property-disallowed-list */
+  margin-right: 2ex; // Add space between the file selector button and the file name
```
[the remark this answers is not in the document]
dlnr: Removed in favor of consistency.

packages/css/src/components/file-input/file-input.scss
dlnr: Do you use a very old version? Safari 3.2?

packages/css/src/components/file-input/file-input.scss
dlnr: `::file-selector-button` is kind of an oddball, it does not inherit rules from the parent.

packages/css/src/components/file-input/file-input.scss
dlnr: Same rules as the secondary button, I've copied them in case your input is on a different color background.

proprietary/tokens/src/components/ams/file-input.tokens.json
dlnr: Very iffy solution, defining a cursor on the input causes it to flicker, the button will not inherit (not seen as child of input) any CSS.

packages/css/src/components/file-input/file-input.scss
alimpens: No, Safari 15.4, on an iPhone 13

packages/css/src/components/file-input/file-input.scss
alimpens: This is not a very clear comment, but I think you mean: adding a border to it will make it look too much like a text input?

If so: not necessarily. The preliminary Figma design uses a dashed border to indicate the clickable area. We could use that here as well.

packages/css/src/components/file-input/file-input.scss
alimpens: This comment is about the `font-size`, not the `font-family`. `font-size` has `inherit` by default.

packages/css/src/components/file-input/file-input.scss
alimpens: Shouldn't the entire input get a white background color then? Same as our other inputs. It doesn't make sense to do this just for the button imo

packages/css/src/components/file-input/file-input.scss
alimpens: Ha, something weird happens here. You forgot to define `--ams-file-input-file-selector-button-disabled-cursor`, but somehow having an undefined var makes `file-selector-button` inherit this value from the input? Weird. 

Anywhoo, you're right, you have to specifically set the cursor. Can you define the var as well?

packages/css/src/components/file-input/file-input.scss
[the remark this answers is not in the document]
alimpens: Removing the space doesn't look very good though. Is there no way to add space on iOS?

packages/css/src/components/file-input/file-input.scss
dlnr: image
This is the border you're not seeing?

packages/css/src/components/file-input/file-input.scss
dlnr: The question mark is because last week you asked me to remove the border.

packages/css/src/components/file-input/file-input.scss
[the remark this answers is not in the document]
dlnr: A very dodgy way with position absolute and padding...

packages/css/src/components/file-input/file-input.scss
dlnr: I've added a dasheb border and padding, I had to remove the `width: 100%` because the width of the button was being added to input. Turns out the buttons is like a shadow-dom iframe-ish thing: https://css-tricks.com/almanac/selectors/f/file-selector-button/#aa-the-button-is-part-of-the-shadow-dom

packages/css/src/components/file-input/file-input.scss
dlnr: It is to preserve the appearance of the button and does not affect anything in a negative way.

proprietary/tokens/src/components/ams/file-input.tokens.json
dlnr: Resolved

packages/css/src/components/file-input/file-input.scss
dlnr: Var is defined

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,62 @@
...
+  box-shadow: var(--ams-file-input-file-selector-button-box-shadow);
+  color: var(--ams-file-input-file-selector-button-color);
+  cursor: var(--ams-file-input-file-selector-button-cursor);
+  font-family: inherit;
+  font-size: inherit; // iOS specific fix
+  font-weight: inherit;
+  margin-inline-end: var(--ams-file-input-file-selector-button-margin-inline-end);
+  padding-block: var(--ams-file-input-file-selector-button-padding-block);
+  padding-inline: var(--ams-file-input-file-selector-button-padding-inline);
+
```
alimpens, line 49: I think we should add a border here in forced colors mode

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,39 @@
+{
+  "ams": {
+    "file-input": {
+      "background-color": { "value": "{ams.color.primary-white}" },
+      "border": { "value": "{ams.border.width.sm} dashed {ams.color.neutral-grey3}" },
+      "color": { "value": "{ams.color.primary-black}" },
+      "cursor": { "value": "{ams.action.activate.cursor}" },
+      "font-family": { "value": "{ams.text.font-family}" },
+      "font-weight": { "value": "{ams.text.font-weight.normal}" },
+      "font-size": { "value": "{ams.text.level.5.font-size}" },
```
alimpens, line 10:
```suggestion
      "font-size": { "value": "{ams.text.level.5.font-size}" },
      "font-weight": { "value": "{ams.text.font-weight.normal}" },
```

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,39 @@
...
+      "line-height": { "value": "{ams.text.level.5.line-height}" },
+      "outline-offset": { "value": "{ams.focus.outline-offset}" },
+      "padding-block": { "value": "{ams.space.inside.md}" },
+      "padding-inline": { "value": "{ams.space.inside.md}" },
+      "disabled": {
+        "color": { "value": "{ams.color.neutral-grey2}" },
+        "cursor": { "value": "{ams.action.disabled.cursor}" }
+      },
+      "file-selector-button": {
```
alimpens, line 19: Can you order the keys alphabetically here?

storybook/src/components/FileInput/FileInput.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import { FileInput } from '@amsterdam/design-system-react/src'
+import { Meta, StoryObj } from '@storybook/react'
+
+const meta = {
+  title: 'Components/Forms/File Input',
+  component: FileInput,
+  args: {
+    accept: '',
```
alimpens, line 13: Don't we want this to be `undefined` initially? It renders an empty `accept` attribute now.

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,71 @@
...
+  background-color: var(--ams-file-input-background-color);
+  border: var(--ams-file-input-border);
+  color: var(--ams-file-input-color);
+  cursor: var(--ams-file-input-cursor);
+  font-family: var(--ams-file-input-font-family);
+  font-size: var(--ams-file-input-font-size);
+  font-weight: var(--ams-file-input-font-weight);
+  line-height: var(--ams-file-input-line-height);
+  max-inline-size: calc(100% - var(--ams-file-input-padding-inline) * 2);
+  outline-offset: 4px; // Double the default focus outline offset to compensate for the dashed border
```
alimpens, line 32:
```suggestion
  outline-offset: 0.25rem; // Double the default focus outline offset to compensate for the dashed border
```

packages/css/src/components/file-input/file-input.scss
```diff
@@ -0,0 +1,71 @@
...
+
+@import "../../common/text-rendering";
+
+@mixin reset-button {
+  border: 0;
+  border-radius: 0; // Reset rounded borders on iOS devices
+  box-sizing: border-box;
+}
+
+@mixin ams-file-input-selector-button-forced-color-mode {
```
alimpens, line 14: Why is this a mixin?

proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -0,0 +1,42 @@
...
+        "disabled": {
+          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.md} {ams.color.neutral-grey2}" },
+          "color": { "value": "{ams.color.neutral-grey2}" },
+          "cursor": { "value": "{ams.action.disabled.cursor}" }
+        },
+        "forced-color-mode": {
+          "border": { "value": "{ams.border.width.md} solid" }
+        }
+      },
+      "font-family": { "value": "{ams.text.font-family}" },
```
alimpens, line 33: The unnested vars should go before the nested vars

packages/css/src/components/file-input/file-input.scss
dlnr: Because I was told in the button to make it a mixin, just copying previous instructions is guess? I'll make it not a mixin.

alimpens: Maybe wait a second. What was the reason this should be a mixin in the button PR?

dlnr: To prevent duplication probably. I thought about reset where anything reset-ish should go in a reset.

dlnr: Or maybe because we wanted to make a util maybe?

alimpens: Partially implementing a util would be kind of weird, imo. I don't think that's the reason. 

The mixin is used multiple times in Button, it's only used once here. Try to think about the reason code is written the way it is, before copying it. That can easily lead to bugs.

alimpens (approved):

PR #1219, opened 2024-05-07
feat: Allow small text for ordered lists

dlnr:
I've opted to use the same sizes as the paragraph because they can be used together.

### DIFF
--- a/packages/react/src/OrderedList/OrderedList.tsx
+++ b/packages/react/src/OrderedList/OrderedList.tsx
@@ -9,16 +9,17 @@ import type { ForwardedRef, OlHTMLAttributes, PropsWithChildren } from 'react'
 import { OrderedListItem } from './OrderedListItem'
 
 export type OrderedListProps = {
-  markers?: boolean
   /** Changes the text color for readability on a dark background. */
   inverseColor?: boolean
+  /** Whether the list items show a marker. */
+  markers?: boolean
   /** The size of the ordered list */
   size?: 'small'
 } & PropsWithChildren<OlHTMLAttributes<HTMLOListElement>>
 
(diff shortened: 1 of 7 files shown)


proprietary/tokens/src/components/ams/ordered-list.tokens.json
```diff
@@ -3,12 +3,22 @@
...
-      "line-height": { "value": "{ams.text.level.5.line-height}" },
       "list-style-type": { "value": "decimal" },
+      "small": {
+        "font-size": { "value": "{ams.text.level.6.font-size}" },
+        "line-height": { "value": "{ams.text.level.6.line-height}" }
+      },
+      "medium": {
+        "font-size": { "value": "{ams.text.level.5.font-size}" },
+        "line-height": { "value": "{ams.text.level.5.line-height}" }
+      },
```
VincentSmedinga, line 17: Please move these back to the root level. These sizes belong to the default appearance of the element, not to a variation – and `medium` is not an explicit option for the `size` prop. This is consistent with `UnorderedList` and I will add a ticket to change this for `Paragraph` as well.

packages/react/src/OrderedList/OrderedList.tsx
```diff
@@ -12,11 +12,13 @@ export type OrderedListProps = {
   markers?: boolean
   /** Changes the text color for readability on a dark background. */
   inverseColor?: boolean
+  /** The size of the ordered list */
+  size?: 'small' | 'large'
```
VincentSmedinga, line 16: I don’t think we need large text for lists.

VincentSmedinga (approved):

PR #1225, opened 2024-05-14
feat: Decrease compact spacing

dlnr:
The compact tokens now have `inside` and `stack` spacing that are reduced by 25%.

(Works for me)


VincentSmedinga:
Let’s update the spacious space values first. Updating the compact ones can be easily done again.

PR #1228, opened 2024-05-15
feat: Add Field component

alimpens:


packages/css/src/components/field/README.md
```diff
@@ -0,0 +1,9 @@
+<!-- @license CC0-1.0 -->
+
+# Field
+
+Wraps around a single input and can indicate if that input has a validation error.
```
RubenSibon, line 5:
```suggestion
A field wraps around an input and may indicate if that input has a validation error.
```

I'm not sure about adding “A field wraps [...]” before the sentence. I see that we are inconsistent with repeating the component's (generic) name in the summary to have full sentences. Often we do start the description with the component's generic or capitalised name, but not always: i.e. “An Alert informs the user [...]” or “The accordion component is a [...]”.

packages/react/src/Field/Field.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+describe('Field', () => {
+  it('renders', () => {
+    const { container } = render(<Field />)
+    const component = container.querySelector(':only-child')
+
+    expect(component).toBeInTheDocument()
+    expect(component).toBeVisible()
+  })
+
+  it('renders a design system BEM class name', () => {
```
RubenSibon, line 15: Why do we care about the particular CSS framework or strategy? Isn't this sort of an implementation detail? What if we want to use something else than BEM at a later point and we refactor that? Is it not better to use some CSS linter that is aware of BEM?

RubenSibon (approved):

packages/react/src/Field/Field.test.tsx
VincentSmedinga: Class names are the essential API of our HTML components. They are less visible for now, but consider each React component an implementation of the bare HTML that it renders. The class names we choose as well as their syntax inform the markup that the JavaScript Libraries on top of our CSS Library must implement. That’s why React components must test that they render the expected class names.

packages/css/src/components/field/field.scss
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-field {
+  break-inside: avoid;
+}
```
VincentSmedinga, line 8: I think we can safely add this already:

```css
  display: flex;
  flex-direction: column;
  gap: var(--ams-field-gap);
```

Plus a token:

```json
"gap": {
  "value": "ams.space.stack.sm"
},
```

packages/css/src/components/field/field.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.ams-field__has-error {
```
VincentSmedinga, line 10: I think `has-error` is a modifier, not an element, so we need dashes here instead of the underscores.

proprietary/tokens/src/components/ams/field.tokens.json
```diff
@@ -0,0 +1,14 @@
+{
+  "ams": {
+    "field": {
+      "has-error": {
+        "border-inline-start": {
+          "value": "3px solid {ams.color.primary-red}"
```
VincentSmedinga, line 6:
```suggestion
          "value": "{ams.border.width.lg} solid {ams.color.primary-red}"
```

packages/react/src/Field/Field.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export type FieldProps = {
+  /** Whether the field has an input with a validation error */
+  hasError?: boolean
```
VincentSmedinga, line 12: Reusing the word `invalid` here prevents introducing a different term that refers to the same concept.

packages/css/src/components/field/field.scss
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-field {
+  break-inside: avoid;
```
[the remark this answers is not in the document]
alimpens: > do we want fields to be printed on to single pages?

Yeah, this is the reason I added this rule. Without it, the label can be on 1 page, with the input on another, which is a little confusing. 

I'm not sure what this rule does for CSS columns wrapping, but I don't really expect that we'll break a form field up into multiple columns. [Multi column form layouts aren't recommended](https://baymard.com/blog/avoid-multi-column-forms)

VincentSmedinga (approved):

PR #1230, opened 2024-05-15
feat: Display component's status badge in story

RubenSibon:
Add status badges to stories to inform the reader about the lifecycle status of the component.

This PR uses the [Status addon](https://storybook.js.org/addons/@etchteam/storybook-addon-status/).

GitHub: [https://github.com/etchteam/storybook-addon-status?tab=readme-ov-file](https://github.com/etchteam/storybook-addon-status?tab=readme-ov-file)

Known issues:

- Currently doesn't show a "traffic light" in the sidebar; this is [a known bug](https://github.com/etchteam/storybook-addon-status/issues/28). Once it's fixed it might be the case that there is [no option to turn it off](https://github.com/etchteam/storybook-addon-status/issues/28#issuecomment-1566135242).
- Does not seem to work in our Docs/Assets (i.e. Icon Gallery). Does anyone know how to fix this? If so, please make suggestions.
- Does it support Storybook 7 and 8 properly? See: https://github.com/etchteam/storybook-addon-status/issues

### DIFF
--- a/storybook/src/docs/icon-gallery.docs.mdx
+++ b/storybook/src/docs/icon-gallery.docs.mdx
@@ -1,11 +1,12 @@
-import { Meta } from "@storybook/addon-docs/blocks";
+import { Meta } from "@storybook/blocks";
 import { AmsterdamIconGallery } from "./components/AmsterdamIconGallery";
+import { Badge } from "../../../packages/react/src/Badge";
 
-{/* Below "parameters" prop does not seem to work; see docs: https://github.com/etchteam/storybook-addon-status?tab=readme-ov-file#story-usage */}
-
-<Meta title="Docs/Assets/Icons" parameters={{ status: { type: "beta" } }} />
+<Meta title="Docs/Assets/Icons" />
 
 # Icons
+
+<Badge label="Beta" color="orange" style={{ position: "absolute", top: "2rem" }} />
 
 An overview of all currently available icons.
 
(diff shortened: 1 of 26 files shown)


storybook/src/docs/icon-gallery.docs.mdx
```diff
@@ -1,7 +1,9 @@
-import { Meta } from "@storybook/blocks";
+import { Meta } from "@storybook/addon-docs/blocks";
 import { AmsterdamIconGallery } from "./components/AmsterdamIconGallery";
 
-<Meta title="Docs/Assets/Icons" />
+{/* Below "parameters" prop does not seem to work; see docs: https://github.com/etchteam/storybook-addon-status?tab=readme-ov-file#story-usage */}
```
alimpens, line 4: If this plugin doesn't work here, and we don't want to show the statusses in the TOC anyway, isn't it easier to just make a custom element and place it in the MDX? The currect badges aren't very easy to see imo

RubenSibon: I updated it to show how it looks with our own Badge component to show the status, but as you can see we then need to deal with the margin above the heading.

The [Badge addon has way more style properties](https://github.com/geometricpanda/storybook-addon-badges?tab=readme-ov-file#preconfigured-badges) which is nice, but I also could not get it to work with the docs. Maybe there is a solution, but I can't figure one out.

RubenSibon: I'm experimenting with our Badge component again. It's now added below the story's heading and absolutely positioned above it. I did that for the Icon Gallery and the Header. What do you think?

Also @VincentSmedinga

VincentSmedinga: Yes, I think it’s fair to say that the conclusion of this experiment is that we can best implement our own status badge. We want it to appear a bit differently than what they provide anyway, and the quality isn’t up to par. Thanks, still :)

PR #1232, opened 2024-05-15
feat!: Rename Breadcrumb Item to Link

VincentSmedinga:
A breadcrumb is a list of links. The ‘item’ level in between is implicit; the link is what matters. This change reflects that. Now the `href` prop is on a link, not an item.

Changes:

- The `Breadcrumb.Item` component has been renamed to `Breadcrumb.Link`
- Rest props and class names get applied to the `a` element instead of the `li`
- The `-item` fragment of the design tokens has been removed

### DIFF
--- a/packages/react/src/Breadcrumb/BreadcrumbLink.tsx
+++ b/packages/react/src/Breadcrumb/BreadcrumbLink.tsx
@@ -5,16 +5,14 @@
 
 import { clsx } from 'clsx'
 import { forwardRef } from 'react'
-import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import type { AnchorHTMLAttributes, ForwardedRef } from 'react'
 
-export type BreadcrumbLinkProps = {
-  href: string
-} & PropsWithChildren<HTMLAttributes<HTMLLIElement>>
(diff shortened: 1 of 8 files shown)


dlnr (dismissed):

RubenSibon (approved):

packages/react/src/Breadcrumb/BreadcrumbLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import { clsx } from 'clsx'
+import { forwardRef } from 'react'
+import type { AnchorHTMLAttributes, ForwardedRef } from 'react'
+
+export type BreadcrumbLinkProps = AnchorHTMLAttributes<HTMLAnchorElement>
+
+export const BreadcrumbLink = forwardRef(
+  ({ children, className, href, ...restProps }: BreadcrumbLinkProps, ref: ForwardedRef<HTMLAnchorElement>) => (
```
alimpens, line 13: We don't need to define `href` here, it's already passed by `restProps`

VincentSmedinga: Done.

PR #1235, opened 2024-05-21
chore: Replace remaining pixel lengths with rems

VincentSmedinga:


packages/css/src/components/card/card.scss
```diff
@@ -14,7 +14,7 @@
   &:has(:focus-visible) {
     outline-color: -webkit-focus-ring-color;
     outline-style: auto;
-    outline-width: 1px;
```
RubenSibon, line 17: That `0.0625rem` leads to a 1px width with a default 16px font-size is not immediately obvious, while `1px` is. So, I wonder if we could use functions that calculate the rem value from a given pixel value?

RubenSibon: Why do we use `rem` and not `em` for margins and paddings?

VincentSmedinga: > That 0.0625rem leads to a 1px width with a default 16px font-size is not immediately obvious, while 1px is. So, I wonder if we could use functions that calculate the rem value from a given pixel value?

I don’t think it’s very important to know to how many pixels something comes down to. Also, 1rem is not always equal to 16px. We want to express lengths as a fraction of the user’s font size. I agree that `0.5rem` is clearer than `0.0625rem`, but give it time and you’ll know them all by heart like me :)

Another reason not to: most values live in tokens, and we don’t want Sass function inside of them, and we can’t wrap them in CSS or we must check whether a value is expressed in pixels or rems – or cqws etc.

VincentSmedinga: > Why do we use rem and not em for margins and padding?

Because margins and paddings are defined on containers and font-size is not, so ems and rems are often the same there, and ems are less predictable.

packages/css/src/components/switch/switch.scss
```diff
@@ -51,10 +53,11 @@
...
-  // The 2px is to account for the 1px padding-inline on the label
-  transform: translate(calc(var(--ams-switch-width) - var(--ams-switch-thumb-width) - 2px));
+  transform: translate(
+    calc(var(--ams-switch-width) - var(--ams-switch-thumb-width) - 2 * var(--ams-switch-track-padding))
+  );
 }
 
 .ams-switch:hover .ams-switch__input:enabled + .ams-switch__label::before {
-  box-shadow: 0 0 0 2px var(--ams-switch-thumb-hover-color);
+  box-shadow: 0 0 0 var(--ams-switch-thumb-hover-box-shadow-spread-radius) var(--ams-switch-thumb-hover-color);
```
alimpens, line 62: Do we want to turn this into a single token? We can use `ams-switch-thumb-hover-color` there as well.

proprietary/tokens/src/common/ams/focus.tokens.json
```diff
@@ -1,7 +1,7 @@
 {
   "ams": {
     "focus": {
-      "outline-offset": { "value": "2px" }
+      "outline-offset": { "value": "{ams.border.width.md}" }
```
alimpens, line 4: It's not really a border width though, it's an offset. Both just happen to be 2px

VincentSmedinga: Oh wait, it’s not the outline itself. Changed to `0.125rem`.

packages/css/src/components/switch/switch.scss
VincentSmedinga: I don’t understand, please elaborate.

packages/css/documentation/coding-conventions.md
```diff
@@ -13,8 +13,8 @@ Still, we define its thickness and offset for the initial state.
 ```css
 .ams-link {
   text-decoration: none;
-  text-decoration-thickness: 2px;
-  text-underline-offset: 3px;
```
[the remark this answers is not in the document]
VincentSmedinga: I’ve updated the `text-underline-offset` to use values in `em`.

packages/css/src/components/switch/switch.scss
alimpens: We could set something like `box-shadow: var(--ams-switch-thumb-hover-box-shadow);` here.

The token could be `0 0 0 0.125rem {ams.switch.thumb.hover.color}`.

That results in somewhat less concatenation.

(Pseudo code)

VincentSmedinga: Ah yes, much better.

alimpens (approved):

RubenSibon (approved):

PR #1236, opened 2024-05-23
feat!: Allow additional background colours for Badge and remove dark blue option

dlnr:
Processed all badge colors from the Figma file, and remove dark-blue which is only used for the hover color of primary blue. This is why we need a semantic color layer.


RubenSibon (dismissed):

VincentSmedinga (approved):
We’ll be deciding on definitive colour names soon, so the options will probably change once more, but this is a good improvement until then. Let’s make Avatar consistent at that later moment.

PR #1237, opened 2024-05-24
feat!: Add invalid prop to Field Set and update Field and Field Set docs

alimpens:


packages/react/src/Fieldset/Fieldset.tsx
VincentSmedinga, line 1: Separating breaking changes from non-breaking ones would help our implementers find the exact changes they need to make when upgrading. Is it worthwhile to extract a separate PR for the `FieldSet` rename? Is everything else non-breaking?

storybook/src/components/Field/Field.docs.mdx
```diff
@@ -10,6 +10,14 @@ import README from "../../../../packages/css/src/components/field/README.md?raw"
 
 <Controls />
 
+## With Description
+
+A Field can have a description.
+Make sure to connect this description to the input in the Field,
+otherwise this won’t be read by a screen reader.
+
```
VincentSmedinga, line 18: Let’s be explicit in the markup people need to add, rather than letting them find it in the Code View.

```suggestion
Add an `aria-describedby` attribute to the Field
and provide the `id` of the describing element as its value.

```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -0,0 +1,49 @@
...
+<Controls />
+
+## Examples
+
+## With Description
+
+A Field Set can have a description.
+Make sure to connect this description to the Field Set or a specific input,
+otherwise this won’t be read by a screen reader.
+
```
VincentSmedinga, line 20:
```suggestion
Add an `aria-describedby` attribute to the Field Set
and provide the `id` of the describing element as its value.

```

storybook/src/components/Field/Field.docs.mdx
VincentSmedinga: I see you’ve already did so with the examples for radio and checkbox.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -0,0 +1,49 @@
...
+Always also set `aria-required` on the individual radio buttons though, to make sure it’s read by screen readers.
+
+<Canvas of={FieldSetStories.RadioGroup} />
+
+### Checkbox group
+
+Use a Field Set to group checkboxes.
+
+Please note: [NVDA has bug](https://github.com/nvaccess/nvda/issues/12718) which causes it to
+not report a description connected to a Field Set when the Field Set contains checkboxes.
```
VincentSmedinga, line 45:
```suggestion
not report a description connected to a Field Set when it contains checkboxes.
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+A Field Set can indicate whether any of the inputs it contains has a validation error.
+
+<Canvas of={FieldSetStories.WithError} />
+
+### Radio group
+
+Use a Field Set to group radio buttons.
+When grouping radio inputs, use `role="radiogroup"` on Field Set to have this grouping explicitly announced as a radio group.
+Field Set has a default role of `group`.
```
VincentSmedinga, line 33:
```suggestion
When grouping radio inputs, use `role="radiogroup"` on Field Set to have this grouping explicitly announced as a radio group (the default role is `group`).
```

packages/react/src/Fieldset/Fieldset.tsx
alimpens: Yeah, that's possible, and it would be clearer. However, doing case only renames is kind of horrible, especially on Windows machines. So if we can avoid it, that'd make me happy 😅

storybook/src/components/Field/Field.docs.mdx
alimpens: Your suggestion isn't quite correct, at least it isn't the current implementation.

```suggestion
Add an `aria-describedby` attribute to the input and provide the `id` of the describing element as its value.

```

packages/react/src/Fieldset/Fieldset.tsx
VincentSmedinga: Next time then :)

storybook/src/components/Field/Field.docs.mdx
VincentSmedinga: Sorry, badly copied and pasted from Field Set.

VincentSmedinga (approved):

PR #1238, opened 2024-05-24
feat: Set dir auto by default for text input form fields

alimpens:


packages/react/src/SearchField/SearchFieldInput.tsx
```diff
@@ -26,6 +26,7 @@ export const SearchFieldInput = forwardRef(
           {...restProps}
           autoComplete="off"
           className={clsx('ams-search-field__input', className)}
+          dir={dir ?? 'auto'}
```
VincentSmedinga, line 29: > `auto` uses a basic algorithm as it parses the characters inside the element until it finds a character with a strong directionality, then applies that directionality to the whole element ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir))

How about we default to `ltr` to prevent that computation from happening unnecessarily?

alimpens: We want that computation to happen though :D If you type RTL text in a text input, you should switch to RTL.

You can try it yourself: If you change your system language to Arabic for example, and type in the SearchFieldInput, it starts from the right. Without `dir=auto`, it doesn't.

RubenSibon (approved):

PR #1240, opened 2024-05-24
feat!: Use invalid prop for most inputs

alimpens:
Questions:

- Does omitting attributes from types constitute a breaking change?
- Should Search Field and Switch get an `invalid` prop? Are there situations in which these can be invalid?


packages/react/src/Checkbox/Checkbox.tsx
```diff
@@ -12,7 +12,7 @@ export type CheckboxProps = {
   invalid?: boolean
   /** Allows being neither checked nor unchecked. */
   indeterminate?: boolean
-} & PropsWithChildren<InputHTMLAttributes<HTMLInputElement>>
+} & PropsWithChildren<Omit<InputHTMLAttributes<HTMLInputElement>, 'aria-invalid' | 'type'>>
```
VincentSmedinga, line 15: > Does omitting attributes from types constitute a breaking change?

I would say so, yes. It is a change in the component’s API, even if they are HTML attributes. We no longer honour any values implementers set and even flag them as type errors, so they must now remove these attributes if they were using them.

VincentSmedinga:
> * Does omitting attributes from types constitute a breaking change?

Copied this to a change, preventing quoting high-level comments.

> * Should Search Field and Switch get an `invalid` prop? Are there situations in which these can be invalid?

Well, there are theoretical situations. Both implement the `HTMLInputElement` interface which has properties like `required` and `minLength` that may fail a validation rule.

We offer Switch to toggle UI states only, so it can never become invalid in practice. We can remove all validation fields from the type, or not even use the `input` interface at all.

A Search Field might have a minimum length of e.g. 3 characters to facilitate meaningful results, although I would not expect it to turn red if the field gets blurred with a content of less than 3 characters. On the other hand, if people press Enter or click the Search Button, there should be a way to inform them of this rule.

So maybe yes for Search Field, and tighten the props for Switch?

packages/css/src/components/select/select.scss
```diff
@@ -37,12 +37,9 @@
   @include reset;
 }
 
+.ams-select:invalid,
 .ams-select[aria-invalid="true"] {
```
VincentSmedinga, line 41: Please add some (initial) documentaion on this API choice and the way people should implement our form fields in this respect.

I would say we can add `storybook/src/docs/forms.docs.mdx` and include it in the ‘Design Guidelines’ group. We’ll tidy up this section soon.

packages/react/src/DateInput/DateInput.test.tsx
```diff
@@ -38,4 +38,31 @@ describe('Date input', () => {
 
     expect(ref.current).toBe(component)
   })
+
+  // invalid state
```
RubenSibon, line 42: Is such a comment necessary? The `it` test cases should be self-documenting.

packages/react/src/DateInput/DateInput.tsx
```diff
@@ -7,11 +7,20 @@ import clsx from 'clsx'
...
+  invalid?: boolean
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'aria-invalid' | 'type'>
 
 export const DateInput = forwardRef(
-  ({ className, ...restProps }: DateInputProps, ref: ForwardedRef<HTMLInputElement>) => (
-    <input {...restProps} ref={ref} className={clsx('ams-date-input', className)} type="date" />
+  ({ className, invalid, ...restProps }: DateInputProps, ref: ForwardedRef<HTMLInputElement>) => (
+    <input
+      {...restProps}
+      aria-invalid={invalid || undefined}
```
RubenSibon, line 19: So the value may only be `true` or `undefined`? `false` is not an option? Then maybe give the optional `invalid?` prop above the type `true` instead of `boolean`? You either set it to `true` or you don't set it at all?

storybook/src/components/Field/Field.stories.tsx
```diff
@@ -23,7 +23,7 @@ export const Default: Story = {
   render: (args) => (
     <Field invalid={args.invalid}>
       <Label htmlFor="input1">Waar gaat het om?</Label>
-      <TextInput id="input1" aria-invalid={args.invalid ? true : undefined} />
+      <TextInput id="input1" invalid={args.invalid} />
```
RubenSibon, line 26: So, again, it's either `true` or `undefined`. Then I'd say replace the `boolean` type with `true` everywhere and always make it optional.

packages/react/src/DateInput/DateInput.test.tsx
alimpens: The reason for the comment is to imply grouping, i.e. the next n tests are about the invalid state.

It makes more sense to use nested `describe`s for that though, I'll change it

packages/react/src/DateInput/DateInput.tsx
alimpens: No, `invalid` can be `true`, `false` or `undefined`. `aria-invalid` can only be `true` or `undefined`. 

Working with a boolean prop is a bit easier than working with an HTML attribute that can only be `true` or `undefined`, which is why the `invalid` prop exists.

packages/css/src/components/select/select.scss
alimpens: Done, I've added it to `packages/css/documentation/coding-conventions.md`

VincentSmedinga (approved):

PR #1241, opened 2024-05-29
fix: Fix issues with inputs on iOS

alimpens:


packages/css/src/components/select/select.scss
```diff
@@ -4,7 +4,7 @@
  */
 
 @mixin reset {
-  appearance: none;
+  appearance: none; // Reset native appearance to hide default chevron
```
RubenSibon, line 7: Why not a webkit prefix here?

RubenSibon: I guess because it's not an input. Yeah.

RubenSibon (approved):
LGTM, I assume this has been thoroughly tested on all supported devices and browsers.

alimpens:
> LGTM, I assume this has been thoroughly tested on all supported devices and browsers.

Thanks! I tested it on Chrome, Firefox and Edge on Windows, Chrome on Android and Safari 15.2, 15.6 and 16.6 on iOS.

packages/css/src/components/select/select.scss
alimpens: `appearance: none` is used here to hide the default chevron. That isn't necessary for Safari < 15.4 (which are the browsers for which we sometimes need the prefix)

VincentSmedinga (approved):

PR #1242, opened 2024-05-31
feat(internal): Indicate beta status for some components, patterns, and assets

RubenSibon:
Lorem ipsum texts have to be replaced with actual beta reasons or removed. Can you help me with that?

### DIFF
--- a/storybook/src/docs/components/StatusBadge.tsx
+++ b/storybook/src/docs/components/StatusBadge.tsx
@@ -1,11 +1,17 @@
 import './status-badge.css'
 import { Badge } from '@amsterdam/design-system-react/src'
 
-export const StatusBadge = ({ status = 'beta', statusReason = '' }: { status: 'beta'; statusReason: string }) => {
-  return (
-    <span className="ams-storybook-status-badge">
-      <Badge label={status} color="orange" />
-      {statusReason && <span>{statusReason}</span>}
-    </span>
-  )
+type StatusBadgeProps = {
+  /** The status of this component. */
+  label?: string
+  /* The explanation of what needs to change before transitioning into another status. */
+  reason: string
 }
+
+/** Indicates the status of a component. Use this to prepare implementers for API changes. */
+export const StatusBadge = ({ label = 'beta', reason }: StatusBadgeProps) => (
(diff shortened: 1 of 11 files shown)


storybook/src/docs/components/StatusBadge.tsx
```diff
@@ -0,0 +1,19 @@
+import './status-badge.css'
+import { Badge } from '@amsterdam/design-system-react/src'
+
+export const StatusBadge = ({
+  status = 'beta',
+  statusReason = '',
+  documentTop = true,
```
alimpens, line 7: This prop is never used, do we need it?

RubenSibon: My reasoning was to allow us to, occasionally, use the badge somewhere else than on the top of document.

Because it's `inline-flex` it could be used inline somewhere as well.

Vincent actually prefers to always have the component name on top

RubenSibon: Prop has been removed.

storybook/src/docs/components/StatusBadge.tsx
```diff
@@ -0,0 +1,11 @@
+import './status-badge.css'
+import { Badge } from '@amsterdam/design-system-react/src'
+
+export const StatusBadge = ({ status = 'beta', statusReason = '' }: { status: 'beta'; statusReason: string }) => {
```
alimpens, line 4: If the `status` can only be the text 'beta', why not just hardcode it?

RubenSibon: Maybe we want to change it in the future? Vincent had suggested another option.

Of course I could refactor this to the BetaBadge. :-)

alimpens: I would generally avoid working ahead. If we never implement more than 1 status, this partial implementation will keep floating around.

I'd just hardcode the 'beta' string for now, and add more statusses when we need them. I wouldn't rename the component though. It is still a status badge, and it will keep on being that in the future.

VincentSmedinga (approved):

PR #1244, opened 2024-05-31
feat!: Change Accordion `section` boolean prop to `sectionAs` enum

alimpens:


RubenSibon (approved):
LGTM

packages/react/src/Accordion/Accordion.tsx
```diff
@@ -28,7 +28,7 @@ const AccordionRoot = forwardRef(
     const { keyDown } = useKeyboardFocus(innerRef, { rotating: true })
 
     return (
-      <AccordionContext.Provider value={{ headingLevel: headingLevel, section: section }}>
+      <AccordionContext.Provider value={{ headingLevel: headingLevel, as: sectionAs }}>
```
VincentSmedinga, line 31: I suggest not renaming at this hierarchical level and only using `as` within `Section`.

packages/react/src/Accordion/AccordionContext.tsx
```diff
@@ -8,12 +8,12 @@ import { HeadingLevel } from '../Heading/Heading'
 
 export type AccordionContextValue = {
   headingLevel: HeadingLevel
-  section?: boolean
+  as?: 'div' | 'section'
```
VincentSmedinga, line 11: Can you reorder these and those for the default values?

packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -46,23 +47,13 @@ export const AccordionSection = forwardRef(
...
-            aria-labelledby={buttonId}
-            className={clsx('ams-accordion__panel', { 'ams-accordion__panel--expanded': isExpanded })}
-          >
-            {children}
-          </div>
-        )}
+        <Tag
+          id={panelId}
+          aria-labelledby={buttonId}
+          className={clsx('ams-accordion__panel', { 'ams-accordion__panel--expanded': isExpanded })}
```
VincentSmedinga, line 53: These can be sorted as well.

```suggestion
          aria-labelledby={buttonId}
          className={clsx('ams-accordion__panel', { 'ams-accordion__panel--expanded': isExpanded })}
          id={panelId}
```

packages/react/src/Accordion/AccordionContext.tsx
alimpens: I renamed `as` to `sectionAs`, so it's in the right order again.

VincentSmedinga (approved):

PR #1246, opened 2024-05-31
chore: Correct token nesting for inverse color

VincentSmedinga:
This doesn’t affect the actual token names, but having `inverse-color` as a level violates the rule that the name of the leaf must be a CSS property.

### DIFF
--- a/proprietary/tokens/src/components/ams/blockquote.tokens.json
+++ b/proprietary/tokens/src/components/ams/blockquote.tokens.json
@@ -6,7 +6,11 @@
       "font-size": { "value": "{ams.text.level.3.font-size}" },
       "font-weight": { "value": "{ams.text.font-weight.bold}" },
       "line-height": { "value": "{ams.text.level.3.line-height}" },
-      "inverse-color": { "value": "{ams.color.primary-white}" }
+      "inverse": {
+        "color": {
+          "value": "{ams.color.primary-white}"
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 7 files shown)


RubenSibon (approved):
LGTM

proprietary/tokens/src/components/ams/blockquote.tokens.json
```diff
@@ -6,7 +6,11 @@
       "font-size": { "value": "{ams.text.level.3.font-size}" },
       "font-weight": { "value": "{ams.text.font-weight.bold}" },
       "line-height": { "value": "{ams.text.level.3.line-height}" },
-      "inverse-color": { "value": "{ams.color.primary-white}" }
+      "inverse": {
+        "color": {
```
alimpens, line 10: This means the modifier name is now 'inverse', right, not 'inverse-color'?

So it should be `ams-paragraph--inverse`, for example?

Or should the token be `ams.paragraph.inverse-color.color`?

alimpens, line 10: Changing the token structure is technically a breaking change, I think

VincentSmedinga:
Closing – doing this right requires changing CSS classes and React props as well, and we’re planning to revisit the entire API around colours soon anyway.

proprietary/tokens/src/components/ams/blockquote.tokens.json
VincentSmedinga: Good point. I’ve mentioned this change in [a broader ticket](https://gemeente-amsterdam.atlassian.net/jira/software/c/projects/DES/boards/1378/backlog?epics=visible&selectedIssue=DES-686&text=inverse) on this API. We can leave this as is until then.

PR #1247, opened 2024-05-31
feat: Add Error Message

alimpens:
I left out `Label` and `ErrorMessage` as child components of `Field` and `FieldSet` for now, for two reasons:

- `Label` doesn't really makes sense as a child of `FieldSet`. It belongs to the input, not the FieldSet. `ErrorMessage` sometimes belongs to the `FieldSet` (`RadioGroupWithError`), sometimes it belongs to the input (`WithError`)
- The displaynames seem to be set globally, so you can either have `Label`, `Field.Label` or `FieldSet.Label` as the displayname, not the first in 1 case and the second in another

I left out `CheckboxGroupWithError` for now, because I haven't yet found a way to get that working with NVDA. I'll write a new ticket to fix that.


packages/css/src/components/error-message/README.md
```diff
@@ -0,0 +1,11 @@
+<!-- @license CC0-1.0 -->
+
+# Error Message
+
+Show an error message when there is a form field validation error.
+In the error message explain what went wrong and how to fix it.
```
RubenSibon, line 6:
```suggestion
Explain in the error message what went wrong and how to fix it.
```

packages/css/src/components/error-message/error-message.scss
```diff
@@ -0,0 +1,22 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@import "../../common/text-rendering";
+
+@mixin reset {
+  box-sizing: border-box;
+  margin-block: 0;
+}
```
RubenSibon, line 11: For my understanding: we add a new reset for every component. Despite there being variations, there's also a lot of duplication. Why are we not handling some of these common resets globally? With different types of resets (i.e. block and inline resets, etc...)?

storybook/src/components/ErrorMessage/ErrorMessage.docs.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
+
+## With a custom prefix
+
+Error messages are automatically prefixed with a visually hidden text, the Dutch phrase "Invoerfout: ".
```
RubenSibon, line 15: Why is visually hidden? This information could be relevant for anyone. Colour blind people might not use a screen reader, but having more explicit information that an there is an issue with their input could be helpful.

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -128,3 +178,29 @@ export const CheckboxGroup: Story = {
...
+//         <Checkbox name="about" value="evenement" invalid={args.invalid} aria-required="true">
+//           Evenement
+//         </Checkbox>
+//         <Checkbox name="about" value="anders" invalid={args.invalid} aria-required="true">
+//           Iets anders
+//         </Checkbox>
+//       </Column>
+//     </FieldSet>
+//   ),
+// }
```
RubenSibon, line 206: Why is there commented-out code here? WIP?

packages/css/src/components/error-message/README.md
alimpens: Do you have a strong reason to want to change the order here? [I copied this sentence from gov.uk](https://design-system.service.gov.uk/components/error-message/), I'm kind of leaning towards the native speakers.

packages/css/src/components/error-message/error-message.scss
alimpens: Good point, we should look at extracting (parts of) these resets. It's a bit of a chore though, there are a lot of variations and it's easy to turn this into something harder to follow than it is now. [I've added a ticket for this](https://gemeente-amsterdam.atlassian.net/browse/DES-791)

storybook/src/components/ErrorMessage/ErrorMessage.docs.mdx
alimpens: True. This is usually combined with an icon for visual users, but we haven't made a decision about that yet

storybook/src/components/FieldSet/FieldSet.stories.tsx
alimpens: I tried to explain that [here](https://github.com/Amsterdam/design-system/pull/1247/files#diff-69cbe174562903ae5ab20b810f196ff8bd28d15cbf97a08f9a91781c6d5db1d8R63). There is a bug with NVDA and fieldsets that use `aria-described`. The commented out story works for other screen readers, but we have to find a workaround for the NVDA bug before we can show this story (imo).

packages/css/src/components/error-message/README.md
RubenSibon: No strong reason, just that the two sentences repeated a bit too much.

packages/react/src/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,27 @@
...
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { VisuallyHidden } from '../VisuallyHidden'
+
+export type ErrorMessageProps = {
+  prefix?: string
+} & PropsWithChildren<HTMLAttributes<HTMLParagraphElement>>
+
+export const ErrorMessage = forwardRef(
+  (
+    { children, className, prefix = 'Invoerfout: ', ...restProps }: ErrorMessageProps,
```
VincentSmedinga, line 17: Should we render the colon ourselves to ensure the screen reader pauses between the prefix and the error message? (Assuming that it does.)

packages/react/src/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { VisuallyHidden } from '../VisuallyHidden'
+
+export type ErrorMessageProps = {
+  prefix?: string
```
VincentSmedinga, line 12:
```suggestion
  /** An accessible phrase that screen readers announce before the error message. Should translate to something like ‘input error’. */
  prefix?: string
```

packages/react/src/ErrorMessage/ErrorMessage.tsx
alimpens: Yeah, it adds a pause. Good idea to render it ourselves!

packages/react/src/ErrorMessage/ErrorMessage.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const ErrorMessage = forwardRef(
+  (
+    { children, className, prefix = 'Invoerfout', ...restProps }: ErrorMessageProps,
+    ref: ForwardedRef<HTMLParagraphElement>,
+  ) => (
+    <p {...restProps} ref={ref} className={clsx('ams-error-message', className)}>
+      <VisuallyHidden>
+        {prefix}
+        {': '}
+      </VisuallyHidden>
```
VincentSmedinga, line 25: Which one do you prefer?

```suggestion
      <VisuallyHidden>{prefix + ': '}</VisuallyHidden>
```

alimpens: I'm not a fan of plus signs for string concatenation :D But it could be something like `${prefix}: `?

VincentSmedinga: That looks even more complicated haha – we’ll leave it on two lines then.

VincentSmedinga (approved):

PR #1252, opened 2024-06-05
feat: Add Form Error List

alimpens:


packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Alert
+        {...restProps}
+        className={clsx('ams-form-error-list', className)}
+        headingLevel={headingLevel}
+        ref={ref}
+        role="alert"
+        severity="error"
+        title={heading}
+      >
+        <LinkList>
```
dlnr, line 42: This could be just me but I associate the chevrons with links that take you away from the page (like in the footer and related links), not anchors on the page. I would've gone with the `UnorderedList` component. And the lack of padding does not do it justice, it's way too close to the title. It needs the inside spacing here.

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes } from 'react'
+import { Alert } from '../Alert'
+import { HeadingLevel } from '../Heading/Heading'
+import { LinkList } from '../LinkList'
+
+export type FormErrorListProps = {
+  errors: { id: string; label: string }[]
```
dlnr, line 14: This is very basic. Do we know how developers want to pass errors to a component like this?

packages/react/src/FormErrorList/FormErrorList.tsx
dlnr, line 1: This seems more a pattern for the usage of the `Alert` component, if you describe it like that developers are free to handle the error any way they want. The can have multiple arrays or add a paragraph of extra explanation. I would not have made another component. And beyond that if you combine this with an error message with individual formfields you can quickly get a lot of extra vertical space.

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -0,0 +1,54 @@
...
+  heading?: string
+  headingLevel?: HeadingLevel
+} & HTMLAttributes<HTMLDivElement>
+
+export const FormErrorList = forwardRef(
+  (
+    {
+      className,
+      errors,
+      heading = 'Verbeter de fouten voor u verder gaat',
```
dlnr, line 24: The heading is optional but you're giving it a default (dutch) value.

packages/css/src/components/form-error-list/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Where to put the Form Error List
+
+Put the Form Error List at the top of the main container, outside of the `<form>`-tag. If your page includes breadcrumbs or a back link, place it below these, but above the `<h1>`.
+
+## Relevant WCAG requirements
+
+Pay extra attention to these parts:
+
+- [WCAG requirement 1.3.1](https://www.w3.org/TR/WCAG21/#info-and-relationships): the heading level of the accordion sections depends on where in the page the accordion is placed, this may differ per page.
```
dlnr, line 30: Accordion?

packages/css/src/components/form-error-list/README.md
```diff
@@ -0,0 +1,35 @@
...
+For questions that require a user to answer using a single field, like a file upload, select, textarea, text input or character count, link to the `id` of that field.
+
+When a user has to enter their answer into multiple fields, such as day, month and year fields, link to the `id` of the first field that contains an error.
+If you do not know which field contains an error, link to the `id` of the first field.
+
+For questions that require a user to select one or more options from a list using Radios or Checkboxes, link to the `id` of the first Radio or Checkbox.
+
+## Where to put the Form Error List
+
+Put the Form Error List at the top of the main container, outside of the `<form>`-tag. If your page includes breadcrumbs or a back link, place it below these, but above the `<h1>`.
```
dlnr, line 24: Can't you say that it needs to go above the `<form>`?

packages/css/src/components/form-error-list/README.md
```diff
@@ -0,0 +1,35 @@
+<!-- @license CC0-1.0 -->
+
+# Form Error List
+
+Use this component at the top of a page to summarise any errors a user has made.
+When a user makes an error, you must show both a Form Error List and an Error Message above each answer that contains an error.
```
dlnr, line 6: You "must"? What about forms that have like 3 fields?

packages/css/src/components/form-error-list/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Guidelines
+
+- Always show a Form Error List when there is a validation error, even if there’s only one.
```
dlnr, line 10: You've answered my previous comment (lol). But I disagree. Imagine you have an email field for a newsletter in a footer and you make a typo like miss the `@`, showing this component and a formfield error would seem like overkill and possible destroy the footer layout.

alimpens: That's a pretty good reason to not have an e-mail field in the footer imo 😁 I think we removed that from the design as well.

This validation pattern has been pretty thoroughly tested, by Adam Silver / Gov.uk, NLDS, the Danish Design system etc. Simply showing an error message above a field is not obvious enough, so we should avoid it.

packages/css/src/components/form-error-list/README.md
alimpens: That doesn't necessarily cover it, depending on whether the `h1` is inside or outside of the form.

packages/css/src/components/form-error-list/README.md
alimpens: Woops, copy-paste error

packages/react/src/FormErrorList/FormErrorList.tsx
alimpens: > This seems more a pattern for the usage of the Alert component

Yeah, we discussed that last Friday. It does add some default attrs and logic though, such as the `role=alert` and looping through the list of errors, so we felt there is enough there to warrant a separate component. 

> And beyond that if you combine this with an error message with individual formfields you can quickly get a lot of extra vertical space.

That depends. The guideline will probably be to show [one thing per page](https://www.smashingmagazine.com/2017/05/better-form-design-one-thing-per-page/), so the amount of possible errors on 1 page will not be that large.

packages/react/src/FormErrorList/FormErrorList.tsx
alimpens: This component should only render a list of links, so it shouldn't be more complicated than this, imo. Or do you expect that more functionality is required?

packages/react/src/FormErrorList/FormErrorList.tsx
alimpens: I don't understand the comment? The component always has a heading, but you can optionally overwrite it.

packages/react/src/FormErrorList/FormErrorList.tsx
alimpens: Yeah, this is just the basic structure using existing components. [I've added a ticket for design to look at it.](https://gemeente-amsterdam.atlassian.net/browse/DES-792)

packages/react/src/FormErrorList/FormErrorList.tsx
dlnr: This sounds like an assumption but ok let's see if this thing will cause issues after release.

packages/react/src/FormErrorList/FormErrorList.tsx
dlnr: There might be, for example there could be an explanation in the form of a paragraph besides a link.

packages/react/src/FormErrorList/FormErrorList.tsx
dlnr: I mean the heading is optional for the alert but you're forcing it here.

dlnr (approved):

VincentSmedinga (dismissed):

VincentSmedinga (approved):

PR #1254, opened 2024-06-05
feat: Change values for tall, wide, and extra wide aspect ratios

VincentSmedinga:
- The `2x-wide` proportion has changed from 32:9 to 16:5
- The `wide` proportion has changed from 5:4 to 4:3
- The `tall` proportion has changed from 4:5 to 3:4


RubenSibon (approved):
LGTM; I checked the Card and Overlap components out in the deployment.

PR #1255, opened 2024-06-07
chore: Add license to Storybook Docs files

VincentSmedinga:
We hadn’t done this yet, but now we found the correct syntax.


dlnr (approved):
Only your pr title doesn't seem to be conventional commitish enough, but otherwise fine by me.

PR #1258, opened 2024-06-12
feat: Allow a specific Tab to be made active

RubenSibon:


packages/react/src/Tabs/Tabs.tsx
```diff
@@ -12,33 +12,38 @@ import { TabsList } from './TabsList'
...
-})
+export type TabsProps = {
+  /** The initial active tab. */
+  activeTab?: number
+} & PropsWithChildren<HTMLAttributes<HTMLDivElement>>
+
+const TabsRoot = forwardRef(
+  ({ activeTab, children, className, ...restProps }: TabsProps, ref: ForwardedRef<HTMLDivElement>) => {
+    const tabsId = useId()
+    const [_activeTab, setActiveTab] = useState(activeTab ?? 0)
```
dlnr, line 23: Why the underscore? It's not private if its an exported prop right?

storybook/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -40,41 +40,50 @@ const tabMeta = {
 type Story = StoryObj<typeof meta>
 type TabStory = StoryObj<typeof tabMeta>
 
+const defaultTabs = [
```
dlnr, line 43: You could've put it in the meta constant, like we do everywhere else if there's more then one story using the props.

packages/react/src/Tabs/Tabs.tsx
RubenSibon: I wanted to call the prop `activeTab` as well so there is a clash between that and this state variable.

storybook/src/components/Tabs/Tabs.stories.tsx
RubenSibon: Agreed, I'll adjust that.

VincentSmedinga: I was taking a stab at refactoring the example tab template into a configuration object:

```
type TabContent = {
  label: string
  body: ReactNode
}

const tabContent: Array<TabContent> = [
  {
    label: 'Gegevens',
    body: <Paragraph>{exampleParagraph()}</Paragraph>,
  },
(…)
```

Then `defaultTabs` could become a function looping over this data.
I thought it might come closer to how implementers build Tabs.
And we can calculate the max value of `tab` and `activeTab` to show exactly the right amount of controls.

Good idea?

storybook/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -22,6 +24,15 @@ const SlowPanel = ({ children }: PropsWithChildren) => {
 const meta = {
   title: 'Components/Containers/Tabs',
   component: Tabs,
+  argTypes: {
```
VincentSmedinga, line 27: This control doesn’t do anything yet.

storybook/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -22,6 +24,15 @@ const SlowPanel = ({ children }: PropsWithChildren) => {
...
+    activeTab: {
+      control: {
+        type: 'number',
+        min: 0,
+        max: 3,
```
VincentSmedinga, line 32: This value can be made dynamic if we store the example content in an object, see other comment.

packages/react/src/Tabs/Tabs.tsx
VincentSmedinga: Yes, starting a variable's name with an underscore often indicates it is private. But I agree that `activeTab` is the best name for the prop. Maybe use `activeTabId` internally?

storybook/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -8,12 +8,14 @@ import { Meta, StoryObj } from '@storybook/react'
 import { PropsWithChildren } from 'react'
 import { exampleParagraph } from '../shared/exampleContent'
 
+const slowPanelDelay = 1000
```
RubenSibon, line 11: Indeed good to have no magic numbers!

storybook/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -5,96 +5,86 @@
...
-      <Tabs.Panel tab={2} key={3}>
+const defaultTabs = [
+  <Tabs.List key="tabsList">
+    {tabsContent.map((content, index) => (
+      <Tabs.Button key={content.label} tab={index}>
+        {content.label}
+      </Tabs.Button>
+    ))}
+  </Tabs.List>,
+  ...(() =>
```
VincentSmedinga, line 52: Curious why an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) is better here? What’s the advantage over ‘just’ looping over `tabsContent`?

packages/react/src/Tabs/Tabs.test.tsx
```diff
@@ -97,7 +97,111 @@ describe('Tabs', () => {
...
+
+    expect(lastTab).toHaveAttribute('aria-selected', 'false')
+    expect(lastTab).toHaveAttribute('tabindex', '-1')
+
+    expect(screen.getByRole('tabpanel')).toHaveTextContent('Content 1')
+  })
+
+  it('should set the last tab as the initially active tab with a too large argument', async () => {
+    render(
+      <Tabs activeTab={Infinity}>
```
VincentSmedinga, line 158: 😅

VincentSmedinga (approved):
Well done!

storybook/src/components/Tabs/Tabs.stories.tsx
RubenSibon: Yeah, I don't know why I did this. I think we can just put the map function in parentheses as well.

VincentSmedinga: I’ve made it so.

PR #1262, opened 2024-06-14
feat: Allow specifiying a time in a date input

VincentSmedinga:
This allows the type of `datetime-local` for Date Inputs.

I copied the docs for disabled and invalid inputs from Text Input. We should probably extract this to some section on forms in general?

### DIFF
--- a/storybook/src/components/DateInput/DateInput.docs.mdx
+++ b/storybook/src/components/DateInput/DateInput.docs.mdx
@@ -27,7 +27,6 @@ In other words, the input allows any valid combination of year, month, day, hour
 ### Invalid
 
 Indicates that the input value does not meet the specified constraints.
-The border of an invalid input is red.
 An [Error Message](?path=/docs/components-forms-error-message--docs) must be displayed above the field.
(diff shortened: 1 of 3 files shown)


storybook/src/components/DateInput/DateInput.docs.mdx
```diff
@@ -12,10 +12,47 @@ import README from "../../../../packages/css/src/components/date-input/README.md
...
+
+The value represents a local date and time, not necessarily the user’s local date and time.
+In other words, the input allows any valid combination of year, month, day, hour, and minute — even if such a combination is invalid in the user’s local time zone.
+
+<Canvas of={DateInputStories.DateTime} />
+
+### Invalid
+
+Indicates that the input value does not meet the specified constraints.
+The border of an invalid input is red.
```
alimpens, line 30: Do we need to describe this in text? It's already shown in the example

RubenSibon (approved):
LGTM

storybook/src/components/DateInput/DateInput.docs.mdx
VincentSmedinga: I do intend to describe design choices more explicitly, but I’ve removed it for now.

PR #1263, opened 2024-06-17
feat: Allow content to visually extrude from a Spotlight

dlnr:
This draft is a PoC for the usage of CSS gradients to allow the content to extrude.


VincentSmedinga:
Thanks for this effort. We’re closing it in favour of #1397.

- The use of gradients makes the height of the Spotlight unrelated to that of the text.
- The dependency on the Aspect Ratio component makes it less broadly applicable. 
- The feature of spanning rows in the Grid has proved useful.

PR #1265, opened 2024-06-18
chore: Document that the header and footer are required

VincentSmedinga:
This follows up input from our design department that we hsould mark the Header and Footer as required for every page and every application because it plays such a big role in making all our websites recognizable as being from the City, which generates trust to use them.

### DIFF
--- a/packages/css/src/components/footer/README.md
+++ b/packages/css/src/components/footer/README.md
@@ -2,13 +2,16 @@
 
 # Footer
 
-The footer is the closing section on every web page.
-Its content provides service information.
-A part is prescribed, and a part is filled based on the website’s objectives.
+Provides service information at the bottom of every page.
 
 ## Guidelines
 
-- The footer is mandatory, and there is one on every page.
+The Footer must be used on all websites and applications for the City of Amsterdam.
+It includes the logo of the City or the organization, the site title (except for the general website), and a menu with links to commonly used pages.
+The Footer is important because it conveys our corporate identity and is the first thing people see.
+Using it consistently helps users recognize and trust the website.
+It is the same on every page of the application.
+
 - The first column focuses on contact information.
   The information is as specific as possible and tailored to the respective website or page (e.g., including a chat option if available).
 - The second column contains links to relevant (online) sites or sources.
--- a/packages/css/src/components/footer/README.md
+++ b/packages/css/src/components/footer/README.md
@@ -6,15 +6,16 @@ Provides service information at the bottom of every page.
 
 ## Guidelines
 
-The Footer must be used on all websites and applications for the City of Amsterdam.
-It includes the logo of the City or the organization, the site title (except for the general website), and a menu with links to commonly used pages.
-The Footer is important because it conveys our corporate identity and is the first thing people see.
-Using it consistently helps users recognize and trust the website.
+The Footer must be used on all websites for the City of Amsterdam.
+For applications, only its menu at the bottom can be sufficient.
 It is the same on every page of the application.
+
+The large blue area offers space for various practical links:
 
 - The first column focuses on contact information.
   The information is as specific as possible and tailored to the respective website or page (e.g., including a chat option if available).
 - The second column contains links to relevant (online) sites or sources.
 - The third column refers to newsletters, social media, etc.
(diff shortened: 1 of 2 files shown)


packages/css/src/components/header/README.md
```diff
@@ -2,13 +2,18 @@
...
-- Use the Header when the site is hosted on amsterdam.nl or one of its subdomains.
-- The page menu can contain a maximum of 5 items, including menu and search.
-- When you have a long subsite title, use no or as few page menu items as possible.
+The Header must be used on all websites and applications for the City of Amsterdam.
+It includes the logo of the City or the organization, the site title (except for the general website), and a menu with links to commonly used pages.
+The Header is important because it conveys our corporate identity and is the first thing people see.
+Using it consistently helps users recognize and trust the website.
+It is the same on every page of the application.
+
+The Page Menu should contain no more than 5 items, including ‘Search’ and ‘Menu’.
```
dlnr, line 16: You could add 5 items with long titles so it could help to give an approximate character length.

packages/css/src/components/footer/README.md
```diff
@@ -2,13 +2,16 @@
...
 
-The footer is the closing section on every web page.
-Its content provides service information.
-A part is prescribed, and a part is filled based on the website’s objectives.
+Provides service information at the bottom of every page.
 
 ## Guidelines
 
-- The footer is mandatory, and there is one on every page.
+The Footer must be used on all websites and applications for the City of Amsterdam.
```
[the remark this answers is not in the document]
VincentSmedinga: Wow I copied the Header docs and forgot to adapt it for Footer. I’ve updated it to mention websites and applications separately.

We can elaborate on ‘public’ versus ‘employee’ websites and the difference between websites and applications later. This PR was a quick update to document that at least the Header is required.

packages/css/src/components/header/README.md
VincentSmedinga: Yea I’m not a fan of these arbitrary limits either. Reworded.

packages/css/src/components/footer/README.md
```diff
@@ -2,16 +2,20 @@
...
-The footer is the closing section on every web page.
-Its content provides service information.
-A part is prescribed, and a part is filled based on the website’s objectives.
+Provides service information at the bottom of every page.
 
 ## Guidelines
 
-- The footer is mandatory, and there is one on every page.
+The Footer must be used on all websites for the City of Amsterdam.
+For applications, only its menu at the bottom can be sufficient.
```
alimpens, line 10: This is a little unclear imo. I think you mean the PageMenu here? Maybe we should move this to a story, so we can show how you do this.

packages/css/src/components/header/README.md
```diff
@@ -2,13 +2,22 @@
 
 # Header
 
-Arranges the City’s logo, the name of the application, and a page menu.
+Displays the City’s logo and a navigation menu at the top of every page.
```
alimpens, line 5: It doesn't necessarily show a navigation menu, right?

packages/css/src/components/header/README.md
```diff
@@ -2,13 +2,22 @@
...
-- The page menu can contain a maximum of 5 items, including menu and search.
-- When you have a long subsite title, use no or as few page menu items as possible.
+The Header must be used on all websites and applications for the City of Amsterdam.
+It includes the logo of the City or the organization, the site title (except for the general website), and a menu with links to commonly used pages.
+
+The Header is important because it conveys our corporate identity and is the first thing people see.
+Using it consistently helps users recognize and trust the website.
+It is the same on every page of the application.
+
+The amount of links in the Page Menu should be limited to a handful.
```
alimpens, line 17: I think we have differing opinions on this: 'a handful' is way too vague imo, max 5 is a more straightforward guideline I think.

VincentSmedinga (commented):
Oh, I forgot to post these comments.

packages/css/src/components/header/README.md
alimpens: mijn.amsterdam.nl and meldingen.amsterdam.nl currently do not have a navigation menu. I think the way it's currently phrased can imply that a navigation menu is mandatory.

alimpens (approved):

PR #1269, opened 2024-06-21
fix: Use visually hidden as utility class only

alimpens:
We're probably going to get rid of the VisuallyHidden React component, and just use it as a utility class. This PR prepares for that, and also fixes some invalid HTML in the Footer examples.

It also fixes an issue with the Avatar in Header story.

### DIFF
--- a/packages/react/src/Avatar/Avatar.tsx
+++ b/packages/react/src/Avatar/Avatar.tsx
@@ -9,7 +9,6 @@ import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes } from 'react'
 import { Icon } from '../Icon'
 import { Image } from '../Image'
-import { VisuallyHidden } from '../VisuallyHidden'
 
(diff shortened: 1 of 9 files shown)


packages/react/src/Avatar/Avatar.tsx
```diff
@@ -9,7 +9,6 @@ import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes } from 'react'
 import { Icon } from '../Icon'
 import { Image } from '../Image'
-import { VisuallyHidden } from '../VisuallyHidden'
```
dlnr, line 12: Is the component itself going to be removed too?

VincentSmedinga (approved):
Yes much better

packages/react/src/Avatar/Avatar.tsx
alimpens: Probably, yeah

PR #1273, opened 2024-06-26
chore: Update documentation on description and Error Message with Field Set with Checkboxes

alimpens:


storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -179,28 +182,40 @@ export const CheckboxGroup: Story = {
...
-//   ),
-// }
+export const CheckboxGroupWithError: Story = {
+  args: {
+    invalid: true,
+    legend: 'Waar gaat uw melding over?',
+  },
+  render: (args) => (
+    <FieldSet
+      aria-labelledby="fieldset2 description7 error4"
```
VincentSmedinga, line 192: These numbers are a bit confusing: fieldset 2, description 7, error 4. This doesn’t help the explanations of how to populate the label.

I understand we need the `id`s to be unique for the entire docs page. So we can’t use field labels. Prefixing with the story name will make them overly long.

How about a letter per story, e.g. `a-fieldset`, `a-description`, `a-error`. And for examples with multiple inputs: `c-input-1`, `c-input-2`. A bit obscure as well maybe, but can’t think of anything better that is still compact.

VincentSmedinga: Also, making the error label dynamic helps prevent copy paste errors by developers.

```suggestion
      aria-labelledby=`fieldset2 description7 {invalid ? `error4` : undefined}`
```

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -179,28 +182,40 @@ export const CheckboxGroup: Story = {
...
+      legend={args.legend}
+    >
+      <Paragraph id="description7" size="small" className="ams-mb--sm">
+        De laatstgenoemde melding.
+      </Paragraph>
+      {args.invalid && (
+        <ErrorMessage className="ams-mb--sm" id="error4">
+          Geef aan waar uw melding over gaat.
+        </ErrorMessage>
+      )}
```
VincentSmedinga, line 204: Let’s add the conditional error (and labelledby id) to the examples without errors as well. Otherwise, people that toggle the `invalid` prop in these stories will see the red borders but not the text.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -58,12 +58,19 @@ Use a Field Set to group checkboxes.
 Please note: [NVDA has a bug](https://github.com/nvaccess/nvda/issues/12718) which causes it to
 not report a description connected to a Field Set when it contains checkboxes.
 
-Try to avoid using descriptions for Field Sets containing checkboxes for this reason.
+For this reason, we add the description text to the label.
+Screen reader users will no longer be able to skip the description,
+but we think that is preferable to missing the description entirely.
+Add an `id` to the Field Set and the description, and add both to the `aria-labelledby` attribute of Field Set.
```
VincentSmedinga, line 64: I’m thinking we should add a snippet with the essential code for this. Just something like:

```
<Fieldset id="fieldset" labelledby="fieldset description" legend="Label">
  <Paragraph id="description">Description
```

This makes it crystal clear for people what markup to add for this guidelines. Currently, people must either read these instructions very closely or compare code of multiple examples and pick what they need.

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -160,7 +160,10 @@ export const CheckboxGroup: Story = {
     legend: 'Waar gaat uw melding over?',
   },
   render: (args) => (
-    <FieldSet legend={args.legend} invalid={args.invalid}>
+    <FieldSet aria-labelledby="fieldset1 description6" id="fieldset1" invalid={args.invalid} legend={args.legend}>
```
RubenSibon, line 163: Does the content of `#description6` really add value to the Field Set label on top of its own legend property? A description seems additive to me and it is already visible right below.

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon, line 192: Here the same comment as earlier.

I'm not sure about updating the label with the error message if it appears. Why would you want to do that?

RubenSibon (dismissed):
A couple of comments/questions, but other than that it LGTM.

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: @VincentSmedinga, don't you want to send an empty string rather than `undefined` to a string list?

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: Yes, was just testing it out. It would be nice for all stories that if the input(s) are made valid by the user, the error state should actually be removed and not just the style.

storybook/src/components/FieldSet/FieldSet.stories.tsx
alimpens: > it is already visible right below

It isn't for screen readers. When you tab to an input in a form, a screen reader only reads whatever is connected to that input. If you don't connect the description, it isn't read. Normally, you would use `aria-describedby` for this, but that doesn't work with NVDA for Field Sets containing Checkboxes because of the bug linked to in the docs.

storybook/src/components/FieldSet/FieldSet.stories.tsx
alimpens: > Also, making the error label dynamic helps prevent copy paste errors by developers.

And it doesn't lead to code that is flagged by the WAVE a11y checker, so I have to change this anyway 😁.

A letter per story is a good idea imo. I'd prefer a syntax like this though, which reads a little easier imo:

`fieldset-a`, `description-a`, `input-b1`, `input-b2`.

alimpens: Same answer: so it's read by screen readers

VincentSmedinga: It is easier to read. Technically, the letter represents the story and you want to express the hierarchy of parts (like we do with tokens). You could get in trouble with multiple examples that have multiple field sets that have multiple inputs, but fine to have the letter later and see how it goes.

VincentSmedinga (approved):

PR #1274, opened 2024-06-26
chore: Add focus management documentation

alimpens:


storybook/src/docs/usability.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+import { Meta } from "@storybook/blocks";
+
+<Meta title="Docs/Design Guidelines/Usability" />
+
+# Usability
+
+## Focus management
+
+In traditional multi-page applications (MPAs) moving to another page means resetting focus to the `<body>` element,
```
VincentSmedinga, line 11:
```suggestion
In traditional multi-page applications (MPAs), navigating to another page resets the focus to the `<body>` element,
```

storybook/src/docs/usability.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+or the first element with an `autofocus` attribute if there is one.
+For single-page applications (SPAs) this isn’t always done consistently.
+For example, Next.js sets the focus on a new page at roughly the same position as it was on the old page,
+resulting in an unexpected focus position.
```
VincentSmedinga, line 15:
```suggestion
For example, Next.js sets the focus on a new page at roughly the same position as it was on the old page.
This can result in an unexpected focus position, e.g. when navigating between pages of a form.
```

storybook/src/docs/usability.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+
+Make sure to correct this focus behaviour in your application.
+The simplest and most consistent way to do this is to set focus on the `<body>` element when moving to another page.
```
VincentSmedinga, line 18:
```suggestion
The simplest and most consistent way to do this is to set focus on the `<body>` element after navigating to another page.
```

Or ‘when loading a page’?

storybook/src/docs/usability.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+For some applications it might make more sense to set the focus somewhere else when moving to a new page,
```
VincentSmedinga, line 19:
```suggestion
For some applications it might make more sense to set the focus somewhere else,
```

storybook/src/docs/usability.docs.mdx
```diff
@@ -0,0 +1,23 @@
...
+but be very careful with this.
+Your assumptions on how your users use your app might not be correct.
```
RubenSibon, line 21:
```suggestion
Your assumptions on how users use the app might not be correct.
```

Too many you's.

storybook/src/docs/usability.docs.mdx
RubenSibon: "after navigating" seems good to me.

storybook/src/docs/usability.docs.mdx
RubenSibon:
```suggestion
For example, Next.js sets the focus on a new page at roughly the same position as it was on the old page.
This can cause the focus to change unexpectedly, like when you move between pages in a form.
```

RubenSibon (changes requested):

storybook/src/docs/usability.docs.mdx
alimpens: > This can cause the focus to change unexpectedly, like when you move between pages in a form.

This kind of changes the meaning of that sentence.

VincentSmedinga (approved):

PR #1309, opened 2024-06-28
feat: Let Form Error List add the error count to the document title

alimpens:
Part of [the GOV.UK validation pattern](https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors) is showing the error count in the document title. Seeing as Form Error List already gets the list of errors, it made sense to add this functionality here.

On the GOV.UK Design System discussion boards someone mentioned that people often forget to implement this when they have to do it manually, so I think it makes sense to incorporate it in a component.


packages/react/src/FormErrorList/useAddErrorCountToPageTitle.tsx
```diff
@@ -0,0 +1,27 @@
...
+) => {
+  const [documentTitle, setDocumentTitle] = useState<string>()
+
+  useEffect(() => {
+    setDocumentTitle(document.title)
+  }, [])
+
+  if (documentTitle) {
+    if (errors.length > 0) {
+      document.title = `(${errors.length} ${errors.length === 1 ? suffix.singular : suffix.plural}) ${documentTitle}`
```
RubenSibon, line 20: Now we have a ternary inside an if-statement nested inside an if-statement. Maybe add an `else` for the plural case to remove the ternary?

packages/react/src/FormErrorList/useAddErrorCountToPageTitle.tsx
```diff
@@ -0,0 +1,27 @@
...
+    singular: 'invoerfout',
+  },
+) => {
+  const [documentTitle, setDocumentTitle] = useState<string>()
+
+  useEffect(() => {
+    setDocumentTitle(document.title)
+  }, [])
+
+  if (documentTitle) {
```
RubenSibon, line 18: Is it possible in this hook to `return null` if `documentTitle` is nullish to avoid if-statement nesting?

packages/css/src/components/form-error-list/README.md
```diff
@@ -23,6 +23,11 @@ For questions that require a user to select one or more options from a list usin
 
 Put the Form Error List directly above the first question on the page. Place it outside of the `<form>`-tag, [to make sure screen readers do not skip it](https://nldesignsystem.nl/richtlijnen/formulieren/meerdere-stappen/#plaats-de-informatie-over-waar-de-gebruiker-is-in-de-stappen-boven-het-formulier).
 
+## Error count in document title
+
+The error count is added to the document title by this component,
```
VincentSmedinga, line 28:
```suggestion
This component adds the error count to the document title,
```

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -25,6 +26,8 @@ export type FormErrorListProps = {
    * Note: this intentionally does not change the font size.
    */
   headingLevel?: HeadingLevel
+  /** The text following the error count. */
+  docTitleSuffix?: { plural: string; singular: string }
```
VincentSmedinga, line 30: It’s not the suffix of the document title. How about `errorText`, `errorLabel`, `errorTerm`? Maybe even `errorCount*` to match the hook’s name.

alimpens: Isn't `error*` a little confusing? This is a prop for `FormErrorList`, which of itself doesn't have anything to do with the document title. Specifying that this prop is about the document title feels necessary imo. How about `docTitleLabel`?

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -34,10 +37,13 @@ export const FormErrorList = forwardRef(
       errors,
       heading = 'Verbeter de fouten voor u verder gaat',
       headingLevel = 2,
+      docTitleSuffix,
```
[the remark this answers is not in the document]
alimpens: Good point, will do when we decide on the name

packages/react/src/FormErrorList/useAddErrorCountToPageTitle.tsx
alimpens: Do you mean an early return? Yeah, that is a bit cleaner

packages/react/src/FormErrorList/useAddErrorCountToPageTitle.tsx
alimpens: Something like this you mean?

```
  if (errors.length === 1) {
    document.title = `(${errors.length} ${suffix.singular}) ${documentTitle}`
  } else if (errors.length > 1) {
    document.title = `(${errors.length} ${suffix.plural}) ${documentTitle}`
  } else {
    document.title = documentTitle
  }
```

packages/react/src/FormErrorList/FormErrorList.tsx
VincentSmedinga: I proposed `error*` from an opposite perspective: letting the component ask for the singular and plural forms of the noun – without disclosing why it needs them, on purpose.

If we consider setting the title as an implementation detail, we can maintain a more generic content model for this component: “I am responsible for listing errors of a form. Please provide me with a heading and its level, the list of errors, and the words to use for ‘error’ and ‘errors’; and I will take care of everything.” The documentation lets people know that the component updates the document title next to rendering a box in the page.

When coupling the words with the document label, we would need a second prop with the same content (or rename it) if the component needs to render these words elsewhere later (visually, audibly…).

I’d prefer thinking from the content the component needs, not where it needs them for.

alimpens: We generally try to use prop names that give the user an indication of what they should enter there, no? A `heading` prop will render the `heading` of a component, so you should enter the heading text there, etc. 

Something like `errorText` does not communicate the same, imo. I'd have to read the docs to know what I'd have to enter there. I'm not sure `docTitleLabel` is the way to go here, but it gives the user a better indication of what they should enter there. 

> When coupling the words with the document label, we would need a second prop with the same content (or rename it) if the component needs to render these words elsewhere later (visually, audibly…).

I don't think we should look ahead too much. For the foreseeable future, this prop is used to set a part of the doc title, so that can be reflected in the name imo.

VincentSmedinga: > We generally try to use prop names that give the user an indication of what they should enter there, no?

Of course. But a name that describes a field of a thing clearly, without mentioning where it appears, does meet that objective.

> Edit: Thought about it some more, maybe `errorCountLabel` together with a JSDoc description that it's used in the doc title is sufficient?

Yes, I can totally go along with that name.

alimpens: Allrighty, updated it!

VincentSmedinga (approved):

PR #1310, opened 2024-06-28
chore: Use Fields in FieldSet docs

alimpens:
What do we want to do with the error bar for FieldSets containing Fields? Should the error bar only be shown for the Field that has an input with an error (so `<Field invalid="true"`)? Or should it be shown for the entire FieldSet (so `<FieldSet invalid="true"`)?

Currently, the second option is shown.

### DIFF
--- a/storybook/src/components/FieldSet/FieldSet.stories.tsx
+++ b/storybook/src/components/FieldSet/FieldSet.stories.tsx
@@ -7,6 +7,7 @@ import {
   Checkbox,
   Column,
   ErrorMessage,
+  Field,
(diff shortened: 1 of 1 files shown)


RubenSibon:
I agree that option 2 is the right choice. The sub-fields or inputs already get the detailed error message that provides the user with clear and actionable feedback on what to do, but if the inputs belong together in a field set then it makes sense that the error indication also belongs to the entire field set.

RubenSibon (approved):

VincentSmedinga (approved):

PR #1319, opened 2024-07-03
feat: Allow a row to wrap its items onto multiple lines

VincentSmedinga:


dlnr (approved):
This got me thinking on utility classes for flex properties but this seems like a fine change.

PR #1326, opened 2024-07-05
feat: Spotlight breakout with margins

dlnr:
Another breakout content test, but now with margins instead of a gradient background.


VincentSmedinga:
Thanks for this effort. We’re closing it in favour of #1397.

- The dependency on the Aspect Ratio component makes it less broadly applicable.
- The same goes for the `breakout` prop of the entire Grid. I’d expect this to be a feature of a Cell.

PR #1328, opened 2024-07-05
feat: Add focus on initial render to Form Error List

alimpens:
This adds focus on initial render to Form Error List, [as prescribed by gov.uk](https://design-system.service.gov.uk/components/error-summary/).

I had to split the component in 2 parts. One part sets the document title, creates a state and returns early if there are no errors.
The second part is what renders when the component does receive a list of errors. 
These parts had to be separated because the early return caused the hooks used (useRef, useImparitiveHandle, useEffect) to be called conditionally, which isn't allowed by [the rules of hooks](https://react.dev/warnings/invalid-hook-call-warning).

The `hasFocusedOnce` state is necessary to prevent the component receiving focus every time it's shown on screen. React Hook Form validates on submit and revalidates on change, which causes the component to appear while your typing. You don't want to move focus from the input you're typing in to the Form Error List in those cases.


packages/css/src/components/form-error-list/form-error-list.scss
```diff
@@ -0,0 +1,4 @@
+.ams-form-error-list {
+  // The offset is doubled to compensate for the Alert border
+  outline-offset: var(--ams-form-error-list-outline-offset);
```
VincentSmedinga, line 3: Why do we need to double the offset if “its line is outside of the element's border” ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/outline#description))?

VincentSmedinga, line 3: We do [the same for File Input](https://github.com/Amsterdam/design-system/blob/develop/packages/css/src/components/file-input/file-input.scss#L24) but without an extra token. Which of the two approaches do you prefer?

packages/css/src/components/form-error-list/README.md
```diff
@@ -28,6 +28,12 @@ Put the Form Error List directly above the first question on the page. Place it
 This component adds the error count to the document title,
 in line with [GOV.UK guidelines for informing users about validation errors](https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors).
 
+## Focus on initial render
+
+This component receives focus on initial render.
```
VincentSmedinga, line 33: This accommodates a wider audience:

```suggestion
This component receives focus as soon as it gets displayed on the screen.
```

packages/css/src/components/form-error-list/README.md
```diff
@@ -28,6 +28,12 @@ Put the Form Error List directly above the first question on the page. Place it
...
+This allows keyboard users to quickly navigate to the errors in the form.
```
VincentSmedinga, line 34: Doesn’t it also help sighted users quickly understand which fields have errors? And users of screen readers as well?

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -24,6 +22,8 @@ export type FormErrorListProps = {
   errorCountLabel?: { plural: string; singular: string }
   /** The list of error messages to display. */
   errors: FormError[]
+  /** Whether the component receives focus on first render */
+  focusOnRender?: boolean
```
VincentSmedinga, line 26: I think we can name this [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus).

packages/react/src/FormErrorList/FormErrorList.tsx
```diff
@@ -34,39 +34,25 @@ export type FormErrorListProps = {
...
-      >
-        <LinkList>
-          {errors.map(({ id, label }) => (
-            <LinkList.Link href={id} key={`${id}-${label}`}>
-              {label}
-            </LinkList.Link>
-          ))}
-        </LinkList>
-      </Alert>
+        hasFocusedOnce={hasFocusedOnce}
```
VincentSmedinga, line 53: Can we reorder this one? :)

packages/react/src/FormErrorList/FormErrorListWithErrors.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ForwardedRef, forwardRef, useEffect, useImperativeHandle, useRef } from 'react'
+import type { FormErrorListProps } from './FormErrorList'
+import { Alert } from '../Alert'
+import { LinkList } from '../LinkList'
+
+type FormErrorListWithErrorsProps = Omit<FormErrorListProps, 'errorCountLabel'> & {
+  /** Whether the component has set focus once. */
+  hasFocusedOnce: boolean
+  /** Callback to let parent component know whether focus has been set once. */
+  setHasFocusedOnce: any
```
VincentSmedinga, line 16: Should we avoid the `any` type?

packages/css/src/components/form-error-list/form-error-list.scss
VincentSmedinga: Additional research indicated we need to add an `outline-width: medium` reset: Safari on Mac seems to have an abnormal initial value of `thick` here, which explains the extra whitespace between border and outline.

VincentSmedinga: I’ve added this line to this component for now. We’ll research this further and expand to other components if necessary, in a new ticket & PR.

VincentSmedinga (dismissed):

VincentSmedinga (approved):

PR #1329, opened 2024-07-06
feat!: Improve sizing behaviour of Dialog

VincentSmedinga:
Aligns the (maximum) block and inline sizes of a Dialog with our spacing system.

A Dialog now gets
- as large as necessary to fit its content,
- with a maximum that leaves a medium grid space of margin at all sides,
- and a horizontal maximum of 48 rems.

The inner layout has been improved as well: making the `<dialog>` a flex container automatically makes its child `<form>` size optimally. Setting heights for the grid rows of the form no longer seems necessary.

It is a breaking change because the `dialog-form-max-block-size` token gets removed.

### DIFF
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -3,14 +3,19 @@
  * Copyright Gemeente Amsterdam
  */
 
+@mixin reset-dialog {
+  inset: 0;
+  padding-block: 0;
+  padding-inline: 0;
+}
+
 .ams-dialog {
   background-color: var(--ams-dialog-background-color);
   border: var(--ams-dialog-border);
-  inset: 0;
   max-inline-size: var(--ams-dialog-max-inline-size);
-  padding-block: 0;
-  padding-inline: 0;
   position: fixed;
+
+  @include reset-dialog;
 
   /* no token because dialog does not inherit from any element */
   &::backdrop {
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -12,6 +12,7 @@
 .ams-dialog {
   background-color: var(--ams-dialog-background-color);
   border: var(--ams-dialog-border);
+  display: flex;
   inline-size: var(--ams-dialog-inline-size);
   max-block-size: var(--ams-dialog-max-block-size);
   max-inline-size: var(--ams-dialog-max-inline-size);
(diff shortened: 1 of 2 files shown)


proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -3,12 +3,13 @@
     "dialog": {
       "background-color": { "value": "{ams.color.primary-white}" },
       "border": { "value": "0" },
-      "max-inline-size": { "value": "min(87.69vw, 45rem)" },
+      "inline-size": { "value": "calc(100vw - 2 * {ams.space.grid.md})" },
+      "max-block-size": { "value": "calc(100vh - 2 * {ams.space.grid.md})" },
```
alimpens, line 7: Shouldn't Dialog just use `%` instead of `vw` / `vh`? It's ok for it to take up a certain portion of space relative to its parent, no? Currently it listens to the viewport, which means it overflows when used in a smaller container (like currently in the Storybook docs on smaller screens). Added benefit: we would be rid of 2 more directional values.

RubenSibon (commented):
I don't have much to add to Aram's comment, but I did notice that [this example of how to open a dialog](https://designsystem.amsterdam/demo-DES-785-dialog-margin/?path=/docs/components-containers-dialog--docs&globals=theme:Spacious#open-dialog-button) doesn't look right.

Schermafbeelding 2024-07-10 140045

proprietary/tokens/src/components/ams/dialog.tokens.json
VincentSmedinga: A percentage width requires a Dialog to be rendered at a position in the DOM where no parents can constrain the width. I don’t think that’s a sacrifice we can make. With viewport units, people can render the Dialog anywhere they want.

VincentSmedinga:
> I don't have much to add to Aram's comment, but I did notice that [this example of how to open a dialog](https://designsystem.amsterdam/demo-DES-785-dialog-margin/?path=/docs/components-containers-dialog--docs&globals=theme:Spacious#open-dialog-button) doesn't look right.

I had noticed it and hoped it would be a local issue. However, I struggle to see how any of the changes in this PR could be responsible. If the issue persists, let’s update Storybook and research further.

proprietary/tokens/src/components/ams/dialog.tokens.json
dlnr: Or `100dvh` to prevent weird overflow problems on mobile devices?

alimpens (approved):

packages/css/src/components/dialog/dialog.scss
```diff
@@ -21,18 +27,14 @@
 .ams-dialog__form {
   display: grid;
   gap: var(--ams-dialog-form-gap);
-  grid-template-rows: auto 1fr auto;
-  max-block-size: var(--ams-dialog-form-max-block-size);
```
dlnr, line 25: This broke the scrolling inside the Dialog (while keeping the action buttons in view).

PR #1330, opened 2024-07-08
feat: Add horizontal and vertical alignment options to Row

VincentSmedinga:
Rows can now align their items horizontally and vertically.

This is useful for:
- right-aligning a single item, e.g. `<Row align="end"><Avatar /></Row>`
- a link at the far end of a heading, e.g. `<Row align="between"><Heading /><Link /></Row>`
- ‘between’ can also be used for ‘previous’ and ‘next‘ buttons if we want them like that
- vertically aligning differently sized texts: `<Row alignVertical="baseline"><Heading /><Paragraph /></Row>`

After merging, I’ll do the same for `Column`.

This touches on an important API choice: Row has alignment and vertical alignment, Column has alignment and horizontal alignment. For both, the regular alignment works in their main direction. Physical directions is conceptually easier to understand than separating ‘justification’ and ‘alignment’ as the flexbox spec does. For CSS, that is the better API. For our DSL, horizontal and vertical is. We translate between them for the best of both worlds.

Horizontal alignment uses ‘start’ and ‘end’ as labels for the extremes; vertical uses ‘top’ and ‘bottom’. Again, these words are easy to guess (as soon as people are used to use ‘start’ instead of ‘left’, that is).


packages/css/src/components/row/row.scss
```diff
@@ -19,3 +19,47 @@
...
+
+.ams-row--align-evenly {
+  justify-content: space-evenly;
+}
+
+.ams-row--align-start {
+  justify-content: flex-start;
+}
+
+.ams-row--align-vertical-bottom {
```
alimpens, line 47: Don't we want to use `end` here, instead of `bottom`? We're already asking are users to familiarize themselves with `start` and `end`, might as well do it consistently, no?

packages/react/src/Row/Row.tsx
```diff
@@ -6,24 +6,43 @@
...
+import type { CrossAlign, MainAlign } from '../common/layout'
 
 export const rowGapSizes: Array<string> = ['extra-small', 'small', 'medium', 'large', 'extra-large']
 
-type RowTag = 'article' | 'div' | 'section'
 type RowGap = (typeof rowGapSizes)[number]
+type RowTag = 'article' | 'div' | 'section'
 
 export type RowProps = {
+  /** The horizontal alignment of the items in the row. */
```
alimpens, line 17: Maybe use `inline` here instead of `horizontal`, to educate our users?

packages/react/src/Row/Row.tsx
```diff
@@ -6,24 +6,43 @@
...
+  (
+    { align, alignVertical, as: Tag = 'div', children, className, gap = 'medium', wrap, ...restProps }: RowProps,
+    ref: any,
+  ) => (
+    <Tag
+      {...restProps}
+      ref={ref}
+      className={clsx(
+        'ams-row',
+        `ams-row--${gap}`,
```
alimpens, line 39: Shouldn't we prefix the modifier with `gap` here? So `ams-row--gap-medium`?

packages/css/src/components/row/row.scss
```diff
@@ -19,3 +19,47 @@
...
+  justify-content: flex-end;
+}
+
+.ams-row--align-evenly {
+  justify-content: space-evenly;
+}
+
+.ams-row--align-start {
+  justify-content: flex-start;
+}
```
alimpens, line 45: This is default, do we need a class and option etc. for that?

packages/css/src/components/row/row.scss
```diff
@@ -19,3 +19,47 @@
...
+  align-items: baseline;
+}
+
+.ams-row--align-vertical-center {
+  align-items: center;
+}
+
+.ams-row--align-vertical-stretch {
+  align-items: stretch;
+}
```
alimpens, line 61: Same here, this is default

storybook/src/components/Row/Row.docs.mdx
```diff
@@ -24,6 +24,21 @@ Wrap a Row around a set of components that need the same amount of white space b
 
 <Controls />
 
+### Horizontal Alignment
+
+Items can be aligned horizontally to the start, center or end of the row.
+Or whitespace can be distributed between, around or evenly around them.
```
alimpens, line 30: Maybe add here that start alignment is default?

storybook/src/components/Row/Row.stories.tsx
```diff
@@ -34,9 +49,25 @@ type Story = StoryObj<typeof meta>
...
+    alignVertical: 'center',
+    children: ThreeParagraphs,
+  },
+}
+
 export const Wrapping: Story = {
   args: {
-    children: FourBoxes,
+    children: Array.from(Array(4).keys()).map((i) => (
+      <Paragraph className="ams-docs-pink-box" key={i} style={{ inlineSize: '16rem' }} />
```
alimpens, line 69: Why did you use empty paragraphs here?

RubenSibon (commented):
I'm not sure that I agree with the rational to differ from how CSS handles alignment versus justification. Why not use `justify` instead of `align` if that is what web developers already are used to? And why not educate the users to think in terms of `start` and `end` and locale agnostic directionality as Aram suggested?

packages/css/src/components/row/row.scss
VincentSmedinga: I think either approach would be justifiable.

I aim for names that are easy to guess and close to common mental models. My main goal for a design system is helping people from different disciplines work together on digital products. One dream is that we’ll have some kind of environment where people with all kinds of roles, not just developers, design in the browser, writing our API.

In that light, I would say reserving ‘start’ and ‘end’ for horizontal and using ‘top’ and ‘bottom’ for vertical is a bit more straightforward. I guess people need to process ‘start’ meaning the top of a column a little bit longer than ‘top’.

packages/react/src/Row/Row.tsx
VincentSmedinga: No, too difficult. Maybe jargon, even 😉 Horizontal is far more natural – even I still need to translate ‘block’ and ‘inline’ to physical directions, let alone non-devs.

packages/react/src/Row/Row.tsx
VincentSmedinga: Yes indeed!

storybook/src/components/Row/Row.stories.tsx
VincentSmedinga: Copy, paste, adapt, I think. Updated to spans.

storybook/src/components/Row/Row.docs.mdx
VincentSmedinga: The robots even advised to write the options out in a list. It is longer, but they still have a point:

> This revision organizes the information into bullet points, making it easier to scan and understand the options available. It also provides a consistent format for listing the alignment types and their descriptions.

VincentSmedinga:
> I'm not sure that I agree with the rational to differ from how CSS handles alignment versus justification. Why not use `justify` instead of `align` if that is what web developers already are used to? And why not educate the users to think in terms of `start` and `end` for the "vertical" axis and locale agnostic directionality as Aram suggested?

I understand. See my replies to those comments. I want to be inclusive for designers and other roles in digital product development. Design systems are multidisciplinary. We as front-end devs are the ones structuring the design for everyonw, but the names we choose are not just for ourselves but the entire team. We feedback our iteration on modeling and naming to designers in order to reach common understanding. Development will always have some “oh that’s too technical, I don’t understand and don’t want to try” about it, and most things we do are best left behind our curtains, but I believe the shared concepts in the design should be easily shared, through common language.

packages/css/src/components/row/row.scss
VincentSmedinga: We’ll go for ‘start’ and ‘end’ in both directions.

packages/css/src/components/row/row.scss
VincentSmedinga: We don’t see a use case for it now, so we’ll be removing the classes setting defaults.

packages/css/src/components/row/row.scss
VincentSmedinga: Updated.

packages/react/src/index.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 /* Append here */
+export * from './common/layout'
```
dlnr, line 7: Do we need to export utils here? We do not do this for the other one (keyboard focus) and it works fine.

dlnr (commented):
The rest seems fine. I do wonder how we would deal with elements that use `align-self`, but if they do they are not added by us.

alimpens (approved):

PR #1308, opened 2024-06-28
feat: Add example pages of an application to assess compact text and space sizes

VincentSmedinga:
This adds two pages for Amopis and a form page for Amsterdam.nl.

It extracts a `Layout` component and a common `meta` object for both sites.

It shows the page on the Docs page as well, although it doesn’t help much with the code view yet.

PR #1331, opened 2024-07-08
chore: Add example pages of an application to assess compact text and space sizes

VincentSmedinga:
This adds two pages for Amopis and a form page for Amsterdam.nl.

It extracts a Layout component and a common meta object for both sites.

It shows the page on the Docs page as well, although it doesn’t help much with the code view yet.

Replaces #1308.


VincentSmedinga:
These examples have served their purpose in the team to visually assess the changes in spacing and typography for Compact Mode. We don’t necessarily promote these specific screens yet, so we’re not merging it into production.

PR #1337, opened 2024-07-09
fix: Typescript fix for storymeta and broken story (CustomTagName)

dlnr:
Tried to fix argTypes with typecasting to ControlType but there is no good way to reuse these argTypes. Perhaps VScode gives other errors then Webstorm in this case?


VincentSmedinga (approved):
Too bad. Yes, let’s just repeat them for now and maybe try again later.

PR #1355, opened 2024-07-10
fix!: Change CSS classes for Heading level for consistency with token names

RubenSibon:
Heading CSS classnames now follow the lead of Heading tokens to include the word "level" to change the visual size, but not the position in the page hierarchy.


VincentSmedinga (approved):

VincentSmedinga:
This is a breaking change, so the PR title requires its prefix to have an exclamation point.

PR #1366, opened 2024-07-12
feat!: Add and rename Pagination props, remove use of `aria-label`

RubenSibon:
# Describe the pull request

The Amsterdam Design System [advices against using `aria-label`](https://github.com/Amsterdam/design-system/blob/main/packages/react/documentation/coding-conventions.md?plain=1#L37). We should then not use the bad practice in our own components.

## What

This branch removes usage of `aria-label` and replaces it with the labels in `span`s with the `ams-visually-hidden` class name.

## Additional notes

- Should we also refactor the properties `nextAriaLabel` and `previousAriaLabel` and thus break the API? **Yes, the props have been changed, thus breaking the API.**
- A [new issue has been made](https://designsystem.amsterdam/?path=/docs/components-forms-switch--docs) for the Switch story which uses the bad practice. Addressing that in this PR was deemed out-of-scope.


packages/react/src/Pagination/Pagination.tsx
```diff
@@ -124,30 +124,21 @@ export const Pagination = forwardRef(
     }
 
     return (
-      <nav {...restProps} className={clsx('ams-pagination', className)} ref={ref} aria-label="Paginering">
+      <nav {...restProps} className={clsx('ams-pagination', className)} ref={ref}>
+        <span className="ams-visually-hidden">Paginering</span>
```
dlnr, line 128: Lets not hardcode dutch in here. There has been an issue from nlds asking us to your props instead of hardcoded strings.

storybook/src/components/Switch/Switch.docs.mdx
```diff
@@ -9,7 +9,7 @@ import README from "../../../../packages/css/src/components/switch/README.md?raw
 <Markdown>{README}</Markdown>
 
 It can be used as a standalone component or as a form control element.
-When used as a standalone component provide an `aria-label` or `aria-labelledby` prop to keep it accessible.
+When used as a standalone component provide an `aria-labelledby` property or a descriptive label in an `ams-visually-hidden` element to keep it accessible.
```
dlnr, line 12: This is kind of confusing, normally the "aria-labelledby" with refer to an ID of something, but the label of the input is unable to receive an id. So labelledby something other then the label??

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -124,30 +124,21 @@ export const Pagination = forwardRef(
     }
 
     return (
-      <nav {...restProps} className={clsx('ams-pagination', className)} ref={ref} aria-label="Paginering">
+      <nav {...restProps} className={clsx('ams-pagination', className)} ref={ref}>
```
dlnr, line 127: You would still be able to supply an aria-labbeledby attribute, which is ok? https://web.dev/articles/website-navigation#add_a_landmark

packages/react/src/Pagination/Pagination.tsx
VincentSmedinga: See #1220

storybook/src/components/Switch/Switch.docs.mdx
RubenSibon: I'm not sure I follow. Can you make a suggestion?

packages/react/src/Pagination/Pagination.tsx
RubenSibon: People can still do that via the `{...restProps}`, can't they?

packages/react/src/Pagination/Pagination.tsx
RubenSibon: Adjusted.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -124,30 +127,21 @@ export const Pagination = forwardRef(
...
+        <span className="ams-visually-hidden">{ariaLabel}</span>
```
alimpens, line 131: This currently isn't read by screen readers. I think you need to connect it to the `nav` with `aria-labelledby`

storybook/src/components/Switch/Switch.docs.mdx
dlnr: This involves the Switch component and I confused by the use of "label" which is in there but does not describe anything.
```jsx
<div className={clsx('ams-switch', className)}>
        <input {...restProps} className="ams-switch__input" id={inputId} ref={ref} type="checkbox" role="switch" />
        <label className="ams-switch__label" htmlFor={inputId}></label>
      </div>
```

dlnr: So I guess another Form Label component you can use to describe the switch is fine.

packages/react/src/Pagination/Pagination.tsx
alimpens: [Yep](https://www.w3.org/TR/WCAG20-TECHS/H97.html#H97-ex2)

storybook/src/components/Switch/Switch.docs.mdx
dlnr: Multiple `for` attributes seems wrong though.

```jsx
<Field>
<Label htmlFor="switch-with-label">
    Label
  </Label>
  <Switch
    id="switch-with-label"
  />
</Field>
```

Where the switch has an empty label with the same htmlFor attribute.

dlnr: Should it be?

```jsx
<Field>
<Label id="switch-label">
    Label
  </Label>
  <Switch
    aria-labelledby="switch-label"
  />
</Field>
```

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -10,6 +10,8 @@ import type { ForwardedRef, HTMLAttributes } from 'react'
 import { Icon } from '../Icon/Icon'
 
 export type PaginationProps = {
```
alimpens, line 12: The changes in this PR warrant a rethink of the Pagination api I think. Terms like `ariaLabel` aren't correct anymore. We discussed this in #ads-frontend, what do you think of the following api + descriptions?

```
export type PaginationProps = {
  /** The accessible name for the Pagination component. */
  visuallyHiddenLabel?: string
  /** The maximum amount of pages shown. Minimum value: 5. */
  maxVisiblePages?: number
  /** The accessible name for next page-button. */
  nextVisuallyHiddenLabel?: string
  /** The visible label for the next page-button. */
  nextLabel?: string
  /** A function to run when the page number changes. */
  // eslint-disable-next-line no-unused-vars
  onPageChange?: (page: number) => void
  /** The current page number. */
  page?: number
  /** The accessible name for the previous page-button. */
  previousVisuallyHiddenLabel?: string
  /** The visible label for the previous page-button. */
  previousLabel?: string
  /** The total amount of pages. */
  totalPages: number
} & HTMLAttributes<HTMLElement>
```

storybook/src/components/Switch/Switch.docs.mdx
alimpens: Maybe we should revert the changes to the Switch docs and focus on Pagination for this PR. That's a lot of work already.

packages/react/src/Pagination/Pagination.tsx
RubenSibon: Yeah, I think we should change the API. Can it be in-scope for this PR though? Maybe better to do it right away?

packages/react/src/Pagination/Pagination.tsx
RubenSibon: `aria-labelledby` has been added. Confirmed that NVDA now reads the label.

storybook/src/components/Switch/Switch.docs.mdx
RubenSibon: Made a ticket for [(aria-)label in Switch](https://gemeente-amsterdam.atlassian.net/browse/DES-924)

packages/react/src/Pagination/Pagination.tsx
RubenSibon: Changed

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -5,28 +5,30 @@
...
   /** The maximum amount of pages shown. Minimum value: 5. */
   maxVisiblePages?: number
-  /** The accessible name for the link to the next page. */
-  nextAriaLabel?: string
-  /** The label for the link to the next page. */
+  /** The visible label for the next page-button. */
   nextLabel?: string
+  /** The accessible name for next page-button. */
```
alimpens, line 17:
```suggestion
  /** The accessible name for the next page-button. */
```

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -118,36 +121,32 @@ export const Pagination = forwardRef(
       onChangePage(currentPage + 1)
     }
 
+    const id = useId()
+    const navLabelId = `nav-label-${id}`
```
alimpens, line 125: We don't need this, right? `useId` already ensures the id is unique?

RubenSibon: Adjusted

alimpens (approved):

dlnr (approved):

PR #1386, opened 2024-07-12
chore: Organize ‘Docs’ section

VincentSmedinga:
# Describe the pull request

## What

Improves the structure of the general documentation.

## Why

Several perspectives had become mixed up. I think we have 3 subjects here:
- a description of our design choices (and later principles)
- guidelines on things to consider when implementing our components
- presentation of our branding / foundation / design tokens

## How

1. Updated the `title` prop of `Meta` for these docs.
2. Renamed a few files to match their new title
3. Improved some texts I came across

## Checklist

Before submitting your pull request, please ensure you have done the following. Check each checkmark if you have done so or if it wasn't necessary:

- [x] Add or update unit tests
- [x] Add or update documentation
- [x] Add or update stories
- [x] Add or update exports in index.\* files
- [x] Start the PR title with a Conventional Commit prefix, [as explained here](https://github.com/Amsterdam/design-system/blob/main/documentation/publishing.md?plain=1#L11).

## Additional notes

- The sorting of the stories seems off. Some of my instructions work, some don’t. Especially the content order of the developer guide is strange – although I did rename ‘breakpoints’ to ‘responsive design’ to ensure ‘Getting started’ comes first.
- Only review the reordering please, not the content or any pages that might be missing.

PR #1387, opened 2024-07-15
feat!: Remove HeadingLevel type, group Heading sizes in story

VincentSmedinga:
- Having all semantic headings in one example makes it easier to visually compare them
- We didn’t have an example of sizes yet; now we do
- I could use this to assess the new text sizes for compact mode


dlnr (approved):

packages/react/src/Heading/index.ts
```diff
@@ -1,2 +1,2 @@
-export { Heading } from './Heading'
+export { Heading, headingLevels, headingSizes } from './Heading'
 export type { HeadingLevel, HeadingProps } from './Heading'
```
alimpens, line 2: User can get to `HeadingLevel` with `HeadingProps["level"]`. Do we want to export them separately?

storybook/src/components/Heading/Heading.stories.tsx
```diff
@@ -32,24 +32,24 @@ type Story = StoryObj<typeof meta>
...
-export const Heading1: Story = {}
-
-export const Heading2: Story = {
-  args: {
-    level: 2,
-  },
+export const Levels: Story = {
+  render: (args) => (
+    <Column gap="small">
+      {headingLevels.map((level) => (
```
alimpens, line 38: I'm not sure I'm a fan of this approach. It makes the code less copy-able for our users.

storybook/src/components/Heading/Heading.docs.mdx
```diff
@@ -14,21 +14,22 @@ import README from "../../../../packages/css/src/components/heading/README.md?ra
...
 ## Examples
 
-### Heading 1
+### Levels
 
-<Canvas of={HeadingStories.Heading1} />
+Four levels of headings are available.
+A page should contain only one Heading with level 1.
+Do not skip heading levels.
+User agents can use heading information to construct a table of contents for a document.
```
alimpens, line 22: Shouldn't this be in the guidelines section? (Most of it is already)

packages/react/src/Heading/Heading.tsx
```diff
@@ -9,8 +9,11 @@ import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
 import { getHeadingElement } from './getHeadingElement'
 
-export type HeadingLevel = 1 | 2 | 3 | 4
-type HeadingSize = 'level-1' | 'level-2' | 'level-3' | 'level-4' | 'level-5' | 'level-6'
+export const headingLevels: Array<number> = [1, 2, 3, 4]
+export const headingSizes: Array<string> = ['level-1', 'level-2', 'level-3', 'level-4', 'level-5', 'level-6']
+
+export type HeadingLevel = (typeof headingLevels)[number]
+export type HeadingSize = (typeof headingSizes)[number]
```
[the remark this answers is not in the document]
VincentSmedinga: Removed the entire type.

packages/react/src/Heading/index.ts
```diff
@@ -1,2 +1,2 @@
-export { Heading } from './Heading'
+export { Heading, headingLevels, headingSizes } from './Heading'
```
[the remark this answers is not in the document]
VincentSmedinga: No, that was not the intention. Also, these separate types are now gone anyway.

packages/react/src/Heading/index.ts
VincentSmedinga: I like how it makes a ‘heading level’ a thing of its own. And using `HeadingProps` imports the entire type object instead of one property, although I don’t know how bad that is.

Slightly out of scope, but I’ve removed the separate type now. That does make this a breaking change; if that’s a reason to leave it in we can revert the commit.

storybook/src/components/Heading/Heading.docs.mdx
VincentSmedinga: Yes, moved up. And rewritten most of the guidelines.

storybook/src/components/Heading/Heading.stories.tsx
VincentSmedinga: Alright.

packages/css/src/components/heading/README.md
```diff
@@ -2,15 +2,14 @@
 
 # Heading
 
-A heading conveys information about the content below it.
-Use headings to allow the user to grasp the structure of the page quickly.
+Communicates the organization of the content on a page.
```
alimpens, line 5: I don't think this is an improvement, it's a bit vague. Why do you want to change it?

packages/css/src/components/heading/README.md
```diff
@@ -2,15 +2,14 @@
...
 
 ## Guidelines
 
-- A heading describes the content beneath it.
-- Do not use a heading for formatting and styling. Use it to represent the page’s structure.
-- Use headings hierarchically, and do not skip heading levels.
-  So, an `h1` heading should be followed by an `h2`, not an `h3`.
+- Every page should contain one Heading with level 1.
+- Do not skip heading levels, e.g. a level 2 Heading should be followed by one with level 3, not level 4.
+- Do not use a Heading for formatting and styling, but represent the page’s structure.
```
alimpens, line 11:
```suggestion
- Do not use a Heading for formatting and styling, but to represent the page’s structure.
```

packages/css/src/components/heading/README.md
```diff
@@ -2,15 +2,14 @@
...
+  User agents may use heading information to generate a table of contents for a document.
```
alimpens, line 12: Headings serve a more important purpose imo, which is to let users of screen readers navigate the hierarchy. Generating a table of contents feels like a somewhat niche use?

packages/css/src/components/heading/README.md
```diff
@@ -2,15 +2,14 @@
 
 # Heading
 
-A heading conveys information about the content below it.
-Use headings to allow the user to grasp the structure of the page quickly.
+Communicates the organization of the content on a page.
 
 ## Guidelines
 
-- A heading describes the content beneath it.
```
alimpens, line 10: Without this item, we don't really describe what a Heading actually does. I'd leave it in.

storybook/src/components/Heading/Heading.docs.mdx
```diff
@@ -14,21 +14,34 @@ import README from "../../../../packages/css/src/components/heading/README.md?ra
...
+<Canvas of={HeadingStories.Level3} />
+
+#### Level 4
+
+<Canvas of={HeadingStories.Level4} />
+
+### Sizes
+
+It may be visually more suitable to display a heading in a smaller font size.
+Six sizes from the set of [Text Levels](?path=/docs/docs-design-guidelines-typography--docs) can be used for this purpose.
```
alimpens, line 41: This url will change following [this PR](https://github.com/Amsterdam/design-system/pull/1386)

packages/css/src/components/heading/README.md
VincentSmedinga: I borrowed it from the [W3C WAI](https://www.w3.org/WAI/tutorials/page-structure/headings/). I think the rationale should not include usage information and should not start with the component name. I intend to update the rationale for all components like this and thought I could take this one along already.

I’ve restored the “A heading describes the content beneath it.” so we won’t miss that as well?

packages/css/src/components/heading/README.md
VincentSmedinga: Yes. updated.

packages/css/src/components/heading/README.md
alimpens: It's still really vague imo. We could just do this, and remove the first guideline:

```suggestion
Describes the content that follows it.
Use Headings to allow the user to grasp the structure of the page quickly.
```

storybook/src/components/Heading/Heading.docs.mdx
VincentSmedinga: Updated.

PR #1390, opened 2024-07-15
chore: Visualise grid lines through a background gradient

VincentSmedinga:
This shows light grey grid lines on the Grid docs page to indicate the columns. It separates the concepts of columns and cells – cells are still pink. Examples are full screen which adds to the experience as well.

It uses a gradient background based on the grid variables. The lines are a bit off in Safari, unfortunately, but in Chrome and Firefox they appear correctly. One comment for it on the code though.

I got the idea while working on the Spotlight with break-out content, for which a separate PR will follow.

~~About the class name and file used: I intend to merge all documentation classes under one namespace, either `.ams-docs-*` or `.ams-storybook-*` and move them into one file shortly.~~


storybook/config/preview-body.html
```diff
@@ -59,4 +59,42 @@
...
+      90deg,
+      var(--column-color) 0%,
+      var(--column-color) var(--column-width),
+      transparent var(--column-width),
+      transparent calc(var(--column-width) + var(--gap-width))
+    );
+    content: "";
+    inset-block: 0;
+    inset-inline: var(--margin-width);
+    position: absolute;
```
VincentSmedinga, line 98: I resorted to this to ensure the background image doesn’t overlap the horizontal grid padding – `background-position` doesn’t work. But it hampers inspecting the element through dev tools, so a better solution is welcome.

storybook/src/components/Grid/Grid.stories.tsx
```diff
@@ -66,7 +69,7 @@ type CellStory = StoryObj<typeof cellMeta>
 const StoryTemplate: Story = {
   decorators: [
     (Story) => (
-      <Screen>
+      <Screen className="ams-storybook-show-grid">
```
VincentSmedinga, line 72: I’d prefer to apply the class name to `<Story />` – but that doesn’t get applied.

Making it an `arg` in `meta` applies it to Grid Cells as well, resulting in an incorrect background pattern.

dlnr (dismissed):

storybook/config/preview-body.html
VincentSmedinga: (This is now in `docs.css`.)

RubenSibon (approved):

VincentSmedinga:
Superseded by #1605.

PR #1391, opened 2024-07-15
feat!: Remove Visually Hidden component

VincentSmedinga:
We’ve started using this as a utility class instead.


storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -3,25 +3,16 @@
...
 import { Meta, StoryObj } from '@storybook/react'
+import type { HTMLAttributes } from 'react'
+
+type VisuallyHiddenProps = HTMLAttributes<HTMLSpanElement>
+
+const VisuallyHidden = ({ children }: VisuallyHiddenProps) => <span className="ams-visually-hidden">{children}</span>
 
 const meta = {
-  title: 'Components/Containers/Visually Hidden',
+  title: 'Utilities/CSS/Visually Hidden',
```
alimpens, line 14: Do we expect other utilities than utility classes? If not, I'd just call this section 'Utility classes'

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -3,25 +3,16 @@
  * Copyright Gemeente Amsterdam
  */
 
-import { Paragraph, VisuallyHidden } from '@amsterdam/design-system-react/src'
 import { Meta, StoryObj } from '@storybook/react'
+import type { HTMLAttributes } from 'react'
+
+type VisuallyHiddenProps = HTMLAttributes<HTMLSpanElement>
+
+const VisuallyHidden = ({ children }: VisuallyHiddenProps) => <span className="ams-visually-hidden">{children}</span>
```
alimpens, line 11: The code view doesn't render well this way.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
VincentSmedinga: Yes, I expect we can offer JS formatting functions for the way Amsterdam wants to display numbers, dates etc.

And removing this level will promote the utility classes to a very visible spot that they don’t really deserve.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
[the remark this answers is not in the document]
VincentSmedinga: Discussed that this is the good approach.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
VincentSmedinga: It does now, through the `render` prop of a Story.

storybook/src/utils/VisuallyHidden/VisuallyHidden.docs.mdx
```diff
@@ -0,0 +1,8 @@
+import { Controls, Markdown, Primary } from "@storybook/blocks";
+import README from "../../../../packages/css/src/components/visually-hidden/README.md?raw";
+
+<Markdown>{README}</Markdown>
```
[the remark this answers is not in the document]
VincentSmedinga: It does now; probably because of the missing `Meta` component.

packages/react/documentation/coding-conventions.md
```diff
@@ -36,7 +36,7 @@ Subcomponents (e.g. `Grid.Cell`) are kept in separate files (e.g. `GridCell.tsx`
 
 ## Text for screen readers only
 
-Use [Visually Hidden](https://designsystem.amsterdam/?path=/docs/components-containers-visually-hidden--docs) or the `ams-visually-hidden` utility class to hide content from sighted users but keep it accessible to non-visual user agents, such as screen readers.
+Use the `ams-visually-hidden` [utility class](http://designsystem.amsterdam/?path=/docs/utilities-css-visually-hidden--docs) to hide content from sighted users but keep it accessible to non-visual user agents, such as screen readers.
 
 Do not use `aria-label`. Tools for automatic translation often [do not translate its value](https://adrianroselli.com/2019/11/aria-label-does-not-translate.html), and it may get overlooked when doing manual translation.
```
dlnr, line 41: You could mention that `aria-labbeledby` is recommended to connect visually hidden tekst to a certain element.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
dlnr: I'd put this under docs too, it is confusing because it's not a component but does have an html code example.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
dlnr: If you are going to create an example why not use the react components and add `className='ams-visually-hidden'` to the Paragraph component. 

And render is not needed to achieve this, you could use `parameters.docs.source.code` to add a static code example.

VincentSmedinga: The public React component gets deleted in this PR. The stories file has a local component instead because a Story requires it.

I updated the example to use React `Paragraph` components, even though these CSS components are usable outside of React as well, e.g. in our future libraries. But for now, mixin React here seems fine.

The render function seems a more realisatic proof of how this class name works than a fragment of static code, or am I missing something? Can you propose an example?

packages/react/documentation/coding-conventions.md
dlnr: I was confusing this with [another comment](https://github.com/Amsterdam/design-system/pull/1366#discussion_r1684243095) where aria-label is replaced by a visually hidden field which is not automatically linked to what you're labelling.

VincentSmedinga: So this can be resolved?

dlnr (approved):

PR #1397, opened 2024-07-15
feat: Add Breakout component that lets media content stick out of a Spotlight

VincentSmedinga:
**Update** (2024/09/13): We decided against making Grid more complex (for now). We copied Grid, including the new row spanning and gap covering features, to a separate Breakout component, then reverted the changes in Grid, maintaining the approach outlined below.

---

tl;dr:

1. This uses a 2 x 2 Grid
2. The Spotlight is separated from the text
3. Grid Cells can now span rows
4. The example still requires two inline styles
5. For the rest, it works! :tada:

---

I had this lingering idea that the feature set of CSS Grid should be rich enough to meet (most of) our ‘Spotlight with break-out content challenge’. We already use stacking grid cells in `Overlap`, like Niels I thought grid rows should play a role, and we need some box alignment features.

With our current approach of a Grid in a Spotlight, we could never let something on the grid pop out of the Spotlight, as it contains the Grid and would grow with it.

So I turned it upside down and put the Spotlight in a Grid. The idea was to have a grid of two columns and two rows, with the Spotlight in the bottom two cells and the image in the two at the right, using overlapping grid cells as mentioned. The realisation then was that it would obfuscate half of the text.

My solution was to separate the Spotlight from the text. It may be a bit controversial, but there’s no way both will become misaligned and the text unreadable because they use the bottom left cell. The empty Spotlight is just a coloured div, and the component API still ensures the correct colours are used. It does need a negative margin to overlap the horizontal and vertical whitespace, but we can reuse the design tokens here and use the `padding` props to make it work with content above and below.

The rows' height is perfect by default—their `auto` size makes the Spotlight exactly as high as the text, even while they are separated in the DOM, and the image takes the rest of the height. 

I’ve extended the existing API with rows. This makes it easy to stack the parts vertically instead of horizontally on narrow screens.

The approach works with any of the aspect ratios, any content even. It is a bit of a puzzle to make this example look good on various widhts – text an images scale totally differently. I had to give up the 50/50 layout on the 8-column grid; check out the commit before ‘Stack image and text vertically in medium grid as well’ to see why. And I wrapped all Spotlight stories in a Screen because the sticking out broke spectacularly on my ultra-wide monitor. (Which might be a good idea for all stories; we can combine it with the language attribute.)

The increased size of the CSS for the Grid component is a downside. We should shorten these class names. An idea is to make them utility classes with abbreviations like we do for ‘margin bottom’. As a bonus, we can put this class name directly on the paragraph or image without needing an intermediary Grid Cell div.

There are still a few things funky in this approach, see my own comments, but I think the aspects that work well – row height, grid placement, responsiveness, general applicability, much automatic stuff – outweigh them. There’s a few more things we could hide into an API.

The example is called ‘Stick out’ because ‘Break out’ seems to be used for sections stretching to the window's edge while other content gets horizontally constrained. ‘Pop out’ could be an alternative, too. I’ve checked with Thomas, and he’s on the same track :)


storybook/src/components/Spotlight/Spotlight.stories.tsx
```diff
@@ -29,6 +36,38 @@ type Story = StoryObj<typeof meta>
...
+  render: ({ as, color }) => (
+    <Grid>
+      <Grid.Cell
+        span="all"
+        rowSpan={{ narrow: 2, medium: 2, wide: 1 }}
+        rowStart={2}
+        style={{
+          marginBlock: 'calc(var(--ams-space-grid-md) * -1)',
+          marginInline: 'calc(var(--ams-space-grid-lg) * -1)',
+        }}
```
VincentSmedinga, line 49: This could become something like an `overlayWhitespace` Boolean prop, although we need a bit more information to correctly select medium and large space lengths.

storybook/src/components/Spotlight/Spotlight.stories.tsx
```diff
@@ -29,6 +36,38 @@ type Story = StoryObj<typeof meta>
...
+      >
+        <Spotlight as={as} color={color} style={{ height: '100%' }} />
```
VincentSmedinga, line 51: We can remove the inline style if we allow the grid cell classes and props on any element – that would make the Spotlight a direct child of the Grid, thus a grid item, stretching by default.

storybook/src/components/Spotlight/Spotlight.stories.tsx
```diff
@@ -29,6 +36,38 @@ type Story = StoryObj<typeof meta>
...
+      </Grid.Cell>
+      <Grid.Cell rowStart={{ narrow: 3, medium: 3, wide: 2 }} span={{ narrow: 4, medium: 8, wide: 6 }} start={1}>
```
VincentSmedinga, line 53: This `start={1}` is often optional, but required here to place the 3 items correctly.

storybook/src/components/Spotlight/Spotlight.stories.tsx
```diff
@@ -29,6 +36,38 @@ type Story = StoryObj<typeof meta>
...
+        <Paragraph inverseColor={!color || !['green', 'yellow'].includes(color)}>{exampleParagraph()}</Paragraph>
+      </Grid.Cell>
+      <Grid.Cell
+        rowSpan={2}
+        rowStart={1}
+        span={{ narrow: 4, medium: 8, wide: 6 }}
+        start={{ narrow: 1, medium: 1, wide: 7 }}
+        style={{ alignSelf: 'end' }}
```
VincentSmedinga, line 61: Should become an alignment prop or utility class (which relates it to #1330).

packages/react/src/Grid/gridCellClasses.ts
```diff
@@ -5,7 +5,12 @@
 
 import type { GridCellProps } from './GridCell'
 
-export const gridCellClasses = (span?: GridCellProps['span'], start?: GridCellProps['start']): string[] => {
+export const gridCellClasses = (
```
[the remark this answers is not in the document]
dlnr: I've noticed (in my PR) that the addition or rows makes the classes way to long:
```
ams-grid__cell ams-grid__cell--colspan-4 ams-grid__cell--colspan-4-medium ams-grid__cell--colspan-5-wide ams-grid__cell--start-1 ams-grid__cell--start-1-medium ams-grid__cell--start-2-wide ams-grid__cell--rowspan-1 ams-grid__cell--rowspan-1-medium ams-grid__cell--rowspan-1-wide ams-grid__cell--row-2 ams-grid__cell--row-2-medium ams-grid__cell--row-2-wide
```
So perhaps an abbreviation is recommendable?

dlnr (commented):
This makes me wonder if grid should be a design system component instead of some guidelines. It's getting overcomplicated and solving this in application specific CSS is going to be a lot lighter/faster.

Just a couple of remarks, not a real review.

packages/css/src/components/grid/grid.scss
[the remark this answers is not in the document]
VincentSmedinga: I would say spanning columns is the default and can remain without a prefix for brevity. Let’s revisit when we try to shorten these props and class names.

packages/react/src/Grid/gridCellClasses.ts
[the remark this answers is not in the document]
VincentSmedinga: Yes, this is getting out of hand, but let’s assess renaming as a separate task.

storybook/config/preview-body.html
```diff
@@ -59,4 +59,42 @@
     margin: -1rem; /* stylelint-disable-line */
     padding: 1rem; /* stylelint-disable-line */
   }
+
```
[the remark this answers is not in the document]
VincentSmedinga: Removed from this branch now.

VincentSmedinga:
Discussed and decided to extract the extra complexity from Grid to a separate Breakout component, e.g.

```jsx
<Breakout>
  <Breakout.Cell colSpan="all" rowSpan={2}>
    <Map />
  </Breakout.Cell>
  <Breakout.Cell coverGap rowStart={2} rowSpan={2}>
    <Spotlight />
  </Breakout.Cell>
  <Breakout.Cell colStart={2} colSpan={8} rowStart={3}>
    <Heading />
    <Paragraph />
  </Breakout.Cell>
</Breakout>
```

packages/react/src/Grid/Grid.tsx
```diff
@@ -7,16 +7,11 @@ import clsx from 'clsx'
...
 export type GridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12
-
-export type GridColumnNumbers = {
-  narrow: GridColumnNumber
-  medium: GridColumnNumber
-  wide: GridColumnNumber
-}
-
 export type GridPaddingSize = 'small' | 'medium' | 'large'
+export type ResponsiveGridValues<T> = T | { narrow: T; medium: T; wide: T }
```
alimpens, line 14: This creates a type with the same column count for all sizes, no? I don't think that's what we want.

packages/react/src/Grid/index.ts
```diff
@@ -1,3 +1,3 @@
 export { Grid } from './Grid'
-export type { GridProps, GridColumnNumber, GridColumnNumbers } from './Grid'
+export type { GridProps, GridColumnNumber, ResponsiveGridValues } from './Grid'
```
alimpens, line 2: Same here, do we want to export all these types to our users?

storybook/src/styles/docs.css
```diff
@@ -95,10 +96,16 @@
   font-family: "Amsterdam Sans", sans-serif;
   font-size: var(--ams-paragraph-small-font-size);
   line-height: var(--ams-paragraph-small-line-height);
-  margin-block: 0;
-  margin-inline: 0;
   padding-block: 1.5rem;
   text-align: center;
+
+  p& {
```
alimpens, line 102: Do we use `ams-docs-item` on a paragraph anywhere?

storybook/src/components/Breakout/Breakout.docs.mdx
```diff
@@ -0,0 +1,41 @@
...
+The text next to it defines the height of the Spotlight.
+On narrower screens, the text moves below the image.
+Ensure that the text is short, or the image tall, to prevent losing the break-out effect.
+
+This component is a copy of `Grid`, so it offers the same props.
+Additionally, `Breakout` allows cells to span multiple columns, and a cell to expand to cover its adjacent gaps and margins.
+
+_The example isn’t very wide on this page.
+And due to how Storybook works, the Grid in the example may incorrectly use 12 columns.
+Both circumstances prevent the image from breaking out if the random example text is long.
```
alimpens, line 21: Isn't this a good reason to not use random text here? The example usually doesn't look the way we want it to, which is a bit of a shame.

storybook/src/components/Breakout/Breakout.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <Screen>
+        <Story />
+      </Screen>
+    ),
+  ],
+} satisfies Meta<typeof Breakout>
+
+export default meta
+
+const cellMeta = {
```
alimpens, line 26: These controls do not seem to be working. `colStart`, `rowSpan` and `rowStart` aren't shown, and only `colSpan` `all` has the desired effect, using a number there doesn't work.

storybook/src/components/Breakout/Breakout.docs.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+This example lets an image pop out.
+The text next to it defines the height of the Spotlight.
+On narrower screens, the text moves below the image.
+Ensure that the text is short, or the image tall, to prevent losing the break-out effect.
+
+This component is a copy of `Grid`, so it offers the same props.
+Additionally, `Breakout` allows cells to span multiple columns, and a cell to expand to cover its adjacent gaps and margins.
+
+_The example isn’t very wide on this page.
```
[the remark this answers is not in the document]
VincentSmedinga: We inherited this guideline from Fabrique, but I don’t think we have actively adopted it yet. You’re right that it’s public now.

I have no problem reading this text in our font and at this size. I like the tone it conveys. But I’ve shortened the example text anyway, so I removed this note as well.

storybook/src/components/Breakout/Breakout.docs.mdx
VincentSmedinga: Yes, it is.

packages/css/src/components/grid/_mixins.scss
```diff
@@ -0,0 +1,110 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@mixin ams-grid {
```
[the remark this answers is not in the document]
alimpens: > For what reason?

Ease of use mostly. Currently, I can't see at a glance what `.ams-grid` does, I have to go to another file and connect the CSS block with the right class in my head. I quite often browse through the Utrecht CSS and this gets a little annoying.

On the other hand, having both in the same files makes that file longer, so that might decrease the readability.

> It also allows consumers to import the mixins without the class names.

This also works when both are in the same file, no? You can import mixins without importing the built CSS?

storybook/src/components/Spotlight/Spotlight.stories.tsx
VincentSmedinga: We’ve merged this with `coverGap` into `has="spotlight"`.

storybook/src/components/Spotlight/Spotlight.stories.tsx
VincentSmedinga: We’ve created `has="figure"` for this.

PR #1389, opened 2024-07-15
feat: Remove fluidity from text and spacing in Compact Mode

VincentSmedinga:
As discussed, having fluid type in Compact Mode makes less sense now that regular text only grows 2 pixels between window widths of 320 and 1600, which comes down to 0.15%. This PR fixates Text Level 5 to 16 pixels.

This required the sizes to grow a bit slower on narrow screens. I’ve tested it with Amopis, looks good to me.

Because space lengths are based on Text Level 5, this also is no longer fluid in Compact Mode.

PR #1418, opened 2024-07-18
chore: Fix documentated values for Compact typography

VincentSmedinga:
I still made a copy paste error in #1389, apparently.


dlnr (approved):
At this point I'll just agree to move this change along even if the compact theme should not use fluid typography for the sake of using fluid typography.

PR #1422, opened 2024-07-19
feat!: Apply default gap in Row and Column without extra class

VincentSmedinga:
# Describe the pull request

## What

Removes the `ams-row|column--gap-medium` class name and applies the default `gap` of `Row` and `Column` in the component class name.

Adds an extra option of ‘none’ to the `gap` property to allow overriding this default if necessary.

## Why

Both `Row` and `Column` apply whitespace between their children through the default `medium` value of the `gap` prop. Laying out children is their main task, and preventing children to bump into each other is a logical default.

The current setup added an extra `ams-row|column--gap-medium` class name to do so, but a true default gets applied by the component classname alone.

## How

- Removed `md` from the set of `ams-sizes` that generate class names
- Removed the default value of `medium` from the `gap` prop of both components
- Added `no` to the set of `ams-sizes` that generate class names

## Checklist

- [x] Add or update stories

## Additional notes


packages/react/src/Column/Column.tsx
```diff
@@ -20,7 +20,7 @@ export type ColumnProps = {
 } & PropsWithChildren<HTMLAttributes<HTMLElement>>
 
 export const Column = forwardRef(
-  ({ as: Tag = 'div', children, className, gap = 'medium', ...restProps }: ColumnProps, ref: any) => (
+  ({ as: Tag = 'div', children, className, gap, ...restProps }: ColumnProps, ref: any) => (
     <Tag {...restProps} ref={ref} className={clsx('ams-column', `ams-column--gap-${gap}`, className)}>
```
alimpens, line 24: This renders `ams-column--gap-undefined` if you don't set `gap`

VincentSmedinga: Ah, yes. Added the condition.

packages/css/src/components/column/column.scss
dlnr, line 1: This is kind of a break from our previous implementations with a gap or padding (see Grid) and I would expect the default to have no gap and no additional class. If you do need medium to be the default it should have the medium gap modifier class. And then the "no" option (which is also odd) should not print out a classname.

packages/react/src/Column/Column.test.tsx
```diff
@@ -19,7 +19,7 @@ describe('Column', () => {
     const component = container.querySelector(':only-child')
 
     expect(component).toHaveClass('ams-column')
-    expect(component).toHaveClass('ams-column--gap-medium')
+    expect(component).not.toHaveClass('ams-column--gap-medium')
```
dlnr, line 22: So to continue my rant. This should be true if medium is the default.

packages/react/src/Column/Column.tsx
```diff
@@ -20,8 +20,8 @@ export type ColumnProps = {
 } & PropsWithChildren<HTMLAttributes<HTMLElement>>
 
 export const Column = forwardRef(
-  ({ as: Tag = 'div', children, className, gap = 'medium', ...restProps }: ColumnProps, ref: any) => (
-    <Tag {...restProps} ref={ref} className={clsx('ams-column', `ams-column--gap-${gap}`, className)}>
+  ({ as: Tag = 'div', children, className, gap, ...restProps }: ColumnProps, ref: any) => (
+    <Tag {...restProps} ref={ref} className={clsx('ams-column', gap && `ams-column--gap-${gap}`, className)}>
```
dlnr, line 24: This is correct and the "no" gap should return false here.

dlnr (commented):
I know you guys talked about this and I was busy helping crying children (...) but I think this the default should not be medium. And certainly not in this way.

packages/css/src/components/column/column.scss
alimpens: > This is kind of a break from our previous implementations with a gap or padding (see Grid)

That's not really true though. The Grid renders a gap without a class by default.

alimpens (dismissed):

packages/react/src/Column/Column.test.tsx
VincentSmedinga: No, this class name no longer exists.

The test is not really necessary to be honest – before today it would have been `ams-column--medium`. Should I update or remove the test?

packages/react/src/Column/Column.tsx
VincentSmedinga: Only if we don’t merge this PR. The idea is that `Row` and `Column`, just like `Grid` render a gap between their children by default.

We add the ‘none’ option to `Row` and `Column` to support (edge) cases where that is unwanted. We could even wait to add that option until people need it.

packages/css/src/components/column/column.scss
VincentSmedinga: > If you do need medium to be the default it should have the medium gap modifier class.

I don’t agree. It’s not a default if you need another class.

packages/react/src/Column/Column.test.tsx
alimpens: It is a little weird to test for something that will never exist. Let's remove it

VincentSmedinga: They’re gone.

alimpens (approved):

PR #1427, opened 2024-07-19
feat: Add documention for Gap and Margin utility classes

VincentSmedinga:
# Describe the pull request

## What

This documents the Gap and Margin Bottom utility classes.

## Why

So that people know when and how to use them – or even that they exist.

## How

It follows the approach of the recently added Visually Hidden docs.

I needed to restore the `md` value in `ams-sizes` again, because these gap and margin utilities do need a class name for this size, so I updated the `Row` and `Column` stylesheets to skip this value.

By the way, I think we should mark (non-trivial) documentation updates as features from now on. Documentation on how to use things is an important part of our design system and they deserve a place in the change log.


packages/css/src/common/size.scss
VincentSmedinga, line 1: The branch for this PR is on top of that for #1391. It must be rebased after that has been merged.

VincentSmedinga: Done.

storybook/src/components/Field/Field.stories.tsx
```diff
@@ -3,7 +3,7 @@
  * Copyright Gemeente Amsterdam
  */
 
-import { ErrorMessage, Label, TextInput } from '@amsterdam/design-system-react'
+import { ErrorMessage, Label, TextInput } from '@amsterdam/design-system-react/src'
```
alimpens, line 6: This isn't really necessary, only the controls for Field are generated, right? We can move the `Paragraph` import to the regular import as well.

storybook/src/utils/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -3,7 +3,7 @@
  * Copyright Gemeente Amsterdam
  */
 
-import { Paragraph } from '@amsterdam/design-system-react'
+import { Paragraph } from '@amsterdam/design-system-react/src'
```
alimpens, line 6: Same here, this is only necessary if we want to autogenerate controls.

packages/css/src/common/size.scss
```diff
@@ -3,11 +3,12 @@
  * Copyright Gemeente Amsterdam
```
alimpens, line 3: Copied from a PR comment: 

> By the way, I think we should mark (non-trivial) documentation updates as features from now on. Documentation on how to use things is an important part of our design system and they deserve a place in the change log.

That creates a kind of weird situation though. The documentation lives in Storybook, which is a package we don't publish. Marking it as a feature without changing something in the packages we do publish doesn't result in a new version being released. And vice versa, if we do release a new version of the packages we publish, it's because there's some other change in that package, not because of changes to the docs.

packages/css/src/components/margin/README.md
```diff
@@ -2,4 +2,14 @@
...
+Adds white space below a single element.
+
+The five sizes of [Component Space](/docs/foundation-design-tokens-space--docs) are available for the length of the margin.
+
+## Guidelines
+
+- Use this utility class to vertically separate one element from the next.
+- It can be used on any element and sets the `margin-block-end` CSS property. This declaration is marked with the `!important` flag to ensure it overrides any other margins.
+- Elements’ top and bottom margins are sometimes collapsed into a single margin. Consult [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Mastering_margin_collapsing) for details.
+- To add equal margins between elements, wrap them in a [Column](/docs/components-layout-column--docs) component instead.
```
alimpens, line 14: Do we want to reference the `gap` util class here as well?

storybook/src/utils/Margin/Margin.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+const lengthOptions = ['xs', 'sm', 'md', 'lg', 'xl'] as const
+
+type MarginProps = {
+  length: (typeof lengthOptions)[number]
+} & HTMLAttributes<HTMLElement>
+
+const Margin = ({ children, length }: MarginProps) => <span className={`ams-mb--${length}`}>{children}</span>
+
+const render = ({ length }: MarginProps) => (
+  <div>
```
alimpens, line 19: We can render an array here, so we don't have to wrap it in a `div`. Same goes for the visually hidden docs.

storybook/src/utils/Gap/Gap.docs.mdx
```diff
@@ -0,0 +1,13 @@
+{/* @license CC0-1.0 */}
+
+import { Controls, Markdown, Meta, Primary } from "@storybook/blocks";
+import * as GapStories from "./Gap.stories.tsx";
+import README from "../../../../packages/css/src/components/gap/README.md?raw";
+
+<Meta of={GapStories} />
+
+<Markdown>{README}</Markdown>
```
alimpens, line 9: Relevant for all css util stories: the thing I'd want most from the docs as a developer is a button to copy just the css class, without any of the other stuff. Can we add that?

packages/css/src/common/size.scss
VincentSmedinga: Fair point. Still I think it’s important to point people to new or updated guidelines. Let’s think of something else.

storybook/src/components/Field/Field.stories.tsx
VincentSmedinga: Ah yes, didn’t realise that distinction. Updated all components in one fell swoop.

storybook/src/utils/Gap/Gap.docs.mdx
VincentSmedinga: Hah, great minds think alike. I had some local commits on top of this branch for such a thing already :) Needs a bit more work though.

storybook/src/utils/Margin/Margin.stories.tsx
VincentSmedinga: Had tried that, but not correctly maybe. Works now.

storybook/src/utils/Gap/Gap.docs.mdx
VincentSmedinga: Added copyable class names to Gap, Margin, and Visually Hidden, along with visual previews of the tokens.

I chose to use divs and classnames and inline styles for them rather than components. Some of these docs are in a README which isn’t MDX.

storybook/src/utils/Margin/Margin.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  title: 'Utilities/CSS/Margin',
+  component: Margin,
+  args: {
+    length: 'xs',
+  },
+  argTypes: {
+    length: {
+      control: 'radio',
+      description: 'The amount of white space to add below the element.',
+      options: ['xs', 'sm', 'md', 'lg', 'xl'],
```
alimpens, line 33: Same here, we can reuse `lengthOptions`

packages/css/src/common/size.scss
alimpens: Robbert once mentioned creating a package for docs, that could be an interesting option. We'd have to decide on the format though.

VincentSmedinga: https://gemeente-amsterdam.atlassian.net/browse/DES-925

storybook/src/utils/Gap/Gap.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { Paragraph } from '@amsterdam/design-system-react'
+import { Meta, StoryObj } from '@storybook/react'
+import type { HTMLAttributes } from 'react'
+
+const lengthOptions = ['xs', 'sm', 'md', 'lg', 'xl'] as const
+
+type GapProps = {
+  length: (typeof lengthOptions)[number]
+} & HTMLAttributes<HTMLElement>
```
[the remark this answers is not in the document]
VincentSmedinga: I can't quickly find why it does. What do you want to do with this. Try to understand? Accept as a happy coincidence? Or explicitly add `PropsWithChildren`?

storybook/src/utils/Gap/Gap.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  title: 'Utilities/CSS/Gap',
+  component: Gap,
+  args: {
+    length: 'xs',
+  },
+  argTypes: {
+    length: {
+      control: 'radio',
+      description: 'The amount of vertical space between items.',
+      options: ['xs', 'sm', 'md', 'lg', 'xl'],
```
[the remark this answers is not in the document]
VincentSmedinga: Yes. I think I was trying to get the options to show up under `Description` in the Controlts table. But either way they don’t, it just says `string`. Been struggling with that in #1428 as well.

storybook/src/utils/Margin/Margin.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type MarginProps = {
+  length: (typeof lengthOptions)[number]
+} & HTMLAttributes<HTMLElement>
+
+const Margin = ({ children, length }: MarginProps) => <span className={`ams-mb--${length}`}>{children}</span>
+
+const render = ({ length }: MarginProps) => [
+  <h1 className={`ams-heading ams-heading--2 ams-mb--${length}`}>This heading has a bottom margin</h1>,
+  <Paragraph>It introduces white space between the heading and this paragraph below.</Paragraph>,
```
alimpens, line 20: Components in an array need a key, otherwise you get a warning.

storybook/src/utils/Gap/Gap.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import { HTMLAttributes, PropsWithChildren } from 'react'
+
+const gapLengthOptions = ['xs', 'sm', 'md', 'lg', 'xl'] as const
+type GapLength = (typeof gapLengthOptions)[number]
```
alimpens, line 9: Why not just specify the options here? We're not exporting `gapLengthOptions` or using it anywhere else.

storybook/src/utils/Gap/Gap.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export type GapProps = {
+  /** The amount of vertical space between items. */
+  length: GapLength
```
alimpens, line 13: Wait, why is this called `length`? And not `gap`?

storybook/src/utils/Margin/Margin.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import type { HTMLAttributes } from 'react'
+
+const marginLengthOptions = ['xs', 'sm', 'md', 'lg', 'xl'] as const
+type MarginLength = (typeof marginLengthOptions)[number]
```
alimpens, line 9: Same here

storybook/src/utils/Margin/Margin.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export type MarginProps = {
+  /** The amount of white space to add below the element. */
+  length: MarginLength
```
alimpens, line 13: Same here, `space` would make more sense, no?

storybook/src/utils/Margin/Margin.tsx
```diff
@@ -0,0 +1,17 @@
...
+} & HTMLAttributes<HTMLElement>
```
alimpens, line 14:
```suggestion
} & PropsWithChildren<HTMLAttributes<HTMLSpanElement>>
```

storybook/src/utils/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,13 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import type { HTMLAttributes } from 'react'
+
+export type VisuallyHiddenProps = HTMLAttributes<HTMLSpanElement>
```
alimpens, line 8:
```suggestion
export type VisuallyHiddenProps = PropsWithChildren<HTMLAttributes<HTMLSpanElement>>
```

storybook/src/utils/Gap/Gap.tsx
```diff
@@ -0,0 +1,17 @@
...
+} & PropsWithChildren<HTMLAttributes<HTMLElement>>
```
alimpens, line 14:
```suggestion
} & PropsWithChildren<HTMLAttributes<HTMLSpanElement>>
```

storybook/src/utils/Gap/Gap.tsx
VincentSmedinga: It’s the length (width or height) of the gap. Not the gap of the Gap :)

storybook/src/utils/Gap/Gap.stories.tsx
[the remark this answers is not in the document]
alimpens: We fixed this when we moved the components to a separate files

storybook/src/utils/Gap/Gap.stories.tsx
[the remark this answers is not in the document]
alimpens: We fixed this by moving the components to a separate file.

storybook/src/utils/Gap/Gap.tsx
alimpens: It still feels a little weird imo. How about 'size'?

VincentSmedinga: Not strongly opposed, but [length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) is a very common term in CSS, and correct here, right?

VincentSmedinga: Updated to `size`.

storybook/src/utils/Margin/Margin.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Meta, StoryObj } from '@storybook/react'
+import type { MarginProps } from './Margin'
+import { Margin } from './Margin'
+
+const render = ({ size }: MarginProps) => (
+  <>
+    <Heading className={`ams-mb--${size}`} level={2}>
+      This heading has a bottom margin
+    </Heading>
+    <Paragraph key={1}>It introduces white space between the heading and this paragraph below.</Paragraph>
```
alimpens, line 16:
```suggestion
    <Paragraph>It introduces white space between the heading and this paragraph below.</Paragraph>
```

alimpens (approved):

PR #1449, opened 2024-07-22
feat: Add initial Password Input component

VincentSmedinga:
# Describe the pull request

## What

A component to enter a password.

This is a bare MVP version. It does not include a button to show or hide the characters – there is no design for it yet; even more so, the need for it hasn’t been discussed yet. I would say the features we have in our other inputs are a good basis for this one. People can start using this component, get the correct styling, and feedback any requirements they may have. I also created a ticket for the feature to show the password.

## Why

It was missing.

## How

- Scaffolded all files for a new component through plop
- Used `DateInput` and `TextInput` as inspiration
- Copied docs and tests from [an earlier approach](https://github.com/Amsterdam/design-system/commit/c5b75e04fded8b9b4caf2d3403c49daa2bc4e08d)

## Checklist

- [x] Add or update stories

## Additional notes


packages/react/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,27 @@
...
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'aria-invalid' | 'type'>
+
+export const PasswordInput = forwardRef(
+  ({ className, invalid, ...restProps }: PasswordInputProps, ref: ForwardedRef<HTMLInputElement>) => (
+    <input
+      {...restProps}
+      aria-invalid={invalid || undefined}
+      className={clsx('ams-password-input', className)}
+      ref={ref}
+      type="password"
```
alimpens, line 22: Should we set `autocapitalize="none"`, `autocorrect="off"` and `spellcheck="false"` here as well? It's a pretty firm guideline, no?

packages/css/src/components/password-input/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Password Input
+
+Helps users enter a password.
+
+Use this component when the input requires sensitive information, like passwords or PINs.
+It ensures that the input is not readable by others who might be looking at the screen.
+The characters entered are hidden, represented by squares.
+
+Consider setting the following attributes:
```
alimpens, line 11: These are guidelines, no? Why the different format?

packages/css/src/components/password-input/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. Allow the user’s password manager to automatically fill the password through `autocomplete="current-password"`.
+   When asking for a new password, use `autocomplete="new-password"` instead.
+2. Add a `minlength` attribute to ensure passwords meet a minimum length requirement.
```
alimpens, line 15:
```suggestion
2. Add a `minlength` attribute to ensure passwords meet a minimum length requirement.
  Describe this in the [Field](?path=/docs/components-forms-field--docs)’s description as well.
```

storybook/src/components/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+      description: 'Prevents interaction. Avoid if possible.',
+    },
+  },
+} satisfies Meta<typeof PasswordInput>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
```
alimpens, line 28: Gov.uk advises against realtime validation. Maybe wrap the input in a `form` with `novalidate`?

packages/react/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,27 @@
...
+  invalid?: boolean
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'aria-invalid' | 'type'>
+
+export const PasswordInput = forwardRef(
+  ({ className, invalid, ...restProps }: PasswordInputProps, ref: ForwardedRef<HTMLInputElement>) => (
+    <input
+      {...restProps}
+      aria-invalid={invalid || undefined}
+      className={clsx('ams-password-input', className)}
+      ref={ref}
```
alimpens, line 21: Did you remove `dir={dir ?? 'auto'}` here on purpose? A password can be in a RTL language as well, no?

packages/react/src/PasswordInput/PasswordInput.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  it('supports ForwardRef in React', () => {
+    const ref = createRef<HTMLInputElement>()
+
+    const { container } = render(<PasswordInput ref={ref} />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(ref.current).toBe(component)
+  })
+
```
alimpens, line 42: Did you remove the following tests from TextInput on purpose? They seem relevant to me.

- should be working in a controlled state
- should not update the value when disabled

packages/react/src/PasswordInput/PasswordInput.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <PasswordInput id="password-field" />
+        </>,
+      )
+
+      const component = screen.getByLabelText(/password/i)
+
+      expect(component).toHaveAttribute('type', 'password')
+    })
+  })
+})
```
alimpens, line 84: If we add `autocapitalize`, `autocorrect` and `spellcheck` by default, we should test for them as well.

packages/css/src/components/password-input/password-input.scss
```diff
@@ -0,0 +1,59 @@
...
+  @include text-rendering;
+  @include reset;
+
+  &:hover {
+    box-shadow: var(--ams-password-input-hover-box-shadow);
+  }
+}
+
+.ams-password-input::placeholder {
+  color: var(--ams-text-input-placeholder-color);
```
alimpens, line 40:
```suggestion
  color: var(--ams-password-input-placeholder-color);
```

The token is missing as well.

packages/css/src/components/password-input/README.md
VincentSmedinga: Copied from Text Input, where it was an example. Adjusted.

packages/css/src/components/password-input/README.md
VincentSmedinga: Updated in a commit – we don’t (need to) use the `path` param.

packages/css/src/components/password-input/password-input.scss
VincentSmedinga: Fixed and added in a commit.

packages/react/src/PasswordInput/PasswordInput.test.tsx
VincentSmedinga: No, not sure where I copied the tests from. Added now.

packages/react/src/PasswordInput/PasswordInput.tsx
VincentSmedinga: No not on purpose. It’s in now.

packages/react/src/PasswordInput/PasswordInput.tsx
VincentSmedinga: Sure thing!

storybook/src/components/PasswordInput/PasswordInput.stories.tsx
VincentSmedinga: Yes, added.

VincentSmedinga (commented):
Thanks, thorough as always! 💪🏻

packages/react/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, InputHTMLAttributes } from 'react'
+
+export type PasswordInputProps = {
+  /** Whether the value fails a validation rule. */
+  invalid?: boolean
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'aria-invalid' | 'type'>
```
alimpens, line 13: We should probably omit `autocapitalize`, `autocorrect` and `spellcheck` here as well, right? So users don't get the impression they can change those?

storybook/src/components/TextInput/TextInput.docs.mdx
```diff
@@ -26,7 +26,7 @@ On some devices, it may show an email-specific keyboard.
 Consider setting the following attributes:
 
 1. Set `autocomplete="email"` to help browsers autofill the user’s email address.
-2. Set `autocapitalize="none"`, `autocorrect="off"` and `spellcheck="false"` to stop browsers automatically changing user input.
+2. Set `autocapitalize="off"`, `autocorrect="off"` and `spellcheck="false"` to stop browsers automatically changing user input.
```
alimpens, line 29: Apple recommends using `none`, they deprecated `off`

storybook/src/components/PasswordInput/PasswordInput.stories.tsx
alimpens: Hm, this doesn't seem to be working, it still validates while you type.

VincentSmedinga: > The issue you’re encountering is likely due to the browser’s built-in CSS styling applied to the input field when the `minlength` attribute is used, regardless of the `novalidate` attribute on the form. The `novalidate` attribute disables the browser's built-in form validation, meaning it won’t prevent form submission or show default validation messages. However, it does not prevent the browser from applying its default styles based on the validation state of the input fields.
> 
> Here's what’s happening:
> 
> 1. The `minlength` attribute specifies a minimum number of characters that the user must enter. If the input value is shorter than this minimum, the input field is considered invalid.
> 2. Despite using `novalidate`, the browser still tracks the validity state of the input element and may apply default styles indicating an invalid state (such as turning the border red).

So I guess the issue here is our styling for `:invalid`, not the example itself.

storybook/src/components/TextInput/TextInput.docs.mdx
VincentSmedinga: Are you an encyclopaedia? (Serious question.) (Fixed.)

alimpens: Ha! No, just crazy Google skills 😁

storybook/src/components/PasswordInput/PasswordInput.stories.tsx
alimpens: Hm, I see. Do we still want to recommend using it then? If it doesn't work the way we want it to? Both React Hook Form and Formio don't use the `minlength` attribute directly, I can imagine that's for this reason.

`minlength` is kind of similar to `maxlength` I guess, in that you can use them in form validation libraries, but not directly as attributes. Both validate too quickly, and in the case of `maxlength` it cuts off the input.

VincentSmedinga: Yes, I think I agree. It still feels like a big leap to denounce browser validation entirely, but GOV.UK makes a compelling case and we all know how bad errors before finishing typing are.

I’ve removed the attribute and overturned the guideline.

alimpens: Yeah, I think that is a good idea for now. We have to work this out in detail in our form guidelines, but that's something for later.

Without any validation attributes, we can get rid of the `<form noValidate>` decorator, I think?

alimpens (approved):

PR #1456, opened 2024-07-25
chore: Improve presentation of design token tables

VincentSmedinga:
I discovered that the space token previews would shrink with the window size. Seems logical because we use the actual tokens for their widt. On the other hand, we display these for their largest width only and the idea is to show the slope of the scale. That’s better done by expressing the largest values in rems, which don’t depend on the window size but do scale with text if people choose to do so.

The tables in the Space docs now show the full token names (`ams-space-xx`) instead of just the size (`xx`) – they are in the ‘Design Tokens’ section after all, and this matches how we displayed it for Borders. Then, having ‘Extra Small’ etc. as row headings didn’t add that much anymore so I merged those.

Some other changes too to make things easier to grasp. And ‘example’ instead of ‘preview’: we use ‘Examples’ for stories as well.

### DIFF
--- a/storybook/src/foundation/design-tokens/border.docs.mdx
+++ b/storybook/src/foundation/design-tokens/border.docs.mdx
@@ -14,7 +14,7 @@ We have 4 border widths:
 
 | Token name            | px  |  rem   | Example                                                                                            |
 | :-------------------- | :-: | :----: | -------------------------------------------------------------------------------------------------- |
-| `ams.border.width.sm` |  1  | 0.0625 | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: '0.0625rem' }} /> |
-| `ams.border.width.md` |  2  | 0.125  | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: '0.125rem' }} />  |
-| `ams.border.width.lg` |  3  | 0.1875 | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: '0.1875rem' }} /> |
-| `ams.border.width.xl` |  4  |  0.25  | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: '0.25rem' }} />   |
+| `ams.border.width.sm` |  1  | 0.0625 | <div className="ams-docs-token-example--border" style={{ borderInlineStartWidth: '0.0625rem' }} /> |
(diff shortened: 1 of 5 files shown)


storybook/src/foundation/design-tokens/border.docs.mdx
```diff
@@ -12,9 +12,9 @@ Elements that have a border, outline or underline use one of these widths.
...
-| **Small**       |  1  | 0.0625 | `ams.border.width.sm` | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: 'var(--ams-border-width-sm)' }} /> |
-| **Medium**      |  2  | 0.125  | `ams.border.width.md` | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: 'var(--ams-border-width-md)' }} /> |
-| **Large**       |  3  | 0.1875 | `ams.border.width.lg` | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: 'var(--ams-border-width-lg)' }} /> |
-| **Extra large** |  4  |  0.25  | `ams.border.width.xl` | <div className="ams-docs-token-preview--border" style={{ borderInlineStartWidth: 'var(--ams-border-width-xl)' }} /> |
+| Token name            | px  |  rem   | Example                                                                                            |
+| :-------------------- | :-: | :----: | -------------------------------------------------------------------------------------------------- |
+| `ams.border.width.sm` |  1  | 0.0625 | <div className="ams-docs-token-example--border" style={{ borderInlineStartWidth: '0.0625rem' }} /> |
```
alimpens, line 17: Why the hardcoded value here, instead of a CSS var?

storybook/src/foundation/design-tokens/space.docs.mdx
```diff
@@ -74,28 +74,21 @@ In Compact Mode, the medium grid space grows from 16 to 40 pixels between window
 <table>
```
alimpens, line 74: Why was this HTML again?

storybook/src/foundation/design-tokens/border.docs.mdx
VincentSmedinga: I think they have always been like this. Updated now.

storybook/src/foundation/design-tokens/space.docs.mdx
VincentSmedinga: Because we used `colspan`, but as of this PR we no longer do :) Converted back to Markdown.

alimpens (approved):

PR #1460, opened 2024-07-26
feat!: Use inline SVGs for Radio icon

alimpens:
# Describe the pull request

## What

It uses an inline SVG for the Radio icon and allows users to replace this icon. It also styles the default icon for correct display in forced colors mode. 

## Why

This makes it easier for external users to change the Icon. They can use their own SVGs and style them however they want to. This PR also fixes the display of Radios in forced colors mode. 

## How

By using an inline SVG with `ams-radio__circle` and `ams-radio__checked-indicator` classes, and styling those.

## Checklist

- [x] Add or update stories


packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +35,158 @@
...
-    color: var(--ams-radio-hover-color);
-    text-decoration-line: underline;
-    text-decoration-thickness: var(--ams-radio-hover-text-decoration-thickness);
-    text-underline-offset: 0.375rem;
+  // Radio button check mark
+  &::after {
+    background-color: var(--ams-radio-button-checked-background-color);
+    block-size: calc(var(--ams-radio-font-size) * var(--ams-radio-line-height));
+    flex-shrink: 0;
+    inline-size: 1.5rem;
```
dlnr, line 45: Where does this static value come from? I would expect a token and something related to the font-size/line-height perhaps. I could be wrong, it looks good.

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +35,158 @@
...
-    .ams-radio__circle::after {
-      border-color: var(--ams-radio-circle-hover-border-color);
-    }
+  // Radio button
+  &::before {
+    background-color: var(--ams-radio-button-background-color);
+    block-size: calc(var(--ams-radio-font-size) * var(--ams-radio-line-height));
+    content: "";
+    flex-shrink: 0;
+    inline-size: 1.5rem;
```
dlnr, line 63: Same here

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +35,158 @@
...
-// Default checked
-.ams-radio__input:checked {
-  + .ams-radio__label .ams-radio__circle::after {
-    background-image: var(--ams-radio-circle-checked-background-image);
+// Default hover
+.ams-radio__label:hover {
+  color: var(--ams-radio-hover-color);
+  text-decoration-line: underline;
+  text-decoration-thickness: var(--ams-radio-hover-text-decoration-thickness);
+  text-underline-offset: 0.375rem;
```
dlnr, line 82: Not related to ams.link-appearance.text-underline-offset?

dlnr (commented):
Looks like an improvement, some non token values that look weird to me.

dlnr:
This vertical alignment of the text seems slightly off:

image

Figma

image

Storybook

packages/css/src/components/radio/radio.scss
alimpens: The radio button is currently a fixed size for all screen widths in the Fabrique designs, it's always 1.5rem. We haven't handled radios on smaller screens in our own designs yet.

Good idea to turn it into a token!

packages/css/src/components/radio/radio.scss
alimpens: I think it should be, I'll change it

alimpens:
> This alignment of the text seems slightly off:
> 
> image
> Figma
> 
> image
> Storybook

Weird, it's exactly the same on my machine on a full width screen. Maybe the font renders slightly different on Macs? Did you check on a full width (or at least > 1600px) screen?

packages/css/src/components/radio/radio.scss
dlnr: Ok, so this will not scale like the font?

packages/css/src/components/radio/radio.scss
dlnr: Sweet

packages/css/src/components/radio/radio.scss
alimpens: Currently, no. Unless the design changes.

dlnr (dismissed):

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +35,158 @@
...
+  text-decoration-thickness: var(--ams-radio-text-decoration-thickness);
+  text-underline-offset: var(--ams-radio-text-underline-offset);
 
-  &:hover {
-    color: var(--ams-radio-hover-color);
-    text-decoration-line: underline;
-    text-decoration-thickness: var(--ams-radio-hover-text-decoration-thickness);
-    text-underline-offset: 0.375rem;
+  // Radio button check mark
+  &::after {
```
VincentSmedinga, line 43: Some circles get arbitrarily cut off in Safari. Can’t pinpoint why.

packages/css/src/components/radio/radio.scss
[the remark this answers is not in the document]
alimpens: The released version has a maximum size of 38.38px. The latest iteration in this PR has the same size.

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
   gap: var(--ams-radio-gap);
   line-height: var(--ams-radio-line-height);
   outline-offset: var(--ams-radio-outline-offset);
+  position: relative;
```
dlnr, line 26: This does not do anything.

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
...
-      border-color: var(--ams-radio-circle-hover-border-color);
-    }
+.ams-radio__icon {
+  circle:first-child {
+    stroke: var(--ams-radio-icon-outer-stroke);
+    stroke-width: 0.15rem;
   }
 
-  @include text-rendering;
+  circle:nth-child(2) {
```
dlnr, line 46: Radio.svg is pretty much static, so you could've also used BEM to classes inside the SVG. For example `ams-radio__circle-inner` or something. Then if another party wants to customize radio for whatever reason they are not bound by the number of children in the SVG but can give classes to the components of the radio icon.

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
...
+
+// Default
+@media (forced-colors: active) {
+  .ams-radio__label,
+  .ams-radio__label:hover,
+  .ams-radio__input[aria-invalid="true"] + .ams-radio__label,
+  .ams-radio__input[aria-invalid="true"] + .ams-radio__label:hover {
+    .ams-radio__icon {
+      circle:first-child {
+        stroke: FieldText;
```
dlnr, line 156: Could this be solved by `currentColor`? So you can just set the `.ams-radio__label` to `color: FieldText`? Just asking maybe you've already tried?

storybook/src/components/Radio/Radio.stories.tsx
```diff
@@ -145,3 +149,9 @@ export const InAFieldSetWithValidation: Story = {
     </FieldSet>
   ),
 }
+
+export const CustomIcons: Story = {
+  args: {
+    icon: <FavouriteIcon />,
```
dlnr, line 155: Too bad it doesn't have children to show a selected state. It would be cool if you are able to put an SVG in here instead of an icon component.

storybook/src/components/Radio/Radio.docs.mdx
```diff
@@ -35,3 +35,12 @@ If the Radio can become invalid, add an Error Message component and its `id` to
 Check [the Field Set docs](/docs/components-forms-field-set--docs) for more information on configuring it.
 
 <Canvas of={RadioStories.InAFieldSetWithValidation} />
+
+### Custom icons
+
+It is possible to change the default Radio icon.
+This is intended for users who do not work within the City of Amsterdam corporate identity.
+If you work for the City of Amsterdam, it is unlikely that you need this.
+If you would like to use a custom icon, please contact designsystem@amsterdam.nl.
```
dlnr, line 44:
```suggestion
If you would like to use a custom icon, please contact [designsystem@amsterdam.nl](mailto:designsystem@amsterdam.nl).
```
You could link it but also, maybe not (spam) if you agree on the other comments in the custom icons story about the svg class named inner/outer stuff. So you add `<svg>` element example in there that is selectable. Don't know if taking these kinds of requests is a good idea, the documentation should be enough.

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
   gap: var(--ams-radio-gap);
   line-height: var(--ams-radio-line-height);
   outline-offset: var(--ams-radio-outline-offset);
+  position: relative;
+  text-decoration-thickness: var(--ams-radio-text-decoration-thickness);
+  text-underline-offset: var(--ams-radio-text-underline-offset);
```
dlnr, line 28:
```suggestion
  text-underline-offset: var(--ams-radio-text-underline-offset);
  -webkit-tap-highlight-color: transparent;
```

Think we've already discussed this once but on an Ipad you get an ugly tap effect (semi-transparent black background) and this get rid of that.

dlnr (commented):
The text alignment stuff of my previous review is resolved.

packages/css/src/components/radio/radio.scss
VincentSmedinga: This is a reset, so let’s include it in the mixin.

Should we check other components for this? Might be applicable with every `touch-action: manipulation;`.

packages/css/src/components/radio/radio.scss
VincentSmedinga: Even more so: if we write CSS for icons, their SVG markup has effectively become API. We need semantic names for the circle and the dot, and use them in SVG, CSS, and tokens.

storybook/src/components/Radio/Radio.stories.tsx
VincentSmedinga: Yes, please give me a heart-shaped dot when I click it! 😍

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
...
 
-  &:hover {
-    color: var(--ams-radio-hover-color);
-    text-decoration-line: underline;
-    text-decoration-thickness: var(--ams-radio-hover-text-decoration-thickness);
-    text-underline-offset: 0.375rem;
+.ams-radio__icon-container {
+  block-size: var(--ams-radio-icon-container-block-size);
+  display: flex;
+  flex-shrink: 0;
```
VincentSmedinga, line 36: Let’s follow this advice from the [specification](https://www.w3.org/TR/css-flexbox-1/#flex-components):

> Authors are encouraged to control flexibility using the `flex` shorthand rather than with its longhand properties directly, as the shorthand correctly resets any unspecified components to accommodate common uses.

```suggestion
  flex: none;
```

proprietary/assets/icons/Radio.svg
VincentSmedinga, line 1: Zooming doesn’t work correctly.

Screenshot 2024-09-19 at 16 34 32

packages/css/src/components/radio/radio.scss
```diff
@@ -44,111 +23,181 @@
...
+  inline-size: var(--ams-radio-icon-container-inline-size);
+}
 
-    .ams-radio__circle::after {
-      border-color: var(--ams-radio-circle-hover-border-color);
-    }
+.ams-radio__icon {
+  circle:first-child {
+    stroke: var(--ams-radio-icon-outer-stroke);
+    stroke-width: 0.15rem;
```
VincentSmedinga, line 43: How do we decide which design choices to store in the SVG and which in CSS?

VincentSmedinga: Does the value of `0.15rem` match the design choice for the width of the strokes? I think the designers decided on 1.5 pixels for a canvas of 24 by 24 pixels.

packages/css/src/components/radio/radio.scss
alimpens: Do we want to remove this styling though? It's an a11y feature.

See also [this](https://github.com/necolas/normalize.css/issues/23) and [this](https://github.com/twbs/bootstrap/pull/22964)

packages/css/src/components/radio/radio.scss
alimpens: > How do we decide which design choices to store in the SVG and which in CSS?

The general idea I used now was SVG for shapes, CSS for style, similar to what we do with HTML and CSS.

> Does the value of 0.15rem match the design choice for the width of the strokes?

Good catch, I'm not sure where `0.15` came from. In the designs [here](https://www.figma.com/design/9IGm6IdPUYizBNGsUnueBd/branch/UzLI9D01t8POzFIpg46CBe/Amsterdam-Design-System?node-id=2927-29187&node-type=canvas&t=wlbVJASvfrUBoIXd-0) stroke width is 2px, so it should be `0.125rem` I think?

packages/css/src/components/radio/radio.scss
alimpens: Good point. I originally tried this approach, but SVGR changes the class names when we convert from SVG to React component, so the final class names were a bit unpredictable. And I was too lazy to look into our SVGR config 😁 

I'll look into it now, I agree it's the better approach

packages/css/src/components/radio/radio.scss
alimpens: You could, but the label text shouldn't really be `FieldText`, but `CanvasText`. So they're 2 different system colors, which is why `currentColor` doesn't really work here, I think.

storybook/src/components/Radio/Radio.docs.mdx
alimpens: Linking explicitly isn't necessary, MDX already does that for you.

Using a custom icon is still kind fiddly, regardless of the approach we use. We could explain it in detail in the docs, but that feels like overkill for something that might never be used? It's kind of a fringe feature imo

proprietary/assets/icons/Radio.svg
alimpens: Ah, good catch, should be pixels indeed

storybook/src/components/Radio/Radio.stories.tsx
alimpens: I updated this. Maybe not quite what you envisioned, but close

storybook/src/components/Radio/Radio.docs.mdx
VincentSmedinga: I’ve updated the copy and removed the suggestion to send an e-mail.

VincentSmedinga (approved):

PR #1499, opened 2024-08-07
feat: Add Hint as component and Label and FieldSet props

RubenSibon:
# Describe the pull request

Mark form fields as not required.

## What

Adds text to a label or a field set's legend if the form element (input, field set, etc.) is explicitly marked as not required. The text communicates that the field is not required for submission of the form.

## Why

There needs to be a way to communicate to end-users what form fields are not required.

## How

Properties are added to Label and Field Set to set them as required. The values should be the same as the associated 

## Checklist

- [x] Add or update stories

## Additional notes

- Documentation needs attention
  - We should link to discussions about best practices around marking fields as "required", "optional" or "not required".


packages/react/src/Label/Label.tsx
```diff
@@ -8,13 +8,27 @@ import clsx from 'clsx'
...
+export const LabelOptionalFlag = ({ notRequiredLabel }: LabelProps) => (
+  <span className="ams-label__not-required">
+    <span className="ams-visually-hidden">,</span> ({notRequiredLabel})
+  </span>
+)
+
 export const Label = forwardRef(
   (
-    { children, className, ...restProps }: PropsWithChildren<LabelHTMLAttributes<HTMLLabelElement>>,
+    { children, className, notRequiredLabel = 'niet verplicht', required, ...restProps }: LabelProps,
```
alimpens, line 26: I'm not sure we should hard-code the `notRequired` part like this. [NLDS suggests either marking all fields that are not required, or all fields that are retuired, with a preference for the first one](https://www.nldesignsystem.nl/richtlijnen/formulieren/voorkom-fouten/verplichte-velden/#vermeld-duidelijk-of-een-veld-wel-of-niet-verplicht-is). If we hard-code it as not required, we lose the second option. 

Maybe we can make it more generic by using a `suffix` or `tag` prop or something, which can be used for both options. What do you think?

alimpens, line 26: Do we need 2 props for this? Couldn't we just use `notRequiredLabel` without a default string, which is shown when it's defined? The prop `required` implies that it does something more than set a suffix, imo.

packages/react/src/Label/Label.tsx
```diff
@@ -8,13 +8,27 @@ import clsx from 'clsx'
...
+export type LabelProps = PropsWithChildren<LabelHTMLAttributes<HTMLLabelElement>> & {
+  /** Text to inform the user that the field is explicitely not required. */
+  notRequiredLabel?: string
+  /** Whether the field set is required. */
+  required?: boolean
+}
+
+export const LabelOptionalFlag = ({ notRequiredLabel }: LabelProps) => (
+  <span className="ams-label__not-required">
+    <span className="ams-visually-hidden">,</span> ({notRequiredLabel})
```
alimpens, line 20: The comma isn't necessary when we use parentheses

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -38,9 +40,16 @@ type Story = StoryObj<typeof meta>
...
-    <FieldSet invalid={args.invalid} legend={args.legend}>
+    <FieldSet
+      invalid={args.invalid}
+      legend={args.legend}
+      required={args.required}
+      notRequiredLabel={args.notRequiredLabel}
+    >
       <Field className="ams-mb--sm">
-        <Label htmlFor="input-a1">Voornaam</Label>
+        <Label htmlFor="input-a1" required={args.required} notRequiredLabel="">
```
alimpens, line 50: `required` and `notRequiredLabel` shouldn't be defined here, right? You set the label on the Field Set legend?

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -93,15 +108,54 @@ export const WithDescription: Story = {
   ),
 }
 
+export const WithNotRequiredLabel: Story = {
```
alimpens, line 111: This currently doesn't show the not required label. I'm not sure we really need this story, the not required label part works the same for all fieldsets. Showing it once is enough imo

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,22 +6,41 @@
 import clsx from 'clsx'
 import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { LabelOptionalFlag } from '../Label/Label'
```
alimpens, line 9: Seeing as we currently define the styling of the legend separately from the styling of the label, maybe we shouldn't reuse the styling for the flag here? It's a bit confusing if only part of the styling is reused imo.

packages/css/src/components/blockquote/blockquote.scss
```diff
@@ -21,17 +21,17 @@
   line-height: var(--ams-blockquote-line-height);
   quotes: "“" "”";
 
+  @include hyphenation;
```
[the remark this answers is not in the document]
VincentSmedinga: I agree; great to keep tidying things up, but preferably in a separate PR.

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,22 +6,41 @@
...
 
 export type FieldSetProps = PropsWithChildren<HTMLAttributes<HTMLFieldSetElement>> & {
   /** Whether the field set has an input with a validation error */
   invalid?: boolean
   /** The text for the caption. */
   legend: string
+  /** Text to inform the user that the field is explicitely not required. */
+  notRequiredLabel?: string
+  /** Whether the field set is required. */
```
VincentSmedinga, line 18:
```suggestion
  /** Whether a value is required for all inputs. */
```

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,22 +6,41 @@
...
 import { forwardRef } from 'react'
 import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { LabelOptionalFlag } from '../Label/Label'
 
 export type FieldSetProps = PropsWithChildren<HTMLAttributes<HTMLFieldSetElement>> & {
   /** Whether the field set has an input with a validation error */
   invalid?: boolean
   /** The text for the caption. */
   legend: string
+  /** Text to inform the user that the field is explicitely not required. */
```
VincentSmedinga, line 16:
```suggestion
  /** Informs the user that providing a value for the inputs is optional. */
```

packages/react/src/Label/Label.tsx
VincentSmedinga: Two props do feel a bit much, but removing the default text would require consumers to provide the same text in many places.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -24,7 +24,15 @@ and provide the `id` of the describing element as its value.
 
 <Canvas of={FieldSetStories.WithDescription} />
 
-## With Error
+### With Not Required Label
+
+A Field Set can indicate whether all of the inputs it contains are <em>not</em> required.
```
VincentSmedinga, line 29:
```suggestion
A Field Set can indicate whether values for all its inputs are optional.
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -24,7 +24,15 @@ and provide the `id` of the describing element as its value.
...
+Make sure to <em>not</em> add a `required` or `aria-required="true"` attribute to <em>any</em> form element in this Field Set.
+Also never set either the `aria-required` attribute on a field set that is not of the `role="radiogroup"` (see: [Radio Group](#radio-group)).
```
VincentSmedinga, line 31: Not sure what this must convey; can you rephrase?

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -24,7 +24,15 @@ and provide the `id` of the describing element as its value.
 
 <Canvas of={FieldSetStories.WithDescription} />
 
-## With Error
+### With Not Required Label
```
VincentSmedinga, line 27: We can use ‘Optional’ instead of ‘Not Required’ in our docs and code. Just not for users.

And in this case, ‘Optional’ is enough instead of ‘With Optional Label’.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -43,15 +51,22 @@ Always also set `aria-required` on the individual radio buttons though, to make
 
 <Canvas of={FieldSetStories.RadioGroup} />
 
-### Radio group with error
+### Radio Group With Not Required Label
+
+A Field Set with a radio button group can also indicate whether all of the radio inputs it contains are <em>not</em> required.
```
VincentSmedinga, line 56: We decided to use ‘Radio’ everywhere in another PR, not ‘radio input’ or ‘radio button’.

packages/react/src/Label/Label.tsx
```diff
@@ -8,13 +8,27 @@ import clsx from 'clsx'
 import { forwardRef } from 'react'
 import type { ForwardedRef, LabelHTMLAttributes, PropsWithChildren } from 'react'
 
+export type LabelProps = PropsWithChildren<LabelHTMLAttributes<HTMLLabelElement>> & {
+  /** Text to inform the user that the field is explicitely not required. */
```
VincentSmedinga, line 12:
```suggestion
  /** Informs the user that providing a value for the associated input is optional. */
```

packages/react/src/Label/Label.tsx
```diff
@@ -8,13 +8,27 @@ import clsx from 'clsx'
...
+  notRequiredLabel?: string
+  /** Whether the field set is required. */
```
VincentSmedinga, line 14:
```suggestion
  /** Whether the associated input is required. */
```

packages/css/src/components/field-set/README.md
```diff
@@ -7,6 +7,8 @@ A component to group related form inputs.
 ## Guidelines
 
 - Use Field Set when you need to show a relationship between multiple form inputs. For example, you may need to group a set of text inputs into a single Field Set when asking for an address.
+- If all the child form elements get a `required` attribute with the same value, then add the same property with that value to the field set. Also make sure to have an empty string as a value for the `notRequiredLabel` property on all the Label components to avoid many, redundant ‟not required” labels.
+- If the various child form elements have mixed `required` attribute values then do the reverse and set the `notRequiredLabel` property on the Field Set to an empty string and set the `required` property on the child labels as needed.
```
VincentSmedinga, line 11: Don’t we advise against client-side validation [like GOV.UK](https://design-system.service.gov.uk/patterns/validation/#turn-off-html5-validation) does? If yes, we advise the opposite with respect to the `required` attribute.

packages/css/src/components/blockquote/blockquote.scss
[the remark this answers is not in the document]
RubenSibon: Yeah agreed. If I remember correctly my IDE's linter was complaining.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -34,7 +42,7 @@ Add an `aria-describedby` attribute to the input and provide the `id` of Error M
 
 <Canvas of={FieldSetStories.WithValidation} />
 
-### Radio group
+## Radio Group
```
RubenSibon, line 45:
```suggestion
### Radio Group
```

packages/react/src/Label/Label.tsx
RubenSibon: We decided to use the concept of "hint".

RubenSibon: We ended up using two, very loosely connected, props. With the `optional` prop the default hint "niet verplicht" appears at the end of the label's text in parentheses, but only if no value for the `hint` prop was provided. If a `hint` is provided it will always display that value.

packages/css/src/components/label/label.scss
```diff
@@ -15,4 +15,9 @@
 
   @include hyphenation;
   @include text-rendering;
+
+  & > .ams-hint {
+    display: inline-flex;
+    flex-wrap: nowrap;
```
VincentSmedinga, line 21: Is this the best approach to prevent wrapping? Why not use `white-space: nowrap` which is specifically aimed at handling text?

packages/css/src/components/label/label.scss
```diff
@@ -15,4 +15,9 @@
 
   @include hyphenation;
   @include text-rendering;
+
+  & > .ams-hint {
```
VincentSmedinga, line 19: Can we generally apply these styles to `.ams-hint` without the parent selector?

packages/react/src/Label/Label.tsx
```diff
@@ -8,15 +8,36 @@ import clsx from 'clsx'
...
+}
+
+type LabelOptionalProp = {
+  /** Whether the associated input is optional. Will append the text '(niet verplicht)' to the label if no hint is provided. */
+  optional?: boolean
+}
+
+export type LabelProps = PropsWithChildren<LabelHTMLAttributes<HTMLLabelElement>> & (LabelHintProp & LabelOptionalProp)
+
+const LabelHint = ({ hint }: LabelHintProp) => <span className="ams-hint">({hint})</span>
```
VincentSmedinga, line 23: Making this a separate `Hint` component matches the CSS component to a React one and prevents repetition between the Label and FieldSet components.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -14,7 +14,7 @@ import README from "../../../../packages/css/src/components/field-set/README.md?
 
 ## Examples
 
-### With description
+### With Description
```
VincentSmedinga, line 17: We chose sentence case for headings in #1565. Only component names in headings still start with uppercase letters, but `Description` is not a component.

```suggestion
### With description
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -24,7 +24,14 @@ and provide the `id` of the describing element as its value.
 
 <Canvas of={FieldSetStories.WithDescription} />
 
-### With validation
+### With Hint Marking Field Set As Required
```
VincentSmedinga, line 27:
```suggestion
### With Hint
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -34,7 +41,7 @@ Add an `aria-describedby` attribute to the input and provide the `id` of Error M
 
 <Canvas of={FieldSetStories.WithValidation} />
 
-### Radio group
+### Radio Group
```
VincentSmedinga, line 44:
```suggestion
### Radio group
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -44,15 +51,22 @@ Even so, always set `aria-required` on the individual Radios as well to ensure t
 
 <Canvas of={FieldSetStories.RadioGroup} />
 
-### Radio group with validation
+### Radio Group With Hint Marking Group As Optional
```
VincentSmedinga, line 54:
```suggestion
### Radio group with Hint
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -44,15 +51,22 @@ Even so, always set `aria-required` on the individual Radios as well to ensure t
...
+
+A Field Set with a Radio group can also indicate whether all of the radios it contains are <em>not</em> required.
+Make sure to add the `required` or `aria-required="true"` attribute to <em>any</em> samely named Radios in this Field Set.
+
+<Canvas of={FieldSetStories.RadioGroupWithHint} />
+
+### Radio Group With Validation
```
VincentSmedinga, line 61:
```suggestion
### Radio group with validation
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -44,15 +51,22 @@ Even so, always set `aria-required` on the individual Radios as well to ensure t
...
 
 A Field Set with a Radio group can also have a validation error.
 In this case, connect the Error Message to the Field Set instead of an input.
 Add an `aria-describedby` attribute to the Field Set and provide the `id` of Error Message as its value.
 
 <Canvas of={FieldSetStories.RadioGroupWithValidation} />
 
-### Checkbox group
+### Checkbox Group
```
VincentSmedinga, line 69:
```suggestion
### Checkbox group
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -79,7 +93,9 @@ The order of the `id`s is the order in which they are read by screen readers.
 
 <Canvas of={FieldSetStories.CheckboxGroup} />
 
-### Checkbox group with validation
+Note that you should not make a Field Set containing inputs of `type="checkbox"` required.
+
+### Checkbox Group With Validation
```
VincentSmedinga, line 98:
```suggestion
### Checkbox group with validation
```

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -38,7 +40,7 @@ type Story = StoryObj<typeof meta>
 
 export const Default: Story = {
   render: (args) => (
-    <FieldSet invalid={args.invalid} legend={args.legend}>
+    <FieldSet invalid={args.invalid} legend={args.legend} hint={args.hint} optional={args.optional}>
```
VincentSmedinga, line 43: Let’s destructure `args` in cases like this to make the implementation easier to read.

packages/css/src/components/label/label.scss
RubenSibon: I'm making a dedicated Hint component that comes with its own styling.

packages/react/src/Label/Label.tsx
RubenSibon: Indeed! Creating a new Hint component.

storybook/src/components/FieldSet/FieldSet.docs.mdx
RubenSibon: Removed the sentence and changed the text at the Radio group story.

packages/react/src/FieldSet/FieldSet.test.tsx
```diff
@@ -54,4 +54,36 @@ describe('FieldSet', () => {
...
+  it('renders the default hint text "niet verplicht" after the field-set\'s legend marked as optional', () => {
+    const { container } = render(<FieldSet legend="Legend" optional={true} />)
+
+    const legend = container.querySelector('legend')
+
+    expect(legend).toHaveTextContent('Legend (niet verplicht)')
+  })
+
+  it('renders the provided hint text "optional" after the field-set\'s legend marked as optional', () => {
+    const { container } = render(<FieldSet legend="Legend" optional={true} hint="optional" />)
```
alimpens, line 75: Maybe we shouldn't use the string 'optional' here, seeing as the prop is called `optional` as well.

packages/react/src/FieldSet/FieldSet.test.tsx
```diff
@@ -54,4 +54,36 @@ describe('FieldSet', () => {
...
+
+    const legend = container.querySelector('legend')
+
+    expect(legend).toHaveTextContent('Legend (optional)')
+  })
+
+  it('renders the provided hint text "required" after the field-set\'s legend marked as not optional', () => {
```
alimpens, line 82:
```suggestion
  it('renders the provided hint text when `optional` is set to `false`', () => {
```

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,25 +6,43 @@
...
 import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { Hint, HintProps } from '../Hint'
 
 export type FieldSetProps = PropsWithChildren<HTMLAttributes<HTMLFieldSetElement>> & {
   /** Whether the field set has an input with a validation error */
   invalid?: boolean
   /** The text for the caption. */
   legend: string
-}
+  /** Whether the associated inputs are optional. Will append the text '(niet verplicht)' to the label if no hint is provided. */
```
alimpens, line 16:
```suggestion
  /** Appends the text '(niet verplicht)' to the label if no hint is provided. Use when the associated inputs are optional. */
```

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,25 +6,43 @@
...
+    { children, className, invalid, legend, hint, optional, ...restProps }: FieldSetProps,
+    ref: ForwardedRef<HTMLFieldSetElement>,
+  ) => {
+    let optionalHint = null
+
+    if (optional) {
+      optionalHint = hint ? <Hint hint={hint} /> : <Hint hint="niet verplicht" />
+    } else if (hint) {
+      optionalHint = <Hint hint={hint} />
+    }
```
alimpens, line 31: Can't we extract this logic to the Hint component itself? Something like this (pseudocode):

```
const getHintText = ({ hint, optional }: HintProps) => {
  if (hint) {
    return <Hint hint={hint} />
  } else if (optional) {
    return <Hint hint="niet verplicht" />
  }

  return null
}

export const Hint = forwardRef(
  ({ className, hint, optional, ...restProps }: HintProps, ref: ForwardedRef<HTMLElement>) => {
    const hintText = getHintText({ hint: hint, optional: optional })

    return (
      hintText && (
        <span {...restProps} ref={ref} className={clsx('ams-hint', className)}>
          ({hintText})
        </span>
      )
    )
  },
)
```

packages/react/src/Hint/Hint.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  })
+
+  it('renders an additional class name', () => {
+    const { container } = render(<Hint className="extra" />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('ams-hint extra')
+  })
+
```
alimpens, line 31: We're missing a test for the `hint` prop

packages/react/src/Hint/Hint.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export type HintProps = PropsWithChildren<HTMLAttributes<HTMLElement>> & {
+  /** Provides additional context to the user on some inline text element such as a Label. For example: a form field can be marked as optional. Keep the text brief, preferably no more than two words, as it doesn't break words to a new line. */
```
alimpens, line 11:
```suggestion
  /** Show a custom hint text. */
```

packages/react/src/Label/Label.test.tsx
```diff
@@ -86,4 +86,52 @@ describe('Label', () => {
 
     expect(ref.current).toBe(div)
   })
+
+  it('renders a hint text after the label not marked as optional', () => {
```
alimpens, line 90: The same comments to the Field Set tests apply here

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -24,7 +24,14 @@ and provide the `id` of the describing element as its value.
 
 <Canvas of={FieldSetStories.WithDescription} />
 
-### With validation
+### With Hint
+
+A Field Set can indicate whether all of the inputs it contains are optional or required.
+Make sure to add a `required` or `aria-required="true"` attribute to <em>all</em> form element in this Field Set.
```
alimpens, line 30: We don't recommend native validation, right?

```suggestion
Make sure to add `aria-required="true"` to <em>all</em> form elements in the Field Set when they are required. 
```

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -79,6 +93,8 @@ The order of the `id`s is the order in which they are read by screen readers.
 
 <Canvas of={FieldSetStories.CheckboxGroup} />
 
+Note that you should not make a Field Set containing inputs of `type="checkbox"` required.
+
```
alimpens, line 97: This is a little confusing. You should set `aria-required` on all Checkboxes in a group that is required. You shouldn't use `required` at all. I'd remove this sentence.

storybook/src/components/FieldSet/FieldSet.docs.mdx
```diff
@@ -44,6 +51,13 @@ Even so, always set `aria-required` on the individual Radios as well to ensure t
 
 <Canvas of={FieldSetStories.RadioGroup} />
 
+### Radio group with Hint
+
+A Field Set with a Radio group can also indicate whether all of the radios it contains are <em>not</em> required.
+Make sure to add the `required` or `aria-required="true"` attribute to <em>any</em> samely named Radios in this Field Set.
+
+<Canvas of={FieldSetStories.RadioGroupWithHint} />
+
```
alimpens, line 60: Is this story necessary? It isn't really any different from the 'With Hint' story, right?

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -93,6 +95,34 @@ export const WithDescription: Story = {
   ),
 }
 
+export const WithHint: Story = {
+  args: { optional: false, hint: 'verplicht' },
+  render: ({ invalid, legend, hint, optional }) => (
+    <FieldSet invalid={invalid} legend={legend} optional={optional} hint={hint}>
+      <Field className="ams-mb--sm">
+        <Label htmlFor="input-a1">Voornaam</Label>
+        {invalid && <ErrorMessage id="error-a1">Vul uw voornaam in.</ErrorMessage>}
```
alimpens, line 104: You shouldn't use the same `id` (`error-a1` for example) more than once on a page.

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -170,7 +240,6 @@ export const RadioGroupWithValidation: Story = {
   render: (args) => (
     <FieldSet
       aria-describedby={`description-e${args.invalid ? ' error-e' : ''}`}
-      aria-required="true"
```
alimpens, line 173: Is this intentional? I'm not sure we want this

storybook/src/components/Hint/Hint.docs.mdx
```diff
@@ -0,0 +1,21 @@
+{/* @license CC0-1.0 */}
```
alimpens, line 1: Do we want to show Hint in Storybook? It's sort of an internal component, no? We could even exclude it from our React export.

packages/css/src/components/hint/hint.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-hint {
+  color: var(--ams-hint-color);
+  white-space: nowrap;
```
alimpens, line 8: This is a little risky for text that can be input by users. Wouldn't `display: inline-block;` achieve the same goal and allow the hint text to wrap when necessary?

RubenSibon: Yeah, that was my solution before (`display: inline-flex` actually), but @VincentSmedinga didn't agree. He suggested `white-space: nowrap`. I've also looked at `text-wrap: nowrap`.

storybook/src/components/Hint/Hint.docs.mdx
RubenSibon: I was considering that, but decided that it could be handy to expose it. Because: why not? What's against giving developers a feature like this?

storybook/src/components/FieldSet/FieldSet.docs.mdx
RubenSibon: Removed

storybook/src/components/FieldSet/FieldSet.docs.mdx
RubenSibon: Removed

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: Fixed

packages/css/src/components/field-set/README.md
alimpens: I suggested to revise and simplify these docs, but yeah, we'll probably go for the same recommendations as gov.uk.

packages/react/src/Hint/Hint.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    expect(ref.current).toBe(component)
+  })
+
+  it('renders the provided hint text', () => {
+    const { container } = render(<Hint hint="hint text" />)
+
+    const hint = container.querySelector('span.ams-hint')
+
+    expect(hint).toHaveTextContent('(hint text)')
+  })
```
RubenSibon, line 48: There seems to be a memory leak with the `Hint` component in Jest DOM.

We're getting the following error:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

Running the tests with the `--logHeapUsage` option reveals that these four individual tests use over 4000MB(!) of RAM. The same goes for the four tests in `Label` and `FieldSet`.

@VincentSmedinga @alimpens @dlnr can you help me figure out what causes it?

packages/react/src/Hint/Hint.test.tsx
RubenSibon: Has been added, see below.

packages/css/src/components/hint/hint.scss
RubenSibon: I will change it back to `flex: inline-block`.

storybook/src/components/Hint/Hint.docs.mdx
alimpens: It's mostly that we do not want them to use it anywhere else at the moment, right? If we show it, it's easier for users to use it incorrectly.

packages/react/src/Hint/Hint.test.tsx
RubenSibon: D'oh!

packages/css/src/components/hint/hint.scss
RubenSibon: Done!

storybook/src/components/Hint/Hint.docs.mdx
RubenSibon: I've removed the story and docs 😢

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: No longer relevant (but doesn't show "outdated" for some reason).

packages/css/src/components/field-set/README.md
RubenSibon: Changed.

storybook/src/components/FieldSet/FieldSet.stories.tsx
alimpens: This is still relevant? `aria-required` is removed, I don't think it should be.

packages/react/src/FieldSet/FieldSet.test.tsx
```diff
@@ -54,4 +54,36 @@ describe('FieldSet', () => {
...
+  it('renders the provided hint text after the legend', () => {
+    render(<FieldSet legend="Legend" hint="hint text" />)
+
+    const component = screen.getByRole('group', { name: 'Legend (hint text)' })
+
+    expect(component).toBeInTheDocument()
+  })
+
+  it('renders the default hint text after the legend', () => {
+    const { container } = render(<FieldSet legend="Legend" optional={true} />)
```
alimpens, line 67: I think we can leave out `={true}`. It doesn't matter much, but maybe users will copy this code. Same for the other tests.

```suggestion
    const { container } = render(<FieldSet legend="Legend" optional />)
```

packages/react/src/FieldSet/FieldSet.test.tsx
```diff
@@ -54,4 +54,36 @@ describe('FieldSet', () => {
...
+
+    const legend = container.querySelector('legend')
```
alimpens, line 69: You've changed the selector for the previous test, but not in all the other tests.

packages/react/src/FieldSet/FieldSet.tsx
```diff
@@ -6,25 +6,33 @@
...
-      {...restProps}
-      ref={ref}
-      className={clsx('ams-field-set', invalid && 'ams-field-set--invalid', className)}
-    >
-      <legend className="ams-field-set__legend">{legend}</legend>
-      {children}
-    </fieldset>
-  ),
+  (
+    { children, className, invalid, legend, hint, optional, ...restProps }: FieldSetProps,
```
alimpens, line 20: Can you sort these alphabetically?

packages/react/src/Hint/Hint.tsx
```diff
@@ -0,0 +1,41 @@
...
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export type HintProps = PropsWithChildren<HTMLAttributes<HTMLElement>> & {
+  /** Show a custom hint text. */
+  hint?: string
+  /** Appends the text '(niet verplicht)' to the label or legend if no hint is provided. Use when the associated inputs are optional. */
+  optional?: boolean
+}
+
+const getHintText = ({ hint, optional }: HintProps) => {
```
alimpens, line 17: Not the most important thing in the world, but: I think this allows all global HTML attributes to be passed to this function. I think it's cleaner to separate the `hint` and `optional` type from `PropsWithChildren<HTMLAttributes<HTMLElement>>`, and only use the first one here.

RubenSibon: Aangepast

packages/react/src/FieldSet/FieldSet.tsx
RubenSibon: Adjusted

packages/react/src/FieldSet/FieldSet.test.tsx
RubenSibon: Made consistent

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: Added `aria-required` to all Field Sets that only have required Fields.

storybook/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -74,7 +66,7 @@ export const WithDescription: Story = {
         {args.invalid && <ErrorMessage id="error-b1">Vul uw voornaam in.</ErrorMessage>}
         <TextInput
           aria-describedby={args.invalid ? 'error-b1' : undefined}
-          aria-required="true"
+          aria-required
```
alimpens, line 69: I'm not sure this in an improvement. React turn this into `aria-required="true"`, so the end result is the same, but it sort of muddles up props and attributes imo. They are different things, and using `aria-required="true"` here makes that difference a bit more clear imo. Same for all the other instances of `aria-required`

packages/react/src/FieldSet/FieldSet.test.tsx
alimpens: This still isn't a `getByRole` selector?

storybook/src/components/FieldSet/FieldSet.stories.tsx
RubenSibon: I've adjusted it.

packages/react/src/FieldSet/FieldSet.test.tsx
RubenSibon: I've switched to `getByRole` and `name`.

packages/react/src/FieldSet/FieldSet.test.tsx
```diff
@@ -54,4 +54,36 @@ describe('FieldSet', () => {
...
+
+    expect(component).toBeInTheDocument()
+  })
+
+  it('renders the default hint text after the legend', () => {
+    render(<FieldSet legend="Legend" optional />)
+
+    const component = screen.getByRole('group', { name: 'Legend (niet verplicht)' })
+
+    expect(component).toHaveTextContent('Legend (niet verplicht)')
```
alimpens, line 71: You can just use `expect(component).toBeInTheDocument()` here. If the text is incorrect, `component` will not exist.

alimpens (approved):

PR #1529, opened 2024-08-21
fix: Remove margin Safari sets on buttons

alimpens:
# Describe the pull request

## What

It removes the margin older Safari sets on buttons

## Why

To have consistent styling between older Safari and other browsers

## How

By setting `margin-block` and `margin-inline` to 0

## Checklist

- [x] Add or update stories

## Additional notes

### DIFF
--- a/packages/css/src/components/accordion/accordion.scss
+++ b/packages/css/src/components/accordion/accordion.scss
@@ -19,9 +19,16 @@
   margin-inline: 0;
 }
 
+@mixin reset-button {
+  border: 0;
+  margin-block: 0; // [1]
+  margin-inline: 0; // [1]
+
+  // [1] Remove the margin in older Safari.
+}
+
 .ams-accordion__button {
   background-color: transparent;
-  border: 0;
   color: var(--ams-accordion-button-color);
   cursor: pointer;
   display: flex;
@@ -34,6 +41,8 @@
   padding-block: var(--ams-accordion-button-padding-block);
   padding-inline: var(--ams-accordion-button-padding-inline);
   text-align: start;
+
+  @include reset-button;
 
   &:focus {
     outline-offset: var(--ams-accordion-button-focus-outline-offset);
--- a/packages/css/src/components/icon-button/icon-button.scss
+++ b/packages/css/src/components/icon-button/icon-button.scss
@@ -5,8 +5,12 @@
 
 @mixin reset {
   border: 0;
+  margin-block: 0; // [1]
+  margin-inline: 0; // [1]
   padding-block: 0;
   padding-inline: 0;
+
+  // [1] Remove the margin in older Safari.
 }
 
 .ams-icon-button {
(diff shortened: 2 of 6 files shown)


dlnr (approved):

packages/css/src/components/accordion/accordion.scss
```diff
@@ -19,9 +19,16 @@
   margin-inline: 0;
 }
 
+@mixin reset-button {
+  border: 0;
+  margin-block: 0; // [1]
+  margin-inline: 0; // [1]
+
+  // [1] Remove the margin in older Safari.
```
VincentSmedinga, line 27: Which versions of Safari render this border, and what is their browser share?

Do newer versions not require the reset at all?

packages/css/src/components/icon-button/icon-button.scss
```diff
@@ -5,8 +5,12 @@
 
 @mixin reset {
```
VincentSmedinga, line 6: Anything against a new guideline that all mixins include the name of the element they target?

VincentSmedinga (changes requested):
Let’s assess for how many people we add these lines of code.

packages/css/src/components/accordion/accordion.scss
alimpens: Safari 15.6 and older render this margin, which is [about 1.6% of browsers used](https://caniuse.com/usage-table). Newer versions do not have this margin. I've seen mentions that older Firefox has this margin as well, but I can't reproduce that.

packages/css/src/components/icon-button/icon-button.scss
alimpens: The 'rule' I currently use is:

- use `reset` when there is a single class in the file
- use `reset-*` when there is more than one class in the file.

Using `reset-button` for Icon Button doesn't add anything imo, it's already clear what class the reset is for.

VincentSmedinga: Yes, same here. I brought it up because it can be helpful to find all resets everywhere that target e.g. a `<button>` or a `<ul>`. Might have saved you some time in this PR, for instance?

packages/css/src/components/accordion/accordion.scss
VincentSmedinga: Alright, that seems enough to include these lines.

VincentSmedinga (approved):

packages/css/src/components/icon-button/icon-button.scss
alimpens: Hm, not really, I just searched for '<button' 😁 If we had this convention, I'd probably still do that, just in case.

I think I'm kind of negative to neutral on this change. I don't think it adds much, except another thing you have to keep in mind. I could live with it though.

VincentSmedinga: Alright, let’s not change it then.

PR #1533, opened 2024-08-23
chore: Make Text Input and Text Area uncontrolled in stories

alimpens:
# Describe the pull request

## What

It changes Text Input and Text Area to uncontrolled components in Storybook. 

## Why

Controlled components do not change automatically when interacting with them, you have to set an `onChange` handler for that. We didn't for Text Input and Text Area, which means you can't change the text. This PR fixes that.

## How

By using `defaultValue` instead of `value`.

## Checklist

- [x] Add or update stories

## Additional notes

### DIFF
--- a/storybook/src/components/TextArea/TextArea.stories.tsx
+++ b/storybook/src/components/TextArea/TextArea.stories.tsx
@@ -13,7 +13,7 @@ const meta = {
   title: 'Components/Forms/Text Area',
   component: TextArea,
   args: {
-    value: paragraph,
+    defaultValue: paragraph,
     disabled: false,
     invalid: false,
   },
@@ -23,6 +23,9 @@ const meta = {
         type: 'number',
       },
       description: 'The width, expressed in the average number of characters.',
+    },
+    defaultValue: {
+      table: { disable: true },
     },
     disabled: {
       description: 'Prevents interaction. Avoid if possible.',
@@ -39,9 +42,6 @@ const meta = {
         type: 'number',
       },
       description: 'The number of lines to show',
-    },
-    value: {
-      description: 'The value of the field.',
     },
   },
 } satisfies Meta<typeof TextArea>
--- a/storybook/src/components/TextInput/TextInput.stories.tsx
+++ b/storybook/src/components/TextInput/TextInput.stories.tsx
@@ -14,6 +14,9 @@ const meta = {
     invalid: false,
   },
   argTypes: {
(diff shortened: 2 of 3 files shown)


storybook/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -24,6 +24,9 @@ const meta = {
       },
       description: 'The width, expressed in the average number of characters.',
     },
+    defaultValue: {
+      table: { disable: true },
+    },
```
VincentSmedinga, line 29: This field doesn’t get hidden for `TextInput`. I agree that hiding is the correct choice because editing the field does nothing to the content. I suggest we make this a global setting as we do for `children` in `storybook/config/preview.tsx`.

storybook/src/components/TextInput/TextInput.stories.tsx
VincentSmedinga, line 1: Not a comment to this file, but I found one stray `defaultValue: { summary: 'false' },` in `OrderedList.stories.tsx` – agree to delete that in this branch as well?

storybook/src/components/TextArea/TextArea.stories.tsx
alimpens: I've hidden `defaultValue` for Text Input as well.

I'm not sure we should make this a global setting. This is only relevant for inputs where you can input text, which currently is only TextInput, TextArea and PasswordInput. It'll probably be more confusing if we hide that in a global setting. 

Should we show the `defaultValue` control for the `disabled` stories? Currently, there is no way to change the text there.

storybook/src/components/TextInput/TextInput.stories.tsx
alimpens: Good catch, removed it!

storybook/src/components/TextArea/TextArea.stories.tsx
VincentSmedinga: > I'm not sure we should make this a global setting. This is only relevant for inputs where you can input text, which currently is only TextInput, TextArea and PasswordInput. It'll probably be more confusing if we hide that in a global setting.

Fair enough.

> Should we show the defaultValue control for the disabled stories? Currently, there is no way to change the text there.

People can switch the `disabled` prop off and on again. I would say that’s easy enough to prevent having to add this exception.

VincentSmedinga (approved):

Issue #1546, opened 2024-08-27
Closing a Dialog with the default "x" icon triggers onSubmit()


joeyamst:
# Describe the issue

When closing a `Dialog` with the default "x" icon, the `onSubmit()` hook that is given as prop to the `Dialog` will trigger. This is unwanted because the form in my dialog gets submitted without the user wanting to submit it.

## Summary

The Dialog component contains a `<form>`. When I create a Dialog and give it a `onSubmit()` prop, this function will be triggered when I press the "x" icon.

## Steps to reproduce

(Provide a clear and concise list of steps to reproduce the issue.)

1. Create a Dialog that can be opened with a button
2. Give the dialog a prop `onSubmit()` with some logging code
3. Close the dialog with the standard provided "x" icon.
4. Observe that the `onSubmit()` function is triggered

Here's a link to a codesandbox: https://codesandbox.io/p/devbox/muddy-forest-n9h9lt?workspaceId=5bdd83a2-36b4-4d1b-acb9-544b42f5ece6

## Environment

- Browser: Chrome and Firefox
- Operating System: Windows
- Version of the design system: v0.11.0

## Checklist

Before submitting your issue, please ensure you have done the following:

- [x] Search the issue tracker to ensure the issue has not already been reported.
- [x] Verify that the issue can be consistently reproduced.
- [x] Check if the issue is already documented.

## Additional notes

It is probably due to how `<form method="dialog">` handles the closing of a dialog, seeing it as a submit.

VincentSmedinga:
Hi @joeyamst, thanks for reporting. That certainly looks like a bug. We will look into it and report back. The link to the CodeSandbox leads to a 404 for me, but we can probably reproduce without it.

VincentSmedinga:
@joeyamst Me neither – I registered with GitHub and added additional authorization scopes, but I’m unfamiliar with that site. No worries.

We’re well underway with a solution in the PR mentioned above. I’ve just invited you to add your feedback, if any.

VincentSmedinga:
@joeyamst We’ve just released [@amsterdam/design-system-react v0.11.1](https://github.com/Amsterdam/design-system/releases/tag/design-system-react-v0.11.1) to address this. The close button now calls the dialog’s `close` method instead of submitting its form. We also offer new `openDialog` and `closeDialog` functions.

No changes on your side are necessary. Please let us know whether this indeed resolves the problem.

dlnr:
[...]

PR #1547, opened 2024-08-28
fix: Close Dialog without submit

alimpens:
# Describe the pull request

## What

This changes the Dialog close button from `type="submit"` to `type="button"`, and uses JS to close the Dialog.

## Why

We received [an issue noting that using the Dialog close button triggers a submit event](https://github.com/Amsterdam/design-system/issues/1546), which it shouldn't.

## How

By changing the Dialog close button from `type="submit"` to `type="button"`, and using and exporting a JS function to close the nearest modal.

## Checklist

- [x] Add or update stories

## Additional notes

- JSDom doesn't fully support `dialog`, so this functionality isn't tested.

### DIFF
--- a/packages/react/src/Dialog/Dialog.tsx
+++ b/packages/react/src/Dialog/Dialog.tsx
@@ -26,7 +26,7 @@ export const Dialog = forwardRef(
     ref: ForwardedRef<HTMLDialogElement>,
   ) => (
     <dialog {...restProps} ref={ref} className={clsx('ams-dialog', className)}>
-      <form method="dialog" className="ams-dialog__form">
+      <form className="ams-dialog__form" method="dialog">
         <header className="ams-dialog__header">
           <Heading size="level-4">{heading}</Heading>
           <IconButton label={closeButtonLabel} size="level-4" type="button" onClick={closeDialog} />
--- a/packages/react/src/Dialog/Dialog.tsx
+++ b/packages/react/src/Dialog/Dialog.tsx
@@ -29,7 +29,7 @@ export const Dialog = forwardRef(
       <form className="ams-dialog__form" method="dialog">
         <header className="ams-dialog__header">
           <Heading size="level-4">{heading}</Heading>
-          <IconButton label={closeButtonLabel} size="level-4" type="button" onClick={closeDialog} />
+          <IconButton label={closeButtonLabel} onClick={closeDialog} size="level-4" type="button" />
         </header>
         <article className="ams-dialog__article">{children}</article>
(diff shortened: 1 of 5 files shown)


storybook/src/components/Dialog/Dialog.docs.mdx
```diff
@@ -24,14 +24,14 @@ Click or tap the button to open a dialog.
 
 <Canvas of={DialogStories.TriggerButton} />
 
-Some basic example code to open and close a dialog:
+You can open the dialog using this code:
 
 ```ts
 const openDialog = () => (document.querySelector("#openDialog") as HTMLDialogElement)?.showModal();
```
VincentSmedinga, line 30: How about complementing our function to close a dialog with one to open it, e.g. `openDialog('#dialogId')`?

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -27,7 +29,7 @@ export const Dialog = forwardRef(
       <form method="dialog" className="ams-dialog__form">
         <header className="ams-dialog__header">
           <Heading size="level-4">{heading}</Heading>
-          <IconButton formNoValidate label={closeButtonLabel} size="level-4" />
+          <IconButton label={closeButtonLabel} size="level-4" type="button" onClick={closeDialog} />
```
VincentSmedinga, line 32: Sorting the props:

```suggestion
          <IconButton label={closeButtonLabel} onClick={closeDialog} size="level-4" type="button" />
```

VincentSmedinga, line 32: Why do we need `type="button"` again?

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -27,7 +29,7 @@ export const Dialog = forwardRef(
       <form method="dialog" className="ams-dialog__form">
```
VincentSmedinga, line 29: #ocd

```suggestion
      <form className="ams-dialog__form" method="dialog">
```

packages/react/src/Dialog/Dialog.tsx
alimpens: A `button` element has `type="submit"` by default. There is a discrepancy there though. We change this default to `type="button"` for our Button component. Do we want to change this for IconButton as well? Makes sense to keep those in sync imo

VincentSmedinga: Yes, I’ll create a separate task for that.

VincentSmedinga (approved):

PR #1582, opened 2024-09-12
feat!: Remove form from Dialog

dlnr:
# Describe the pull request

The form inside the dialog should be optional because there is a need to built larger forms in the popup. Besides that the vertical scroll (for long content) was broken by moving CSS to the dialog instead of the form wrapping the content. I've taking the liberty to recover some of that work but then rename the element: "form > content" and "actions > footer". The latter to illustrate that you are adding children to a (sticky) footer component.

## What

Actions become footer and some changes in tokens.

## Why

To allow developers to work with content other then `<form method="dialog">` and to document how to use it if you want. 

## How

- Rolling back the max-block-size in the dialog itself.
- Removing the form element and adding it to a story.
- Using a `form` attribute on the submit button when it's in the footer.

## Checklist

- [X] Add or update stories

## Additional notes

- I realise that making a row in the grid scrollable is weird but this is the most robust solution.

### DIFF
--- a/storybook/src/components/Dialog/Dialog.docs.mdx
+++ b/storybook/src/components/Dialog/Dialog.docs.mdx
@@ -21,6 +21,7 @@ This closes the Dialog when submitting the form.
 Pass the submit Button to the `footer` prop,
 and link it to the form by passing its `id` to the Buttons `form` attribute.
 The Dialog returns the value of the submit Button, so you can check which Button was clicked.
+For more information, see [Handling the return value from the dialog (MDN)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#handling_the_return_value_from_the_dialog).
 
 <Canvas of={DialogStories.FormDialog} />
 


storybook/src/components/Dialog/Dialog.docs.mdx
```diff
@@ -14,6 +14,18 @@ import README from "../../../../packages/css/src/components/dialog/README.md?raw
...
 
+### Form dialog
+
+The `<form method="dialog">` element is a practical way to manage dialogs.
+It simplifies the process by automatically handling the closure and data submission
+when a user interacts with a submit or cancel button. For more complex scenarios,
+developers can opt for custom handling by adding the form as a child element.
+Additionally, creating sticky buttons in a dialog's footer is achievable by placing
+a submit button outside the `<form>` and associating it with the form's ID using the
+`form` attribute.
```
alimpens, line 25:
```suggestion
### Form in a Dialog

Set `method="dialog"` when using a form in Dialog.
This closes the Dialog when submitting the form.
Pass the submit Button to the `footer` prop,
and link it to the form by passing its `id` to the Buttons `form` attribute.
```

storybook/src/components/Dialog/Dialog.docs.mdx
```diff
@@ -14,6 +14,18 @@ import README from "../../../../packages/css/src/components/dialog/README.md?raw
...
+
+<Canvas of={DialogStories.FormDialog} className="ams-dialog-story" />
```
alimpens, line 27: This class seems to be unused.

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -65,9 +62,46 @@ export const Default: Story = {
   },
 }
 
+export const FormDialog: Story = {
+  args: {
+    open: true,
+    footer: (
+      <>
+        <Button type="submit" form="dialog1" value="continue">
```
alimpens, line 70: Do we need a `value` here? We don't really explain that anywhere

proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -3,11 +3,11 @@
...
       "background-color": { "value": "{ams.color.primary-white}" },
       "border": { "value": "0" },
-      "inline-size": { "value": "calc(100% - 2 * {ams.space.grid.md})" },
-      "max-block-size": { "value": "calc(100% - 2 * {ams.space.grid.md})" },
+      "inline-size": { "value": "calc(100dvw - 2 * {ams.space.grid.md})" },
       "max-inline-size": { "value": "48rem" },
-      "form": {
+      "content": {
         "gap": { "value": "{ams.space.md}" },
+        "max-block-size": { "value": "calc(100dvh - 4 * {ams.space.grid.md})" },
```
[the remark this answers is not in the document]
dlnr: Pretty great browser support? Usage relative is 92% and when is anything 100%? Even 'width' is 96%... 

DVH has great benefits: https://web.dev/blog/viewport-units
Onscreen keyboard behaviour is different per browser/device: https://developer.chrome.com/blog/viewport-resize-behavior#viewport_resize_behavior and I think we should test an input in the Dialog to test how this will work out before drawing conslusions. But it's possibly another reason not to touch the default display behaviour of the dialog.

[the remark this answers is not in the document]
dlnr: Also 100% does not seem to work well in safari iOS.
Simulator Screenshot - iPhone 15 - 2024-09-13 at 14 45 58

storybook/src/components/Dialog/Dialog.docs.mdx
dlnr: Do we have to capitalise every mention of a component?

storybook/src/components/Dialog/Dialog.stories.tsx
dlnr: Ah I should explain though, thats how form method dialog passes an optional value in `returnValue`.

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -23,18 +23,18 @@ const openDialog = (id: string) => (document.querySelector(id) as HTMLDialogElem
...
     ref: ForwardedRef<HTMLDialogElement>,
   ) => (
     <dialog {...restProps} ref={ref} className={clsx('ams-dialog', className)}>
-      <form className="ams-dialog__form" method="dialog">
+      <div className="ams-dialog__content">
         <header className="ams-dialog__header">
           <Heading size="level-4">{heading}</Heading>
           <IconButton label={closeButtonLabel} onClick={closeDialog} size="level-4" type="button" />
         </header>
         <article className="ams-dialog__article">{children}</article>
```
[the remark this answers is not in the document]
dlnr: Seemed like the way to go since it's possibly an indpendant item of content.

proprietary/tokens/src/components/ams/dialog.tokens.json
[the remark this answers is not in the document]
alimpens: I think the question we should answer here is: Do we want Dialog to change size when the keyboard is shown? I think it's a little odd, squashing the readable text when the keyboard appears. You'll especially notice that with longer texts.

storybook/src/components/Dialog/Dialog.docs.mdx
alimpens: [Yup](https://github.com/Amsterdam/design-system/blob/develop/documentation/storybook.md#structure:~:text=Component%20names%20are%20in%20title%20case%20%E2%80%93%20starting%20each%20word%20with%20a%20capital%20letter%20%E2%80%93%20to%20make%20people%20recognise%20them%20as%20such.)

storybook/src/components/Dialog/Dialog.stories.tsx
alimpens: I know, but we don't say that to our users. So now it's something we set without explaining why. And do we need it in this case? The default is an empty string, which is fine here because we don't use `returnValue`

proprietary/tokens/src/components/ams/dialog.tokens.json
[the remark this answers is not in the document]
dlnr: It does not seem to change size, opening the keyboard does not change the height of elements. The content of dialog wil continue to determine the size. Or am I missing something here? If you can show me a case where the dialog is squashed I can have a go at it.

storybook/src/components/Dialog/Dialog.stories.tsx
dlnr: We don't use it but if you're working on something functional you might need to know if the user clicks yes or no.

packages/react/src/Dialog/Dialog.test.tsx
```diff
@@ -66,13 +66,13 @@ describe('Dialog', () => {
...
+    const { getByText } = render(<Dialog heading="Test heading" footer={<button>Click Me</button>} />)
 
     expect(getByText('Click Me')).toBeInTheDocument()
   })
 
-  it('does not render actions when not provided', () => {
+  it('does not render footer when not provided', () => {
     const { queryByText } = render(<Dialog heading="Test heading" />)
 
     expect(queryByText('Click Me')).not.toBeInTheDocument()
```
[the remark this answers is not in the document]
dlnr: Good find

storybook/src/components/Dialog/Dialog.stories.tsx
alimpens: My point is: we should either explain what it does, or not use it at all. Using it without an explanation is confusing.

storybook/src/components/Dialog/Dialog.docs.mdx
```diff
@@ -14,11 +14,21 @@ import README from "../../../../packages/css/src/components/dialog/README.md?raw
 
 ## Examples
 
+### Form in a Dialog
+
+Set `method="dialog"` when using a form in Dialog.
+This closes the Dialog when submitting the form.
+Pass the submit Button to the `footer` prop,
+and link it to the form by passing its `id` to the Buttons `form` attribute.
+The Dialog returns the value of the submit Button, so you can check which Button was clicked.
```
alimpens, line 23: This is still pretty unclear. Maybe we should link to MDN?

```suggestion
The Dialog returns the value of the submit Button, so you can check which Button was clicked.
For more information, see [Handling the return value from the dialog (MDN)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#handling_the_return_value_from_the_dialog).
```

alimpens (approved):

PR #1593, opened 2024-09-20
feat!: Replace Aspect Ratio component with utility classes and prop on Image

RubenSibon:
# Describe the pull request

Simplifies giving (media) elements fixed two dimensional aspect ratios.

## What

Converts the Aspect Ratio component to a just tokens, CSS classes and an `aspectRatio` property on `Image`.

## Why

Simplification: less components and overhead by using component properties, CSS classes or tokens.

## How

Removing the Aspect Ratio React component with exception of the types.

## Checklist

- [x] Add or update stories

## Additional notes

- For now we are only placing the `aspectRatio` property on `Image`. It may be added to more (media) elements, such as `Video` in the future.
- The tokens and SCSS mixins may also come in handy in the upcoming `Figure` component.


packages/react/src/Image/Image.tsx
```diff
@@ -6,15 +6,25 @@
 import clsx from 'clsx'
 import { forwardRef } from 'react'
 import type { ForwardedRef, ImgHTMLAttributes } from 'react'
+import { AspectRatioProps } from '../AspectRatio'
 
-export type ImageProps = {
+export type ImageProps = AspectRatioProps & {
```
VincentSmedinga, line 11: Hmm… an image should always have an aspect ratio configured, right? It’s not like it’s okay to ‘forget’ choosing one and using an image of arbitrary size – then the entire idea of having a fixed set of allowed ratios is gone.

Let’s make 16/9 the default.

packages/css/src/components/image/README.md
```diff
@@ -15,7 +15,7 @@ Displays an image.
   For example, provide small, medium, and large variants for various screen sizes.
   This prevents unnecessary downloading of large files.
   Do this especially for the main image of a page, where the difference between sizes on a narrow and wide screen is most significant.
-- Ensure that the aspect ratio of the image is supported by the [Aspect Ratio](/docs/components-layout-aspect-ratio--docs) component.
+- Ensure that the aspect ratio of the image is supported by the [aspect ratio token](/docs/brand-design-tokens-aspect-ratio--docs). Alternatively, use the `cover` Image prop or the `object-fit: contain` CSS property to display the image correctly.
```
VincentSmedinga, line 18: Something like this? (Following from ideas in other comments.)

```suggestion
- The image has an aspect ratio of 16:9 by default. [Other options](/docs/brand-design-tokens-aspect-ratio--docs) are available.
```

VincentSmedinga (commented):
Looks like proof for the approach to me!

packages/css/src/components/image/image.scss
```diff
@@ -3,12 +3,20 @@
...
+
 .ams-image {
   block-size: auto; /* [1] */
   font-style: italic; /* [3] */
   max-inline-size: 100%; /* [1] */
   vertical-align: middle; /* [2] */
 
+  &[class*="ams-image-aspect-ratio--"] {
+    @include ams-aspect-ratio;
+  }
```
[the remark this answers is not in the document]
RubenSibon: I actually agree. Was just implementing the criteria literally.

packages/css/src/components/image/image.scss
```diff
@@ -3,12 +3,20 @@
...
+
+  @include generate-aspect-ratio-classes("-aspect-ratio");
+
   &--cover {
     object-fit: cover;
```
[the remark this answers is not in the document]
RubenSibon: Yeah, and I feel that adding `object-fit: contain` would be a good use-case as well. Why wasn't that added?

packages/css/src/components/aspect-ratio/_mixins.scss
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@mixin ams-aspect-ratio {
+  block-size: auto;
+  inline-size: fit-content;
+  overflow: hidden;
+  position: relative;
```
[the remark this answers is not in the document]
VincentSmedinga: Hopefully all of them 🤞🏼

packages/css/src/components/image/image.scss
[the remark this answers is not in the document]
VincentSmedinga: That will render the image smaller than its Grid Cell, which makes the alignment look off. Or do you see other use cases where this would be useful?

storybook/src/foundation/design-tokens/aspect-ratio.docs.mdx
```diff
@@ -0,0 +1,22 @@
+{/* @license CC0-1.0 */}
```
alimpens, line 1: Can you also add (probably very similar) docs to Utilities/CSS/Aspect Ratio?

RubenSibon: I was considering adding either this or both. It does feel as if it is doubling. Currently, we don't do that (have token and CSS docs). Any thoughts about that @VincentSmedinga?

packages/react/src/Image/Image.tsx
RubenSibon: Does everyone agree, @alimpens @dlnr?

This will make that `object-fit: cover` is also always applied as per [this comment](https://github.com/Amsterdam/design-system/pull/1593#discussion_r1813504108).

packages/css/src/components/aspect-ratio/_mixins.scss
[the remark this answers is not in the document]
RubenSibon: All of them have been move to `Image` or removed.

packages/react/src/Image/Image.tsx
alimpens: I think I agree. It has the consequence that images will be distorted if they do not have one of our allowed ratio's. But maybe that's preferable to not setting a ratio at all.

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { Paragraph } from '@amsterdam/design-system-react'
+import tokens from '@amsterdam/design-system-tokens/dist/index.json'
+import { Meta, StoryObj } from '@storybook/react'
+import { AspectRatio } from './AspectRatio'
+import type { AspectRatioProps } from './AspectRatio'
+
+const render = ({ aspectRatio }: AspectRatioProps) => (
+  <div className={`ams-aspect-ratio--${aspectRatio}`}>
+    <Paragraph
```
dlnr, line 14: The Paragraph as example is kind of confusing to me. I would expect an image since this is the only code example for aspect ratio and the most common usage is on an image. I would also like to know (personally) if it's ok to use the util to extend the `Image` class or that it still needs a wrapper.

dlnr (commented):
My comment also reflect changes that are needed in Figma, if an image can receive an aspect-ratio class Figmas image component will need an aspect ratio property.

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
VincentSmedinga: Changed to use a pink box.

dlnr (dismissed):

packages/css/src/components/image/README.md
```diff
@@ -4,6 +4,12 @@
 
 Displays an image.
 
+## Design
+
+Every Image must display in one of the [available aspect ratios](https://designsystem.amsterdam/?path=/docs/brand-design-tokens-aspect-ratio--docs).
```
alimpens, line 9:
```suggestion
Every Image should be shown in one of the [available aspect ratios](https://designsystem.amsterdam/?path=/docs/brand-design-tokens-aspect-ratio--docs).
```

storybook/src/foundation/design-tokens/aspect-ratio.docs.mdx
```diff
@@ -0,0 +1,32 @@
...
+# Aspect Ratio
+
+Constrains media content to a supported aspect ratio.
+
+## Design
+
+Media content like images, videos, and maps take up sizable portions of a screen.
+Displaying them in a limited set of aspect ratios keeps the visual impression of a page composed.
+The default aspect ratio is 16:9.
+Content with intrinsic dimensions that do not match an aspect ratio will be cropped without distorting its appearance.
```
alimpens, line 16:
```suggestion
Content with intrinsic dimensions that do not match an aspect ratio will be cropped.
```

packages/css/src/components/image/README.md
```diff
@@ -4,6 +4,12 @@
...
+The default is 16:9.
+If the intrinsic dimensions of the source do not match an aspect ratio, the image will get cropped to cover the intended area without distorting its appearance.
```
alimpens, line 11:
```suggestion
If the intrinsic dimensions of the source do not match an aspect ratio, the image will get cropped to cover the intended area.
```

VincentSmedinga (approved):

PR #1595, opened 2024-09-23
feat: Image slider

dlnr:
# Describe the pull request

The same image slider but with an object to define the images, this improves the rendering of thumbnails. Also disabled the scrolling of thumbnails, instead it will turn into a wrapped grid when you add more then 5 images. 

## What

- Object driven component
- Keyboard navigation for thumbnails
- multiple rows grid for thumbnails
- Improved aria roles

## Why

Having a slider with children complicated the markup and dupliating the thumbnails as buttons was an ugly solution.

## How

The slides object is fairly simple and the source of the image was now easy to use in a background-image for the button
The aspect-ratio class usage might change after Ruben's changes to the component.

## Checklist

- [X] Add or update stories

## Additional notes


packages/css/src/components/image-slider/image-slider.scss
```diff
@@ -63,30 +60,39 @@
...
-  overscroll-behavior-x: contain;
-  scroll-snap-type: x mandatory;
-  scrollbar-width: none;
+
+  // grid-auto-flow: column;
+  // grid-auto-columns: minmax(58px, 188px);
+  // overflow-x: auto;
+  // overscroll-behavior-x: contain;
+  // scroll-snap-type: x mandatory;
+  // scrollbar-width: none;
```
VincentSmedinga, line 72: Can we remove these?

```suggestion
```

packages/react/src/ImageSlider/ImageSlider.test.tsx
```diff
@@ -22,35 +20,52 @@ window.IntersectionObserver = jest.fn(() => ({
 }))
 
 describe('Image slider', () => {
+  const slides: SlideProps[] = [
```
VincentSmedinga, line 23: These are images. Let’s create a good distinction between Slides and Images – so rename ‘slide’ to ‘image’ where possible.

packages/react/src/ImageSlider/ImageSlider.test.tsx
```diff
@@ -22,35 +20,52 @@ window.IntersectionObserver = jest.fn(() => ({
...
-    const { container } = render(
-      <ImageSlider>
-        <ImageSlider.Item slideId={0}>child</ImageSlider.Item>
-      </ImageSlider>,
-    )
+  it('renders slides', () => {
+    const { container } = render(<ImageSlider slides={slides} />)
+
+    const slideElements = container.querySelectorAll('.ams-image-slider__item')
+    const slideArray = Array.from(slideElements)
```
VincentSmedinga, line 54: These could go without splitting?

```const slides = Array.from(container.querySelectorAll('.ams-image-slider__item'))```

File: packages/react/src/ImageSlider/ImageSlider.tsx [hunk 4]
```diff
@@ -5,40 +5,57 @@
...
 import { ImageSliderItem } from './ImageSliderItem'
 import { ImageSliderScroller } from './ImageSliderScroller'
 import { ImageSliderThumbnails } from './ImageSliderThumbnails'
+import { Ratio } from '../AspectRatio'
 import { IconButton } from '../IconButton'
+import { Image } from '../Image/Image'
+
+export type SlideProps = {
+  src: string
+  /** Prove a URL to the image */
```
VincentSmedinga, line 20: JSDoc comments should go above the prop they describe :)

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -5,40 +5,57 @@
...
 import { ImageSliderContext } from './ImageSliderContext'
 import { ImageSliderItem } from './ImageSliderItem'
 import { ImageSliderScroller } from './ImageSliderScroller'
 import { ImageSliderThumbnails } from './ImageSliderThumbnails'
+import { Ratio } from '../AspectRatio'
 import { IconButton } from '../IconButton'
+import { Image } from '../Image/Image'
+
+export type SlideProps = {
+  src: string
```
VincentSmedinga, line 19: All these props are passed into `Image` and `AspectRatio` components. We should reuse their types to prevent future changes introducing issues here, e.g. `src: ImageProps['src']`.

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -5,40 +5,57 @@
...
+  /** Label for the image if you need to translate the alt text */
+  imageLabel?: string
+} & HTMLAttributes<HTMLDivElement>
 
 export const ImageSliderRoot = forwardRef(
   (
     {
-      children,
       className,
+      slides,
```
VincentSmedinga, line 52: Please alphabetically sort all lists like this, such as fields in types, restructuring assignments, function parameters, etc.

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -69,19 +86,64 @@ export const ImageSliderRoot = forwardRef(
...
+      const handleKeyDown = (event: KeyboardEvent) => {
+        switch (event.key) {
+          case 'ArrowLeft':
+            goToPreviousSlide()
+            break
+          case 'ArrowRight':
+            goToNextSlide()
+            break
+        }
+      }
```
VincentSmedinga, line 98: Can we use a lookup object instead of a switch?

```suggestion
      const keyActionMap = {
        ArrowLeft: goToPreviousImage,
        ArrowRight: goToNextImage,
      }

      const handleKeyDown = (event: KeyboardEvent) => {
        const action = keyActionMap[event.key as keyof typeof keyActionMap]
        action?.()
      }
```

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -69,19 +86,64 @@ export const ImageSliderRoot = forwardRef(
...
+
       if (targetRef.current) {
         const slides = targetRef.current.children
         const slidesArray = Array.from(slides)
```
VincentSmedinga, line 102: This can be combined as well:

```const slides = Array.from(targetRef.current.children)```

File: packages/react/src/ImageSlider/ImageSlider.tsx [hunk 9]
```diff
@@ -69,19 +86,64 @@ export const ImageSliderRoot = forwardRef(
...
     const synchronise = () => {
       updateControls()
     }
 
+    const handleThumbsKeyDown = (event: ReactKeyboardEvent<HTMLElement>) => {
+      const target = event.target as HTMLElement
+      const element = target.parentElement?.children[currentSlideId]
+
+      if (event.key === 'ArrowRight') {
+        const next = element?.nextElementSibling as HTMLElement | null
```
VincentSmedinga, line 126: If we give all element variables a suffix of `El` or `Element,` the code will be easier to read. (Lots of occurrences.)

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -126,30 +188,23 @@ export const ImageSliderRoot = forwardRef(
...
             controls && 'ams-image-slider--controls',
             scrollbar && 'ams-image-slider--scrollbar',
-            snapstop && 'ams-image-slider--snapstop',
             thumbnails && 'ams-image-slider--thumbnails',
             className,
           )}
         >
           {controls && (
-            <div className="ams-image-slider__controls" onKeyDown={handleKeyDown}>
+            <div className="ams-image-slider__controls">
```
VincentSmedinga, line 207: We could extract a small component for these buttons as well, just like for the scroller and the thumbnails.

packages/react/src/ImageSlider/ImageSlider.tsx
```diff
@@ -111,11 +173,11 @@ export const ImageSliderRoot = forwardRef(
 
     const goToPreviousSlide = () => {
```
VincentSmedinga, line 174: Can some or all of these functions be moved outside of the component? They keep getting redefined like this. Some will need an extra parameter.

packages/react/src/ImageSlider/ImageSlider.test.tsx
dlnr: A `ImageSliderItem` is a wrapper for a image, and i would like to prevent naming it "ImageSliderImageProps", the parent is called a "Slider" and a child would be a "Slide"? Which is also weird, so I settled on "Item".

Looking at the naming and hierarchy I would suggest using:

- ImageSlider
  - ImageSliderScroller
    - ImageSliderItem
        - Image (import) 
  - ImageSliderThumbnails
      - ImageSliderThumbnail 

I tried to be consistent in naming, using the component name a prefix for its children (excluding imported children). I'm not a fan of diving in and renaming one thing because of semantics.

Or rename the whole thing to what it is, a Carousel. [Here is an article](https://www.smashingmagazine.com/2023/02/guide-building-accessible-carousels/) on a carousel but they name the children "Slide".

packages/react/src/ImageSlider/ImageSlider.tsx
dlnr: This doesn't improve readability.

packages/react/src/ImageSlider/ImageSlider.tsx
dlnr: The observers are in `useEffect` to prevent rerenders. And alle these functions are in here to prevent juggling that ref around. I'm not sure how `useMemo` would help. This is to cache results between re-renders?

The component would have been much cleaner in vanilla javascript, "react-izing" caused a lot of headache so I'm open to suggestions.

packages/react/src/ImageSlider/ImageSlider.tsx
dlnr: An ImageSliderControlsButton you mean or the whole controls component?

packages/react/src/ImageSlider/ImageSlider.tsx
VincentSmedinga: Well, it’s not that difficult to read (it appears that the cast wasn’t needed, so I removed it), while it does improve maintainability. And the actions aren’t undefined; maybe you tried after renaming the functions? I updated these as well.

packages/react/src/ImageSlider/ImageSlider.test.tsx
VincentSmedinga: It’s not a Carousel – it has no text content and doesn’t automatically slide.

`ImageSliderItem` is fine.

dlnr:
Having some trouble moving slides to a child component, also not sure if moving all everything to a shared context is more performant. We discussed this and decided to optimise this later.

packages/css/src/components/image-slider/image-slider.scss
```diff
@@ -47,6 +48,10 @@
   grid-column: 1/-1;
   grid-row: 1;
   justify-content: space-between;
+
+  @media (pointer: coarse) and (max-width: $ams-breakpoint-medium) {
+    display: none;
+  }
```
RubenSibon, line 54: Interesting find. Something to remember.

dlnr: The breakpoint is something we don't use elsewhere, we use `min-width`, but it saves a bunch of lines.

dlnr: Is this something I need to resolve?

packages/react/src/ImageSlider/ImageSlider.tsx
VincentSmedinga: Done.

VincentSmedinga (approved):
Awesome!

PR #1607, opened 2024-09-27
feat!: Update behaviour of Header and related components

VincentSmedinga:
This implements the updated design for the Header.

It’s a significant PR that contains breaking changes for various components. It might be best to review this PR globally for the general approach and then extract smaller, focused PRs. This makes the detailed reviews more pleasant, updates the change log correctly, and helps our users understand and implement the changes.

Review my comment at the bottom to see the three approaches I considered and how I chose the winner.

UX agrees with the changes this PR introduces to the design – most notably, the generous white space – and will update Figma accordingly.

The big chunks that could be the new PRs:
- Allow Grid and Grid Cell to render more tags
- Allow the Logo to grow a bit
- Introduce Item and Button for Page Menu
- Implement new Header design
- Use the new Header in Home Page example

These things remain to be done, preferably in separate PRs as well:
- complete a11y – at least some accessible names are missing
- render a narrow logo of just the St. Andrew’s Crosses
- extract the `AppHeader` in the Amsterdam.nl home page to a pattern
- test the updated Header design with the menus of existing sites
- remove the Mega Menu component

Other tasks:
- Include missing license headers (test files and some others)
- Use HTMLElement and HTMLDivElement correctly
- Decorate all stories with a Screen


storybook/src/components/Header/Header.stories.tsx
```diff
@@ -3,14 +3,27 @@
...
 const meta = {
   title: 'Components/Containers/Header',
   component: Header,
+  decorators: [
+    (Story) => (
+      <Screen>
+        <Story />
+      </Screen>
+    ),
+  ],
```
dlnr, line 57: To prevent the body from having padding, you can't use viewport settings now to see phone screens for example.

```suggestion
  decorators: [
    (Story) => (
      <Screen>
        <Story />
      </Screen>
    ),
  ],
  parameters: {
    layout: 'fullscreen',
  },
```

storybook/src/components/Header/Header.stories.tsx
```diff
@@ -3,14 +3,58 @@
...
+        Mijn Amsterdam
+      </PageMenu.Link>
+    </PageMenu.Item>
+  </>
+)
+
 const meta = {
   title: 'Components/Containers/Header',
   component: Header,
+  argTypes: {
```
dlnr, line 45: Thought we agreed on the previous title to test the longest possible site name. It's also nicer to test the linebreak.

storybook/src/components/Header/Header.stories.tsx
```diff
@@ -3,14 +3,58 @@
...
+    as: {
+      control: { type: 'radio' },
+      options: ['header', 'div'],
+    },
```
dlnr, line 49: In case you every want to add a story that opens the menu? 

```suggestion
export const WithPageMenu: Story = {
```

storybook/src/components/Header/Header.stories.tsx
```diff
@@ -29,64 +73,62 @@ export const ForSubBrand: Story = {
...
+        <PageMenu.Item>
+          <PageMenu.Link href="#">Menu link 1</PageMenu.Link>
+        </PageMenu.Item>
+        <PageMenu.Item>
+          <PageMenu.Link href="#">Menu link 2</PageMenu.Link>
+        </PageMenu.Item>
+        <PageMenu.Item>
+          <PageMenu.Link href="#">Menu link 3</PageMenu.Link>
+        </PageMenu.Item>
       </PageMenu>
```
dlnr, line 93: Perhaps you could use real links like the other stories? 
```suggestion
        <PageMenu.Link href="#">Contact</PageMenu.Link>
        <PageMenu.Link href="#">Mijn Amsterdam</PageMenu.Link>
        <PageMenu.Link href="#">Nieuws</PageMenu.Link>
```
And it might help to test 5 links, which is the advised maximum amount of links in this navigation element.

storybook/src/components/Header/Header.stories.tsx
```diff
@@ -29,64 +73,62 @@ export const ForSubBrand: Story = {
...
+export const WithMenu: Story = {
   args: {
-    links: (
-      <PageMenu alignEnd wrap={false}>
-        <PageMenu.Link href="#">Contact</PageMenu.Link>
-        <PageMenu.Link href="#">Mijn Amsterdam</PageMenu.Link>
-        <PageMenu.Link href="#" icon={SearchIcon}>
-          Zoeken
-        </PageMenu.Link>
+    menu: (
```
dlnr, line 82: Very unfortanate Storybook shows it all on one line in the code preview.

dlnr (commented):
Some comments on the Header stories

packages/react/src/Header/Header.tsx
```diff
@@ -11,54 +11,49 @@ import { Logo } from '../Logo'
...
+  /** A secondary menu of links and buttons. Use a Page Menu. Can include buttons to show the Mega Menu or a Search Field. */
   menu?: ReactNode
-} & HTMLAttributes<HTMLElement>
+} & HTMLAttributes<HTMLDivElement>
 
 export const Header = forwardRef(
   (
     {
       appName,
+      as: Tag = 'div',
```
dlnr, line 32: This is not always a `<header>`? There are also a11y factors to consider, like the definition of the banner role. So you would need a `<div role="banner">` to replace it correctly.

packages/react/src/Header/Header.tsx
```diff
@@ -11,54 +11,49 @@ import { Logo } from '../Logo'
...
+    ref: ForwardedRef<HTMLDivElement>,
+  ) => (
+    <Tag {...restProps} ref={ref} className={clsx('ams-header', className)}>
+      <div className="ams-header__section">
+        <a className="ams-header__logo-link" href={logoLink}>
+          <span className="ams-visually-hidden">{logoLinkTitle}</span>
+          <Logo brand={logoBrand} />
+        </a>
+        {appName && (
+          <Heading level={1} size="level-5">
```
dlnr, line 49: The size 'level-5' does not resemble the design and moves the logo down when the text wraps to a new line.

packages/react/src/Header/Header.tsx
```diff
@@ -11,54 +11,49 @@ import { Logo } from '../Logo'
...
-            </div>
-          )}
-        </header>
-      </>
-    )
-  },
+    ref: ForwardedRef<HTMLDivElement>,
+  ) => (
+    <Tag {...restProps} ref={ref} className={clsx('ams-header', className)}>
+      <div className="ams-header__section">
```
dlnr, line 43: Do you really need a wrapping div? You could give the logo, app-name and menu their own flex properties to achieve the same effect. Perhaps some `display: contents` magic?

packages/css/src/components/page-menu/page-menu.scss
```diff
@@ -15,6 +16,17 @@
   box-sizing: border-box;
 }
 
+@mixin reset-button {
```
dlnr, line 19: This might be a good reason the make a reusable mixin for buttons?

packages/css/src/components/page-menu/page-menu.scss
```diff
@@ -28,14 +40,19 @@
...
+.ams-page-menu__item {
+  &--secondary {
+    @media screen and (not (min-width: $ams-breakpoint-medium)) {
+      display: none;
+    }
+  }
 }
 
-@mixin page-menu-item {
+@mixin ams-page-menu-link-or-button {
```
dlnr, line 55: I understand you want to improve the semantics but you've only renamed this mixin, the tokens are still called `-ams-page-menu-item`, so it increases the confusion for me. Ie. Are we targetting the li or the button/link? And what if you ever add a username dropdown to the mix?

packages/css/src/components/header/header.scss
```diff
@@ -6,81 +6,19 @@
...
-  @media screen and (min-width: $ams-breakpoint-wide) {
-    column-gap: var(--ams-header-column-gap);
-    flex-wrap: nowrap;
-  }
+.ams-header__section {
+  align-items: center;
+  column-gap: var(--ams-header-section-column-gap);
+  display: flex;
+  flex: auto;
 }
```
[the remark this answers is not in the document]
VincentSmedinga: I'm unsure what you mean by ‘aligned on the left side’. Both sections can grow, so either they’ll sit next to each other or be stacked 100% wide.

I agree that the two halves can look off if they only overlap a bit in the middle, but it does conform to the current design. We may iterate later.

packages/css/src/components/page-menu/page-menu.scss
VincentSmedinga: It might. I’ll add a separate ticket for it.

packages/css/src/components/page-menu/page-menu.scss
VincentSmedinga: Hmm, yes, the token names… The links and buttons share many styles, so keeping the mixin seemed worthwhile. Items (`li`) have always been the parent element of both, so it was a bit confusing already.

We could go with `ams-page-menu-element` to target links, buttons, and future elements that look the same?

packages/react/src/Header/Header.tsx
VincentSmedinga: No, this approach separates the `Header` and `Mega Menu` components entirely and leaves it to the app to let the menu button in the header toggle the menu's visibility.

And because the menu should be part of the HTML `header`, the `Header` component can no longer render that element – unless there is no Mega Menu, in which case people can do `<Header as="header">`.

The example Home Page already uses this approach, and I’ll document it further.

packages/react/src/Header/Header.tsx
VincentSmedinga: It seems more robust to encode the two sections that the header clearly has. I tried having logo, app name and page menu as direct children of the menu and concluded:

- This allows the app name to wrap below the logo instead of wrapping its text
- It requires left margins on the app name and the page menu, which couples whitespace to semantic elements. Coupling it to structural sections allows for different elements while keeping spacing intact.

packages/react/src/Header/Header.tsx
dlnr: This could solve it for now, Grid also defaults to div so not needed.
```suggestion
      as: Tag = 'header',
```

packages/react/src/Header/Header.tsx
VincentSmedinga: Bas and I changed this together; he will update the design.

I don’t see the logo moving down?

storybook/src/components/Header/Header.stories.tsx
VincentSmedinga: The ‘English’ and ‘Mijn Amsterdam’ items not disappearing was due to a different bug. I’d rather not use fullscreen layout here because it removes the padding that the Header will have in real life (even though its sizes will be different) and because this `Header` is not the entire header – it’s just a container that still needs a parent grid for the correct layout. The media query may kick in 20 pixels late or early, but that’s acceptable.

packages/css/src/components/logo/logo.scss
```diff
@@ -6,6 +6,7 @@
 .ams-logo {
   block-size: var(--ams-logo-block-size);
   display: block;
+  min-block-size: var(--ams-logo-min-block-size);
```
dlnr, line 9: Shall we try a container query to show just the crosses on narrow screens?

packages/react/src/Grid/Grid.tsx
```diff
@@ -29,17 +29,28 @@ type GridPaddingTopAndBottomProps = {
 }
 
 export type GridProps = {
+  /** The HTML element to use. */
+  as?: 'header' | 'div' | 'footer'
```
dlnr, line 33: This breaks the forwardRef test because it's only for a div. Perhaps we needs something like this: https://medium.com/@yishayhaz/how-to-use-as-prop-in-react-typescript-2c310eb8ee1e

packages/react/src/Header/Header.tsx
VincentSmedinga: The typical setup will combine a `Header` with a `MegaMenu`. Making `header` the default for the former sets people up for mistakes.

The Grid wrapping the Header and the Mega Menu will need to render the `header`.

Only if a site doesn’t use a Mega Menu (or Search button in the Page Menu, or anything else) must the Header render a `header`. Having the least common option as the default will set people up for mistakes.

packages/react/src/Grid/Grid.tsx
VincentSmedinga: I’ve updated to type the ref as `any` now, as we [decided a while ago](https://github.com/Amsterdam/design-system/pull/1140/files), but it looks like we can give the approach in that article a try.

packages/react/src/Header/Header.tsx
dlnr: I understand, in my previous attempts I solved this with grid areas.

packages/css/src/components/page-menu/page-menu.scss
dlnr: Sure.

storybook/src/components/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -16,7 +16,9 @@ const meta = {
   decorators: [
     (Story) => (
       <Screen>
-        <Story />
+        <Grid paddingVertical="medium">
```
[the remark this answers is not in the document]
VincentSmedinga: Hmm, tricky. I’ve reverted the change to the Mega Menu stories. This required adding an extra Grid Cell to the Home Page example, but that’s not how we want it.

packages/react/src/Header/Header.tsx
VincentSmedinga: > And because the menu should be part of the HTML header

It turns out this is not required – the Header can render a `<header>` and the Mega Menu a `<nav>` below it.

packages/css/src/components/logo/logo.scss
VincentSmedinga: Sounds interesting, but what would you want the container query to test that a ‘regular’ window width query can’t? Is it to consider the presence of an app title?

packages/react/src/Header/Header.tsx
dlnr: Perhaps it's still in my old PR's, I was told both the menu button, page-menu and mega-menu should reside in the same nav element, like gov.uk does it. This presented my with a lot of challenges so the PR was dropped. I'm curious what changed? @alimpens could you elaborate?

dlnr: Found the comment with the link: https://github.com/Amsterdam/design-system/pull/992#discussion_r1445836227

packages/css/src/components/header/header.scss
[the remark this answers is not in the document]
dlnr: Commit [8220373](https://github.com/Amsterdam/design-system/commit/8220373993d07de6f9808ec0cb994ab49af4de04) fixed it.

packages/css/src/components/logo/logo.scss
dlnr: The header has a max width so it would make sense I suppose. You could use a `:has()` selector in combination with a container (or normal) query to hide the logo text. So probably not in this file but that defies the purpose of having isolated components, so targeting logo in the header CSS is wrong but needed? That's probably why it should be a pattern and not a component?

storybook/src/components/MegaMenu/MegaMenu.stories.tsx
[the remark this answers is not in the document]
dlnr: Ah did not even check the homepage example. I must confess I'm not sure how you want to structure the a page now, it might help to know. Does this come close?

```jsx

<Screen>
  <Header {...props}>
    <MegaMenu>
       <Grid>
          <MenuItems />
       </Grid>
    </MegaMenu>
  </Header>
  <Grid>
    <Content />
  </Grid>
  <Footer>
    <FooterMenuItems />
  </Footer>
</Screen>

```

The polymorphic grid is going to throw (developer) people off, a Grid should not nest another Grid right?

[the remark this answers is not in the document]
dlnr: This is the reason, you are going to get misaligned grids:

image

VincentSmedinga:
I’ve considered 3 approaches for the structure of HTML and components for the header, mega menu and search bar. For simplicity, the examples below do not contain the search bar, and only the menu button is specified for the Page Menu.

My requirements were:
1. Optimal HTML, accessibility and SEO
2. Use existing components where possible
3. A ‘natural’ API; not too many tricks

**Approach 1**: one explicit grid to contain everything

```tsx
<Grid paddingVertical="medium">
  <Grid.Cell span="all">
    <Header menu={<PageMenu.Button>Menu</PageMenu.Button>} />
  </Grid.Cell>
  <nav className="ams-visually-hidden ams-display-contents">
    <Grid.Cell>
      <Columns>…</Columns>
    </Grid.Cell>
    <Grid.Cell>…</Grid.Cell>
  </nav>
</Grid>
```

\+ One grid, no `as` props
\+ Mega Menu is a pattern, not a component
\+/- An independent component for CSS multi-columns
\- Requires `display: contents`, unsure about browser support

**Approach 2**: two explicit grids

```tsx
<Grid paddingVertical="medium">
  <Grid.Cell span="all">
    <Header menu={<PageMenu.Button>Menu</PageMenu.Button>} />
  </Grid.Cell>
</Grid>,
<Grid as="nav" className="ams-visually-hidden" paddingBottom="large">
  <Grid.Cell>
    <Columns>…</Columns>
  </Grid.Cell>
  <Grid.Cell>…</Grid.Cell>
</Grid>
```

\+ Existing and logical API
\+ Mega Menu is a pattern, not a component
\+ Doesn’t need `display: contents`
\+ Easy to increase bottom padding for menu alone
\+/- An independent component for CSS multi-columns
\- A bit more HTML (just one tag actually)

**Approach 3**: two implicit grids

```tsx
<Header menu={<PageMenu.Button>Menu</PageMenu.Button>} />,
<MegaMenu className="ams-visually-hidden ams-display-contents" key={2}>
  <Grid.Cell>
    <Columns>…</Columns>
  </Grid.Cell>
  <Grid.Cell>…</Grid.Cell>
</MegaMenu>,
```

`Header` renders `<Grid><Grid.cell span="all"><header class="ams-header" ref><Logo /><Heading />{menu}</header></Grid.Cell></Grid>`

`MegaMenu` renders `<Grid as="nav" className="ams-mega-menu" paddingBottom="large">{children}</Grid>`

\+ Simple outer API: Header and MegaMenu
\+ MegaMenu.Section could implement CSS multi-col locally
\- Two components render a Grid as well; dependencies, weird classes and refs
\- Strange API: Grid.Cell as children of MegaMenu 
\- Requires `display: contents`, unsure about browser support

I chose approach 2.

storybook/src/components/MegaMenu/MegaMenu.stories.tsx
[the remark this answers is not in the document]
dlnr: Fixed after the latest commits 👍 

The only issue is the wrapping `<nav>` element but this is hard.

storybook/src/components/MegaMenu/MegaMenu.stories.tsx
```diff
@@ -28,47 +25,44 @@ type Story = StoryObj<typeof meta>
...
-            </LinkList>
-          </MegaMenu.ListCategory>
-        </Grid.Cell>
-      </Grid>
-    ),
+    children: [
+      <Heading level={1} size="level-2">
+        Alle onderwerpen
+      </Heading>,
+      <MegaMenu.ListCategory>
```
dlnr, line 32: MegaMenu is deprecated??

storybook/src/pages/amsterdam/common/AppHeader.tsx
```diff
@@ -1,13 +1,74 @@
...
+
+type Feature = 'menu' | 'search'
+
+export const AppHeader = () => {
+  const [visibleFeature, setVisibleFeature] = useState<Feature | undefined>(undefined)
+
+  const toggleFeature = (feature: Feature) =>
+    setVisibleFeature((currentFeature) => (feature === currentFeature ? undefined : feature))
+
+  const visibilityClass = (feature: Feature) => (feature === visibleFeature ? null : 'ams-visually-hidden')
```
dlnr, line 19: Is `ams-visually-hidden` the way to go? Yes you'll hide it but for the visually impaired, they will get the menu read out loud to them everytime. It should probably just use `display: none` and the menu (toggle) button should have `aria-controls="mega-menu-id"` so that the menu gets read to them when they toggle the menu.

storybook/src/pages/amsterdam/common/AppHeader.tsx
```diff
@@ -1,13 +1,74 @@
...
+            Menu
+          </PageMenu.Button>
+        </PageMenu.Item>
+      </PageMenu>
+    ),
+    [toggleFeature],
+  )
+
+  const Menu = () => (
+    <Grid as="nav" className={visibilityClass('menu')} paddingBottom="large">
```
dlnr, line 46: Getting some ts errors here, no `nav` in `as` property and possible `null` for `className`.

proprietary/tokens/src/components/ams/logo.tokens.json
```diff
@@ -1,8 +1,9 @@
 {
   "ams": {
     "logo": {
-      "block-size": { "value": "2.5rem" },
+      "block-size": { "value": "{ams.space.grid.md}" },
       "emblem": { "color": { "value": "{ams.color.primary-red}" } },
+      "min-block-size": { "value": "2.5rem" },
```
dlnr, line 6: This is a conflict with `block-size` which has a min size of 1rem.

storybook/src/pages/amsterdam/common/AppHeader.tsx
dlnr: Even Copilot agrees:

Using "visually hidden" to hide the megamenu may be a poor decision due to the following reasons:

1. Accessibility Issues: "Visually hidden" is intended for content that is hidden from sighted users but accessible to screen readers. Using it to hide interactive elements like a megamenu can confuse screen reader users as they might access the menu without visual context.
2. User Experience: Sighted users may face unexpected behavior if the megamenu is hidden visually but still interactable. This can lead to confusion and a poor user experience.
3. Best Practices: The repository guidelines suggest visually available content should also be accessible to non-visual user agents and vice versa. Hiding the megamenu only visually contradicts this practice.

dlnr: Pretty cool it even refers to your own documentation

storybook/src/components/MegaMenu/MegaMenu.stories.tsx
VincentSmedinga: Yes. It has never done much; see the low amount of tokens and CSS. The outer `nav` will now be rendered through a `Grid`; its only remaining feature is the CSS columns, for which we can make an independent `Columns` component.

proprietary/tokens/src/components/ams/logo.tokens.json
VincentSmedinga: The two are intended to cooperate: the minimum height remains ‘40px’ as it was, but the logo is allowed to grow to ‘56px’. The result is 33% top padding, 33% logo, and 33% bottom padding (vertically).

storybook/src/pages/amsterdam/common/AppHeader.tsx
VincentSmedinga: The first should be cache; see `Grid.tsx`. Fixed the second.

storybook/src/pages/amsterdam/common/AppHeader.tsx
VincentSmedinga: Correct, I haven’t implemented all accessibility features yet; I wanted to get the team’s feedback on this approach and API first. We will ensure that both sighted and non-sighted people can navigate properly. Screen reader users will want to hear the items, or they can use the Skip Link.

Someone suggested the Mega Menu must be in the DOM for search engines. That could well be – we should research that. Of course, we could employ other strategies like XML sitemaps, but I don’t expect they’ll allow us to hide the main menu for crawlers entirely.

I do remove the Search bar from the DOM, so I have considered it.

> Pretty cool it even refers to your own documentation

Yes, it is!

packages/react/src/Header/Header.tsx
```diff
@@ -30,35 +28,29 @@ export const Header = forwardRef(
...
+          <span className="ams-visually-hidden">{logoLinkTitle}</span>
+          <Logo brand={logoBrand} />
+        </a>
+        {appName && (
+          <Heading className="ams-header__app-name" level={1} size="level-5">
+            {appName}
+          </Heading>
+        )}
+      </div>
+      <div className="ams-header__section">{menu}</div>
```
alimpens, line 51: This should be a `nav`, I think. The main navigation section is the Page Menu and the Mega Menu together.

storybook/src/components/Header/Header.stories.tsx
```diff
@@ -29,64 +65,62 @@ export const ForSubBrand: Story = {
...
   args: {
-    appName: 'Aan de Amsterdamse grachten',
-    menu: <button className="ams-header__menu-button">Menu</button>,
+    appName: 'Onderzoek en Statistiek',
+    menu: <PageMenu alignEnd>{MenuButton}</PageMenu>,
   },
 }
 
-export const WithAppNameLinksAndMenuButton: Story = {
+export const WithAppNameMenuAndMenuButton: Story = {
```
alimpens, line 115: Do we want to add a story here for the header + mega menu? It'd be nice to see the whole thing together, and I think I would look for it here.

storybook/src/pages/amsterdam/common/AppHeader.tsx
```diff
@@ -1,13 +1,74 @@
...
+
+type Feature = 'menu' | 'search'
+
+export const AppHeader = () => {
+  const [visibleFeature, setVisibleFeature] = useState<Feature | undefined>(undefined)
+
+  const toggleFeature = (feature: Feature) =>
+    setVisibleFeature((currentFeature) => (feature === currentFeature ? undefined : feature))
+
+  const visibilityClass = (feature: Feature) => (feature === visibleFeature ? undefined : 'ams-visually-hidden')
```
alimpens, line 19: I just tested this with a screen reader, it's kinda of annoying that it always reads out the mega menu, even if it's collapsed. It also messes up your tab index. I think we should just use `display: none` here, same as gov.uk. Do you have a link to the comment where someone said this was a good idea?

packages/react/src/Header/Header.tsx
VincentSmedinga: Yes, but I think it would be best to let `PageMenu` render a `nav` as its root tag. This would add that semantics to the footer menu as well.

I think the Page Menu and the Main Menu should be separate navigation sections. Screen reader users can then select which menu they want to explore.

storybook/src/components/Header/Header.stories.tsx
VincentSmedinga: I want to make that the first entry in a new ‘Patterns’ section. Header + Mega Menu is a composition with some state: a pattern.

(The confusion may be that our Header component is just Logo + Heading + Page Menu, while visually the Main Menu and the Grid whitespace around it would conceptually belongs to the (application) header as well. I considered renaming Header to Masthead or Navigation Bar or something.)

Another reason is that this will give us room to present a couple of variations for larger and smaller sites, based on the navigation of some of our current sites.

storybook/src/pages/amsterdam/common/AppHeader.tsx
VincentSmedinga: The ‘SEO’ comment is kind of [concise](https://codefornl.slack.com/archives/C01DAT4TRPF/p1727947636520699?thread_ts=1727946073.769069&cid=C01DAT4TRPF) :)

Robbert has [more on making header and navigation accessible](https://codefornl.slack.com/archives/C01DAT4TRPF/p1727959649493419?thread_ts=1727946073.769069&cid=C01DAT4TRPF).

packages/react/src/Header/Header.tsx
alimpens: > I think the Page Menu and the Main Menu should be separate navigation sections

I think that's a little odd, semantically. The entire thing is our main navigation section imo, I don't really see why we should separate them. The mega menu is just a sub menu of the main menu.

It also goes against the whole '[use landmark roles sparingly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#accessibility_concerns)' thing.

See also these examples, although the last 2 use the `nav` tag a bit less sparingly than I would:

- https://www.gov.uk/
- https://adobe-accessibility.github.io/Accessible-Mega-Menu/
- https://design-system.w3.org/code/header-nav.html
- https://www.rvo.nl/

storybook/src/components/Header/Header.stories.tsx
alimpens: I think this depends on where we'll end up with our semantic definition of the Header. I think our Header contains our main navigation, which is comprised of the Page Menu and the Mega Menu. If we go with that definition, then the Mega Menu is part of the Header.

storybook/src/pages/amsterdam/common/AppHeader.tsx
alimpens: Ha, that's a little hard to put into context. One thing I can comment on: Using `display: none` is not the same as not having the menu in the DOM. I agree it should be in the DOM, but visually hidden causes too many a11y issues imo.

storybook/src/pages/amsterdam/common/AppHeader.tsx
dlnr: Search engines execute javascript nowadays and are able to crawl the site without the help of the navigation. The child links of the Google search results are based on relevance, not the sitemap or main nav. Forcing visually impaired users to always use the skip link is not the way to go.

VincentSmedinga:
Superseded by #1784.

PR #1628, opened 2024-10-02
fix: Disallow invalid number of columns for narrow and medium Grids

RubenSibon:
# Describe the pull request

There is a undesirable feature in `Grid` that allows setting a higher number of columns than is possible for that specific breakpoint.

## What

Do not allow developers to set an invalid number of columns per media breakpoint on a `Grid.Cell` in a `Grid` via TypeScript.

## Why

It restricts the freedom of developers to write invalid code.

## How

Added separate `GridNarrowColumnNumber` and `GridMediumColumnNumber` types.

## Checklist

- [x] Add or update stories

## Additional notes

- Documentation or stories do not need to be updated because8 they already state the valid number of columns per breakpoint.
- I couldn't think of a way to use unit tests to test this.


VincentSmedinga:
The CSS classes we render from these props can receive the same treatment. Although, the last time I tried, I ran into a gotcha that I can’t remember now. Please perform sufficient visual tests on different window widths with high and low values, single and triplets, etc.

VincentSmedinga:
Oh, and please do the same for `SpotlightGrid` after #1629 has been merged.

RubenSibon:
See: https://css-tricks.com/preventing-a-grid-blowout/

packages/react/src/Breakout/BreakoutCell.tsx
```diff
@@ -21,9 +26,9 @@ type BreakoutCellSpanAllProp = {
 
 type BreakoutCellSpanAndStartProps = {
   /** The amount of grid columns the cell spans. */
-  colSpan?: 'all' | GridColumnNumber | GridColumnNumbers
+  colSpan?: 'all' | GridNarrowColumnNumber | GridMediumColumnNumber | GridWideColumnNumber | GridColumnNumbers
```
VincentSmedinga, line 29: This has overlap: 

We discussed `Pick<>`, but that’s for object types, not literals.

This may be an option:

```
type Subset<T, U> = T extends U ? T : never

export type GridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12

export type GridColumnNumbers = {
  narrow: Subset<GridColumnNumber, 1 | 2 | 3 | 4>
  medium: Subset<GridColumnNumber, 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8>
  wide: GridColumnNumber
}
```

Types that accept all three variations can keep using `GridColumnNumber`.

VincentSmedinga: This is an alternative, but seems overkill:

```ts
type Range<N extends number, Result extends Array<unknown> = []> = Result['length'] extends N
  ? Result[number]
  : Range<N, [...Result, Result['length']]>

export type GridColumnNumber = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12

export type GridColumnNumbers = {
  narrow: Range<4>
  medium: Range<8>
  wide: GridColumnNumber
}
```

VincentSmedinga: Subset seems best?

RubenSibon: Well, I like how simple it is to use `Range` even if the definition is more complex.

I did something similar in the experimental branch: https://github.com/Amsterdam/design-system/blob/9cad4c476dba00f5ec985a4b6fb538758ca0304f/packages/react/src/Grid/Grid.tsx#L12

RubenSibon: In the end chose the KISS solution. :-)

VincentSmedinga (approved):

PR #1654, opened 2024-10-09
feat: Allow Button to display an icon without a label

dlnr:
# Describe the pull request

While the story wants the tertiary button to be able to replace the IconButton component I've tied the [Figma button](https://www.figma.com/design/2epLNj5v1HrQ1lbOjAa7GU/DES-859-%7C-Icon-button?node-id=15700-53370&m=dev) API to the code. Because narrowing the "icon only" variant to just the tertiary button would require it be a separate component or have dodgy conditionals. Also the map wil need a secondary icon only variant or a button group of icon buttons might need a call to action.

## What

I've reverted the changes to the IconPosition enum (not using iconStart and iconEnd) because using code connect this will match the React code snippet ,it saves a couple of properties and prevents the ability to have two icons in a button. It does introduce more variants in Figma, but this should be fine.

The label property was also not needed and the "icon only" variant introduces a span around the children with a visually hidden class.

## Why

Previously an icon would have to be manually added to the children of the button, the size of button would be up to the developer, as well as adding a square attribute. There were plenty of ways to introduce errors in the rendering of the button. This fixes that while also having an API to change the position of the icon.

## How

This PR introduces two props (icon and iconPosition) and conditionals to render the children.

## Checklist

- [X] Add or update stories

## Additional notes

- This PR is tied to the linked Figma component
- After merge a code connect file will be uploaded to display the correct code snippet


packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,42 @@
...
   /** The level of prominence. Use a primary button only once per page or section. */
   variant?: 'primary' | 'secondary' | 'tertiary'
-} & PropsWithChildren<ButtonHTMLAttributes<HTMLButtonElement>>
+} & (ButtonIcon | ButtonIconStart) &
+  PropsWithChildren<ButtonHTMLAttributes<HTMLButtonElement>>
 
 export const Button = forwardRef(
   (
-    { children, className, type, disabled, variant = 'primary', ...restProps }: ButtonProps,
+    { children, className, type, disabled, variant = 'primary', icon, iconStart, iconOnly, ...restProps }: ButtonProps,
```
VincentSmedinga, line 31: Please sort the props

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -12,14 +11,16 @@ const meta = {
   title: 'Components/Buttons/Button',
   component: Button,
   args: {
-    children: 'Button label',
+    children: 'Actietekst',
```
VincentSmedinga, line 14: ‘Button label’ may not be the best (for one, because it’s more English than Dutch), but ‘Actietekst’ isn’t realistic content either.

Let’s use ‘Versturen’ as the default label for a primary button and ‘Annuleren’ for a secondary button. Not sure about the tertiary, though.

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,42 @@
 import clsx from 'clsx'
 import { forwardRef } from 'react'
 import type { ButtonHTMLAttributes, ForwardedRef, PropsWithChildren } from 'react'
+import { Icon } from '../Icon'
+
+type ButtonIcon = {
+  icon?: Function
```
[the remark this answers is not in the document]
dlnr: This requires a change in Icon, I'm already working on Icon i a different story. So perhaps we can create a new story after these are merged?

packages/css/src/components/button/button.scss
```diff
@@ -95,3 +95,17 @@
...
+}
+
+// TODO: Proposition
+// .ams-button--icon-start {
+//   padding-inline-start: var(--ams-button-icon-start-padding-inline-start);
+// }
+
+// .ams-button--icon-end {
+//   padding-inline-end: var(--ams-button-icon-end-padding-inline-end);
+// }
```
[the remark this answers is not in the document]
dlnr: The inline text does not have a wrapping element, this makes it impossible to use `has` or `+`.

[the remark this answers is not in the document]
dlnr: Also, this seems like a design choice, so I would like to suggest a Figma tickets first.

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,42 @@
...
+  iconStart?: never
```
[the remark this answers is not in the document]
dlnr: It's reverted

packages/react/src/Button/Button.tsx
dlnr: Sorted, this is something I forget when the props are formatted (stylelint) on one line.

storybook/src/components/Button/Button.stories.tsx
dlnr: I've opted for "Terug" for tertiary, I remember seeing it used for this purpose. I've also considered "Lees meer" like NLDS uses in button stories.  Buttons are getting quite short like this though.

packages/react/src/Button/Button.test.tsx
```diff
@@ -104,10 +105,22 @@ describe('Button', () => {
...
 
-    const { container } = render(<Button ref={ref} />)
+    const { container } = render(<Button ref={ref}>Click me!</Button>)
 
     const button = container.querySelector(':only-child')
 
     expect(ref.current).toBe(button)
   })
+
+  it('renders a button with an icon', () => {
```
VincentSmedinga, line 115: Can we test the two `iconPosition` values as well? E.g. `start` should actually render it as the first child.

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,31 @@
 import clsx from 'clsx'
 import { forwardRef } from 'react'
 import type { ButtonHTMLAttributes, ForwardedRef, PropsWithChildren } from 'react'
+import { Icon } from '../Icon'
 
 export type ButtonProps = {
   /** The level of prominence. Use a primary button only once per page or section. */
   variant?: 'primary' | 'secondary' | 'tertiary'
+  icon?: Function
+  iconPosition?: 'start' | 'only'
```
VincentSmedinga, line 15: Reordered and added prop descriptions:

```suggestion
  /** An icon to add to the button. */
  icon?: Function
  /** The position of the icon. The default is after the label. */
  iconPosition?: 'start' | 'only'
  /** The level of prominence. Use a primary button only once per page or section. */
  variant?: 'primary' | 'secondary' | 'tertiary'
```

VincentSmedinga, line 15: Should we do ‘before’ instead of ‘start’?

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -12,18 +11,17 @@ const meta = {
   title: 'Components/Buttons/Button',
   component: Button,
   args: {
-    children: 'Button label',
-    variant: 'primary',
+    children: 'Versturen',
     disabled: false,
+    variant: 'primary',
   },
   argTypes: {
```
VincentSmedinga, line 18: Please add options and labels for `iconPosition`, where the `undefined` option shows as ‘after’, so people can reset it after choosing an option.

VincentSmedinga (commented):
I’ve added descriptions to the examples.

packages/css/src/components/button/button.scss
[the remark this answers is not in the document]
dlnr: Also this is not how it works:
`.ams-button:has(.ams-button__icon-start)`
It should be `.ams-button.ams-button__icon-start` because .ams-button__icon-start is not a child of the button. .ams-icon is though.

packages/react/src/Button/Button.tsx
dlnr: I don't think so I've been using the inline/block start/end terminology everywhere. Before reminds me of pseudo elements instead of logical properties.

packages/react/src/Button/Button.tsx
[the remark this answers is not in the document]
dlnr: I don't think you should change this, like the `Icon` it accepts the svg as a Function, this seems to work great with our iconset.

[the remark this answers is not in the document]
VincentSmedinga: Alright, not sure if I forgot some reasoning from our discussion, but looking at it now we can update this later if individual `*Icon` components will accept props.

@alimpens?

packages/css/src/components/button/button.scss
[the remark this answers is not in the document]
VincentSmedinga: This is the complete picture; I do think it would work. But we decided against changing the padding anyway.

```tsx
return (
  <button … className={clsx('ams-button', …)}>
    {iconPosition === 'start' && <Icon className="ams-button__icon-start" … />}
    …
  </button>
)
```

```html
<button class="ams-button">
  <svg class="ams-icon ams-button__icon-start" />
  Button label
</button>
```

```css
.ams-button:has(ams-button__icon-start) {
  padding-inline-start: …;
}

File: packages/react/src/Button/Button.test.tsx (hunk 6 above)
Comment (dlnr) [reply packages/react/src/Button/Button.test.tsx:115]:
```html
<button>
 Click me!
 <icon />
</button>
```

In this scenario `.ams-button > .ams-icon:first-child` would apply to the icon too. As well as `last-child`.

VincentSmedinga: Oh yeah. Bummer.

VincentSmedinga (approved):

packages/react/src/Button/Button.tsx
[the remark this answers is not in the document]
VincentSmedinga: Sorry, I’m entirely mistaken. `Button` renders the `Icon` component itself, so you’re right that we don’t need an entire component in the `icon` prop.

storybook/src/components/Button/Button.docs.mdx
```diff
@@ -16,28 +16,45 @@ import README from "../../../../packages/css/src/components/button/README.md?raw
 
 ### Primary
 
-The most important call-to-action.
-One primary Button may be used per screen.
+A primary button draws attention to the most important call to action.
+Only one primary Button should be used per screen or, in applications, large section.
```
alimpens, line 20: Application have screens as well, no? Why the exception for large sections here?

storybook/src/components/Button/Button.docs.mdx
```diff
@@ -16,28 +16,45 @@ import README from "../../../../packages/css/src/components/button/README.md?raw
...
 
 <Canvas of={ButtonStories.Primary} />
 
 ### Secondary
 
-Use the secondary Button for less prominent calls to action.
-It is possible to use more than 1 secondary Button.
+The secondary is the default type of Button.
```
alimpens, line 26: This is currently not true

packages/react/src/Button/Button.tsx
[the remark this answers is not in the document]
alimpens: When we make those changes to the Icon component, we do need to pass an SVG React component to Button as well. This has the benefit that you can add classes and other props to the SVG React component.

I think it makes sense to do all that work (changes to Icon and changes to all components that use Icon) in a separate PR though

storybook/src/components/Button/Button.docs.mdx
VincentSmedinga: As an example, each panel or drawer of an application could have a primary button. I assume the guideline to have max 1 per page is thought up from the perspective of a web page.

But it might be too early to mention this in docs.

VincentSmedinga:
```suggestion
Only one primary Button should be used per screen.
```

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,33 @@
...
+    { children, className, disabled, icon, iconPosition, type, variant = 'primary', ...restProps }: ButtonProps,
     ref: ForwardedRef<HTMLButtonElement>,
   ) => {
     return (
       <button
         {...restProps}
         ref={ref}
         disabled={disabled}
-        className={clsx('ams-button', `ams-button--${variant}`, className)}
+        className={clsx('ams-button', `ams-button--${variant}`, `ams-button--icon-${iconPosition}`, className)}
```
VincentSmedinga, line 30: Do we need to render classes for ‘start’ and ‘end’? The stylesheets do not contain selectors for them.

We can do this instead:

```
iconPosition === 'only' && `.ams-button--icon-position-only`
```

And the class should contain `-position`.

This does remove our option to unit test the rendering of an icon at the start or end.

storybook/src/components/Button/Button.docs.mdx
VincentSmedinga:
```suggestion
Use a secondary button for other actions.
```

packages/react/src/Button/Button.tsx
dlnr: I did previously but it can be removed now, also fixes the undefined.

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -12,18 +11,28 @@ const meta = {
...
+      table: { disable: true },
+    },
+    iconPosition: {
+      if: { arg: 'icon', neq: null },
+      control: {
+        type: 'inline-radio',
+        labels: { undefined: 'end', start: 'start', only: 'only' },
+      },
+      options: [undefined, 'start', 'only'],
+    },
```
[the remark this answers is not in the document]
alimpens: > The primary story is a button without an icon so it makes sense.

I don't think it does. The primary controls should show all possible controls for a component, even if they aren't used in the primary story. You should preferably be able to recreate all stories with the primary controls. 

I'm not sure what you mean with 'it messes up the padding', it shouldn't, right? 

We have a example of how to create an icon picker in the [Icon controls](https://designsystem.amsterdam/?path=/docs/components-media-icon--docs)

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -35,24 +44,39 @@ export const Primary: Story = {}
...
   args: {
+    children: 'Terug',
     variant: 'tertiary',
   },
 }
 
 export const WithIcon: Story = {
   args: {
-    children: ['Button label', <Icon key="icon" svg={ShareIcon} size="level-5" />],
+    children: 'Bel 14 020',
```
alimpens, line 61: This isn't the best example, because this should be a link, not a button... How about a close icon with the label 'Sluiten'?

packages/react/src/Button/Button.test.tsx
dlnr: Fixed the `undefined` and created tests for all icon positions using a `span` for the child.

storybook/src/components/Button/Button.stories.tsx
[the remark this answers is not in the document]
dlnr: Thanks, good tip. When you set the iconPosition to only you get different inline padding (to make it square) so it does "mess up" the button if an icon is not in there. This fixes is for storybook, I could set iconPosition to `never` in the button. 

The icon fix is good but the code example still does not work out, it shows you `icon={() => {}}`.

storybook/src/components/Button/Button.stories.tsx
dlnr: Someone keeps tinkering with my PR's 😬. I changed it until it happens again.

VincentSmedinga: Sorry, thought I was helping.

dlnr: I'm a big fan of you doing the documenting, but perhaps it's more convenient to do it in a docs specific PR. To prevent the second reviewer requesting changes the first reviewer is doing.

proprietary/tokens/src/components/ams/button.tokens.json
```diff
@@ -56,6 +56,10 @@
           "background-color": { "value": "transparent" },
           "color": { "value": "{ams.color.neutral-grey2}" }
         }
+      },
+      "icon-only": {
```
VincentSmedinga, line 60: I missed this before: the token should have ‘position’ in it as well.

I would split it already: `"icon-position": { "only": …`

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,41 @@
...
+          'ams-button',
+          `ams-button--${variant}`,
+          iconPosition === 'only' && `ams-button--icon-position-only`,
+          className,
+        )}
         type={type || 'button'}
       >
-        {children}
+        {icon && (iconPosition === 'start' || iconPosition === 'only') && (
+          <Icon svg={icon} size="level-5" square={iconPosition === 'only'} />
```
dlnr, line 40: I've changed the `square` prop to prevent the padding from growing too big when a label is shown. The padding around the icon should resemble the padding around the inline text.

packages/react/src/Button/Button.tsx
```diff
@@ -6,26 +6,41 @@
...
 import { forwardRef } from 'react'
 import type { ButtonHTMLAttributes, ForwardedRef, PropsWithChildren } from 'react'
+import { Icon } from '../Icon'
+import type { IconProps } from '../Icon'
 
 export type ButtonProps = {
+  /** An icon to add to the button. */
+  icon?: IconProps['svg']
+  /** The position of the icon. The default is after the label. */
+  iconPosition?: 'start' | 'only'
```
VincentSmedinga, line 16: Should we disallow providing an `iconPosition` without an `icon`? If so, through a discriminative union type like e.g. `Grid` does for padding. After that, the rendering conditions below can be simplified.

packages/react/src/Button/Button.tsx
VincentSmedinga: Gets me thinking… Should it even stay a public prop? We only want the icon container to become square if the icon is in a button and alone, right?

packages/react/src/Button/Button.tsx
dlnr: I've tried it before: https://github.com/Amsterdam/design-system/blob/8d3e9226aa1ea841e5b211d7e06c344223991fc3/packages/react/src/Button/Button.tsx

VincentSmedinga: This is not about padding, but about preventing `<Button iconPosition="start">Label</Button>`.

This should do it:

```ts
type IconButtonProps = {
  /** An icon to add to the button. */
  icon: IconProps['svg']
  /** The position of the icon. The default is after the label. */
  iconPosition?: 'start' | 'only'
}

type TextButtonProps = {
  icon?: never
  iconPosition?: never
}

export type ButtonProps = {
  /** The level of prominence. Use a primary button only once per page or section. */
  variant?: 'primary' | 'secondary' | 'tertiary'
} & PropsWithChildren<ButtonHTMLAttributes<HTMLButtonElement>> &
  (IconButtonProps | TextButtonProps)
```

packages/react/src/Button/Button.tsx
VincentSmedinga: https://gemeente-amsterdam.atlassian.net/browse/DES-997

storybook/src/components/Button/Button.stories.tsx
VincentSmedinga: Yes, let’s talk a bit about this next Wednesday.

VincentSmedinga (approved):

storybook/src/components/Button/Button.stories.tsx
```diff
@@ -3,27 +3,39 @@
...
   component: Button,
   args: {
-    children: 'Button label',
-    variant: 'primary',
+    children: 'Versturen',
     disabled: false,
+    variant: 'primary',
+    /* This is the only was storybook will honor the conditional in the iconPosition argType (line 29) */
+    // @ts-ignore
+    icon: null,
```
alimpens, line 19: Is this still necessary now we always show the icon controls?

dlnr: No, missed it.

VincentSmedinga (approved):

alimpens (approved):

PR #1656, opened 2024-10-14
feat!: Iconset update

dlnr:
# Describe the pull request

Updated iconset and workflow. This might break your application because there are some icons missing, missing icons in our stories have been copied.

## What

Replaced iconset with new icons.

## Why

New iconset.

## How

- Figma branch where iconset is exportable
- SVGO script that only removes svg markup that does not break the icons (like `fill-rule`)
- Updated Icons docs
- Replaced renamed icons in stories
- Copied missing icons to figma for designers to consider.

## Checklist

- [X] Add or update stories

## Additional notes

- [Figma branch updated icons](https://www.figma.com/design/9IGm6IdPUYizBNGsUnueBd/branch/zxGuonKxzqqvtvkAIaY8iX/Amsterdam-Design-System?node-id=26181-18306&node-type=frame&t=3D2Fu2YMVHsuhBvU-11)


VincentSmedinga (changes requested):
This doesn’t seem ready for a review. Many icons aren’t square and don’t come in filled and solid couples.

dlnr:
> Many icons aren’t square and don’t come in filled and solid couples.

- The square part comes as a prop in the Icon component, I did not see the need to square every viewbox. Which is also impossible in Figma, the svg export resizes the viewbox to the bounds of the shape(s).
- The iconset duplicated a lot of icons that did not have an outline counterpart (like the chevron or menu) in order to not break the fill property in Figma. Moving away from the fill prop and having separate components (or shapes) for all icons (like House and HouseOutline) removed the need to have duplicate icons. ie. It would be redundant to have ChevronDown and ChevronDownOutline if they are identical.

PR #1659, opened 2024-10-14
chore: configure pnpm version for corepack

Robbert:
# Describe the pull request

Add code to support `corepack enable`

## What

Support for `corepack enable`

## Why

```
Your pnpm version is incompatible with "/Users/bobby/Sites/github.com/Amsterdam/design-system".

Expected version: ^9
Got: 8.15.8

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.
```

## How

Add `packageManager` property to `package.json`

## Checklist

- [ ] Add or update stories

## Additional notes

- [corepack enable](https://github.com/nodejs/corepack?tab=readme-ov-file#corepack-enable--name) docs


package.json
```diff
@@ -81,5 +81,6 @@
     "watch:react": "npm run --workspace packages/react build:watch",
     "watch:storybook": "npm run --workspace storybook start",
     "watch:tokens": "npm run --workspace proprietary/tokens watch"
-  }
+  },
+  "packageManager": "pnpm@9.12.1"
```
alimpens, line 85: Thanks for the PR! Seems like a fine addition to me. 

One question: [the docs strongly recommend adding a hash here](https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages), any reason not to?

Robbert: IIRC @matijs found that the hash could vary per platform, so the hash for one version would be different for different for various CPU architectures, or something like that.

alimpens (approved):

PR #1662, opened 2024-10-15
feat: Let Grid and Grid Cell render any structural tag

VincentSmedinga:
# Describe the pull request

## What

This provides more options for Grid and Grid Cell to render a semantic HTML element instead of a `div`.

## Why

People should be able to apply the features of these components to any HTML element. Otherwise, they must add an unnecessary wrapper element.

## How

Extracted from #1607. Added options to the `as` prop, extracted a configuration object for it, included all of them in the unit tests.

## Checklist

- [x] Add or update stories

## Additional notes


packages/react/src/Grid/Grid.test.tsx
```diff
@@ -87,8 +88,23 @@ describe('Grid', () => {
     })
   })
 
+  gridTags.forEach((tag) => {
+    it(`renders with a custom ${tag} tag`, () => {
+      const { container } = render(<Grid as={tag} aria-label={tag === 'section' ? 'Accessible name' : undefined} />)
```
alimpens, line 93: Is the `aria-label` necessary?

packages/react/src/Grid/Grid.test.tsx
```diff
@@ -87,8 +88,23 @@ describe('Grid', () => {
...
+
+      let component: HTMLElement | null
+      if (tag === 'div') {
+        component = container.querySelector(tag)
+      } else {
+        component = screen.queryByRole(ariaRoleForTag[tag])
```
alimpens, line 99: Why not use `getByRole` here?

packages/react/src/Grid/Grid.tsx
```diff
@@ -29,17 +32,28 @@ type GridPaddingTopAndBottomProps = {
...
+      as: Tag = 'div',
+      children,
+      className,
+      gapVertical,
+      paddingBottom,
+      paddingTop,
+      paddingVertical,
+      ...restProps
+    }: GridProps,
+    ref: any,
```
alimpens, line 54: Can't we use `ForwardedRef<HTMLElement>` here?

packages/react/src/common/accessibility.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
alimpens, line 1: Currently this is only used in Grid, shouldn't we keep it there for the time being?

packages/react/src/Grid/Grid.test.tsx
VincentSmedinga: It allows selecting by role below, for all tags except for the `div`. Would you prefer removing `aria-label` and selecting by `:only-child`?

packages/react/src/Grid/Grid.test.tsx
VincentSmedinga: Wasn’t aware of the difference – updated.

packages/react/src/Grid/Grid.tsx
VincentSmedinga: No, because it can be a `div` and `HTMLDivElement` requires the `align` attribute.

packages/react/src/Grid/Grid.test.tsx
alimpens: Ah, right. No, selecting by role has preference

packages/react/src/Grid/Grid.tsx
alimpens: Maybe `ForwardedRef<any>`? So we at least check if the ref is of type `MutableRefObject`?

VincentSmedinga: Sure!

alimpens (approved):

packages/react/src/common/accessibility.ts
VincentSmedinga: https://gemeente-amsterdam.atlassian.net/browse/DES-1033

PR #1693, opened 2024-10-21
feat!: Make terms bold instead of descriptions in Description List

VincentSmedinga:
# Describe the pull request

## What

Following a [request from our community](https://github.com/Amsterdam/design-system/issues/1580), We’ve revised our earlier decision to display the descriptions of a Description List with bold text. Instead, we want the terms to be bold. 

## Why

Regular text is easier to read, especially if it is long as in form summaries. Making terms bold matches the header in forms. All communities in NLDS use bold for the terms, not the descriptions.

## How

- Updated the relevant tokens and CSS.
- Removed a list-wide font-weight token that wasn’t used. This makes the change breaking.
- Improved the content of the examples, sourcing from [our editorial guide](https://www.amsterdam.nl/schrijfwijzer/inclusieve-taal-richtlijnen-tips/inclusieve-woordenlijst/).
- Changed ‘details’ to ‘description’ in docs – because [MDN does so](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd)  for `<dd>` and it does not clash with `<details>`.

## Checklist

- [x] Add or update stories

## Additional notes

- We should rename `DescriptionList.Details` to `DescriptionList.Description` and the `details` tokens and classes. Let’s do this in a separate PR.
- An inline style in the ‘Multiple terms’ example hides a bug (a ticket exists for it).
- Customizing the widths of the columns, another request from the same source, also has its own ticket.


proprietary/tokens/src/components/ams/description-list.tokens.json
```diff
@@ -4,15 +4,17 @@
...
       "line-height": { "value": "{ams.text.level.5.line-height}" },
       "row": {
         "gap": { "value": "{ams.space.sm}" }
       },
+      "term": {
+        "font-weight": { "value": "{ams.text.font-weight.bold}" }
+      },
       "details": {
-        "font-weight": { "value": "{ams.text.font-weight.bold}" },
+        "font-weight": { "value": "{ams.text.font-weight.normal}" },
```
RubenSibon, line 17: Is it necessary to set the `font-weight` if it's the default (`normal`)?

This token is not being used, correct? Or is this for legacy purposes?

RubenSibon (approved):

RubenSibon (dismissed):

packages/css/src/components/description-list/description-list.scss
```diff
@@ -33,20 +32,22 @@
...
 .ams-description-list__term {
+  font-weight: var(--ams-description-list-term-font-weight);
+
   @media screen and (min-width: $ams-breakpoint-medium) {
     grid-column-start: 1;
   }
 }
 
-@mixin reset-details {
+@mixin reset-dd {
```
alimpens, line 42: We're using `dd`, `details` (in the class name) and `description` (in the README) now. Shouldn't it all be `description`?

packages/react/src/DescriptionList/DescriptionListDetails.test.tsx
```diff
@@ -3,7 +3,7 @@ import { createRef } from 'react'
 import { DescriptionList } from './DescriptionList'
 import '@testing-library/jest-dom'
 
-describe('Description list details', () => {
+describe('Description List Description', () => {
```
alimpens, line 6: This means the subcomponent should be called `DescriptionList.Description` as well, no?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -3,25 +3,27 @@
...
   decorators: [inverseColorDecorator],
   args: {
     children: [
-      <DescriptionList.Term key={1}>Gebied</DescriptionList.Term>,
-      <DescriptionList.Details key={2}>Gemeente Amsterdam</DescriptionList.Details>,
-      <DescriptionList.Term key={3}>Stadsdeel</DescriptionList.Term>,
-      <DescriptionList.Details key={4}>West</DescriptionList.Details>,
-      <DescriptionList.Term key={5}>Opmerkingen</DescriptionList.Term>,
-      <DescriptionList.Details key={6}>{paragraph}</DescriptionList.Details>,
+      <DescriptionList.Term key={1}>het hoger onderwijs</DescriptionList.Term>,
```
alimpens, line 17: Why did you change this example? I kind of prefer the old one

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -33,16 +35,48 @@ type Story = StoryObj<typeof meta>
 
 export const Default: Story = {}
 
-export const MultipleDetails: Story = {
+export const MultipleDescriptions: Story = {
+  args: {
+    children: [
+      <DescriptionList.Term key={1}>blinde, slechtziende</DescriptionList.Term>,
```
alimpens, line 41: I think you've changed the convention here, and I'm not sure that's what we want. Previously, we always started terms and descriptions with a capital, in these examples you sometimes do and sometimes don't.

I think starting with a capital is more visually appealing.

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: Agreed – you may have missed this in the PR description:

> We should rename `DescriptionList.Details` to `DescriptionList.Description` and the ‘details’ tokens and classes. Let’s do this in a separate PR.

I propose using the element name in our mixins.

packages/react/src/DescriptionList/DescriptionListDetails.test.tsx
VincentSmedinga: Yes, happy to create a PR after merging this one.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
VincentSmedinga: It was a bit contrived. ‘Gebied’ for Gemeente Amsterdam? Mentioning 4 city districts while we have 8, some with incorrect names.

I wanted a good example of the generic meaning of a description list. This content is straight from our writing guide and matches that goal: it *describes terms* that we preferred often-used ones.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
VincentSmedinga: I don’t think we had that convention in place. The first two examples concern phrases that often sit in the middle of a sentence. They will not be capitalized there, which is a good reason not to do it here.

The [source document](https://www.amsterdam.nl/schrijfwijzer/inclusieve-taal-richtlijnen-tips/inclusieve-woordenlijst/), written by content specialists, uses lower case as well.

The third example uses brand names and complete sentences, thus capital letters and punctuation.

packages/css/src/components/description-list/description-list.scss
alimpens: Ah, missed that part. A separate PR for `Details` -> `Description` sounds like a good idea to me.

> I propose using the element name in our mixins.

That would be a new convention, right? Makes sense to me, but maybe we should do that in a separate PR for all relevant mixins, instead of just here?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
alimpens: Maybe we should specify or link to the convention (in a separate PR)? [The convention for print is a bit simpler](https://amsterdam.nl/stijlweb/basiselementen/typografie/#hb251a4e3-5384-43be-a7ee-829d14c49fea), and leads to a more appealing result, imo.

Wasn't there anything on this on the old Stijlweb site?

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: Sure, although we could have let this one slip in already? Reverted it.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
VincentSmedinga: Well, look at that:

> losse woorden, grammaticaal complete en incomplete zinnen worden met beginkapitaal gezet

Doesn’t sound like this is specific for print, so I’ve changed them.

VincentSmedinga: Created [a new epic](https://gemeente-amsterdam.atlassian.net/browse/DES-1008) to keep track of the content guidelines we’ll once add to our docs.

PR #1695, opened 2024-10-23
feat!: Use new Button features for SearchField

VincentSmedinga:
# Describe the pull request

Comments in Search Field suggested reusing Icon Button. Now that we’re replacing that with an icon-only button, I wondered if that could be a drop-in replacement. And… it is! I found no visual or other regressions.

## What

Use Icon Button for Search Field Button.

## Why

Reuse existing components.

## How

- Replaced the custom HTML button with Icon Button and remove associated styles and tokens.
- Decided to keep `SearchField.Button` as a better API than letting consumers use IconButton explicitly in SearchField.
- Decided to disallow the `icon`, `iconPosition` and `variant` props on Search Field Button, although having a visible ‘Zoeken’ label doesn’t seem bad either.

## Checklist

- [x] Add or update stories

## Additional notes

- Should we refactor `SearchField.Input` to `SearchInput` and make this thing a pattern?


packages/react/src/SearchField/SearchFieldButton.tsx
dlnr, line 1: What would be the reason to still use a `SearchField.Button` component instead of using `Button` directly?

dlnr (commented):
Seems like a logical change, which can also be done in Figma. But i don't get the extra layer of SearchField.Button.

packages/react/src/SearchField/SearchFieldButton.tsx
RubenSibon: I guess a reason is that developers do not have to think about what icon to use. It should always be the search/magnifying glass icon, right?

RubenSibon (dismissed):
LGTM, but do address @dlnr's comment about using the `Button` component directly.

alimpens (approved):

packages/react/src/SearchField/SearchFieldButton.tsx
VincentSmedinga: It is indeed, as I wrote in the description:

> Decided to keep SearchField.Button as a better API than letting consumers use IconButton explicitly in SearchField.

packages/react/src/SearchField/SearchFieldButton.tsx
```diff
@@ -4,24 +4,18 @@
...
-// TODO: replace this with IconButton when that's done
-// TODO: discuss if IconButton is the right component to replace this
 export const SearchFieldButton = forwardRef(
-  ({ label = 'Zoeken', className, ...restProps }: SearchFieldButtonProps, ref: ForwardedRef<HTMLButtonElement>) => (
-    <button {...restProps} ref={ref} className={clsx('ams-search-field__button', className)}>
-      <span className="ams-visually-hidden">{label}</span>
-      <Icon svg={SearchIcon} size="level-5" square />
-    </button>
+  ({ children = 'Zoeken', ...restProps }: SearchFieldButtonProps, ref: ForwardedRef<HTMLButtonElement>) => (
+    <Button {...restProps} icon={SearchIcon} iconBefore ref={ref}>
```
alimpens, line 16:
```suggestion
    <Button {...restProps} icon={SearchIcon} iconOnly ref={ref}>
```

alimpens, line 16: Don't we want to pass the classname here as well? So you can still override the styling if you need to?

VincentSmedinga: That’s what I thought at first. Then I remembered your suggestion last week not to offer class names that we don’t use ourselves. People can still add their own class name: `<SearchField.Button className="custom">`.

alimpens: Ah, of course, `className` is passed through the `restProps`

alimpens (approved):

PR #1728, opened 2024-10-28
chore: Wrap all stories in a Screen

VincentSmedinga:
# Describe the pull request

## What

This wraps all stories of Storybook in a `Screen` component.

I’ve not made an exception for our page examples to keep the Skip Link spanning the entire page width. I think it’s best to limit that link to the Screen's width to prevent it from growing overly large on very wide screens. This doesn’t change its functionality. Few screens will be affected in practice.

## Why

To ensure no example is wider than a `Screen`, which should not happen as we demand every page to be wrapped in such a component.

We could override the decorator for application screens later, but there’s no need now.

## How

- Replaced the `div lang="nl"` that wrapped all stories, keeping the language attribute
- Removed decorators wrapping a Screen for component stories

## Checklist

- [x] Add or update stories

## Additional notes

–

### DIFF
--- a/storybook/config/preview.tsx
+++ b/storybook/config/preview.tsx
@@ -4,6 +4,7 @@ import '@amsterdam/design-system-assets/font/index.css'
 import '@amsterdam/design-system-css/dist/index.css'
 import '../src/styles/docs.css'
 import '../src/styles/overrides.css'
+import { Screen } from '@amsterdam/design-system-react'
 import { withThemeByClassName } from '@storybook/addon-themes'
 import { viewports } from './viewports'
 
@@ -13,12 +14,12 @@ export const argTypes = {
   },
 }
 
-// Set language to Dutch for Canvas and Stories
+// Wrap in Screen, set language to Dutch for Canvas and Stories
 export const decorators = [
   (Story: any) => (
-    <div lang="nl">
+    <Screen lang="nl">
       <Story />
-    </div>
+    </Screen>
   ),
   withThemeByClassName({
     themes: {
@@ -57,7 +58,4 @@ export const parameters = {
     viewports,
   },
   viewMode: 'docs',
-  html: {
-    root: 'div[lang="nl"]',
-  },
 }
--- a/storybook/src/components/Breakout/Breakout.stories.tsx
+++ b/storybook/src/components/Breakout/Breakout.stories.tsx
@@ -3,20 +3,13 @@
  * Copyright Gemeente Amsterdam
  */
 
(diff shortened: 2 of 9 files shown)


storybook/config/preview.tsx
```diff
@@ -13,12 +14,12 @@ export const argTypes = {
   },
 }
 
-// Set language to Dutch for Canvas and Stories
+// Wrap in Screen, set language to Dutch for Canvas and Stories
 export const decorators = [
   (Story: any) => (
-    <div lang="nl">
+    <Screen lang="nl">
```
VincentSmedinga, line 20: Should we add an `ams-theme` class to be explicit? The [NLDS Candidate Storybook](https://nl-design-system.github.io/candidate/?path=/story/paragraph--default) does it (see HTML tab) and seems like a good idea.

People can learn (and be instructed) to copy things _inside_ that element, not the wrapper. This is current practice with the `<div lang="nl">` – the new variant `<div class="ams-theme ams-screen" lang="nl">` may make that even clearer.

dlnr: The `lang` attribute is a workaround right? Normally it would go on the html tag.

Also, `Screen` seems to get a default modifier class `ams-screen--wide` where it is a default. So you would not need it, only if you specify x-wide? Which could be a boolean then.

alimpens: > Should we add an ams-theme class to be explicit?

What are we making explicit? [We currently don't use `ams-theme` in Storybook](https://github.com/Amsterdam/design-system/blob/develop/storybook/config/preview.tsx#L1), I don't quite understand why we would want to add an unused class.

storybook/config/preview.tsx
```diff
@@ -57,7 +58,4 @@ export const parameters = {
     viewports,
   },
   viewMode: 'docs',
-  html: {
-    root: 'div[lang="nl"]',
```
alimpens, line 61: Why did you remove this? We do not want to show the decorators in the HTML tab, right?

storybook/config/preview.tsx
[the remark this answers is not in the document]
VincentSmedinga: I had erroneously removed the config to strip the screen div from the HTML panel; it’s been restored.

I don’t mind that the example for Gap gets centred. All other docs pages and stories are centred and I wouldn’t want to make an exception for utils alone. It only occurs in very wide windows.

alimpens (approved):

storybook/config/preview.tsx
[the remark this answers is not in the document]
VincentSmedinga: Discussed elsewhere.

PR #1736, opened 2024-11-01
fix: Always give icon a key

dlnr:
# Describe the pull request

## What

The content function with a switch produces children without a key.

## Why

To prevent an error.

## How

Added key to single child which should not be needed, It must have something to do with the wrapping content function that produces content as a Reactnode. . 

## Checklist

## Additional notes

### DIFF
--- a/packages/react/src/Button/Button.tsx
+++ b/packages/react/src/Button/Button.tsx
@@ -48,7 +48,7 @@ export const Button = forwardRef(
         case !icon:
           return children
         case iconBefore:
-          return [<Icon svg={icon} size="level-5" />, children]
+          return [<Icon key={1} svg={icon} size="level-5" />, children]
         case iconOnly:
           return [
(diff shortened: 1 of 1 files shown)


RubenSibon (dismissed):
Easy!

packages/react/src/Button/Button.tsx
```diff
@@ -57,7 +57,7 @@ export const Button = forwardRef(
             </span>,
           ]
         default:
-          return [children, <Icon svg={icon} size="level-5" />]
+          return [children, <Icon key={2} svg={icon} size="level-5" />]
```
VincentSmedinga, line 60: Doesn’t have to be 2 in a different array.

```suggestion
          return [children, <Icon key={1} svg={icon} size="level-5" />]
```

VincentSmedinga (approved):

alimpens (approved):

PR #1738, opened 2024-11-01
feat: Rename reset mixins to include element name

RubenSibon:
# Describe the pull request

## What

Rename Sass reset mixins to include the HTML or common name of the element they reset. 

## Why

This refactor is not really necessary, but makes it somewhat easier to understand what each reset mixin does. It also helps with global searches on the word "reset".

## How

Global search on `@mixin reset ` and then changing those names by inluding the element name after a dash.

## Checklist

## Additional notes

- There are some instances where an HTML name was not used: `paragraph` instead of `p`, `list` instead of `ol` or `ul` and `heading` instead of `h1` to `h6`. Please check if this makes sense.
- One reset applies to both a `dd` and a `label` element. In this case the reset is named `reset-label-or-description`.


packages/css/src/components/blockquote/blockquote.scss
```diff
@@ -6,7 +6,7 @@
 @import "../../common/hyphenation";
 @import "../../common/text-rendering";
 
-@mixin reset {
```
alimpens, line 9: Is this the convention we want? The ticket isn't very clear on that. The convention we had (undocumented) was: call the mixin `reset` if there's only one, and `reset-name-of-thing` when there are multiple resets.

Maybe we should define and document the convention first, before implementing it?

VincentSmedinga: The ticket is concice, but indicates actual HTML element names: “Dus `reset-button`, `reset-ul`, `reset-dd` etc.”. So `reset-blockquote` matches that, while some others in this PR do not.

packages/css/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -5,7 +5,7 @@
 
 @import "../../common/text-rendering";
 
-@mixin reset {
+@mixin reset-list {
```
VincentSmedinga, line 8: Please use the element names. There’s no need for a difference with the type selector for which we reset the user agent styles. That `reset-ul` and `reset-ol` will possibly be the same is okay.

packages/css/src/components/error-message/error-message.scss
```diff
@@ -5,7 +5,7 @@
 
 @import "../../common/text-rendering";
 
-@mixin reset {
+@mixin reset-paragraph {
```
VincentSmedinga, line 8: Just `reset-p` etc.

packages/css/src/components/heading/heading.scss
```diff
@@ -6,7 +6,7 @@
 @import "../../common/hyphenation";
 @import "../../common/text-rendering";
 
-@mixin reset {
+@mixin reset-heading {
```
VincentSmedinga, line 9: I understand this is a bit difficult. Options are `reset-hx`, `reset-hn`, and `reset-h`. Let’s go with the latter.

packages/css/src/components/screen/screen.scss
```diff
@@ -3,7 +3,7 @@
  * Copyright Gemeente Amsterdam
  */
 
-@mixin reset {
+@mixin reset-div {
```
VincentSmedinga, line 6: Hmm… what to do with this one. You could argue this is not even a reset. We’re not overriding some opiniated design of the browser, we’re just setting a different value for this keyword. Just include it in the main selector?

packages/css/src/components/text-area/text-area.scss
```diff
@@ -5,7 +5,7 @@
 
 @import "../../common/text-rendering";
 
-@mixin reset {
+@mixin reset-text-area {
```
VincentSmedinga, line 8: Just `textarea` without the dash, please, like HTML.

packages/css/src/components/top-task-link/top-task-link.scss
```diff
@@ -15,7 +15,7 @@
   text-decoration: none;
 }
 
-@mixin reset {
+@mixin reset-label-or-description {
```
VincentSmedinga, line 18: Like before, let’s not call this a reset.

VincentSmedinga (approved):

alimpens (approved):

PR #1747, opened 2024-11-08
chore: Merge decorators

VincentSmedinga:
Merging the colour decorators with the global one hides their resulting markup from the HTML element, which may confuse users.

Switching the order of the class selectors prevents invalid appearances of components for which both the `contrastColor` and `inverseColor` have been set to true.


storybook/src/styles/docs.css
```diff
@@ -57,14 +57,14 @@
...
 
-.ams-docs-dark-background {
-  background-color: #004699; /* Dark blue */
-}
-
 .ams-docs-light-background {
   background-color: #ffe600; /* Yellow */
 }
 
+.ams-docs-dark-background {
```
alimpens, line 64: I assume this change is to make sure dark background always beats light background? This is a little error prone imo. Shouldn't we fix this in the global decorator?

VincentSmedinga: You don’t need to assume – it plainly says so in the PR description :)

Source order matters in CSS, so this is not necessarily not robust. It is a valid way to implement our intent. I’ve added a comment to prevent people from switching the selectors around in the future.

Repairing this in the decorator only ‘fixes’ the display in Storybook, our components remain unaffected. I’ll add a ticket to move these colour props into an enum.

alimpens: > it plainly says so in the PR description

Ah, totally did not read that 😁 Adding a comment is a good way to make our intent explicit imo.

PR #1751, opened 2024-11-13
feat: Add File List component to display below a File Input

dlnr:
# Describe the pull request

## What

This is an optional file list to add to a file input or uploader component.

## Why

Our designers often use a file preview when uploading documents.

## How

A separate component, to which you can use click handlers or context or whatever method you need to display files.
It also introduced utils to format file type and file size.

## Checklist

- [X] Add or update stories

## Additional notes


packages/react/src/common/index.ts
VincentSmedinga, line 1: We prefer not to use barrel files to minimize the chance of cyclical dependencies.

storybook/src/components/FileList/FileInputWithFileList.tsx
```diff
@@ -0,0 +1,33 @@
+import { FileInput, FileList } from '@amsterdam/design-system-react'
+import { useRef, useState } from 'react'
+
+export const FileInputWithFileList = () => {
+  const inputRef = useRef<HTMLInputElement>(null)
+  const [files, setFiles] = useState<FileList | null>(null)
+
+  const handleFilesChange = () => {
```
VincentSmedinga, line 8: I think function names like `changeFiles` and `removeFile` are more straightforward.

storybook/src/components/FileList/FileList.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {}
+
+export const WithInput: Story = {
+  parameters: {
+    docs: {
+      source: {
+        code: `
```
VincentSmedinga, line 40: Can we prevent duplicating the component’s source code? Would a ‘raw’ import work like we do for Markdown files? Or something else?

VincentSmedinga (commented):
This is going to be a nice addition!

packages/react/src/FileList/FileList.test.tsx
```diff
@@ -0,0 +1,45 @@
+import { render } from '@testing-library/react'
+import { createRef } from 'react'
+import { FileList } from './FileList'
+import '@testing-library/jest-dom'
+
+var files = [
+  new File(['sample1'], 'sample1.txt', { type: 'text/plain', lastModified: Date.now() }),
+] as unknown as FileList // This is a workaround because jest-dom does not support DataTransfer
+
+describe('FileList', () => {
```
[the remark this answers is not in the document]
dlnr: I've tried in various ways just to find out DataTransfer is not in the jest-dom. I'm actually for a storybook interaction test instead. What are your thoughts? I do think another PR would be in order, it's a new feature.

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,51 @@
...
+              {file.type.includes('image') ? (
+                <img src={URL.createObjectURL(file)} alt={file.name} width={50} height="auto" />
+              ) : (
+                <Icon svg={DocumentIcon} size="level-3" square />
+              )}
+            </div>
+            <div className="ams-file-list__file-name">
+              {file.name}
+              <div className="ams-file-input__file-details">
+                ({formatFileType(file.type)}, {formatFileSize(file.size)})
```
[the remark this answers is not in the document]
dlnr: It falls back to the mimetype

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const FileList = forwardRef(
+  ({ files, onDelete, className, ...restProps }: FileListProps, ref: ForwardedRef<HTMLDivElement>) => {
+    return (
+      <div {...restProps} ref={ref} className={clsx('ams-file-list', className)}>
+        {Array.from(files).map((file, index) => (
+          <div key={index} className="ams-file-list__file">
+            <div className="ams-file-list__file-preview">
+              {file.type.includes('image') ? (
+                <img src={URL.createObjectURL(file)} alt={file.name} width={50} height="auto" />
```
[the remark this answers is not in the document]
dlnr: Left it there to discuss, then forgot :) I would've liked to discuss what component is responsible for the width? The Image or the wrapper, like I've done now in file-list-scss?

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,51 @@
...
+              {file.name}
+              <div className="ams-file-input__file-details">
+                ({formatFileType(file.type)}, {formatFileSize(file.size)})
+              </div>
+            </div>
+            {onDelete && (
+              <Button variant="tertiary" onClick={() => onDelete(index)}>
+                Verwijder
+              </Button>
+            )}
```
dlnr: Need to wrap in a div to prevent 100 height caused by flex.

storybook/src/components/FileList/FileList.stories.tsx
dlnr: Also a tryout, it did not work out loading as a raw file but I've hidden the sourceState in canvas and used the Source doc block. This might be a good and recognisable solution for all composition type of examples.

dlnr (commented):
This is part 1 of 2, will round it up later.

packages/css/src/components/file-list/file-list.scss
```diff
@@ -0,0 +1,39 @@
...
+  font-family: var(--ams-file-list-file-font-family);
+  font-size: var(--ams-file-list-file-font-size);
+  font-weight: var(--ams-file-list-file-font-weight);
+  gap: var(--ams-file-list-file-gap);
+  line-height: var(--ams-file-list-file-line-height);
+}
+
+.ams-file-list__file-preview {
+  display: grid;
+  flex: 0 0 50px;
```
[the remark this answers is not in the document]
dlnr: 4rem would be too big for small screen, I've applied a clamp rule.

packages/react/src/FileList/FileList.tsx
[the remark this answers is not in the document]
dlnr: The we would also need something you can translate? I'll change it to 'Document' for now.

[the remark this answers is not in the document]
dlnr: Also very weird it's not catched by the default switch.

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,51 @@
...
+export type FileListProps = {
+  files: FileList
+  // eslint-disable-next-line no-unused-vars
+  onDelete?: (index: number) => void
+} & HTMLAttributes<HTMLDivElement>
+
+export const FileList = forwardRef(
+  ({ files, onDelete, className, ...restProps }: FileListProps, ref: ForwardedRef<HTMLDivElement>) => {
+    return (
+      <div {...restProps} ref={ref} className={clsx('ams-file-list', className)}>
```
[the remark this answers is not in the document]
alimpens: Shouldn't it be an `ul`? The meaning of the list is the same when you change the order, I think.

packages/react/src/common/formatFileSize.test.tsx
```diff
@@ -0,0 +1,27 @@
+import { formatFileSize } from './formatFileSize'
+
+describe('formatFileSize', () => {
+  it('formats bytes correctly', () => {
+    expect(formatFileSize(500)).toBe('500B')
+  })
+
+  it('formats kilobytes correctly', () => {
+    expect(formatFileSize(1024, 1)).toBe('1KB')
```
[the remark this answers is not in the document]
dlnr: The default precision is 3, the tests are mainly to test the units.

packages/css/src/components/file-list/file-list.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.ams-file-list {
+  display: flex;
+  flex-direction: column;
+  gap: var(--ams-file-list-gap);
+  padding-block: var(--ams-file-list-padding-block);
+
+  @include text-rendering;
+  @include reset-ol;
```
VincentSmedinga, line 20:
```suggestion
@mixin reset-ul {
  margin-block: 0;
  padding-inline: 0;
}

.ams-file-list {
  display: flex;
  flex-direction: column;
  gap: var(--ams-file-list-gap);
  padding-block: var(--ams-file-list-padding-block);

  @include text-rendering;
  @include reset-ul;
```

packages/react/src/FileList/FileList.tsx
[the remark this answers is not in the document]
VincentSmedinga: Yes, containers are responsible for limiting size; components should be flexible.

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,52 @@
...
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes } from 'react'
+import { Button } from '../Button'
+import { Icon } from '../Icon'
+import { formatFileSize } from '../common/formatFileSize'
+import { formatFileType } from '../common/formatFileType'
+
+export type FileListProps = {
+  files: FileList
```
VincentSmedinga, line 16: Let’s try the subcomponent approach here that we often use, as suggested in [Slack](https://datalab-amsterdam.slack.com/archives/C06B35WFZM2/p1733223833899999).

```
<FileList>
  <FileList.Item />
  <FileList.Item />
</FileList>
```

dlnr: ALso if you do not require a user to use FileList.Item baecause it's a loop in this component??

VincentSmedinga: Yes, we do so with various other components. Likewise, we could have done something like `<DescriptionList items={[term, description, ...]} />`. Object props are just not very clear, even though it’s a bit different for File List because we get the object straight from the browser.

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,52 @@
...
+  // eslint-disable-next-line no-unused-vars
+  onDelete?: (index: number) => void
+} & HTMLAttributes<HTMLOListElement>
+
+export const FileList = forwardRef(
+  ({ files, onDelete, className, ...restProps }: FileListProps, ref: ForwardedRef<HTMLOListElement>) => (
+    <ul {...restProps} ref={ref} className={clsx('ams-file-list', className)}>
+      {Array.from(files).map((file, index) => (
```
dlnr, line 24: @alimpens and @VincentSmedinga if I understand correctly, you want user to build this loop themselves?

VincentSmedinga: Yes, please. Subcomponents offers them more control, and it relieves us from the object prop.

packages/react/src/FileList/FileList.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <li key={index} className="ams-file-list__file">
+          <div className="ams-file-list__file-preview">
+            {file.type.includes('image') ? (
+              <img src={URL.createObjectURL(file)} alt={file.name} />
```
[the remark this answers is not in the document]
VincentSmedinga: Forgot to publish this comment.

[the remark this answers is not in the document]
dlnr: It receives a mimetype from the selected blob, so I do not think this is necessary.

packages/react/src/FileList/FileListItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+export type FileListItemProps = {
+  file: File
+  onRemove?: () => void
+} & HTMLAttributes<HTMLLIElement>
+
+export const FileListItem = forwardRef(
+  ({ file, onRemove, className, ...restProps }: FileListItemProps, ref: ForwardedRef<HTMLLIElement>) => (
+    <li {...restProps} ref={ref} className={clsx('ams-file-list__file', className)}>
+      <div className="ams-file-list__file-preview">
+        {file.type.includes('image') ? (
```
VincentSmedinga, line 24: Is this a better expression of what we mean here? An `includes` can result in false positives for edge cases.

```suggestion
        {file.type.startsWith('image/') ? (
```

VincentSmedinga (approved):

PR #1757, opened 2024-11-19
feat!: Change colours for Alert

VincentSmedinga:
# Describe the pull request

## What

1. Gives all alerts a background colour now that we have tints for it
2. Changes the colour of the informational Alert from blue to azure
3. Removes some tokens that were no longer in use

## Why

1. To make Alerts stand out more.
2. To use blue only for interactive elements
3. Because changing them didn’t do anything anyway


dlnr (approved):
Looks a lot better

PR #1763, opened 2024-11-26
fix: Correct layout of Description List if multiple terms share a description

dlnr:
# Describe the pull request

## What

An example inline style to prevent a row skip when using multiple terms.

## Why

Looks weird

## How

An inline style is the most straightforward solution, I've experimenten with sibling selectors and grid-area but it requires too much CSS and you'll never fix it for large lists. This solution also allows the next "normal" item to not need an inline style.

## Checklist

- [X] Add or update stories

## Additional notes


storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -46,10 +46,17 @@ export const MultipleDescriptions: Story = {
...
-      <DescriptionList.Description key={3}>Persoon met een visuele handicap</DescriptionList.Description>,
-      <DescriptionList.Description key={4}>Persoon die blind is</DescriptionList.Description>,
-      <DescriptionList.Description key={5}>Persoon die slechtziend is</DescriptionList.Description>,
+      <DescriptionList.Description key={1.2}>Persoon met een visuele beperking</DescriptionList.Description>,
+      <DescriptionList.Description key={1.3}>Persoon met een visuele handicap</DescriptionList.Description>,
+      <DescriptionList.Description key={1.4}>Persoon die blind is</DescriptionList.Description>,
+      <DescriptionList.Description key={1.5}>Persoon die slechtziend is</DescriptionList.Description>,
+      <DescriptionList.Term key={2}>Doof</DescriptionList.Term>,
+      <DescriptionList.Term key={3}>Slechthorend</DescriptionList.Term>,
+      <DescriptionList.Description key={2.1} style={{ gridRowStart: 5 }}>
```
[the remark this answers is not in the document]
dlnr: Well you actually only need to supply a row number for the tablet breakpoint, you need to "auto" rows part to work otherwise.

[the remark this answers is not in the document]
dlnr: Also the column count for grid goes up to 12, which is managable in CSS. If you have a description list of 50+ items it gets pretty bulky.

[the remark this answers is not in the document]
VincentSmedinga: > Well you actually only need to supply a row number for the tablet breakpoint, you need to "auto" rows part to work otherwise. Also the column count for grid goes up to 12, which is managable in CSS. If you have a description list of 50+ items it gets pretty bulky.

I’m not sure I understand. Let’s talk this through, unless the `div` approach is a good solution.

> A more pragmatic approach would be to wrap multiple terms in a div.

Interesting. I thought all related `dt` and `dd` elements should be wrapped in one `div`. You only wrap the `dt`s and leave the `dd` as a sibling. Is that valid? Does it work correctly in assistive technologies?

I consulted the [WHATWG spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element).

The content model seems to require the former approach:

> Content model:
> - Either: Zero or more groups each consisting of one or more dt elements followed by one or more dd elements, optionally intermixed with script-supporting elements.
> - Or: One or more div elements, optionally intermixed with script-supporting elements.

While this information allows the latter as well:

> The dl element represents an association list consisting of zero or more name-value groups (a description list). A name-value group consists of one or more names (dt elements, possibly as children of a div element child) followed by one or more values (dd elements, possibly as children of a div element child), ignoring any nodes other than dt and dd element children, and dt and dd elements that are children of div element children.

Please test whether wrapping only the `dt`s in a `div` is valid HTML and whether screen readers announce the terms and descriptions correctly.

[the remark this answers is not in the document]
dlnr: You're right, not valid html, which is a bummer.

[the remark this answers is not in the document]
dlnr: I've push a solution using groups. Unfortunately `subgrid` would be very useful here but the group class has been added to the various modifiers that set the width of the first column.

Only caveat is that it does need a `minWith` prop set to `md` to look any good.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,33 @@ export const MultipleDescriptions: Story = {
...
+  args: {
+    termsWidth: 'md',
+    children: [
+      <DescriptionList.Group key={1}>
+        <DescriptionList.Term key={1}>Achternaam</DescriptionList.Term>
+        <DescriptionList.Description key={2}>
+          De naam die een persoon van zijn of haar ouders krijgt
+        </DescriptionList.Description>
+      </DescriptionList.Group>,
+      <DescriptionList.Group key={2}>
```
VincentSmedinga, line 67: Grouping related terms and descriptions is a common and valid approach. Too bad that it prevents the list from automatically sizing its terms column because the terms are no longer grid items.

Multiple terms sharing one description may be an edge case. We might accept that people must choose a width for their terms in this situation, if we get it to work completely. Let’s discuss during our front-end refinement meeting.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,33 @@ export const MultipleDescriptions: Story = {
...
+        <DescriptionList.Term key={1}>Naam</DescriptionList.Term>
+        <DescriptionList.Term key={2}>Voornaam</DescriptionList.Term>
+        <DescriptionList.Term key={3}>Roepnaam</DescriptionList.Term>
+        <DescriptionList.Term key={4}>Bijnaam</DescriptionList.Term>
+        <DescriptionList.Description key={5}>
+          De naam waarmee een persoon wordt aangesproken
+        </DescriptionList.Description>
```
VincentSmedinga, line 74: This description sits next to the fourth term. I think we want it next to the first one. Giving a description in a group a `grid-row-start: 1` will fix this.

VincentSmedinga: Which makes me wonder. We don’t need the divs if we could ‘move up’ a description a couple of rows. CSS can count preceding nodes with the next-sibling combinator: (e.g. through `dt + dt + dd { grid-row: ‘move up 1’ }`, `dt + dt + dt + dd { grid-row: ‘move up 2’ }`. Do you see an approch for this? We can’t use negative row numbers because the term may not be the first in the list.

dlnr: Or just use subgrid on the group? It has a [93% support in the netherlands](https://caniuse.com/mdn-css_properties_grid-template-rows_subgrid) at the moment, it's use would only apply to larger screens and you could fallback (the group) to a single column like the native DL does.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -25,19 +25,30 @@
...
+  .ams-description-list--terms-width-md .ams-description-list__group {
     grid-template-columns: 1fr 2fr;
   }
 
-  .ams-description-list--terms-width-lg {
+  .ams-description-list--terms-width-lg,
+  .ams-description-list--terms-width-lg .ams-description-list__group {
     grid-template-columns: 1fr 1fr;
   }
 }
```
VincentSmedinga, line 54: A description wrapping over multiple lines will push the second and consecutive terms down. Can you check whether something like this could address that?

```scss
.ams-description-list__group {
  .ams-description-list__description {
    grid-row: 1 / -1;
  }
}
```

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
```diff
@@ -20,6 +20,14 @@ A term may have multiple descriptions.
 
 <Canvas of={DescriptionListStories.MultipleDescriptions} />
 
+### Multiple terms
+
+In the case of multiple consequtive terms for a description you need to
+wrap all `DescriptionList.Term` and `DescriptionList.Description` pairs in `DescriptionList.Group`.
+And it will also require the minWidth property to be set to anything but `auto` to prevent the misalignment of different groups.
```
VincentSmedinga, line 27:
```suggestion
If multiple terms share a single description, group them in a `DescriptionList.Group` to ensure proper grid layout.
Note that this approach is incompatible with the automatic width of the terms column; select a fixed width option instead.
```

dlnr: > Wait a minute, we can adjust the grid column widths automatically if a group is present 🤯🎉
> 
> ```css
> .ams-description-list__group:has(.ams-description-list__group) {
>   grid-template-columns: 1fr 2fr;
> }
> ```
> 
> Maybe as a separate declaration block to prevent invalidating the entire block for browsers that don’t recognise the `:has` pseudo-class?

`.ams-description-list__group:has(.ams-description-list__group)` Wouldn't work, it implies that the group contains another group. The first column width is something I'm not happy about, i'll check if it works without breakpoints.

packages/css/src/components/description-list/description-list.scss
dlnr: This will also cause overlap when there are multiple descriptions.

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
VincentSmedinga: Sorry big typo of course – I meant

```
.ams-description-list:has(.ams-description-list__group) {
  grid-template-columns: 1fr 2fr;
}
```

“If the list has a group, apply the 33/67 column widths to it instead of auto.”

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,32 @@ export const MultipleDescriptions: Story = {
...
+        <DescriptionList.Description key={6}>
+          De naam waarmee een persoon wordt aangesproken
+        </DescriptionList.Description>
```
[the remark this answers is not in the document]
dlnr: This is to test multiple descriptions.

packages/css/src/components/description-list/description-list.scss
dlnr: The grid rows need to be explicit for this to work. I would love to do this:

```css
  .ams-description-list__group {
    dd:only-of-type {
      grid-row: 1 / -1;
    }
  }
}
```

But it won't work if the rows are set to auto.

packages/react/src/DescriptionList/DescriptionListGroup.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright Gemeente Amsterdam
+ */
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+
+export type DescriptionListGroupProps = PropsWithChildren<HTMLAttributes<HTMLDivElement>>
+
+export const DescriptionListGroup = forwardRef(
```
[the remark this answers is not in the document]
dlnr: Renamed component.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,32 @@ export const MultipleDescriptions: Story = {
...
+    children: [
+      <DescriptionList.Term key={1}>Achternaam</DescriptionList.Term>,
+      <DescriptionList.Description key={2}>
+        De naam die een persoon van zijn of haar ouders krijgt
+      </DescriptionList.Description>,
+      <DescriptionList.Group key={2}>
+        <DescriptionList.Term key={1}>Naam</DescriptionList.Term>
+        <DescriptionList.Term key={2}>Voornaam</DescriptionList.Term>
+        <DescriptionList.Term key={3}>Roepnaam</DescriptionList.Term>
+        <DescriptionList.Term key={4}>Bijnaam</DescriptionList.Term>
```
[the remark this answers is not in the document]
dlnr: Done

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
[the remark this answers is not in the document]
dlnr: Removed

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,32 @@ export const MultipleDescriptions: Story = {
...
+        <DescriptionList.Description key={5}>
+          De naam waarmee een persoon wordt aangesproken
+        </DescriptionList.Description>
+        <DescriptionList.Description key={6}>
+          De naam waarmee een persoon wordt aangesproken
+        </DescriptionList.Description>
+      </DescriptionList.Group>,
+      <DescriptionList.Term key={1}>Geboortedatum</DescriptionList.Term>,
+      <DescriptionList.Description key={2}>De datum waarop een persoon is geboren</DescriptionList.Description>,
```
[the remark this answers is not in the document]
dlnr: I wanted to clearly develop the case where another DT + DD followed the multiple terms in order to prevent any bugs related to grid row or otherwise.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -54,6 +54,32 @@ export const MultipleDescriptions: Story = {
...
 
+export const MultipleTerms: Story = {
+  args: {
+    termsWidth: 'md',
+    children: [
+      <DescriptionList.Term key={1}>Achternaam</DescriptionList.Term>,
+      <DescriptionList.Description key={2}>
+        De naam die een persoon van zijn of haar ouders krijgt
+      </DescriptionList.Description>,
+      <DescriptionList.Group key={2}>
```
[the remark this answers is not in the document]
dlnr: Done

packages/css/src/components/description-list/description-list.scss
dlnr: Proposed a solution with inline style to set rows.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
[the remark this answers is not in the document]
VincentSmedinga: Alright

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: I've come to understand that the specification deliberately excludes implicit grids for negative indexes of grid item placement to prevent unpredictability and performance issues.

@dlnr @alimpens How about `grid-row: 1 / 999` as an alternative? This streches the cell to all three rows as well, doesn’t seem to cause any rendering issues in my Safari and Firefox and removes the need for inline styles.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -25,23 +25,42 @@
...
+  .ams-description-list--terms-width-lg .ams-description-list__section {
     grid-template-columns: 1fr 1fr;
   }
 }
 
+// Solution for browsers that support subgrid, the group will align to the parent grid
+// This selector is to increase specificity to override the previously set grid-template-columns
+@supports (grid-template-columns: subgrid) {
+  .ams-description-list .ams-description-list__section {
+    grid-template-columns: subgrid;
```
VincentSmedinga, line 60: Haha well done, you have your subgrid after all! 👍🏼

packages/css/src/components/description-list/description-list.scss
```diff
@@ -25,23 +25,42 @@
...
+  .ams-description-list--terms-width-lg,
+  .ams-description-list--terms-width-lg .ams-description-list__section {
     grid-template-columns: 1fr 1fr;
   }
 }
 
+// Solution for browsers that support subgrid, the group will align to the parent grid
+// This selector is to increase specificity to override the previously set grid-template-columns
+@supports (grid-template-columns: subgrid) {
+  .ams-description-list .ams-description-list__section {
```
VincentSmedinga, line 59: Increasing specificity is acceptable – but what if we used `@supports not` for the other selectors? We wouldn’t need the extra selector in that case.

dlnr: You would need a lot more lines of css because the other selectors include the parent list.

packages/css/src/components/description-list/description-list.scss
dlnr: Well not mine personally, just modern CSS.

packages/css/src/components/description-list/description-list.scss
VincentSmedinga: Not _very_ hacky.

Multiple terms sharing multiple declarations is an extreme edge case.

Inline styles can still be added and will override the 999 row span.

Let’s see which way Aram wants to go.

packages/react/src/DescriptionList/DescriptionListSection.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    expect(component).toBeInTheDocument()
+    expect(component).toBeVisible()
+  })
+
+  it('renders a design system BEM class name', () => {
+    const { container } = render(<DescriptionList.Section>Test</DescriptionList.Section>)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('ams-description-list__group')
```
VincentSmedinga, line 21: Group has been renamed to Section.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -54,8 +62,22 @@
...
 
-@mixin reset-dd {
-  margin-inline: 0;
+.ams-description-list__section {
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    column-gap: var(--ams-description-list-column-gap);
+    display: grid;
+    grid-column: span 2;
+
+    dd:only-of-type {
```
VincentSmedinga, line 71: Let’s target our classes instead of the element names.

```suggestion
    .ams-description-list__description:only-of-type {
```

dlnr: This won't work, `:only-of-type` applies to the tag not the classname.

VincentSmedinga: I don’t think it does – yes it selects elements by their tag name, but that doesn’t mean it can’t work together with a class selector.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -54,8 +62,22 @@
...
 
-@mixin reset-dd {
-  margin-inline: 0;
+.ams-description-list__section {
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    column-gap: var(--ams-description-list-column-gap);
+    display: grid;
+    grid-column: span 2;
+
+    :only-of-type {
```
VincentSmedinga, line 71: This would apply to all descendants of the section; we intend to target the terms and descriptions.

```suggestion
    > :only-of-type {
```

packages/css/src/components/description-list/description-list.scss
```diff
@@ -54,8 +62,22 @@
...
+      grid-row: 1 / 9;
+    }
+  }
+}
+
+// Solution for browsers that support subgrid, the group will align to the parent grid
+// This selector is to increase specificity to override the previously set grid-template-columns
```
VincentSmedinga, line 78:
```suggestion
// Aligns terms and descriptions in a section to the grid of the list.
// The extra class selector increases specificity to match earlier declarations.
```

VincentSmedinga (approved):

PR #1773, opened 2024-12-04
feat!: Make active Tab look less like a primary Button

VincentSmedinga:
# Describe the pull request

## What

Implements the updated design of Tabs.

## Why

The active Tab Button looked too much like a primary Button, making the latter less prominent.

## How

Tokens and CSS. I considered leaving some tokens in that we don’t need ourselves but could be useful for other communities to set background and border colors. However, I decided against it (see second commit) as we should first determine whether we want to adopt that approach at all. Even while it would make this change non-breaking.


RubenSibon (approved):
LGTM

packages/css/src/components/tabs/tabs.scss
```diff
@@ -54,8 +53,9 @@
   }
 
   &[aria-selected="true"] {
-    background-color: var(--ams-tabs-button-selected-background-color);
+    box-shadow: var(--ams-tabs-button-selected-box-shadow);
     color: var(--ams-tabs-button-selected-color);
+    font-weight: var(--ams-tabs-button-selected-font-weight);
```
alimpens, line 58: This causes a layout shift. [Here's a workaround for that](https://css-tricks.com/bold-on-hover-without-the-layout-shift/)

packages/css/src/components/tabs/tabs.scss
```diff
@@ -54,11 +54,26 @@
...
     }
   }
 }
+
+.ams-tabs__button-label,
+.ams-tabs__button-label-hidden {
+  grid-area: 1 / 1;
+}
+
+// This hidden label is used to prevent a layout shift when the mega menu is opened
```
alimpens, line 72:
```suggestion
// This hidden label is used to prevent a layout shift when the tab is selected
```

packages/react/src/Tabs/TabsButton.tsx
```diff
@@ -33,7 +33,12 @@ export const TabsButton = forwardRef(
         role="tab"
         tabIndex={activeTab === tab ? 0 : -1}
       >
-        {children}
+        <span aria-hidden="true" className="ams-tabs__button-label-hidden">
+          {children}
+        </span>
+        <span aria-hidden="true" className="ams-tabs__button-label">
```
alimpens, line 39:
```suggestion
        <span className="ams-tabs__button-label">
```

alimpens (approved):

Issue #1774, opened 2024-12-05
Tabs break when there is only one tab in <Tabs.List>


joeyamst:
# Describe the issue

When using the Tabs, it breaks with the error `Uncaught TypeError: children[0].props.children.map is not a function` in the `<Tabs>` component.

## Summary

```
            <Tabs activeTab={0}>
                <Tabs.List>
                    <Tabs.Button tab={0}>Tab 1</Tabs.Button>
                    {/*<Tabs.Button tab={1}>Tab 2</Tabs.Button>*/}
                </Tabs.List>
                <Tabs.Panel tab={0}>
                    <Heading level={3}>Tab 1 Title</Heading>
                </Tabs.Panel>
            </Tabs>
```
This breaks with the error. Uncommenting line 4 and it works fine again and shows 2 tabs.

## Environment

- Browser: Chrome
- Operating System: Windows with WSL
- Version of the design system: v0.13.1

## Checklist

Before submitting your issue, please ensure you have done the following:

- [ x ] Search the issue tracker to ensure the issue has not already been reported.
- [ x ] Verify that the issue can be consistently reproduced.
- [ x ] Check if the issue is already documented.

## Additional notes

I agree having only 1 tab is useless, but for developing sometimes you build one tab first, and add more later. Maybe tabs can also be dynamic on user input, creating a situation where it's possible to only having one tab.

RubenSibon:
A fix for this is incoming.

VincentSmedinga:
This has been resolved but has yet to be published. We expect a new release to ship in the second half of January. If it would be useful to have it earlier, let us know, and we can make it a hotfix.

VincentSmedinga:
@joeyamst The fix has been published in release 0.14.

PR #1775, opened 2024-12-05
fix!: Make Tabs correctly handle a single Tab Button

RubenSibon:
# Describe the pull request

A fix for issue #1774 that Tabs List gives an breaking error when there is only one, non-iterable element in the Tabs List.

## What

A Tabs List does not always expect multiple React Node's.

## Why

Tabs (List) should be able to handle a single Tab button.

## How

Remove the map method and just returning the value of `children`.

## Checklist

- [x] Add or update stories (tested with stories)
- [x] Manually tested in Prototypes (Amopis)

## Additional notes

- Issue: #1774


packages/react/src/Tabs/Tabs.tsx
```diff
@@ -25,7 +25,7 @@ const TabsRoot = forwardRef(
 
     const allTabs = useMemo(() => {
       if (!Array.isArray(children)) return []
-      return (children[0].props.children as ReactNode[]).map((child) => child)
+      return children[0].props.children
```
VincentSmedinga, line 28: Do we remember why this casting was done in the first place? (I don’t.)

RubenSibon: Nope, I don't know either. Removing it doesn't seem to be break anything.

RubenSibon:
Tested it again in the Tabs Story and in the existing Tabs in Amopis in Prototypes.

packages/react/src/Tabs/Tabs.tsx
VincentSmedinga: We might be bumping into React internals here. This change does indeed fix it. We discussed that we should not let `Tabs` interfere with children of `Tabs.List`; I’ll add a ticket for it.

VincentSmedinga (approved):

PR #1777, opened 2024-12-09
feat: Replace deprecated import statements in Sass with new API

dlnr:
# Describe the pull request

## What

Applied the SASS recommended migrator to fix the @import deprecation.

## Why

Warnings in build output.

## How

`npm install -g sass-migrator`
`sass-migrator module --migrate-deps index.scss`

## Checklist

- [ ] Add or update stories

## Additional notes


packages/css/src/components/accordion/accordion.scss
```diff
@@ -3,14 +3,14 @@
...
-@import "../../common/text-rendering";
+@use "../../common/text-rendering";
 
 .ams-accordion {
   display: flex;
   flex-direction: column;
   gap: var(--ams-accordion-gap);
 
-  @include text-rendering;
+  @include text-rendering.text-rendering;
```
VincentSmedinga, line 13: This looks weird. Should we do something about it?

I see `@use "…" as *;` is an option to remove the namespace. Duplicate names would throw an arror, but we can control that within our own library.

We could group related mixins into a single file with a conventional name to get e.g. `typography.text-rendering`.

Or we could accept this.

packages/css/src/components/breakout/breakout.scss
```diff
@@ -3,105 +3,105 @@
...
-@import "../grid/mixins";
+@use "../../common/breakpoint";
+@use "../grid/grid";
+@use "../grid/mixins";
 
 $ams-breakout-row-span-max: 4;
 
 .ams-breakout {
-  @include ams-grid;
+  @include mixins.ams-grid;
```
VincentSmedinga, line 13: There is a significant chance of conflicts here if multiple components expose a file called `mixins`.

packages/css/src/components/breakout/breakout.scss
```diff
@@ -3,105 +3,105 @@
...
 
   .ams-breakout__cell--col-start-#{$i} {
-    @include ams-grid__cell--start($i);
+    @include mixins.ams-grid__cell--start($i);
   }
 }
 
-@media screen and (min-width: $ams-breakpoint-medium) {
-  @for $i from 1 through $ams-grid-column-count {
+@media screen and (min-width: breakpoint.$ams-breakpoint-medium) {
```
VincentSmedinga, line 85: We could try to lose one of the two mentions of `breakpoint`.

packages/css/src/components/breakout/breakout.scss
dlnr: It should only import grid as *, only column count should be moved and included as a sass var like so `grid.$column-count`. Is that an option?

packages/css/src/components/breakout/breakout.scss
dlnr: Do they need the prefix? `breakpoints.$medium` looks better.

packages/css/src/components/blockquote/blockquote.scss
```diff
@@ -21,7 +21,7 @@
   line-height: var(--ams-blockquote-line-height);
   quotes: "“" "”";
 
-  @include hyphenation;
+  @include hyphenation.hyphenation;
   @include text-rendering;
```
VincentSmedinga, line 25: Mixin two approaches isn’t ideal.

packages/css/src/components/breakout/breakout.scss
VincentSmedinga: It depends on whether we consider the mixins public API and whether they may introduce conflicts if people use our library together with others. If either is true, we should add a prefix for functional and technical reasons. If not, we can consider esthetic motives.

VincentSmedinga (commented):
Thanks for the refactoring. I didn’t anticipate the namespace it would add; let’s decide on a good approach to keep the mixin names user-friendly.

packages/css/src/components/blockquote/blockquote.scss
dlnr: I agree. It might be better to bundle our mixins like so:

```css
@include text.hyphenation;
@include text.rendering;
```

packages/css/src/components/breakout/breakout.scss
VincentSmedinga: We could also export the mixins from `grid.scss`. Let’s discuss tomorrow.

VincentSmedinga (approved):

PR #1803, opened 2025-01-07
chore: update Node.js version to 22 in .nvmrc and package.json

dlnr:
Update to Node 22, the new LTS


.nvmrc
```diff
@@ -1 +1 @@
-v20
+22
```
alimpens, line 1: Can you add a newline here?

alimpens (approved):

PR #1814, opened 2025-01-13
feat: Remove Beta status from Header and deprecate Mega Menu

VincentSmedinga:
The Header component is no longer in Beta.

The Mega Menu component is deprecated. I’ve updated the Status Badge component to allow customizing the status label and making it red for a deprecation.

### DIFF
--- a/packages/react/src/MegaMenu/MegaMenu.tsx
+++ b/packages/react/src/MegaMenu/MegaMenu.tsx
@@ -20,4 +20,5 @@ const MegaMenuRoot = forwardRef(
 
 MegaMenuRoot.displayName = 'MegaMenu'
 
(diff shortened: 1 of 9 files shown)


packages/react/src/MegaMenu/MegaMenu.tsx
```diff
@@ -20,4 +20,5 @@ const MegaMenuRoot = forwardRef(
 
 MegaMenuRoot.displayName = 'MegaMenu'
 
+/** @deprecated Use child components in Header instead. */
```
VincentSmedinga, line 23: Do we want this deprecation to appear in the change log? If so, should we list it as a ‘feature’?

dlnr (approved):

PR #1819, opened 2025-01-16
feat(Accordion, Alert, Dialog, Form Error List, Heading)!: Require explicit heading levels

VincentSmedinga:
# Describe the pull request

## What

Requires providing a `level` for Heading and a `headingLevel` for Accordion, Alert, Dialog, and Form Error List.

## Why

The heading level has no sensible default; it can appear anywhere in the document’s hierarchy. We need people to consider which level fits best in each situation.

## How

- Made the appropriate props required
- Updated tests and stories

## Checklist

- [x] Add or update stories

## Additional notes

n/a


packages/react/src/Accordion/AccordionContext.tsx
```diff
@@ -4,11 +4,12 @@
...
+import type { AccordionProps } from './Accordion'
 import type { HeadingProps } from '../Heading/Heading'
 
 export type AccordionContextValue = {
   headingLevel: HeadingProps['level']
-  sectionAs?: 'div' | 'section'
+  sectionAs?: AccordionProps['sectionAs']
 }
 
 const defaultValues: AccordionContextValue = {
```
VincentSmedinga, line 15: There’s still a value of 1 for `headingLevel` in this object. (How) can we remove it?

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -28,7 +28,9 @@ const DialogRoot = forwardRef(
   ) => (
     <dialog {...restProps} ref={ref} className={clsx('ams-dialog', className)}>
       <header className="ams-dialog__header">
-        <Heading size="level-4">{heading}</Heading>
+        <Heading level={2} size="level-4">
```
alimpens, line 31: [A thread on which heading level to use for modal dialogs](https://github.com/mui/material-ui/issues/34250). I think `h1` makes most sense here.

packages/css/src/components/heading/README.md
```diff
@@ -2,16 +2,16 @@
 
 # Heading
 
-Describes the content that follows it.
-Use Headings to allow the user to grasp the structure of the page quickly.
```
alimpens, line 6: Why did you remove this line? I like it

packages/css/src/components/heading/README.md
```diff
@@ -2,16 +2,16 @@
...
 ## Guidelines
 
-- A heading describes the content beneath it.
-- Every page should contain one Heading with level 1.
-- Do not skip heading levels, e.g. a level 2 Heading should be followed by one with level 3, not level 4.
-- Do not use a Heading for formatting and styling, but to represent the page’s structure.
-  Users of screen readers use headings to navigate the page – an incorrect hierarchy can confuse them.
+- All headings together represent the document’s structure. Users of screen readers may use headings to navigate the page – an incorrect hierarchy will confuse them.
+- Every page should contain one heading with level 1.
+  The level of subsequent headings should be appropriate: a subsection must use one level higher than the section it belongs to.
```
alimpens, line 11: This is less clear than the previous line 12 imo.

packages/css/src/components/heading/README.md
```diff
@@ -2,16 +2,16 @@
...
+  The `level` prop doesn’t have a default value; the correct level should be determined for each instance.
+- Search engines give more weight to text in headers.
```
alimpens, line 13: Do we want to explicitly mention this here? I feel like this opens the door for SEO keyword dark pattern stuff.

packages/css/src/components/heading/README.md
```diff
@@ -2,16 +2,16 @@
...
+- Don’t use a Heading only to produce its appearance.
```
alimpens, line 14: This is less clear than the previous line 13 imo

packages/react/src/Alert/Alert.tsx
```diff
@@ -20,10 +20,10 @@ export type AlertProps = {
   /** The text for the Heading. */
   heading?: string
   /**
-   * The hierarchical level of the Heading within the document.
-   * Note: this intentionally does not change the font size.
```
alimpens, line 24: Don't we want to keep this line? I can imagine this is confusing for users.

packages/react/src/Heading/Heading.test.tsx
```diff
@@ -15,91 +15,54 @@ describe('Heading', () => {
...
-    })
-    const h3 = screen.getByRole('heading', {
-      name: 'H3',
-      level: 3,
-    })
-    const h4 = screen.getByRole('heading', {
-      name: 'H4',
-      level: 4,
-    })
+    const h2 = screen.getByRole('heading', { level: 2 })
```
alimpens, line 27: Using a name here is kind of useful I think. If `level={3}` sets a level 4 heading, and vice versa for example, this current test will still pass, even though it shouldn't.

packages/react/src/Heading/getHeadingTag.ts
```diff
@@ -1,14 +1,3 @@
...
-      return 'h2'
-    case 3:
-      return 'h3'
-    case 4:
-      return 'h4'
-    default:
-      return 'h1'
-  }
-}
+export const getHeadingTag = (level: HeadingProps['level']): `h${HeadingProps['level']}` => `h${level}`
```
alimpens, line 3: Without Typescript you can enter anything as `level`. The old function keeps working in that case, the new one doesn't.

packages/css/src/components/heading/README.md
VincentSmedinga: I want to separate the component’s rationale (or goal) from its usage instructions. I restored it into the ‘guidelines’ section.

packages/css/src/components/heading/README.md
VincentSmedinga: Restored.

packages/css/src/components/heading/README.md
VincentSmedinga: Well, SEO in itself isn’t bad. But let’s not open the subject now. Removed the line.

packages/css/src/components/heading/README.md
VincentSmedinga: Opted for middle ground: “Don’t use a Heading just for its visual appearance.”

packages/react/src/Alert/Alert.tsx
VincentSmedinga: Rephrased to “Note: the font size is intentionally the same for every level.” and included it for Alert, Accordion, Form Error List and Table of Contents.

packages/react/src/Dialog/Dialog.tsx
VincentSmedinga: Good find! [This comment](https://github.com/mui/material-ui/issues/34250#issuecomment-1244006106) is indeed convincing.

packages/react/src/Heading/Heading.test.tsx
VincentSmedinga: I don’t understand. How can `level={3}` set a level 4 heading?

packages/react/src/Heading/getHeadingTag.ts
VincentSmedinga: True, but the idea of this PR is that we can’t set a default. Only after I changed the `default` case to `case 1:` did I decide to refactor for simplicity.

What should this function return in JavaScript-land if a level of `5` of `a` was provided?

packages/react/src/Accordion/AccordionContext.tsx
VincentSmedinga: I’ve added `null` to the type in Context.

packages/react/src/Alert/Alert.tsx
alimpens: Ha, this is probably nitpicky, but... I like the old phrasing better. Saying this prop does not change the font size is clearer imo than saying the font size is the same for every level. Because it's not, different levels can have different font sizes, if the `size` prop is also set. `level` just doesn't change anything there.

packages/react/src/Heading/Heading.test.tsx
alimpens: If we make a mistake in the future 😁 Tests should catch mistakes, the current test would not catch that mistake.

packages/react/src/Heading/getHeadingTag.ts
alimpens: Hm, tricky question. Do we want to allow invalid HTML or an incorrect heading structure? And which is more easily detected as erroneous?

[Utrecht uses a default of `h1`](https://github.com/nl-design-system/utrecht/blob/de1c443072f724f825d432b41d2aacefc83c2402/packages/component-library-react/src/Heading.tsx#L29) even though `level` is required, so they've opted to not allow invalid HTML. 

I think I prefer that route. Invalid HTML is almost completely unaccessible, an incorrect heading structure is still somewhat usable. Invalid HTML is easier to spot though.

packages/react/src/Alert/Alert.tsx
VincentSmedinga: No worries, reverted to the initial wording.

packages/react/src/Heading/Heading.test.tsx
VincentSmedinga: I still don’t get how it helps, but I’ve restored the names.

packages/react/src/Heading/getHeadingTag.ts
VincentSmedinga: Yes, seems fair! I restored the function and added a comment.

VincentSmedinga: And allowed `null` as an argument to support that default value in Context.

packages/react/src/Alert/Alert.tsx
```diff
@@ -20,10 +20,11 @@ export type AlertProps = {
   /** The text for the Heading. */
   heading?: string
   /**
-   * The hierarchical level of the Heading within the document.
+   * This property represents the hierarchical level of the Alert’s Heading within the document.
```
alimpens, line 23:
```suggestion
   * The hierarchical level of the Alert’s Heading within the document.
```

alimpens (dismissed):

PR #1821, opened 2025-01-17
feat(Pagination)!: Use links instead of buttons

alimpens:
# Describe the pull request

## What

It makes Pagination use links instead of buttons. 

## Why

Links make more sense here. Pagination is a `nav` section, which has to contain links. 
In our guidelines we mention that clicking a Pagination link should navigate to the top of a new page.
That is default link behaviour.
Using links also allows users to open pages in a new tab, which isn't possible with buttons. 

## How

Using `a` instead of `button`, and getting rid of all the state stuff.

## Checklist

- [x] Add or update stories


packages/react/src/Pagination/Pagination.tsx
```diff
@@ -97,89 +54,59 @@ export const Pagination = forwardRef(
...
-
-    const navLabelId = useId()
+    const range = getRange(page, totalPages, maxVisiblePages)
 
     // Don't show pagination if you only have one page
     if (totalPages <= 1) {
       return null
     }
 
     return (
```
VincentSmedinga, line 67: The render function still looks pretty overwhelming. I think extracting local `PreviousLink`, `PageLink`, `Spacer`, and `NextLink` components would clarify.

storybook/src/components/Pagination/Pagination.docs.mdx
```diff
@@ -11,3 +11,38 @@ import README from "../../../../packages/css/src/components/pagination/README.md
...
+const linkTemplate = (page) => `/?pagina=${page}&categorie=video&thema=bestuur`
+
+return <Pagination linkTemplate={linkTemplate} />
+```
+
+### With a custom link component
+
+You can pass a custom component to render the Pagination links.
+This can be used to integrate Pagination into the routing library for your application.
+For example, you can use Pagination with the Next.js link component like so:
```
VincentSmedinga, line 34: I feel we should avoid addressing ‘you’ too explicitly. And it seems we can use less words here as well:

```suggestion
Pass a custom link component to use Pagination with your application’s routing library.
Here’s an example with the Link component of Next.js:
```

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -5,89 +5,46 @@
...
 import clsx from 'clsx'
-import { forwardRef, useId, useMemo, useState } from 'react'
-import type { ForwardedRef, HTMLAttributes } from 'react'
+import { forwardRef } from 'react'
+import type { AnchorHTMLAttributes, ComponentType, ForwardedRef, HTMLAttributes } from 'react'
+import { getRange } from './getRange'
 import { Icon } from '../Icon/Icon'
 
 export type PaginationProps = {
+  /** The id of the accessible label. */
```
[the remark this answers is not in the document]
alimpens: 'An identifying string' should be obvious for any dev, I think? I like the rest.

I just thought of something though: this breaks the users ability to set an `id` on the `nav` itself, so we should probably rename it. `visuallyHiddenLabelId`?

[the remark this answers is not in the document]
VincentSmedinga: > 'An identifying string' should be obvious for any dev, I think?

Yes, hopefully. Removed it from the suggestion.

> I just thought of something though: this breaks the users ability to set an id on the nav itself, so we should probably rename it. visuallyHiddenLabelId?

Ah, that’s right. And yes, that’s a suitable name.

packages/react/src/Pagination/Pagination.tsx
alimpens: As discussed, I've extracted `LinkItem` and `Spacer`. `PreviousLink` and `NextLink` would not add much.

VincentSmedinga (approved):

PR #1822, opened 2025-01-17
feat!: Combine contrast and inverse color options into a single prop

dlnr:
# Describe the pull request

## What

Every component that uses contrastColor and/or inverseColor props now have one `color` prop so we can prevent both props to be true and allow for scalability.

## Why

Unexpected results 

## How

Refactor components, stories, tests and css to use the color prop and correctly name the selectors.

## Checklist

- [x] Add or update stories

## Additional notes


packages/css/src/components/blockquote/blockquote.scss
```diff
@@ -34,6 +34,6 @@
   }
 }
 
-.ams-blockquote--inverse-color {
-  color: var(--ams-blockquote-inverse-color);
+.ams-blockquote--color-inverse {
```
VincentSmedinga, line 37: We don’t include prop names in our classes. This should be `.ams-blockquote-inverse`. And likewise for other components.

Compare: `.ams-button-primary` (variant), `.ams-paragraph-large` (size) etc.

(Let’s discuss and document the rules for this in our next dev refinment.)

packages/css/src/components/blockquote/blockquote.scss
```diff
@@ -34,6 +34,6 @@
...
+  color: var(--ams-blockquote-color-inverse);
```
VincentSmedinga, line 38: The name of component tokens must end with the CSS property they apply to. This should be `--ams-blockqoute-inverse-color`: block, modifier, property ([NLDS](https://nldesignsystem.nl/handboek/design-tokens/)). And likewise for other components.

proprietary/tokens/src/components/ams/blockquote.tokens.json
```diff
@@ -6,7 +6,7 @@
       "font-size": { "value": "{ams.text.level.3.font-size}" },
       "font-weight": { "value": "{ams.text.font-weight.bold}" },
       "line-height": { "value": "{ams.text.level.3.line-height}" },
-      "inverse-color": { "value": "{ams.color.text.inverse}" }
+      "color-inverse": { "value": "{ams.color.text.inverse}" }
```
VincentSmedinga, line 9: This must be nested. ‘inverse’ is the modifier, ‘color’ the CSS property. So:

```json
  "inverse": {
    "color": { "value": "{ams.color.text.inverse}" }
  }
```

And likewise for other components.

storybook/src/components/Blockquote/Blockquote.stories.tsx
```diff
@@ -14,13 +14,19 @@ const meta = {
...
-    inverseColor: false,
   },
   argTypes: {
     children: {
       description: 'The text for the quote.',
       table: { disable: false },
     },
+    color: {
+      control: {
+        labels: { inverse: 'inverse', undefined: 'default' },
```
VincentSmedinga, line 25: Repeating the label for a property with the same name is unnecessary.

```
labels: { undefined: 'default' },
```

VincentSmedinga (commented):
This isn’t a full review yet; I’ve noticed a few issues that need to be resolved for all components. I believe it’s best to tackle those first before evaluating the rest.

storybook/src/components/Blockquote/Blockquote.stories.tsx
dlnr: This is to allow you disable the inverse radio  button.

VincentSmedinga: Indeed – my point is that the mapping `inverse: 'inverse'` can be left out.

packages/css/src/components/blockquote/blockquote.scss
dlnr: This is confusing me, I had to drop "color" from the token name, while it was there before (`"inverse-color": { "color": {}}` will become `"inverse": { "color": {}}`) but the modifier will remain to be `--inverse-color`? Whilst a button variant like "primary" is not called "variant-primary".

storybook/src/components/Figure/Figure.stories.tsx
```diff
@@ -15,7 +15,13 @@ const meta = {
...
   args: {
     children: caption,
-    inverseColor: false,
+    color: {
+      control: {
+        labels: { undefined: 'default' },
+        type: 'radio',
+      },
+      options: [undefined, 'inverse'],
+    },
```
dlnr, line 24: @VincentSmedinga I'm having some difficulty with the color prop, typescript does not seem to help getting the FigureCaptionProps.

VincentSmedinga (approved):
I had some syntactic remarks but I fixed them myself. Nice work!

PR #1824, opened 2025-01-20
fix(Avatar): Remove incorrect aspect ratio for image

VincentSmedinga:
# Describe the pull request

## What

Restores the correct display of an Image in an Avatar.

## Why

The new default aspect ratio for the Image component (introduced in #1593) forces it to 16:9.

## How

I’m resetting the aspect ratio for an `.ams-image` in an `.ams-avatar`. We don’t know its dimensions, and the user can’t easily provide them. We already ensured to show the center of the image; I’ve made that explicit in the docs.

I decided against using a clean HTML `img` element because we still want to apply the other base styles of our image component.

## Checklist

- [x] Add or update stories

## Additional notes

None.

### DIFF
--- a/packages/css/src/components/avatar/avatar.scss
+++ b/packages/css/src/components/avatar/avatar.scss
@@ -15,10 +15,6 @@
   padding-inline: var(--ams-avatar-padding-inline);
   place-content: center;
 
-  .ams-image {
-    aspect-ratio: initial;
-  }
-
   svg {
     fill: currentColor;
   }
--- a/packages/css/src/components/avatar/avatar.scss
+++ b/packages/css/src/components/avatar/avatar.scss
@@ -15,6 +15,10 @@
   padding-inline: var(--ams-avatar-padding-inline);
   place-content: center;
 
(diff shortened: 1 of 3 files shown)


packages/css/src/components/avatar/avatar.scss
```diff
@@ -15,6 +15,10 @@
   padding-inline: var(--ams-avatar-padding-inline);
   place-content: center;
 
+  .ams-image {
```
alimpens, line 18: You wrote: 

> I decided against using a clean HTML img element because we still want to apply the other base styles of our image component.

Do we? All the aspect ratio stuff in there is irrelevant, because we reset it here. The phantom whitespace thing is not an issue, because this image is a child of a wrapper with an aspect ratio. And the alt text of this image is an empty string. Not much left I think?

VincentSmedinga: Alright, let’s just use an HTML `img` element, then.

alimpens (approved):

PR #1820, opened 2025-01-16
feat!: Change all brand colour tokens to use semantic names

VincentSmedinga:
# Describe the pull request

Thank you for contributing to the project!
Please use this template to help us handle your PR smoothly.

## What

- This gives all brand colour tokens semantic names and groups them.
- I’ve updated the appearance of our colour palette docs and added design token tables for each group.
- A few ‘visited link’ styles and tokens have been removed; they’ve always been half-baked, and removing them is better than adding a correct implementation to this refactoring branch.

## Why

To help us and others apply colour correctly.

## How

Several discussions and experiments over the last couple of months have led to the categorization on [this page](https://designsystem.amsterdam/demo-DES-873-brand-color-tokens/?path=/docs/brand-design-tokens-colour--docs). All new colour tokens have been applied to components.

## Additional notes

- Various PRs will follow to complete the set of colours, especially the ‘interactive’ group.

PR #1825, opened 2025-01-20
feat(Avatar)!: Remove some colour options and make purple the default

VincentSmedinga:
# Describe the pull request

## What

This removes the possibility of making an Avatar blue (it is not interactive), red (it can’t communicate an error or an invalid interactive state), or neutral (we don’t have tokens for these options). The remaining colours are in the ‘highlight’ group that #1820 will introduce.

## Why

See above.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages/css/src/components/avatar/README.md
+++ b/packages/css/src/components/avatar/README.md
@@ -6,10 +6,11 @@ A circular badge representing a person.
 
 ## Design
 
-The avatar contains 1 or 2 initial letters from the person's full name, a picture, or a generic icon.
-The default background colour is dark blue.
+The avatar contains 1 or 2 initial letters from the person’s full name, a picture, or a generic icon.
+To personalize the avatar, the user can be allowd to choose one of the highlight colours.
+The default background colour is purple.
 
 ## Guidelines
 
-- Display an avatar for the person currently using the application,
+- Display an avatar for the person who is using the application,
   or to associate a person with a content item.
--- a/packages/css/src/components/avatar/README.md
+++ b/packages/css/src/components/avatar/README.md
@@ -7,10 +7,10 @@ A circular badge representing a person.
 ## Design
 
 The avatar contains 1 or 2 initial letters from the person’s full name, a picture, or a generic icon.
-To personalize the avatar, the user can be allowd to choose one of the highlight colours.
+To personalize the avatar, the user can be allowed to choose one of the highlight colours.
 The default background colour is purple.
 
 ## Guidelines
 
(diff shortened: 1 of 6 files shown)


packages/css/src/components/avatar/README.md
```diff
@@ -6,10 +6,11 @@ A circular badge representing a person.
 
 ## Design
 
-The avatar contains 1 or 2 initial letters from the person's full name, a picture, or a generic icon.
-The default background colour is dark blue.
+The avatar contains 1 or 2 initial letters from the person’s full name, a picture, or a generic icon.
+To personalize the avatar, the user can be allowd to choose one of the highlight colours.
```
alimpens, line 10:
```suggestion
To personalize the avatar, the user can be allowed to choose one of the highlight colours.
```

packages/css/src/components/avatar/README.md
```diff
@@ -6,10 +6,11 @@ A circular badge representing a person.
...
+The default background colour is purple.
 
 ## Guidelines
 
-- Display an avatar for the person currently using the application,
+- Display an avatar for the person who is using the application,
```
alimpens, line 15:
```suggestion
- Show an avatar for the user of the application,
```

(Sidenote: I often use [this AI text rewriter](https://ahrefs.com/writing-tools/paragraph-rewriter) on the 'Simplified' setting, that generates pretty good results imo.)

packages/react/src/Avatar/Avatar.tsx
```diff
@@ -56,7 +42,10 @@ export type AvatarProps = {
 } & HTMLAttributes<HTMLSpanElement>
 
 export const Avatar = forwardRef(
-  ({ label, imageSrc, className, color = 'blue', ...restProps }: AvatarProps, ref: ForwardedRef<HTMLSpanElement>) => {
+  (
+    { label, imageSrc, className, color = defaultColor, ...restProps }: AvatarProps,
```
alimpens, line 46: Wrong line, but: the component still has a `ams-avatar--purple` class by default, we can get rid of that

VincentSmedinga: Good catch, it’s gone. We don’t even need the explicit default value that isn’t an actual value anywhere anyway.

VincentSmedinga: Yes, done. To reuse the `avatarColors` array, I need to import it through an elaborate path. Is that okay?

alimpens (approved):

Issue #1846, opened 2025-02-02
V0.14.0 Inconsistent MouseEvent Target in Tabs Component


remyvdwereld:
**Inconsistent MouseEvent Target in Tabs Component**

**Summary:**  
Since updating to version 14 of `@amsterdam/design-system-react`, the `Tabs` component is passing different `MouseEvent` targets depending on where the user clicks. The event target is sometimes a `<span>` and other times a `<button>`.

**Steps to Reproduce:**
1. Implement the `Tabs` component with an `onClick` function.
2. Log `e.target` in the `onClick` function.
3. Click on different parts of the `Tabs.Button`.
4. Observe that the event target alternates between a `<span>` and a `<button>`.

**Expected Behavior:**  
The `MouseEvent.target` should consistently refer to the same element type (ideally the `Tabs.Button` component) regardless of where the user clicks.

**Actual Behavior:**  
- Clicking on the button text results in `e.target` being a `<span>`.
- Clicking on the button's outer area results in `e.target` being a `<button>`.

**Code Sample:**
```tsx
const onClick = (e: React.MouseEvent<HTMLElement>) => {
  console.log("e.target:", e.target)
}

<Tabs onClick={ onClick }>
  <Tabs.List>
    <Tabs.Button tab={0}>Content 0</Tabs.Button>
    <Tabs.Button tab={1}>Content 1</Tabs.Button>
  </Tabs.List>
</Tabs>
```

**Environment:**
- `@amsterdam/design-system-react` version: 14.x
- Browser: Chrome

**Why do I need this?**
I want to perform actions based on the clicked Tab.

dlnr:
Discussed with @remyvdwereld and it turns out he adds the onClick event to a tab button to be able to save the active tab to the state. In case you return to this page, te previously visited tab should be active.

The `TabsContext` is not accessible outside the component. I'm tinkering around to find the best way.

@VincentSmedinga I don't think the PR is needed.

VincentSmedinga:
@remyvdwereld We’ve just released [@amsterdam/design-system-react v0.14.1](https://www.npmjs.com/package/@amsterdam/design-system-react) to add the `onChange` prop to `Tabs` and execute an `onClick` callback on a `Tabs.Button`, although your use case probably only requires the former.

You can use `event.currentTarget` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)) instead of `event.target` to always get the `button` element of `Tabs.Button`.

[...]

PR #1847, opened 2025-02-03
fix(Tabs): Execute click handler on Button

VincentSmedinga:
# Describe the pull request

## What

Disallows children of a `Tabs.Button` to become the target of pointer events.

## Why

To ensure the click event always returns the button as its target.

## How

Through CSS.

## Checklist

- [x] Add or update stories

## Additional notes

- Adresses #1846.
- We can release this as a 0.14.1 hotfix. This is why it merges into a release branch.
- I considered adding a Storybook interaction test, but decided against it to keep the bugfix minimal. We should make a start with the subject, though.
- Test this by adding `onClick: (e: MouseEvent<HTMLElement>) => console.log('e.target:', e.target)` to the `args` of the Tabs stories while importing `MouseEvent` from React.

### DIFF
--- a/packages/css/src/components/tabs/tabs.scss
+++ b/packages/css/src/components/tabs/tabs.scss
@@ -62,6 +62,10 @@
       background-color: SelectedItem;
     }
   }
+
+  * {
+    pointer-events: none;
+  }
 }
 
 .ams-tabs__button-label,
--- a/packages/css/src/components/tabs/tabs.scss
+++ b/packages/css/src/components/tabs/tabs.scss
@@ -62,10 +62,6 @@
       background-color: SelectedItem;
     }
   }
-
-  * {
-    pointer-events: none;
-  }
 }
 
 .ams-tabs__button-label,
--- a/packages/react/src/Tabs/TabsButton.tsx
+++ b/packages/react/src/Tabs/TabsButton.tsx
@@ -5,7 +5,7 @@
 
 import clsx from 'clsx'
 import { forwardRef, startTransition, useContext } from 'react'
-import type { ButtonHTMLAttributes, ForwardedRef, PropsWithChildren } from 'react'
+import type { ButtonHTMLAttributes, ForwardedRef, MouseEvent, PropsWithChildren } from 'react'
 import { TabsContext } from './TabsContext'
 
 export type TabsButtonProps = {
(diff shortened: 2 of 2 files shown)


dlnr (approved):

packages/css/src/components/tabs/tabs.scss
```diff
@@ -62,6 +62,10 @@
       background-color: SelectedItem;
     }
   }
+
+  * {
+    pointer-events: none;
+  }
```
RubenSibon, line 68: I feel this should be sorted before the state selectors on line 41 as it selects all child elements.

Could it target `span` instead to be more specific and explicit?

RubenSibon (approved):
LGTM; one comment about the wildcard CSS selector.

packages/react/src/Tabs/TabsButton.tsx
```diff
@@ -14,21 +14,24 @@ export type TabsButtonProps = {
 } & PropsWithChildren<ButtonHTMLAttributes<HTMLButtonElement>>
 
 export const TabsButton = forwardRef(
-  ({ children, className, tab = 0, ...restProps }: TabsButtonProps, ref: ForwardedRef<HTMLButtonElement>) => {
+  ({ children, className, onClick, tab = 0, ...restProps }: TabsButtonProps, ref: ForwardedRef<HTMLButtonElement>) => {
     const { activeTab, tabsId, updateTab } = useContext(TabsContext)
 
+    const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+      onClick?.(event)
```
dlnr, line 21: I thought we would return `activeTab` here.

packages/css/src/components/tabs/tabs.scss
VincentSmedinga: > I feel this should be sorted before the state selectors on line 41 as it selects all child elements.

No, having the default and other variations of the component next to each other, before adressing other components (within it) is more logical. (Actually, a separate declaration block for `.ams-tabs__button *` would make this even clearer.)

> Could it target `span` instead to be more specific and explicit?

Our intention is precisely to let any child of a tabs button ignore its pointer events. The universe selector _is_ specific and explicit here.

alimpens, line 68: I think we're solving a problem that has already been solved here. If you need to access the element that has an `onClick` handler, you can use `event.currentTarget`. If you need to access the element that has been clicked, you can use `event.target`.

Adding CSS for that is unnecessary imo, and it breaks the second case.

VincentSmedinga: Oh my, good point. Reverted it and changed the PR title.

alimpens (approved):

PR #1862, opened 2025-02-13
chore(docs): Add mega menu to example pages

VincentSmedinga:
This implements the Mega Menu in our example pages.

### DIFF
--- a/storybook/src/pages/amsterdam/common/AppHeader.tsx
+++ b/storybook/src/pages/amsterdam/common/AppHeader.tsx
@@ -1,25 +1,41 @@
-import { Header, LinkList } from '@amsterdam/design-system-react'
+import { Grid, Header, Heading, LinkList } from '@amsterdam/design-system-react'
+import { headerMenuLinks, megaMenuLinks } from './menu'
 
 export const AppHeader = () => (
   <Header
-    menuItems={[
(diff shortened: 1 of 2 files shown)


storybook/src/pages/amsterdam/common/AppHeader.tsx
```diff
@@ -1,25 +1,41 @@
...
-      <Header.MenuLink href="#" key="2">
-        Mijn Amsterdam
-      </Header.MenuLink>,
-      <Header.MenuLink fixed href="#" key="3">
-        Zoeken
-      </Header.MenuLink>,
-    ]}
-    noMenuButtonOnWideWindow
+    menuButtonText="Alle onderwerpen"
+    menuItems={headerMenuLinks.map(({ fixed, href, label, lang }) => (
```
VincentSmedinga, line 7: Should we upgrade this function to a component for people to (optionally) use? Same for rendering the Link List Links below.

alimpens: Hm, feels a little over-engineered to me. You have `Header.MenuLink` which is typed, so you already know which props there are and which ones are required etc. I would probably not use a loop here in an application, for readability. So not really necessary, imo.

alimpens (approved):

PR #1876, opened 2025-02-17
feat(Dialog): Change background colour of backdrop and add token for it

VincentSmedinga:
# Describe the pull request

## What

- Changes the background colour to use our new black. The designers have looked into the opacity but left it unchanged.
- Adds a token for the background colour of the backdrop of a Dialog. Support for using custom properties in `::backdrop` is 89.35% in The Netherlands – hence the fallback.
- Adds example content behind the Dialog in two examples, to more realistically see the effect of the backdrop colour.

## Why

- Update the colour to match the design
- Use a token because support for it has increased.

## How

- n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -9,6 +9,9 @@
       "max-inline-size": { "value": "48rem" },
       "padding-block": { "value": "{ams.space.grid.md}" },
       "padding-inline": { "value": "{ams.space.grid.lg}" },
+      "backdrop": {
+        "background-color": { "value": "rgb(24 24 24 / 60%)" }
```
VincentSmedinga, line 13: I’ve requested this to be changed to 62.5% :grin:

storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -86,10 +92,12 @@ export const OpenAndClose: Story = {
...
   decorators: [
     (Story) => (
-      <div>
-        <Button onClick={() => Dialog.open('#ams-dialog')}>Open Dialog</Button>
+      <Column alignHorizontal="start">
+        <Heading level={1}>{heading}</Heading>
+        <Paragraph>{paragraph}</Paragraph>
+        <Button onClick={() => Dialog.open('#ams-dialog')}>Open dialoog</Button>
         <Story />
-      </div>
```
RubenSibon, line 92: Is it possible to use the actual backdrop from the component in the story? When you close the dialog now, it doesn't remove the backdrop.

proprietary/tokens/src/components/ams/dialog.tokens.json
RubenSibon: Why this specific, minor change? Personal preference?

VincentSmedinga: We use powers of 2 in many places, and 62.5% is 5/8. We also use 12.5% transparency for the hover of Icon Button.

storybook/src/components/Dialog/Dialog.stories.tsx
VincentSmedinga: Yea that would be an improvement. Displaying a Dialog is a bit more difficult to than simple flow components. I’m also under the impression that both backdrop colours are different.

We could use the ‘Open and Close’ story as the default one, and trigger a click on the button when loading the story.

A complication, as always, is that stories on the Docs page aren’t in an iframe.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -24,9 +24,12 @@ so do not apply these styles without an `open` attribute. */
 
   @include reset-dialog;
 
-  /* No token because dialog does not inherit from any element in FireFox and Safari */
+  /*
+   * Fallback for browsers where ::backdrop does not inherit from its originating element.
```
VincentSmedinga, line 28:
```suggestion
   * The fallback value is for browsers where ::backdrop does not inherit from its originating element.
```

RubenSibon (dismissed):

storybook/src/styles/docs.css
```diff
@@ -67,9 +67,11 @@
 }
 
 .ams-docs-backdrop {
-  background-color: #0006;
+  background-color: var(--ams-dialog-backdrop-background-color);
   block-size: 100%;
   inline-size: 100%;
+  padding-block: 1rem;
+  padding-inline: 1rem;
```
alimpens, line 74: This padding makes the container overflow on Chrome, I don't think we want that?

VincentSmedinga: We don’t. I applied border-box sizing.

alimpens (approved):

PR #1878, opened 2025-02-18
fix(gap, margin): Use regular space tokens for gap and margin utilities

VincentSmedinga:
# Describe the pull request

## What

It uses regular space tokens for the margin and gap utilities.

## Why

Because using grid space tokens here has been unintentional all along – we’re puzzled how this got through our code reviews.

This will decrease whitespace where these tokens are used. We apologise to users who must update many instances in their application.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/proprietary/tokens/src/components/ams/gap.tokens.json
+++ b/proprietary/tokens/src/components/ams/gap.tokens.json
@@ -1,11 +1,11 @@
 {
   "ams": {
     "gap": {
-      "xs": { "value": "{ams.space.grid.xs}" },
-      "sm": { "value": "{ams.space.grid.sm}" },
-      "md": { "value": "{ams.space.grid.md}" },
-      "lg": { "value": "{ams.space.grid.lg}" },
-      "xl": { "value": "{ams.space.grid.xl}" }
+      "xs": { "value": "{ams.space.xs}" },
+      "sm": { "value": "{ams.space.sm}" },
+      "md": { "value": "{ams.space.md}" },
+      "lg": { "value": "{ams.space.lg}" },
+      "xl": { "value": "{ams.space.xl}" }
     }
   }
 }
--- a/proprietary/tokens/src/components/ams/margin.tokens.json
+++ b/proprietary/tokens/src/components/ams/margin.tokens.json
@@ -1,11 +1,11 @@
 {
   "ams": {
     "margin": {
-      "xs": { "value": "{ams.space.grid.xs}" },
-      "sm": { "value": "{ams.space.grid.sm}" },
-      "md": { "value": "{ams.space.grid.md}" },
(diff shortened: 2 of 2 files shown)


RubenSibon (commented):
Easy enough. But isn't this a breaking change?

VincentSmedinga:
> Easy enough. But isn't this a breaking change?

I must admit I have been thinking about that. Technically, not from an API perspective. But from a visual perspective, we might mark it as such. What do you think @alimpens?

alimpens:
> > Easy enough. But isn't this a breaking change?
> 
> I must admit I have been thinking about that. Technically, not from an API perspective. But from a visual perspective, we might mark it as such. What do you think @alimpens?

Yeah, I think we should mark it as a breaking change, in the sense that devs have to change something in their app to keep it working as it did. TBH, that's why I also think we shouldn't merge this now. We're doing more [work on our space tokens](https://gemeente-amsterdam.atlassian.net/browse/DES-1078) soon, which will probably impact these util classes as well. We shouldn't make users have to fix this twice.

VincentSmedinga:
> TBH, that's why I also think we shouldn't merge this now. We're doing more [work on our space tokens](https://gemeente-amsterdam.atlassian.net/browse/DES-1078) soon, which will probably impact these util classes as well. We shouldn't make users have to fix this twice.

I don’t want to publish a new release until all of the spacing work is done anyway. I started a branch for [this related spacing ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-616), had it as its first commit, then thought it would be good to mention this change separately in the change log. Hence this PR, clearer for everyone to follow then if it gets buried in a larger refactoring.

alimpens:
> > Yeah, I think we should mark it as a breaking change, in the sense that devs have to change something in their app to keep it working as it did.
> 
> Agreed.
> 
> > TBH, that's why I also think we shouldn't merge this now. We're doing more [work on our space tokens](https://gemeente-amsterdam.atlassian.net/browse/DES-1078) soon, which will probably impact these util classes as well. We shouldn't make users have to fix this twice.
> 
> I don’t want to publish a new release until all of the spacing work is done anyway. I started a branch for [this related spacing ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-616), had it as its first commit, then thought it would be good to mention this change separately in the change log. Hence this PR, clearer for everyone to follow then if it gets buried in a larger refactoring.

I think it's a bad idea to merge stuff into `develop` that isn't releasable. It makes it easier to accidentally release stuff we don't want to. And I'm not sure it's better to have separate entries in the changelog for this. As a user, I think I'd rather have 1 entry that says 'the space util classes have changed' than multiple entries where I have to piece together what has happened.

Issue #1828, opened 2025-01-23
Dialog body overflow and padding


dlnr:
The Dialog component has a div that holds the `children`, but the padding applies to the parent of that div which causes an overflow issue. This is because `ams-dialog__body` sets overflow-y to `auto` for scrolling.

<img width="204" alt="Image" src="https://github.com/user-attachments/assets/6d106930-835a-4944-b0a1-4d710dc4ca64" />

_A form in Dialog using the compact theme_

We could try a full-bleed boolean on the dialog that moves the padding from `ams-dialog` to `ams-dialog__body` (and other siblings) or apply it there permanently.

Another issue as you can see on this screenshot is that the block padding grows less then the inline padding causing inconsistency across screen sizes:

```css
	--ams-dialog-padding-block: var(--ams-space-grid-md);
	--ams-dialog-padding-inline: var(--ams-space-grid-lg);
```

This issue was reported to me by @remyvdwereld and I've noticed it myself when using form elements in the Dialog component.

VincentSmedinga:
Thanks! Because the body is a scroll container, the browser creates a block formatting context for it, and it can’t draw anything outside of that.

A fix is underway: https://github.com/Amsterdam/design-system/pull/1879. It moves the padding from the `Dialog` container to its header, body and footer, so that the body has horizontal space to draw the entire focus indicator.

VincentSmedinga:
This fix been released in version 0.15 of the design system packages.

Please close this issue if it resolves the problem you encountered.

PR #1879, opened 2025-02-18
fix(Dialog)!: Prevent a focus indicator from being cut off

VincentSmedinga:
# Describe the pull request

## What

This addresses https://github.com/Amsterdam/design-system/issues/1828. It moves padding from `Dialog` down to its header, body and footer sections.

## Why

To prevent focus indicators at the edge of the body to become cut off. This could happen because the body is a scroll container without any padding. The browser creates a new block formatting context for it, and it can’t draw anything outside of that.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- Header and footer now have zero bottom and top padding – this white space is managed by the gap of the Dialog. It uses component space, while padding uses grid space, and I didn’t want to change that now.


packages/css/src/components/dialog/dialog.scss
```diff
@@ -13,14 +15,12 @@ so do not apply these styles without an `open` attribute. */
   background-color: var(--ams-dialog-background-color);
   border: var(--ams-dialog-border);
   box-sizing: border-box;
-  display: flex; // Using flex here, because grid works strangely in Safari
```
RubenSibon, line 16: Why was this comment removed, but nothing was changed?

Apparently, there was the desire to use Grid here, but we couldn't because of Safari.

VincentSmedinga:
> Is it worth the wait for the release of correct spacing tokens?

The space tokens are currently correct as well. And if we change them, we’ll come across the Dialog layout. So, no, let’s not wait.

packages/css/src/components/dialog/dialog.scss
alimpens: Our convention is to use `display: grid` instead of `display: flex` and `flex-direction: column`. We diverge from that convention here, we should explain why imo.

dlnr: We prefer grid over flex? That's news to me, both serve different purposes. Setting up these rules for yourself instead of picking the right options seems weird to me.

dlnr (dismissed):
This solves the overflow bug and I don't think it's breaking anything either.

RubenSibon (dismissed):

packages/css/src/components/dialog/dialog.scss
RubenSibon: I was also not aware of grid being the preference and I agree that flex is perfectly fine to use for 1 dimensional layouts. The comment doesn't clarify anything meaningful or add much value. I'm resolving the conversation.

PR #1880, opened 2025-02-19
feat: Add responsive version of Amsterdam Logo in Header

alimpens:
# Describe the pull request

## What

It adds a smaller logo (emblem only) to the Header, which is shown on narrow screens. 

## Why

It is in the original design, we didn't implement it yet. 

## How

By adding an emblem-only version of the Amsterdam logo to our Logo component, and showing and hiding that in our Header component. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages/css/src/components/header/header.scss
+++ b/packages/css/src/components/header/header.scss
@@ -30,6 +30,22 @@
   user-select: none; // The hidden branding section should not be selectable
 }
 
+// Increase the specificity to override the default logo styles
+.ams-header__logo.ams-header__logo {
+  display: none;
+
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    display: block;
+  }
+}
+
+// Increase the specificity to override the default logo styles
+.ams-header__logo--emblem-only.ams-header__logo--emblem-only {
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    display: none;
+  }
+}
+
 .ams-header__logo-link {
   outline-offset: var(--ams-header-logo-link-outline-offset);
 }
--- a/packages/react/src/Header/Header.tsx
+++ b/packages/react/src/Header/Header.tsx
@@ -15,10 +15,12 @@ import { HeaderMenuIcon } from './HeaderMenuIcon'
 import { HeaderMenuLink } from './HeaderMenuLink'
 import useIsAfterBreakpoint from '../common/useIsAfterBreakpoint'
 
-const LogoComponent = ({ logoBrand }: { logoBrand: LogoBrand }) => (
+const LogoComponent = ({ logoBrand, responsiveLogo }: { logoBrand: LogoBrand; responsiveLogo: boolean }) => (
   <>
-    <Logo brand={logoBrand} className="ams-header__logo" />
-    {logoBrand === 'amsterdam' && <Logo brand="amsterdam-emblem-only" className="ams-header__logo--emblem-only" />}
+    <Logo brand={logoBrand} className={clsx(logoBrand === 'amsterdam' && responsiveLogo && 'ams-header__logo')} />
+    {logoBrand === 'amsterdam' && responsiveLogo && (
+      <Logo brand="amsterdam-emblem-only" className="ams-header__logo--emblem-only" />
+    )}
   </>
 )
 
@@ -26,7 +28,7 @@ export type HeaderProps = {
   /** The name of the application. */
   brandName?: string
   /** The name of the brand for which to display the logo. */
-  logoBrand?: LogoBrand
+  logoBrand?: Exclude<LogoBrand, 'amsterdam-emblem-only'>
   /** The url for the link on the logo. */
   logoLink?: string
   /** The accessible text for the link on the logo. */
@@ -39,6 +41,8 @@ export type HeaderProps = {
   navigationLabel?: string
   /** Whether the menu button is visible on wide screens.  */
   noMenuButtonOnWideWindow?: boolean
+  /** Whether the Amsterdam logo should be responsive. */
+  responsiveLogo?: boolean
 } & HTMLAttributes<HTMLElement>
 
 const HeaderRoot = forwardRef(
@@ -54,6 +58,7 @@ const HeaderRoot = forwardRef(
       menuItems,
       navigationLabel = 'Hoofdnavigatie',
       noMenuButtonOnWideWindow,
+      responsiveLogo = true,
       ...restProps
     }: HeaderProps,
     ref: ForwardedRef<HTMLElement>,
@@ -74,7 +79,7 @@ const HeaderRoot = forwardRef(
         <div className="ams-header__branding">
           <a className="ams-header__logo-link" href={logoLink}>
             <span className="ams-visually-hidden">{logoLinkTitle}</span>
-            <LogoComponent logoBrand={logoBrand} />
+            <LogoComponent logoBrand={logoBrand} responsiveLogo={responsiveLogo} />
           </a>
           {brandName && (
             <Heading className="ams-header__brand-name" level={1} size="level-5">
@@ -91,7 +96,7 @@ const HeaderRoot = forwardRef(
             {/* The branding section is recreated here, to make sure the page menu breaks at the right spot */}
             <div aria-hidden className="ams-header__branding ams-header__branding--hidden">
               <div className="ams-header__logo-link">
-                <LogoComponent logoBrand={logoBrand} />
+                <LogoComponent logoBrand={logoBrand} responsiveLogo={responsiveLogo} />
               </div>
               {brandName && (
                 <span className="ams-heading ams-heading--level-5 ams-header__brand-name">{brandName}</span>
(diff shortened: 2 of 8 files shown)


packages/react/src/Header/Header.tsx
```diff
@@ -32,6 +41,8 @@ export type HeaderProps = {
   navigationLabel?: string
   /** Whether the menu button is visible on wide screens.  */
   noMenuButtonOnWideWindow?: boolean
+  /** Whether the Amsterdam logo should be responsive. */
+  responsiveLogo?: boolean
```
VincentSmedinga, line 45: I’m not sure we should make this a setting. Doesn’t it complicate things too much? Leaves the Header design up to implementers?

A full logo will often not fit with a few menu items and some zooming. I’d say always hide the wordmark in narrow windows.

alimpens: Also for amsterdam.nl? They don't have `brandName` text, so I can imagine they'd want to show the entire logo. I'm not sure where those designs are though

RubenSibon: Maybe make more clear in the documentation and prop comment that most of the time developers do not have to touch this setting and can leave it at the default?

Generally, I think that an option like this is nice to have and doesn't add too much complication.

VincentSmedinga: I’ve asked the branding team whether always hiding the wordmark on narrow windows is acceptable. I’ve copied the text into the ticket in Jira.

Until then, we can derive the value for `responsiveLogo` from the other props. It must be false if the brandName is not ‘amsterdam’ or if the brandName is ‘amsterdam’ and a ‘brandName’ has been set. That keeps the word mark visible for the main website, which is probably what we want. The prop could then be removed.

packages/react/src/Header/Header.tsx
```diff
@@ -84,7 +96,7 @@ const HeaderRoot = forwardRef(
             {/* The branding section is recreated here, to make sure the page menu breaks at the right spot */}
             <div aria-hidden className="ams-header__branding ams-header__branding--hidden">
               <div className="ams-header__logo-link">
-                <Logo brand={logoBrand} />
+                <LogoComponent logoBrand={logoBrand} responsiveLogo={responsiveLogo} />
```
dlnr, line 99: The nice thing about the logo svg was that you could've just hidden the text path.

packages/css/src/components/header/header.scss
dlnr, line 1: My initial idea to do this would be to use a named container query on the header branding div. If it would exceed the width of the (full) logo you could hide the logo text. Using a grid stack in the header prevents this opportunity though.

packages/react/src/Header/Header.tsx
VincentSmedinga: The Communication department has confirmed that the main amsterdam.nl website must always display the full logo.

packages/react/src/Header/Header.tsx
VincentSmedinga: That doesn’t make the SVG any narrower. How would you approach that?

dlnr: I've tried to use CSS modules to do this:
```css
@container (max-width: 36rem) {
    .header div:first-child > a {
        max-inline-size: 12px;
    }
    .header div:first-child > a svg path:nth-child(1) {
        opacity: 0;
    }
}
```

dlnr: My problem with the breakpoint is that you do not anticipate a long or short brandname. So a tablet size with a short title will hide the logo text way too soon.

I would prefer a named container for the brandname but a grid stack has been used so it won't work.

alimpens: > The dimensions are pretty much fixed

They aren't though. The logo grows from 40 to 56 px in height. I'm not sure your approach works with a fluid logo.

packages/css/src/components/header/header.scss
```diff
@@ -83,7 +94,6 @@
 
 @mixin header-menu-action {
   color: var(--ams-header-menu-item-color);
-  font-family: var(--ams-header-menu-item-font-family);
```
VincentSmedinga, line 86: The menu button no longer uses our font; buttons inherit styles selectively.

packages/css/src/components/header/header.scss
```diff
@@ -13,29 +13,40 @@
   * Display grid is used to let both branding sections overlap.
   */
   display: grid;
+  font-family: var(--ams-header-font-family);
   padding-block: var(--ams-header-padding-block);
   padding-inline: var(--ams-header-padding-inline);
 }
 
-.ams-header__branding {
+.ams-header__logo-link {
```
VincentSmedinga, line 21: The link has turned into a grid item, which means it now takes up the full width and its clickable area covers the entire header. The other element with this class name is not a grid item, which introduces differences between the two.

packages/css/src/components/header/header.scss
```diff
@@ -13,29 +13,40 @@
...
   align-items: center;
   align-self: start; // To align the branding section to the top of the header when it wraps
```
VincentSmedinga, line 23: We no longer have a branding section. We could either restore it around the link to fix its layout, or change the comments mentioning it.

packages/css/src/components/header/header.scss
```diff
@@ -13,29 +13,40 @@
...
-  column-gap: var(--ams-header-branding-column-gap);
-  display: flex;
+  column-gap: var(--ams-header-logo-link-column-gap);
+  display: inline-flex;
```
VincentSmedinga, line 25: This doesn’t make a difference for the visible link, I think.

packages/css/src/components/header/header.scss
```diff
@@ -13,29 +13,40 @@
...
 
-.ams-header__branding--hidden {
+.ams-header__logo-link--hidden {
   opacity: 0%;
   user-select: none; // The hidden branding section should not be selectable
 }
 
-.ams-header__logo-link {
-  outline-offset: var(--ams-header-logo-link-outline-offset);
+.ams-header__logo-container {
```
VincentSmedinga, line 36: Don’t these styles work directly on the svg?

packages/css/src/components/header/header.scss
alimpens: Sharp! Do we want to add `font-family: inherit` to `.ams-header__mega-menu-button`? Or do we want to allow users to specify a different font for the button only?

packages/css/src/components/header/header.scss
alimpens: No, the SVG has a `viewBox`, which makes it responsive. Setting the `inline-size` shrinks the logo, it doesn't cut it off. And we need that responsive behaviour because the logo grows in size in wider screens.

packages/react/src/Header/Header.tsx
```diff
@@ -64,31 +76,19 @@ const HeaderRoot = forwardRef(
...
-        </div>
+        <a className="ams-header__logo-link" href={logoLink}>
+          <span className="ams-visually-hidden">{logoLinkTitle}</span>
+          <LogoLinkContent brandName={brandName} logoBrand={logoBrand} />
+        </a>
         {(children || menuItems) && (
           <nav aria-labelledby="primary-navigation" className="ams-header__navigation">
             <h2 className="ams-visually-hidden" id="primary-navigation">
               {navigationLabel}
             </h2>
```
alimpens, line 87: Do we want to keep this `h2` here? That would mean the `h2` comes before the `h1` on the page. That's allowed in HTML, but maybe it's a little confusing?

Gov.uk does do it, but I can't find any other examples of this. Most examples I found use `aria-label`

packages/css/src/components/header/header.scss
VincentSmedinga: A different font for the button would be crazy, let’s go with inheritance :)

packages/react/src/Header/Header.tsx
VincentSmedinga: As you mentioned today, assistive tech users can navigate by headings or landmarks. Would they want to come across the navigation if they navigate by headings? I would tend to think not, given that they have the choice. So `aria-label`?

packages/react/src/Header/Header.tsx
```diff
@@ -64,29 +76,19 @@ const HeaderRoot = forwardRef(
...
 
             {/* The branding section is recreated here, to make sure the page menu breaks at the right spot */}
```
VincentSmedinga, line 89:
```suggestion
            {/* The logo link section is recreated here, to make sure the page menu breaks at the right spot */}
```

packages/react/src/Header/Header.tsx
```diff
@@ -64,29 +76,19 @@ const HeaderRoot = forwardRef(
...
               {navigationLabel}
             </h2>
 
-            {/* The branding section is recreated here, to make sure the page menu breaks at the right spot */}
-            <div aria-hidden className="ams-header__branding ams-header__branding--hidden">
-              <div className="ams-header__logo-link">
-                <Logo brand={logoBrand} />
-              </div>
-              {brandName && <span className="ams-heading ams-heading--5 ams-header__brand-name">{brandName}</span>}
+            {/* The logo link section is recreated here, to make sure the page menu breaks at the right spot */}
```
VincentSmedinga, line 89:
```suggestion
            {/* The logo link section is recreated here, to make sure the header menu wraps at the right spot */}
```

VincentSmedinga (approved):

PR #1898, opened 2025-03-01
fix(Card): Prevent other links than the Card Link from stretching

VincentSmedinga:
# Describe the pull request

## What

It lets only the Card Link stretch to cover the entire Card; not arbitrary other links.

## Why

Other links would make the Card difficult to use.

## How

- Replace two instances of the type selector with a class selector
- Group link styles for clarity and to prevent descending specificity
- Clarify comments and prevent them from being output in the build

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages/css/src/components/card/card.scss
```diff
@@ -50,8 +30,23 @@
...
+    inset-inline: 0;
+    position: absolute;
+  }
+
   &:hover {
     color: var(--ams-card-link-hover-color);
     text-decoration-line: var(--ams-card-link-hover-text-decoration-line);
   }
+
+  // Remove the outline from the link; it’s already on the Card.
```
alimpens, line 48: I think we should mention the reason for the `:has` selector here. Line 14 and this line work in tandem. 

```suggestion
  // Remove the outline only if :has is supported,
  // which means the outline on the whole card is set
```

VincentSmedinga: Alright, but I still rephrased it.

PR #1904, opened 2025-03-04
fix(Heading)!: Remove ‘level’ from class name

VincentSmedinga:
# Describe the pull request

## What

Changes the class names from `ams-heading--level-x` to `ams-heading--x`.

## Why

We never include the name of the variation in our classes and tokens, only the values. We also just removed ‘level’ from the names of the typography tokens.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages/react/src/Heading/Heading.tsx
+++ b/packages/react/src/Heading/Heading.tsx
@@ -21,11 +21,11 @@ export type HeadingProps = {
 export const Heading = forwardRef(
   ({ children, className, color, level, size, ...restProps }: HeadingProps, ref: ForwardedRef<HTMLHeadingElement>) => {
     const Tag = getHeadingTag(level)
-    const sizeOrLevel = size ?? `level-${level}`
+    const levelValue = size ? size.slice(-1) : level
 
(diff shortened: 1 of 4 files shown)


packages/react/src/Heading/Heading.tsx
```diff
@@ -21,11 +21,11 @@ export type HeadingProps = {
 export const Heading = forwardRef(
   ({ children, className, color, level, size, ...restProps }: HeadingProps, ref: ForwardedRef<HTMLHeadingElement>) => {
     const Tag = getHeadingTag(level)
-    const sizeOrLevel = size ?? `level-${level}`
+    const levelValue = size ? size.slice(-1) : level
```
alimpens, line 24: This isn't quite the level value I think, it's the visual size. So `visualSize`?

VincentSmedinga: Sure! Or `visualLevel`? Size is already visual, and the `size` prop reuses the (hierarchical) `level`s of Heading.

PR #1907, opened 2025-03-05
chore(Footer): Fix heading hierarchy

alimpens:
# Describe the pull request

## What

This removes the Heading level 1 from our Footer examples. It also changes a `section` to a `div`, [a section should not be used for styling purposes only](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#usage_notes).

## Why

According to [these docs](https://www.a11yproject.com/posts/how-to-accessible-heading-structure/), there should only be 1 `h1` on the page. This means our docs were incorrect. 

## How

Removed the heading, changed the tag. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/storybook/src/components/Footer/Footer.stories.tsx
+++ b/storybook/src/components/Footer/Footer.stories.tsx
@@ -71,7 +71,7 @@ export const Default: Story = {
             </LinkList>
           </Grid.Cell>
           <Grid.Cell span={4} start={{ narrow: 1, medium: 1, wide: 9 }}>
-            <section className="ams-mb--md">
+            <div className="ams-mb--md">
               <Heading className="ams-mb--xs" color="inverse" level={2} size="level-4">
                 Nieuwsbrief
               </Heading>
@@ -80,8 +80,8 @@ export const Default: Story = {
                   Inschrijven
                 </LinkList.Link>
               </LinkList>
-            </section>
-            <section>
+            </div>
+            <div>
               <Heading className="ams-mb--xs" color="inverse" level={2} size="level-4">
                 Volg ons
               </Heading>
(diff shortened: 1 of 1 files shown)


storybook/src/components/Footer/Footer.stories.tsx
```diff
@@ -74,7 +71,7 @@ export const Default: Story = {
             </LinkList>
           </Grid.Cell>
           <Grid.Cell span={4} start={{ narrow: 1, medium: 1, wide: 9 }}>
-            <section className="ams-mb--md">
+            <div className="ams-mb--md">
```
VincentSmedinga, line 74: > It also changes a section to a div, [a section should not be used for styling purposes only](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#usage_notes).

But we don’t apply styling based on the tag name. We have content with a heading here, why not mark that up as a section?

VincentSmedinga (commented):
Let’s consider marking updates to our recommended markup as features instead of chores. If people have implemented this markup before, they should be aware of our changes to it and apply them in their project.

alimpens:
> Let’s consider marking updates to our recommended markup as features instead of chores. If people have implemented this markup before, they should be aware of our changes to it and apply them in their project.

This wouldn't have any effect. This change touches the Storybook package, which we don't release. So it will not be added to any changelog.

storybook/src/components/Footer/Footer.stories.tsx
alimpens: The elements function seems to be to vertically separate 2 sections in a `Grid.Cell`. That's a purely visual thing.

[Using too many landmarks can create 'noise' for screen readers](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/region_role#accessibility_concerns).

VincentSmedinga (approved):

PR #1910, opened 2025-03-06
feat!: Consolidate Grid Space and Component Space into one set of spacing tokens

VincentSmedinga:
# Describe the pull request

## What

- Removes the Grid Space tokens, replacing:
  - `var(--ams-space-grid-sm)` with `var(--ams-space-lg)`
  - `var(--ams-space-grid-md)` with `var(--ams-space-xl)` while changing its value from 48 to 56.
  - `var(--ams-space-grid-lg)` with `var(--ams-space-2xl)`; this is a new token
  - The other options weren’t in use
- Removes tokens for gap and margin; their utility classes can use the space tokens directly.
- Makes space tokens for Compact Mode responsive again to save more space on mobile phones. The value for `var(--ams-space-xl)` has also changed here, from 32 to 36.
- Changes the padding tokens of Grid to grow from one size in narrow windows to another on wide ones. This is what we needed to ensure the Grid can take its space on large screens, while not using so much screen space on small screens. It also removes the incorrect coupling between this kind of white space and the grid gaps. Changing one for the better always made the other worse, but these custom pairs seem the correct modelling for the difference in responsive behaviour that has always been there in the background. An advantage of these pairs is also that they reuse existing spacing lengths instead of introducing new ones.
- This also addresses the problem where Grid Space seemed too small on mobile. Vertical space around grids is now 27 instead of 16, and horizontal is 27 as well, instead of 24.
- I think having 27px whitespace between the grid’s content and the edges of the screen (at 320 pixels wide, so it’s around 30 on actual phones) is the best compromise between keeping enough horizontal room for the content and communicating our spacious branding.
- Updates the documentation accordingly.

To help understand how the values changed, here are screenshots from the new (left) and previous (right) token tables, for Spacious (above) and Compact mode (below):

1  Changes in Spacious Mode
2  Changes in Compact Mode

Update: the custom pairs have since been deleted from brand tokens; their values are now in Grid.

## Why

This is meant to make our spacing system easier to use. Many people find it confusing to have both ‘grid spacing’ and ‘regular spacing’. It also lets us set the gap between components without needing separate grid cells.

Usually, choosing a size between `xs` and `xl` will be sufficient for most people.

If they think they need one of the custom pairs, which are a bit more complex, knowing that they range between existing spacing lengths will help them understand.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- UX proposes to make `space-2xl` the default vertical padding on Grid, just like its horizontal padding, replacing `space-xl`. We can apply this in a follow-up PR, where we also remove the other options, and the `padding*` props from Grid, maybe except for a ‘none’ option.


proprietary/tokens/src/components/ams/grid.tokens.json
```diff
@@ -2,17 +2,17 @@
...
       "column-count": { "value": "4" },
-      "column-gap": { "value": "{ams.space.grid.md}" },
+      "column-gap": { "value": "{ams.space.xl}" },
       "padding-block": {
-        "sm": { "value": "{ams.space.grid.sm}" },
-        "md": { "value": "{ams.space.grid.md}" },
-        "lg": { "value": "{ams.space.grid.lg}" }
+        "sm": { "value": "{ams.space.md-lg}" },
+        "md": { "value": "{ams.space.lg-xl}" },
+        "lg": { "value": "{ams.space.xl-2xl}" }
```
VincentSmedinga, line 9: I’m not sure we still need these three options – 50%, 100% or 150% of a grid gap as padding above or below it. I think we can just keep the medium value, like we do for horizontal (inline) padding.

If people want more than a grid gap of whitespace below something, they can add a bottom margin through or `.ams-mb-*` utility classes.

proprietary/tokens/src/components/ams/grid.tokens.json
```diff
@@ -2,17 +2,29 @@
...
+          "value": "clamp(1.6875rem, 1.2344rem + 2.2656vw, 3.5rem)"
+        },
+        "lg": {
+          "comment": "Grows from ams.space.xl to ams.space.2xl.",
+          "value": "clamp(2.25rem, 1.5rem + 3.75vw, 5.25rem)"
+        }
+      },
+      "padding-inline": {
+        "comment": "Grows from ams.space.lg to ams.space.2xl.",
+        "value": "clamp(1.6875rem, 0.7969rem + 4.4531vw, 5.25rem)"
```
alimpens, line 22: I still don't really get why we choose to diverge from our spacing system here. If you look at a screen on 832px, you'll see spaces of 20.4px (md), 30.6px (lg) and 44px (xl) everywhere. Except for our grid, which has 38.6px block padding and 49.8px inline. Why do we choose to add this inconsistency? It makes the spacing system less of a complete system, imo.

proprietary/tokens/src/components/ams/header.tokens.json
```diff
@@ -2,11 +2,15 @@
   "ams": {
     "header": {
       "font-family": { "value": "{ams.typography.font-family}" },
-      "padding-block": { "value": "{ams.space.grid.sm}" },
+      "padding-block": { "value": "{ams.space.lg}" },
```
alimpens, line 5: This is a design change, that shouldn't be part of a tokens PR imo.

proprietary/tokens/src/components/ams/header.tokens.json
```diff
@@ -2,11 +2,15 @@
...
       "padding-inline": {
         "value": "{ams.grid.padding-inline}",
         "comment": "Must be the Grid inline padding, to make sure Header and Grid line up"
       },
+      "branding": {
+        "column-gap": { "value": "{ams.space.md}" },
+        "row-gap": { "value": "{ams.space.md}" }
+      },
```
alimpens, line 13: This does not exist anymore

packages/css/src/components/gap/gap.scss
```diff
@@ -14,7 +14,7 @@
   @if $size != "no" {
     .ams-gap--#{$size} {
       /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
-      grid-gap: var(--ams-gap-#{$size}) !important;
+      grid-gap: var(--ams-space-#{$size}) !important;
```
dlnr, line 17: If `grid-gap` is not a css rule why define it here? If you're going to use it as a util for gap then use `gap:`, it will also work with flexbox.

proprietary/tokens/src/brand/ams/space.compact.tokens.json
```diff
@@ -1,21 +1,12 @@
...
-        },
-        "lg": { "value": "clamp(1.5rem, calc(2.34375vw - 0.09375rem), 3.75rem)" },
-        "xl": { "value": "clamp(2rem, calc(3.125vw - 0.125rem), 5rem)" }
-      }
+      "xs": { "value": "clamp(0.1875rem, 0.1719rem + 0.0781vw, 0.25rem)" },
+      "sm": { "value": "clamp(0.375rem, 0.3438rem + 0.1563vw, 0.5rem)" },
+      "md": { "value": "clamp(0.75rem, 0.6875rem + 0.3125vw, 1rem)" },
+      "lg": { "value": "clamp(1rem, 0.875rem + 0.625vw, 1.5rem)" },
+      "xl": { "value": "clamp(1.25rem, 1rem + 1.25vw, 2.25rem)" },
+      "2xl": { "value": "clamp(1.5rem, 1.125rem + 1.875vw, 3rem)" }
```
dlnr, line 9: My only issue here is that it should be `xxl`, the actual value needs to prove itself in apps and Figma. 
I know you said `2xl` because of scalability but i would stick to t-shirt sizes if you opt to use t-shirt sizes.

packages/css/src/components/gap/gap.scss
dlnr: Also, `!important` should no be needed if you use this as a class right?

storybook/src/components/Footer/Footer.stories.tsx
```diff
@@ -112,12 +112,12 @@ export const OnderzoekEnStatistiek: Story = {
   args: {
     children: [
       <Footer.Spotlight key="footer-spotlight">
-        <Grid gapVertical="large" paddingVertical="medium">
```
alimpens, line 115: Removing `paddingVertical` breaks the design

Bondt007 (dismissed):
as discussed this morning

proprietary/tokens/src/components/ams/header.tokens.json
```diff
@@ -53,7 +57,7 @@
       },
       "navigation": {
         "column-gap": { "value": "{ams.space.lg}" },
-        "row-gap": { "value": "{ams.space.grid.md}" }
+        "row-gap": { "value": "{ams.space.sm}" }
```
alimpens, line 60: Is this change on purpose?

alimpens (approved):

PR #1913, opened 2025-03-07
feat!: Update aspect ratio tokens and CSS utility classes

alimpens:
# Describe the pull request

## What

Use aspect ratios in the names of our aspect ratio tokens and CSS utility classes.

## Why

When you use these tokens or utility classes on an image, you need the explicit aspect ratios in order to pick an image with that same ratio. Using t-shirt sizing obscures the ratio's, which is a bit annoying. 

## How

Find and replace, etc.

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages/css/src/components/aspect-ratio/aspect-ratio.scss
+++ b/packages/css/src/components/aspect-ratio/aspect-ratio.scss
@@ -4,7 +4,7 @@
  */
 
 .ams-aspect-ratio {
-  @each $name in ("x-tall", "tall", "square", "wide", "x-wide", "2x-wide") {
+  @each $name in ("9-16", "3-4", "1-1", "4-3", "16-9", "16-5") {
     &--#{$name} {
       /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
       aspect-ratio: var(--ams-aspect-ratio-#{$name}) !important;
--- a/packages/react/src/common/types.ts
+++ b/packages/react/src/common/types.ts
@@ -1,4 +1,4 @@
-export const aspectRatioOptions = ['x-tall', 'tall', 'square', 'wide', 'x-wide', '2x-wide'] as const
+export const aspectRatioOptions = ['9-16', '3-4', '1-1', '4-3', '16-9', '16-5'] as const
 export type AspectRatioProps = {
   /** The aspect ratio to display media content in. */
   aspectRatio?: (typeof aspectRatioOptions)[number]
--- a/storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
+++ b/storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
@@ -3,6 +3,7 @@
  * Copyright Gemeente Amsterdam
  */
 
+import { aspectRatioOptions } from '@amsterdam/design-system-react/src/common/types'
 import { Meta, StoryObj } from '@storybook/react'
 import { AspectRatio } from './AspectRatio'
 
@@ -10,12 +11,12 @@ const meta = {
   title: 'Utilities/CSS/Aspect Ratio',
   component: AspectRatio,
   args: {
-    aspectRatio: 'x-wide',
+    aspectRatio: '16-9',
   },
   argTypes: {
     aspectRatio: {
       control: 'radio',
-      options: ['2x-wide', 'x-wide', 'wide', 'square', 'tall', 'x-tall'],
+      options: aspectRatioOptions,
     },
   },
   render: ({ aspectRatio }) => (
--- a/packages/css/src/components/aspect-ratio/README.md
+++ b/packages/css/src/components/aspect-ratio/README.md
@@ -9,14 +9,14 @@ Constrains media content to a supported aspect ratio.
 Each available aspect ratio has an associated class name.
 The class can be applied to any component or element.
 
-| Class name                  | Example                                                                     |
-| :-------------------------- | :-------------------------------------------------------------------------- |
-| `ams-aspect-ratio--x-tall`  | <div className="ams-docs-token-example--space ams-aspect-ratio--x-tall" />  |
-| `ams-aspect-ratio--tall`    | <div className="ams-docs-token-example--space ams-aspect-ratio--tall" />    |
-| `ams-aspect-ratio--square`  | <div className="ams-docs-token-example--space ams-aspect-ratio--square" />  |
-| `ams-aspect-ratio--wide`    | <div className="ams-docs-token-example--space ams-aspect-ratio--wide" />    |
-| `ams-aspect-ratio--x-wide`  | <div className="ams-docs-token-example--space ams-aspect-ratio--x-wide" />  |
-| `ams-aspect-ratio--2x-wide` | <div className="ams-docs-token-example--space ams-aspect-ratio--2x-wide" /> |
+| Class name               | Example                                                                  |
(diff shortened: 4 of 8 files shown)


packages/css/src/components/aspect-ratio/aspect-ratio.scss
```diff
@@ -4,7 +4,7 @@
  */
 
 .ams-aspect-ratio {
-  @each $name in ("x-tall", "tall", "square", "wide", "x-wide", "2x-wide") {
+  @each $name in ("9-16", "3-4", "1-1", "4-3", "16-9", "16-5") {
     &--#{$name} {
       /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
       aspect-ratio: var(--ams-aspect-ratio-#{$name}) !important;
```
VincentSmedinga, line 10: Technically, they’re no longer names.

```suggestion
  @each $value in ("9-16", "3-4", "1-1", "4-3", "16-9", "16-5") {
    &--#{$value} {
      /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
      aspect-ratio: var(--ams-aspect-ratio-#{$value}) !important;
```

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
```diff
@@ -3,19 +3,20 @@
...
   component: AspectRatio,
   args: {
-    aspectRatio: 'x-wide',
+    aspectRatio: '16-9',
   },
   argTypes: {
     aspectRatio: {
       control: 'radio',
-      options: ['2x-wide', 'x-wide', 'wide', 'square', 'tall', 'x-tall'],
+      options: aspectRatioOptions,
```
VincentSmedinga, line 19: The ‘human friendly’ labels still use slashes, like in the tokens table. Can you add a list of ‘9 / 16’ etc. here and to the prop on `Image`?

packages/css/src/components/aspect-ratio/aspect-ratio.scss
dlnr: Would it not become "ratio" then?

packages/css/src/components/aspect-ratio/aspect-ratio.scss
```diff
@@ -4,7 +4,7 @@
  */
 
 .ams-aspect-ratio {
-  @each $name in ("x-tall", "tall", "square", "wide", "x-wide", "2x-wide") {
+  @each $name in ("9-16", "3-4", "1-1", "4-3", "16-9", "16-5") {
     &--#{$name} {
```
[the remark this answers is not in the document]
alimpens: I think it's easier for our users to find and replace all utility classes in one go, so I'd bundle that change.

packages/css/src/components/aspect-ratio/aspect-ratio.scss
alimpens: I think both `value` and `ratio` are techically correct. `aspect-ratio-#{$ratio}` is a bit harder to read though.

packages/react/src/common/types.ts
```diff
@@ -1,4 +1,7 @@
-export const aspectRatioOptions = ['x-tall', 'tall', 'square', 'wide', 'x-wide', '2x-wide'] as const
+// This array is not used to create the aspect ratio CSS classes.
```
VincentSmedinga, line 1: Good idea. Let’s add a matching comment to the Sass list as well.

storybook/src/components/ImageSlider/ImageSlider.stories.tsx
```diff
@@ -14,27 +14,27 @@ const meta = {
     images: [
       {
         alt: 'Bridge',
-        aspectRatio: 'x-wide',
+        aspectRatio: '16:9',
```
VincentSmedinga, line 17: This is the default aspect ratio for the Image component – all of these can be removed.

packages/react/src/ImageSlider/ImageSlider.test.tsx
```diff
@@ -23,17 +23,17 @@ describe('Image Slider', () => {
   const images: ImageSliderImageProps[] = [
     {
       alt: 'Bridge',
-      aspectRatio: 'x-wide',
+      aspectRatio: '16:9',
```
VincentSmedinga, line 26: This is the default aspect ratio for the Image component – all of these can be removed.

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
```diff
@@ -3,24 +3,25 @@
...
     },
   },
   render: ({ aspectRatio }) => (
     <div className="ams-docs-column ams-docs-aspect-ratio">
-      <div className={`ams-docs-item ams-aspect-ratio--${aspectRatio}`}></div>
+      <div className={`ams-docs-item ams-aspect-ratio--${aspectRatio?.replace(':', '-')}`}></div>
```
VincentSmedinga, line 24: Even though this is a mock component, this could end up in a class of `ams-aspect-ratio--undefined`.

The same goes for the `Image` used in Image Slider. Can you replace that class with the prop?

storybook/src/utils/AspectRatio/AspectRatio.tsx
```diff
@@ -4,12 +4,7 @@
...
+import { HTMLAttributes } from 'react'
 
-/** Renders examples in Storybook. Not for reuse. */
-export const AspectRatio = ({
-  aspectRatio,
-  children,
-}: AspectRatioProps & PropsWithChildren<HTMLAttributes<HTMLSpanElement>>) => (
-  <span className={`ams-aspect-ratio--${aspectRatio}`}>{children}</span>
-)
+/** Mock component to render examples in Storybook. Not for reuse. */
```
VincentSmedinga, line 9: If you update this comment, please do so for all utility classes (margin, gap, visually hidden).

packages/react/src/ImageSlider/ImageSlider.test.tsx
alimpens: Hm, not quite. I think we have a bug here. Aspect ratio is used twice in the ImageSlider, for the Image and for the thumbnail button. 

We gave Image a default aspect ratio, but we didn't do the same for the thumbnail button. Both Image and the thumbnail button use the utility class, so they're broken by this PR anyway. I think we should pass a prop to Image in ImageSlider, and turn the thumbnail button into a component which has the same props as Image and a default aspect ratio.

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
alimpens: This mock component is only used in Storybook, connected to a radio control with a default value. That makes it impossible for it to be undefined, I think? 

It is an issue for Image in ImageSlider, but I think we should solve that in a different way.

packages/react/src/ImageSlider/ImageSlider.test.tsx
alimpens: I've done this in my last commits

storybook/src/utils/AspectRatio/AspectRatio.stories.tsx
alimpens: I've used `generateAspectRatioClass` here to handle this

VincentSmedinga (approved):

PR #1914, opened 2025-03-07
feat!: Update hyphenation tokens

VincentSmedinga:
# Describe the pull request

## What

1. Moves the `hyphenate-limit-chars` token into the typography group.
2. Removes the tokens for `hyphens` and `overflow-wrap`
3. Updates docs

## Why

1. Because it fits there.
2. Because these values aren’t intended to be customized.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


dlnr (approved):
Makes sense. Also, `text-wrap: pretty` is something to consider, it will prevent multiple lines that end in hyphens (which is ugly): https://developer.chrome.com/blog/css-text-wrap-pretty

VincentSmedinga:
> Also, `text-wrap: pretty` is something to consider, it will prevent multiple lines that end in hyphens (which is ugly)

We have already reduced the chance of successive hyphens by only breaking words of at least 12 characters long. Pretty text wrapping still has a performance penalty, so I think `hyphenate-limit-chars` (which is just landing in Firefox, while Chrome and Edge already support it) should suffice:

> Note that pretty has a negative effect on performance, so it should be only used for longer blocks of text when the layout is more important than speed. – [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap#description)

packages/css/src/common/hyphenation.scss
```diff
@@ -4,7 +4,7 @@
  */
 
 @mixin hyphenation {
-  hyphenate-limit-chars: var(--ams-hyphenation-hyphenate-limit-chars);
-  hyphens: var(--ams-hyphenation-hyphens);
-  overflow-wrap: var(--ams-hyphenation-overflow-wrap);
+  hyphenate-limit-chars: var(--ams-typography-hyphenate-limit-chars);
+  hyphens: auto;
+  overflow-wrap: break-word;
```
alimpens, line 9:
```suggestion
  // Break overflowing words for browsers that don’t support CSS hyphens.
  overflow-wrap: break-word;
```

storybook/src/brand/design-tokens/typography.docs.mdx
```diff
@@ -90,6 +90,17 @@ If it is not available, we use Arial or the generic sans-serif instead.
 | :---------------------------------- | :------------------------------------ |
 | `var(--ams-typography-font-family)` | `'Amsterdam Sans', Arial, sans-serif` |
 
+## Hyphenation
+
+Words that are too long for the end of a line will be split with a hyphen and continue on the next line.
```
alimpens, line 95:
```suggestion
Words that do not fit on a single line will be split with a hyphen and continue on the next line.
```

alimpens (approved):

PR #1923, opened 2025-03-14
feat(Tabs)!: Convert `tab` prop to string

alimpens:
# Describe the pull request

## What

It changes the `tab` prop to type string. Along the way I've also refactored the code a bit and added comments. 

## Why

It was requested by our users,  and we decided a string makes more sense here than a number.

## How

Changed type, fixed tests, added comments. 

## Checklist

- [ ] Add or update stories


packages/react/src/Tabs/Tabs.tsx
```diff
@@ -13,37 +13,53 @@ import { TabsPanel } from './TabsPanel'
...
-    const [activeTabId, setActiveTabId] = useState(0)
+    const [activeTabId, setActiveTabId] = useState<string>()
 
-    const allTabs = useMemo(() => {
+    const allTabIds = useMemo(() => {
       if (!Array.isArray(children)) return []
-      return children[0].props.children
+
+      // The first child of Tabs should be TabsList
+      const tabsList = children[0]
```
VincentSmedinga, line 32: Shouldn’t we then query for a class of `.ams-tabs__list`?

VincentSmedinga (commented):
Nice refactoring and comments, Mr. Boy Scout! :+1:

packages/react/src/Tabs/Tabs.tsx
alimpens: Using something like `querySelector` would move you out of the React context, no? We're currently traversing React props, not DOM nodes. 

Tbh, I think we should rethink this approach at some point, it's a little brittle.

VincentSmedinga: Of course 🤦🏼‍♂️ and yes, agreed.

PR #1924, opened 2025-03-14
fix(Header): Only use matchMedia when there is a mega menu

alimpens:
# Describe the pull request

## What

Checks if Header has a mega menu and only then calls `useIsAfterBreakpoint`.

## Why

`useIsAfterBreakpoint` uses `matchMedia`, which you have to mock in unit tests. If we don't do this check, users that do not have a mega menu still have to mock matchMedia in their tests, although it isn't necessary. 

## How

Add a condition.

## Checklist

- [x] Add or update stories


VincentSmedinga (approved):
Thoughtful!

PR #1933, opened 2025-03-19
fix(Icon): Restore square appearance

VincentSmedinga:
# Describe the pull request

## What

Makes an icon square again if the `square` prop has been set.

## Why

It no longer did so, probably since #1905. This broke the appearance of the Search Field Button, the buttons in an Image Slider, and the close button of Alert and Dialog.

## How

I tried several approaches. The `aspect-ratio: 1 / 1` doesn’t seem to do anything anymore; maybe the icon container doesn’t get a calculated height or width due to the inline flex container, center alignment, an anonymous line box, a zero-width space, or any combination of these.

A width of `1lh` on `.ams-icon` seemed very nice, but browser support is only at 92%.

The solution is an `inline-size` of font size times line height. Not that elegant, but straightforward. I used a local custom property to prevent repeating the `inline-size: calc(…)` three times. I wanted to call it `--line-height`, but it must start with `--ams-`.

I tested a square Icon on its own, a Button with an Icon in it, and various Buttons in an Action Group. The story for the latter:

```
export const Combinations: Story = {
  args: {
    children: [
      <Button icon={ChevronLeftIcon} iconBefore key="back" variant="secondary">
        Terug
      </Button>,
      <Button icon={StopIcon} iconOnly key="cancel" variant="tertiary">
        Annuleren
      </Button>,
      <Button key="save">Opslaan</Button>,
      <Button icon={ChevronRightIcon} key="forward" variant="secondary">
        Verder
      </Button>,
      <Link download href="#" key="download" variant="standalone">
        Downloaden
      </Link>,
    ],
  },
}
```

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages/css/src/components/icon/icon.scss
```diff
@@ -4,10 +4,12 @@
  */
```
alimpens, line 4: Maybe out of scope, but the Icon Button currently isn't square. It's 40x44px on wide screens.

VincentSmedinga: > Maybe out of scope, but the Icon Button currently isn't square. It's 40x44px on wide screens.

I’ve made an Icon Button `display: inline-flex` to remove the phantom white space that `inline-block` introduces. Our Button already did the same.

> And in Alert and Dialog it's wider than it's tall, for some reason?

Not sure why that is, I’ll make a note for it.

alimpens (approved):

PR #1972, opened 2025-04-02
feat(Card)!: Replace gap with block-end margins in Card

RubenSibon:
# Describe the pull request

## What

In Cards, replace blanket gaps with margins at the end of the block (i.e. `margin-bottom`) it's common and direct child elements such as the Card Heading (Group) or a Card Image.

## Why

We want to leave it up to the developer to decide on what vertical spacing to use in the body of the Card, but give some standard spacing on some of the common Card children.

## How

Remove `gap` from `.ams-card` and add `margin-block-end` to the aforementioned children via `.has(+ *)` selectors.

Also update stories to make multiple children in the body look as much as possible as before.

## Checklist

- [x] Add or update stories

## Additional notes

- Should we add documentation on how to deal with vertical spacing in the Card? Such as possible overrides or using a layout component like Column?

### DIFF
--- a/packages/css/src/components/card/card.scss
+++ b/packages/css/src/components/card/card.scss
@@ -5,7 +5,6 @@
 
 .ams-card {
   display: grid;
-  gap: var(--ams-card-gap);
   outline-offset: var(--ams-card-outline-offset);
   position: relative; // Allows stretching the card link below.
   touch-action: manipulation;
@@ -16,12 +15,24 @@
     outline-style: auto;
     outline-width: 0.0625rem;
   }
+
+  & > .ams-heading:has(+ *) {
+    margin-block-end: var(--ams-space-s);
+  }
+
+  & > .ams-image:has(+ *) {
+    margin-block-end: var(--ams-space-s);
+  }
 }
 
 .ams-card__heading-group {
   display: flex;
   flex-direction: column-reverse;
   gap: var(--ams-card-heading-group-gap);
+
+  &:has(+ *) {
+    margin-block-end: var(--ams-space-s);
+  }
 }
 
 .ams-card__link {
--- a/storybook/src/components/Card/Card.stories.tsx
+++ b/storybook/src/components/Card/Card.stories.tsx
@@ -3,7 +3,7 @@
  * Copyright Gemeente Amsterdam
  */
 
-import { Heading, Image, Paragraph } from '@amsterdam/design-system-react'
+import { Column, Heading, Image, Paragraph } from '@amsterdam/design-system-react'
 import { Card } from '@amsterdam/design-system-react/src'
 import { Meta, StoryObj } from '@storybook/react'
 import { exampleTopTask } from '../shared/exampleContent'
@@ -68,12 +68,12 @@ export const WithImage: Story = {
           <Card.Link href="/">Nederlands eerste houten woonwijk komt in Zuidoost</Card.Link>
         </Heading>
       </Card.HeadingGroup>,
-      <Paragraph key={3}>
-        We bouwen een levendige, groene en duurzame woonbuurt tussen de Gooiseweg en het Nelson Mandelapark.
-      </Paragraph>,
-      <Paragraph key={4} size="small">
-        {today}
-      </Paragraph>,
+      <Column gap="small" key={3}>
+        <Paragraph>
+          We bouwen een levendige, groene en duurzame woonbuurt tussen de Gooiseweg en het Nelson Mandelapark.
+        </Paragraph>
+        <Paragraph size="small">{today}</Paragraph>
+      </Column>,
     ],
   },
 }


packages/css/src/components/card/card.scss
```diff
@@ -16,12 +15,24 @@
     outline-style: auto;
     outline-width: 0.0625rem;
   }
+
+  & > .ams-heading:has(+ *) {
+    margin-block-end: var(--ams-space-s);
```
VincentSmedinga, line 20: Please add intermediate component tokens (e.g. `--ams-card-heading-margin-block-end`) so that teams using our components can choose their own spacing values.

packages/css/src/components/card/card.scss
```diff
@@ -16,12 +15,24 @@
     outline-style: auto;
     outline-width: 0.0625rem;
   }
+
+  & > .ams-heading:has(+ *) {
```
VincentSmedinga, line 19: Refering to the class of the child components was our initial approach; later in the refinement we decided to add `Card.Heading` and `Card.Image` components to decrease coupling in the stylesheet.

`Card.Heading` will have the props of `Heading`, render a `Heading`, but add a class of `ams-card__heading` on which we set the bottom margin.

packages/css/src/components/card/card.scss
```diff
@@ -5,7 +5,6 @@
 
 .ams-card {
   display: grid;
-  gap: var(--ams-card-gap);
```
VincentSmedinga, line 8: We can now delete the token as well.

RubenSibon: Verwijderd.

RubenSibon:
Still need to:

- Add more documentation;
- Check if the tests for the new sub-components cover all sensible scenarios.

packages/react/src/Card/CardHeading.tsx
```diff
@@ -0,0 +1,24 @@
...
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { Heading, HeadingProps } from '../Heading'
+
+export type CardHeadingProps = Partial<HeadingProps> & PropsWithChildren<HTMLAttributes<HTMLHeadingElement>>
+
+export const CardHeading = forwardRef(
+  (
+    { children, className, level = 1, size = 'level-4', ...restProps }: CardHeadingProps,
```
alimpens, line 15: Do we want to set a default level here? We don't really know where Card will be in the document structure I think.

alimpens, line 15: The type here can just be `HeadingProps`, I think

packages/react/src/Card/CardHeading.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render } from '@testing-library/react'
+import { createRef } from 'react'
+import { CardHeading } from './CardHeading'
+import '@testing-library/jest-dom'
+
+describe('Card Heading', () => {
+  it('renders', () => {
+    const { container } = render(<CardHeading />)
+
+    const component = container.querySelector(':only-child')
```
alimpens, line 10: We can use `getByRole` here

packages/react/src/Card/CardHeading.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  })
+
+  it('renders an additional class name', () => {
+    const { container } = render(<CardHeading className="extra" />)
+
+    const component = container.querySelector(':only-child')
+
+    expect(component).toHaveClass('ams-card__heading extra')
+  })
+
```
alimpens, line 31: We're missing a test for the `size` prop here. And `level`, if we keep it.

packages/react/src/Card/CardImage.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { Image, ImageProps } from '../Image'
+
+export type CardImageProps = ImageProps & PropsWithChildren<HTMLAttributes<HTMLImageElement>>
+
+export const CardImage = forwardRef(
+  ({ children, className, ...restProps }: CardImageProps, ref: ForwardedRef<HTMLImageElement>) => (
+    <Image {...restProps} className={clsx('ams-card__image', className)} ref={ref}>
+      {children}
```
alimpens, line 16: An `img` should not have children. I'm not sure why TypeScript didn't catch this

packages/react/src/Card/CardImage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import clsx from 'clsx'
+import { forwardRef } from 'react'
+import type { ForwardedRef, HTMLAttributes, PropsWithChildren } from 'react'
+import { Image, ImageProps } from '../Image'
+
+export type CardImageProps = ImageProps & PropsWithChildren<HTMLAttributes<HTMLImageElement>>
+
+export const CardImage = forwardRef(
+  ({ children, className, ...restProps }: CardImageProps, ref: ForwardedRef<HTMLImageElement>) => (
```
alimpens, line 14: The type can just be `ImageProps` I think

storybook/src/components/Card/Card.docs.mdx
```diff
@@ -14,16 +14,17 @@ import README from "../../../../packages/css/src/components/card/README.md?raw";
...
 
 <Canvas of={CardStories.WithTagline} />
 
 ### With image
 
-A card often displays the [image](/docs/components-media-image--docs) of the detail page.
-It must use one of the available [aspect ratios](/docs/brand-design-tokens-aspect-ratio--docs).
+A Card often displays the image of the article it links to.
+The Card Image subcomponent renders an [Image](/docs/components-media-image--docs) component and accepts its properties, e.g. to set an aspect ratio or provide a set of image sources.
+An `alt` text is almost never necessary for the image of a card.
```
alimpens, line 27: Maybe we should be a bit more strict here. If the image does have an `alt` text, it will not be read by screen readers when you navigate to the link or heading. Maybe we should even hardcode `alt=""` in CardImage and omit `alt` from the type? I'd keep this text though, you can also use the Next image component here.

```suggestion
Only use decorative images with `alt=""` here. Screen readers will not read the `alt` text of non-decorative images when navigating to a Card. 
```

storybook/src/components/Card/Card.stories.tsx
```diff
@@ -37,9 +37,9 @@ type Story = StoryObj<typeof meta>
 export const Default: Story = {
   args: {
     children: [
-      <Heading key={1} level={1} size="level-4">
+      <Card.Heading key={1} level={1} size="level-4">
```
alimpens, line 40: We don't need to set the `size` in these stories if we set a default in CardHeading

packages-proprietary/tokens/src/components/ams/card.tokens.json
```diff
@@ -1,9 +1,15 @@
 {
   "ams": {
     "card": {
-      "gap": { "value": "{ams.space.s}" },
+      "heading": {
+        "margin-block-end": { "value": "{ams.space.s}" }
```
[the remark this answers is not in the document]
VincentSmedinga: I expect them to become different soon. @Bondt007 and I ran out of time yesterday before reaching this subject to decide on. Do you want me to propose my changes in this PR?

packages/react/src/Card/CardImage.tsx
VincentSmedinga: Because it’s a valid property – `HTMLImageElement` extends `HTMLElement`. I’ve added #1996 to address this.

packages/react/src/Card/CardHeading.test.tsx
VincentSmedinga: Card Heading outsources its rendering to Heading, and Heading tests its size prop. That seems enough, right?

packages-proprietary/tokens/src/components/ams/card.tokens.json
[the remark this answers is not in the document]
alimpens: It's not quite YAGNI, but if we expect these values to change it's fine to already set this up. No need to propose the changes.

packages/react/src/Card/CardHeading.test.tsx
alimpens: Not really. `CardHeading` has a `size` prop, which has it's own logic (albeit very simple logic 😁). That logic isn't tested now. If we mess up passing the value of `size` to the Heading in the future, our tests would not tell us.

VincentSmedinga: Yes, you’re right. Added it.

alimpens (approved):

PR #1975, opened 2025-04-03
feat(Dialog): Decrease amount of inside and outside spacing

VincentSmedinga:
# Describe the pull request

## What

Decreases all whitespace inside and outside a Dialog to:
- `large` on narrow windows
- `x-large` on medium and  wide windows

Previously, for all window sizes:
- Horizontal padding was `x-large`
- Vertical padding was `large`
- Horizontal and vertical margins were `x-large`

## Why

1. To save space on small screens
2. To make whitespace at the corners square; rectangular space seemed to misalign the title and buttons.

## How

Add tokens and apply them with breakpoints.

## Checklist

- [x] Add or update stories

## Additional notes

- Can we use a margin on `dialog` instead of subtracting space from the maximum width and height? This would simplify both tokens and styles. It does break our rule that components don’t manage their outer whitespace – but that’s to prevent collapsing or duplication margins, both of which won’t happen with a Dialog.
- We should updated these spacing values in Figma, too.

### DIFF
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -26,6 +26,16 @@ so do not apply these styles without an `open` attribute. */
 
   @include reset-dialog;
 
+  @media screen and (min-width: $ams-breakpoint-medium) {
+    inline-size: var(--ams-dialog-medium-inline-size);
+    max-block-size: var(--ams-dialog-medium-max-block-size);
+  }
+
+  @media screen and (min-width: $ams-breakpoint-wide) {
+    inline-size: var(--ams-dialog-wide-inline-size);
+    max-block-size: var(--ams-dialog-wide-max-block-size);
+  }
+
   /*
    * The fallback value is for browsers where ::backdrop does not inherit from its originating element.
    * @see https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -3,6 +3,8 @@
  * Copyright Gemeente Amsterdam
  */
 
+@use "../../common/breakpoint" as *;
+
 @mixin reset-dialog {
   inset-block: 0;
   padding-block: 0;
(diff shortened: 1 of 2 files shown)


dlnr (approved):
Big improvement, especially by dropping the "2XL".

bas-kroese (approved):
Top. Stuk beter inderdaad.
Ik heb de aanpassinge direct ook gedaan in Figma.
@dlnr ik heb ook de code syntax hierbij meegenomen. Alles staat in 'Breakpoints' collection, onder: Components/Dialog/Header, Body, Footer
image

proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -4,24 +4,54 @@
...
+      "max-block-size": { "value": "calc(100dvh - 2 * {ams.space.l})" },
       "max-inline-size": { "value": "48rem" },
+      "medium": {
+        "inline-size": { "value": "calc(100% - 2 * {ams.space.xl})" },
+        "max-block-size": { "value": "calc(100dvh - 2 * {ams.space.xl})" }
+      },
+      "wide": {
+        "inline-size": { "value": "calc(100% - 2 * {ams.space.xl})" },
+        "max-block-size": { "value": "calc(100dvh - 2 * {ams.space.xl})" }
+      },
```
alimpens, line 17: Do we need to set values for `wide`, if they're the same as the values for `medium`?

proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -4,24 +4,54 @@
       "background-color": { "value": "{ams.color.background}" },
```
alimpens, line 4: Can you move this file to `packages-proprietary`?

packages/css/src/components/dialog/dialog.scss
```diff
@@ -3,6 +3,8 @@
  * Copyright Gemeente Amsterdam
```
alimpens, line 3: > Can we use a margin on dialog instead of subtracting space from the maximum width and height? This would simplify both tokens and styles. It does break our rule that components don’t manage their outer whitespace – but that’s to prevent collapsing or duplication margins, both of which won’t happen with a Dialog.

I think it's fine to break the outer whitespace rule for Dialog, but I'm not sure we can achieve a center aligned Dialog with margins on the outside this way. The Dialog is centered by the UA with `margin: auto`. If we override that with a fixed value, the Dialog wouldn't be center aligned anymore.

Bondt007 (commented):
seen, discussed, approved

Bondt007 (dismissed):

packages/css/src/components/dialog/dialog.scss
VincentSmedinga: Ah yes, of course 💡

alimpens (approved):

PR #1981, opened 2025-04-07
feat(Page Header): Add vertical space around collapsible menu

VincentSmedinga:
# Describe the pull request

## What

1. Adds XL vertical space at the top and bottom of the mega menu container.
2. Updates our examples to indicate a Grid child no longer needs to set padding.
3. Separates describing design choices from prescribing guidelines in the docs.
4. Explains how to use a Grid in the collapsible menu.

## Why

1. Otherwise, all our uses would need to configure this whitespace themselves.
2. To update our examples for the change to the Page Header component.
3. To clarify.
4. To help people lay out the collapsible menu and to prevent accessibility problems.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/page-header.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/page-header.tokens.json
@@ -24,6 +24,7 @@
         "text-wrap": { "value": "{ams.typography.heading.text-wrap}" }
       },
       "mega-menu": {
+        "padding-block": { "value": "{ams.space.xl}" },
         "button": {
           "cursor": { "value": "{ams.cursor.interactive}" },
           "label": {
--- a/packages/css/src/components/page-header/README.md
+++ b/packages/css/src/components/page-header/README.md
@@ -8,16 +8,16 @@ Includes the name of the application if it is not the general website.
 ## Design
 
 - The Page Header includes the logo of the City or the organization, the site title (except for the general website), and a menu with links to commonly used pages.
-- An optional 'Menu' button opens a collapsible menu, which has room for more links.
+- An optional ‘Menu’ button opens a collapsible menu, which has room for more links.
 - On narrow windows, links can move from the inline menu to the collapsible one.
 - For websites with a brand name that feature the standard Amsterdam logo, only the emblem is shown on narrow screens.
 
 ## Guidelines
 
 - The Page Header must be used on all websites and applications for the City of Amsterdam.
-- The Page Header is important because it conveys our corporate identity and is the first thing people see.
-  Using it consistently helps users recognize and trust the website.
-- The Page Header is the same on every page of the application, although full-screen pages may hide it, e.g. a video or a map.
+  It’s often the first thing people perceive.
+  As it conveys our corporate identity, using it consistently helps users recognize and trust the website.
+- The Page Header should be the same on every page, although full-screen pages may hide it, e.g. a video or a map.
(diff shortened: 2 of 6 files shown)


packages-proprietary/tokens/src/components/ams/page-header.tokens.json
```diff
@@ -24,6 +24,7 @@
         "text-wrap": { "value": "{ams.typography.heading.text-wrap}" }
       },
       "mega-menu": {
+        "padding-block": { "value": "{ams.space.xl}" },
```
VincentSmedinga, line 27: This results in 92 pixels of bottom padding (at 1600 wide): `xl` from mega menu + `lg` from Header = 56 + 36. @Bondt007 says this would be fine, but we could also set `padding-block-end: calc(var(--ams-space-xl) - var(--ams-space-lg))` to get 84 pixels, which is the intended `--ams-space-2xl`.

VincentSmedinga, line 27: At the top, the padding gets complemented with the row gap of `.ams-page-header__navigation` which can be seen as a bug. That gap is intended to separate the logo from a wrapping header menu, but it also adds space between said menu and the mega menu.

dlnr: This is not the solution I was expecting, if you look at the story there is a Grid component in the Page Header:

```jsx
<Grid gapVertical="small" paddingBottom="large">
  <PageHeader.GridCellNarrowWindowOnly span="all">
    <LinkList>
```

If you change it to 

```jsx
<Grid gapVertical="small" paddingVertical="large">
  <PageHeader.GridCellNarrowWindowOnly span="all">
    <LinkList>
```

It fixes the problem and you still allow internal apps to reduce the padding here without overrides. There is also a variant where the mega menu has a heading ("Alle onderwerpen") which is also possible because you can reduce the padding on the top. Heading line height + gap + padding would be a bit much.

packages/css/src/components/page-header/README.md
```diff
@@ -5,20 +5,27 @@
...
+- The Page Header should be the same on every page, although full-screen pages may hide it, e.g. a video or a map.
 - The inline menu can contain a maximum of 5 items.
   The last two will often be ‘Search’ and ‘Menu’.
-- The 'Menu' button opens a collapsible menu, which has room for more links.
-- On narrow windows, links can move from the inline menu to the collapsible one.
 - Labels should be short to help the inline menu fit on a single line whenever possible.
 - An icon can be added to the end of a link to make its destination easier to guess.
-- For websites with a brand name that feature the standard Amsterdam logo, only the emblem is shown on narrow screens.
+- Wrap the collapsible menu in a [Grid](/docs/components-layout-grid--docs).
+  Don’t set its padding props – Page Header adds the correct space above and below.
```
dlnr, line 26: Page Header also sets the grids inline padding to 0, which is helpful because it allows you to use it as some sort of subgrid (with the browser support). Shouldn't we consider allowing users to set an `paddingHorizontal` prop as well? Because we once stated that components should not  manipulate child components? This would allow you to remove that CSS.

VincentSmedinga:
We discussed this before and decided that letting Page Header add this whitespace was the better option. However, our current approach for navigation in applications does not have whitespace here. This could justify the approach @dlnr advocates. Let’s wait to add it a bit longer until we are certain.

PR #1990, opened 2025-04-10
chore(test): Make all unit tests for an extra class name consistent

VincentSmedinga:
# Describe the pull request

## What

- Renamed the test
- Used a class name of ‘extra’ everywhere (some used ‘large’)
- Merged the component and the extra class into a single line
- Added design system tests for Button and Heading

Resisted the temptation to sort the test consistently.

## Why

I found some inconsistencies and I don’t like those.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages/react/src/Heading/Heading.test.tsx
```diff
@@ -4,6 +4,14 @@ import { Heading } from './Heading'
 import '@testing-library/jest-dom'
 
 describe('Heading', () => {
+  it('renders a design system BEM class name', () => {
```
alimpens, line 7: We usually place this test after the 'it renders'-test. And `getByRole` is preferred to `querySelector`

VincentSmedinga: I’m aware, and was almost on it, but decided to prevent scope creep. It was just that while starting another ticket, I came across some tests using ‘large’ instead of ‘extra’ so I thought let’s just make them consistent. Then I replaced ‘additional’ with ‘extra’ in the test description to match, but that should be all, I think. I can create a ticket for these additional improvements.

alimpens (approved):

PR #1992, opened 2025-04-10
chore(docs): Fix tuple type for icon sizes prop

VincentSmedinga:
# Describe the pull request

## What

We forgot the `as const` piece for the icon sizes array that defines the related prop.

## Why

Restoring it lets TypeScript treat the array as a tuple with literal types instead of just an array of strings. This improves type checking and displays each of the options in the Storybook Controls.

When checking all the instances, I found some of them have an empty line between them, while most don’t. I deleted these.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages/react/src/Avatar/Avatar.tsx
+++ b/packages/react/src/Avatar/Avatar.tsx
@@ -10,7 +10,6 @@ import type { ForwardedRef, HTMLAttributes } from 'react'
 import { Icon } from '../Icon'
 
 export const avatarColors = ['azure', 'green', 'lime', 'magenta', 'orange', 'yellow'] as const
-
 type AvatarColor = (typeof avatarColors)[number]
 
 type AvatarContentProps = {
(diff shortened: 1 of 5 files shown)


packages/react/src/Icon/index.ts
```diff
@@ -1,2 +1,2 @@
 export { Icon } from './Icon'
-export type { IconProps } from './Icon'
+export type { IconProps, IconSize } from './Icon'
```
alimpens, line 2: We don't really need to export `IconSize` I think, we already have `IconProps['size']`

storybook/src/components/Heading/Heading.stories.tsx
```diff
@@ -102,7 +103,7 @@ export const WithIcon: Story = {
 
     return (
       <Row gap="small">
-        <Icon color={args.color} size={iconSize} svg={EmailIcon} />
+        <Icon color={args.color} size={iconSize as IconSize} svg={EmailIcon} />
```
alimpens, line 106: What's the reason for this cast?

alimpens (approved):

PR #1997, opened 2025-04-12
chore: Remove buggy examples for Dialog

VincentSmedinga:
# Describe the pull request

## What

This consolidates the examples for Dialog into two: the ‘Open and Close’ one, and a new Default that doesn’t even use a form.

## Why

We tried to show a Dialog inline and had to add a fake backdrop for it, but forgot to remove the backdrop after people closed the Dialog. We’ve had earlier bugs for the Dialog examples because of the unnatural way we rendered it, e.g. #1839.

The first and second example were largely the same – but the first was shown inline (fake) and the second after clicking a trigger button (real). The content of the first wasn’t right either – an Alert would fit it better. The third example was a combination of the first two, with documentation on the form element. I’ve moved the docs into the README. The fourth example only showed how tall content would scroll, but that’s to be expected, and it shows content that we would just display on a page, not in a Dialog.

So, leaving just two Dialog examples doesn’t make people miss anything and it prevents all the bugs.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


storybook/src/components/Dialog/Dialog.stories.tsx
```diff
@@ -57,127 +49,15 @@ export default meta
...
   args: {
     id: 'ams-dialog',
   },
   decorators: [
     (Story) => (
-      <Column alignHorizontal="start">
-        <Heading level={1}>{heading}</Heading>
-        <Paragraph>{paragraph}</Paragraph>
+      <>
         <Button onClick={() => Dialog.open('#ams-dialog')}>Open dialoog</Button>
```
RubenSibon, line 58: "Open dialoog" sounds strange in Dutch. Users coming from Microsoft Windows might be more familiar with "dialoogvenster".

See: [dialoog](https://nl.wiktionary.org/wiki/dialoog) and [dialoogvenster](https://nl.wiktionary.org/wiki/dialoogvenster) on WikiWoordenboek (Wiktionary).

packages/css/src/components/dialog/README.md
```diff
@@ -10,6 +10,25 @@ A Dialog allows the user to focus on one task or a piece of information by poppi
...
+Add `method="dialog"` to let the browser close the Dialog automatically when the form is submitted.
+
+Wrap the buttons in an [Action Group](/docs/components-layout-action-group--docs) and place it in the `footer`.
+This ensures correct whitespace and scrolling behaviour.
+At the same time, this will position the buttons outside the `form` element.
+Create an `id` for the form and add it to the submit Button’s `form` attribute to connect the two.
+
+If the Action Group must be in the `form`, implement the whitespace and scrolling behaviour as well.
+Add a medium bottom margin (`ams-mb-m`) to the element before it.
+Make sure the content of the form scrolls if necessary, while the Action Group is visible at the bottom at all times.
```
RubenSibon, line 28: I find this section confusing. Why would it be required to have the Action Group in the `form`? Is it necessary to have this exceptional case in the main documentation?

dlnr:
I'm missing the point of this. The storybook Dialog page is to inspect (ie. to work on) the Dialog component, the "fake" backdrop was added on purpose to simulate a working Dialog. Building a working Dialog was done on one story where you have an open Dialog button to describe the method in which to call the Dialog. 

You could've also just replace the `Dialog.close` function with a log to storybook actions in these demo stories.

storybook/src/components/Dialog/Dialog.stories.tsx
VincentSmedinga: As discussed, updated to just ‘Open’.

packages/css/src/components/dialog/README.md
VincentSmedinga: As discussed, I moved this back to a second example and made this explanation more compact.

VincentSmedinga:
> I'm missing the point of this. The storybook Dialog page is to inspect (ie. to work on) the Dialog component, the "fake" backdrop was added on purpose to simulate a working Dialog. Building a working Dialog was done on one story where you have an open Dialog button to describe the method in which to call the Dialog. The Dialog with a scrollbar story is also missing now.

As mentioned in the description, the fake backdrop posed more problems than it gave benefits. That the Dialog scrolls overflowing content is not surprising to warrant an example; certainly not an example with content we don’t want in a Dialog.

> You could've also just replace the Dialog.close function with a log to storybook actions in these demo stories.

Good idea, I’ve added actions.

alimpens (approved):

PR #1998, opened 2025-04-13
feat!: Use border property for borders and box shadows only for hovers

VincentSmedinga:
# Describe the pull request

## What

This uses the CSS `border` property to draw borders. Until now, we used `box-shadow` for them, because borders had been drawn on top of padding in Figma.

We now use `box-shadow` only to make borders stronger on hover.

## Why

Using `box-shadow` presented various problems:
- The syntax of its CSS values is more complicated and less easy to override
- The actual padding is less than specified because the ‘border’ covers part of it
- Box shadows disappear in forced colors, requiring us to still use border _and_ subtract its width from the paddings

Just using `border` does change the design a bit, but is much more practical.

Additionally, it allows us to override only the color for other states (hover and disabled) instead of redeclaring unchanged values for style and width.

## How

Find and replace, test visually.

## Checklist

- [x] Add or update stories

## Additional notes

- The width of the focus outline token has been doubled to achieve its pre-existing appearance. I can’t exactly explain why this is necessary. It’s probably because we let the browser draw the outline – as soon as I add an explicit `outline: 2px solid transparent`, the offset is predicatable, and it doesn’t have rounded corners anymore.
- The original appearance of the Switch thumb can’t be replicated with borders, so it has changed by a pixel. This does make it adhere more to the latest design, which we haven’t implemented yet.


packages/css/src/components/alert/alert.scss
```diff
@@ -5,13 +5,8 @@
 
 .ams-alert {
   background-color: var(--ams-alert-background-color);
-  box-shadow: var(--ams-alert-box-shadow);
+  border: var(--ams-alert-border);
```
alimpens, line 8: We could split this shorthand, so we would only have to set `border-color` in the variant classes

packages-proprietary/tokens/src/components/ams/avatar.tokens.json
```diff
@@ -3,6 +3,7 @@
     "avatar": {
       "aspect-ratio": { "value": "{ams.aspect-ratio.1-1}" },
       "background-color": { "value": "{ams.color.highlight.purple}" },
+      "border": { "value": "{ams.border.width.m} solid {ams.color.highlight.purple}" },
```
alimpens, line 6: Same here

packages-proprietary/tokens/src/components/ams/button.tokens.json
```diff
@@ -13,41 +13,42 @@
...
-          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.m} {ams.color.interactive.hover}" }
+          "border": { "value": "{ams.border.width.m} solid {ams.color.interactive.hover}" },
+          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.s} {ams.color.interactive.hover}" }
         }
       },
       "secondary": {
         "background-color": { "value": "{ams.color.background}" },
         "color": { "value": "{ams.color.interactive.default}" },
-        "box-shadow": { "value": "inset 0 0 0 {ams.border.width.m} {ams.color.interactive.default}" },
+        "border": { "value": "{ams.border.width.m} solid {ams.color.interactive.default}" },
```
alimpens, line 33: We could not set a `border-color` here. It used the defined `color` in that case. Makes it a bit easier for hover and disabled states.

packages-proprietary/tokens/src/components/ams/button.tokens.json
```diff
@@ -13,41 +13,42 @@
...
         "disabled": {
           "background-color": { "value": "{ams.color.interactive.disabled}" },
-          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.m} {ams.color.interactive.disabled}" }
+          "border": { "value": "{ams.border.width.m} solid {ams.color.interactive.disabled}" }
         },
         "hover": {
           "background-color": { "value": "{ams.color.interactive.hover}" },
-          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.m} {ams.color.interactive.hover}" }
+          "border": { "value": "{ams.border.width.m} solid {ams.color.interactive.hover}" },
+          "box-shadow": { "value": "inset 0 0 0 {ams.border.width.s} {ams.color.interactive.hover}" }
```
alimpens, line 27: I don't think our primary button needs this?

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
```diff
@@ -2,7 +2,7 @@
   "ams": {
     "date-input": {
       "background-color": { "value": "{ams.color.background}" },
-      "box-shadow": { "value": "inset 0 0 0 {ams.border.width.s} {ams.color.interactive.secondary}" },
+      "border": { "value": "{ams.border.width.s} solid {ams.color.interactive.secondary}" },
```
alimpens, line 5: We could remove the color here, which makes the border the currentColor.

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
```diff
@@ -19,7 +19,7 @@
       },
       "disabled": {
         "background-color": { "value": "{ams.color.background}" },
-        "box-shadow": { "value": "inset 0 0 0 {ams.border.width.s} {ams.color.interactive.disabled}" },
+        "border": { "value": "{ams.border.width.s} solid {ams.color.interactive.disabled}" },
```
alimpens, line 22: If we remove the color, we can get rid of this token I think

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
```diff
@@ -29,17 +29,18 @@
         }
       },
       "hover": {
-        "box-shadow": { "value": "inset 0 0 0 {ams.border.width.m} {ams.color.interactive.secondary}" },
+        "box-shadow": { "value": "inset 0 0 0 {ams.border.width.s} {ams.color.interactive.secondary}" },
```
alimpens, line 32: We could remove the color here as well

packages-proprietary/tokens/src/components/ams/switch.tokens.json
```diff
@@ -7,10 +7,12 @@
       "inline-size": { "value": "3.5rem" },
       "thumb": {
         "background-color": { "value": "{ams.color.background}" },
-        "inline-size": { "value": "1.75rem" },
         "block-size": { "value": "1.75rem" },
+        "border": { "value": "{ams.border.width.s} solid transparent" },
```
alimpens, line 11: Maybe we shouldn't change this component. In both the new and old design, a `box-shadow` is more appropriate imo. And we already fixed this component in forced colors mode.

Bondt007 (dismissed):
Aram was already thorough :)

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
VincentSmedinga: I’m not sure this always works. Let’s see tomorrow.

packages-proprietary/tokens/src/components/ams/switch.tokens.json
VincentSmedinga: Alright. Also saves us from a glitchy hover effect that it introduced. Reverted, but maintained using a border width token and sorting ‘inline-size’ correctly.

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
VincentSmedinga: Done.

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
VincentSmedinga: Done.

packages-proprietary/tokens/src/components/ams/date-input.tokens.json
VincentSmedinga: Updated to use `currentColor` as discussed.

packages-proprietary/tokens/src/components/ams/button.tokens.json
VincentSmedinga: It does need a border as it influences the size of the button. Removed the box shadow.

packages-proprietary/tokens/src/components/ams/button.tokens.json
VincentSmedinga: Discussed: we build the components for ourselves for now. We can add options we don’t use ourselves as soon as people start asking for it.

VincentSmedinga (commented):
Thanks for testing! Addressed most but not all feedback.

packages/css/src/components/button/button.scss
```diff
@@ -38,42 +38,33 @@
...
-    border: var(--ams-button-forced-color-mode-border); // add border because forced colors changes box-shadow to none
-  }
-}
-
 .ams-button--primary {
   background-color: var(--ams-button-primary-background-color);
-  box-shadow: var(--ams-button-primary-box-shadow);
+  border-color: var(--ams-button-primary-border-color);
+  border-style: var(--ams-button-border-style);
+  border-width: var(--ams-button-border-width);
```
alimpens, line 45: These tokens are moved to the root, the rules should be moved there as well.

packages/css/src/components/select/select.scss
```diff
@@ -5,13 +5,14 @@
...
   border-radius: 0; // Reset rounded borders for Safari on MacOS
 }
 
 .ams-select {
   background-color: var(--ams-select-background-color);
-  box-shadow: var(--ams-select-box-shadow);
+  border-color: var(--ams-select-border-color);
+  border-style: var(--ams-select-border-style);
+  border-width: var(--ams-select-border-width);
   color: var(--ams-select-color);
```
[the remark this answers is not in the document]
alimpens: We first have to decide if we want to do it, before we decide when we want to do it I guess.

packages/css/src/components/button/button.scss
VincentSmedinga: Of course.

packages/css/src/components/date-input/date-input.scss
```diff
@@ -56,7 +57,6 @@
 
 .ams-date-input:disabled {
```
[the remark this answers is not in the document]
VincentSmedinga: I’ve updated the `&:hover` selectors to `&:not(:disabled):hover`. I also had to move them to the bottom of the stylesheet for specificity reasons…

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
VincentSmedinga: If we decide the question is out of scope, we don’t need to answer it yet :)

I would expect users to associate Selects more with Inputs than with Buttons, so I’m not inclined to give it a pointer cursor.

packages/css/src/components/date-input/date-input.scss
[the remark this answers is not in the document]
VincentSmedinga: Ticket: https://gemeente-amsterdam.atlassian.net/browse/DES-1035

packages/css/src/components/text-area/text-area.scss
```diff
@@ -8,14 +8,15 @@
...
   border-radius: 0; // Reset rounded borders on iOS devices
   margin-block: 0;
 }
 
 .ams-text-area {
   background-color: var(--ams-text-area-background-color);
-  box-shadow: var(--ams-text-area-box-shadow);
+  border-color: var(--ams-text-area-border-color);
+  border-style: var(--ams-text-area-border-style);
+  border-width: var(--ams-text-area-border-width);
```
alimpens, line 19: Wrong line, but: a disabled text area still gets a box shadow on hover

packages/css/src/components/select/select.scss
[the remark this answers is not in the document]
alimpens: [I've added a ticket to discuss this. ](https://gemeente-amsterdam.atlassian.net/browse/DES-1280)

packages/css/src/components/text-area/text-area.scss
VincentSmedinga: Ah sorry, I seem to have forgotten that one in the last commit. Added now.

alimpens (approved):

PR #2001, opened 2025-04-15
chore(Card): Improve whitespace

VincentSmedinga:
# Describe the pull request

## What

Updates values for spacing tokens of Card.

This is a follow-up for #1972, hence the ‘chore’ prefix – this doesn’t need to go into the change log.

## Why

We want:
- No extra space between the Heading and the following Paragraph
- XS space between a tagline (metadata) and the Heading
- S space below the image

## How

Visually, together with @Bondt007.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/card.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/card.tokens.json
@@ -2,11 +2,11 @@
   "ams": {
     "card": {
       "heading": {
-        "margin-block-end": { "value": "0" }
+        "margin-block-end": { "value": "{ams.space.xs}" }
       },
       "heading-group": {
         "gap": { "value": "{ams.space.xs}" },
-        "margin-block-end": { "value": "0" }
+        "margin-block-end": { "value": "{ams.space.xs}" }
       },
       "image": {
         "margin-block-end": { "value": "{ams.space.s}" }
--- a/packages-proprietary/tokens/src/components/ams/card.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/card.tokens.json
@@ -2,11 +2,11 @@
   "ams": {
     "card": {
       "heading": {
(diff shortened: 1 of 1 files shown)


packages-proprietary/tokens/src/components/ams/card.tokens.json
```diff
@@ -2,11 +2,11 @@
   "ams": {
     "card": {
       "heading": {
-        "margin-block-end": { "value": "{ams.space.s}" }
+        "margin-block-end": { "value": "0" }
```
alimpens, line 5: This feels a little cramped, doesn't really fit with the spacious feel of our design imo

VincentSmedinga: The line height of our text does enough here. And this makes a shared rhythm of vertical whitespace between multiple lines of headings and paragraphs, which unifies them.

@Bondt007 and I did consider `space.xs` here but decided that `0` was better.

alimpens: > And this makes a shared rhythm of vertical whitespace between multiple lines of headings and paragraphs, which unifies them.

I assumed we would want to avoid that? Headings are a group which should be separated from paragraphs as a group, with a different space value than the lines within those groups. Gestalt and all.

VincentSmedinga: Unifiying and separating visual aspects can go hand in hand. The blue colour, the large font and the stronger font weight make it easy to recognize where the Card Heading ends and where body text begins, while said rhythm shows that they still belong together and helps reading them as one.

VincentSmedinga: As discussed with the entire team, we’ve restored the xs space below the Heading.

alimpens (approved):

PR #2002, opened 2025-04-16
fix(Accordion, Tabs): Only allow keyboard focus on relevant buttons

alimpens:
# Describe the pull request

## What

It removes disabled buttons from the Tabs focusable elements selector, and it adds a specific selector for Accordion. It also adds unit tests to Accordion and Tabs to check if this functionality keeps working.

## Why

When adding a disabled Tabs button, the keyboard focus could not pass it. When adding a regular button to an Accordion Section, it would erroneously receive focus when using the arrow keys. 

## How

Using the class name as selectors, and adding `:not([disabled])`

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages/react/src/Tabs/TabsList.tsx
+++ b/packages/react/src/Tabs/TabsList.tsx
@@ -18,7 +18,7 @@ export const TabsList = forwardRef(
     useImperativeHandle(ref, () => innerRef.current as HTMLDivElement)
 
     const { keyDown } = useKeyboardFocus(innerRef, {
-      focusableElements: ['button[role="tab"]'],
+      focusableElements: ['.ams-tabs__button:not([disabled])'],
       horizontally: true,
       rotating: true,
     })
--- a/packages/react/src/Accordion/Accordion.test.tsx
+++ b/packages/react/src/Accordion/Accordion.test.tsx
@@ -1,8 +1,8 @@
-import { fireEvent, render, screen } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import { createRef } from 'react'
 import { Accordion } from './Accordion'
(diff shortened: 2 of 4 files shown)


packages/react/src/Tabs/TabsList.tsx
```diff
@@ -18,7 +18,7 @@ export const TabsList = forwardRef(
     useImperativeHandle(ref, () => innerRef.current as HTMLDivElement)
 
     const { keyDown } = useKeyboardFocus(innerRef, {
-      focusableElements: ['button[role="tab"]'],
+      focusableElements: ['.ams-tabs__button:not([disabled])'],
```
alimpens, line 21: We can do this a couple of different ways. Tabs looks for a React component with a prop called `tab` in `Tabs.List`, so you can't easily use anything other than `Tabs.Button` there. Using the class name here tightly couples `Tabs.List` and `Tabs.Button`, which is ok in this case imo.

In the previous iteration we allowed any `button[role="tab"]` here, which works for the keyboard focus, but you can't easily use your own button in Tabs. 

We could also get rid of `focusableElements` entirely here. `useKeyboardFocus` would use [its default list](https://github.com/Amsterdam/design-system/blob/develop/packages/react/src/common/useKeyboardFocus.tsx#L17), which included `button:not([disabled])`. `useKeyboardFocus` is used on `Tabs.List` now, so the original bug with a button in a Tab Panel would still be fixed, and we wouldn't tightly couple `Tabs.List` and `Tabs.Button`.

alimpens: As discussed, we'll use the selector as is. I'll add tests to Accordion and Tabs to check whether the `useKeyboardFocus` function is working as expected.

VincentSmedinga (approved):

PR #2009, opened 2025-04-18
fix(Tabs): Remember tab after rerendering component

### LINKED ISSUE #2008: Active tab lost after rerender
When rerendering a component that contains tabs, the tab is reset to the initial tab.

Test:
```
  it('should remember tab after rerender', async () => {
    const user = userEvent.setup()

    const {rerender} = render(
      <Tabs>
        <Tabs.List>
          <Tabs.Button tab="one">Tab 1</Tabs.Button>
          <Tabs.Button tab="two">Tab 2</Tabs.Button>
        </Tabs.List>
        <Tabs.Panel tab="one">Content 1</Tabs.Panel>
        <Tabs.Panel tab="two">Content 2</Tabs.Panel>
      </Tabs>,
    )

    const tabOne = screen.getByRole('tab', { name: 'Tab 1' })
    const tabTwo = screen.getByRole('tab', { name: 'Tab 2' })

    if (tabTwo) {
      await user.click(tabTwo)
    }

    rerender(
      <Tabs>
        <Tabs.List>
          <Tabs.Button tab="one">Tab 1</Tabs.Button>
          <Tabs.Button tab="two">Tab 2</Tabs.Button>
        </Tabs.List>
        <Tabs.Panel tab="one">Content 1</Tabs.Panel>
        <Tabs.Panel tab="two">Content 2</Tabs.Panel>
      </Tabs>,
    )

    expect(tabOne).toHaveAttribute('aria-selected', 'false')
    expect(tabOne).toHaveAttribute('tabindex', '-1')
    expect(tabTwo).toHaveAttribute('aria-selected', 'true')
    expect(tabTwo).toHaveAttribute('tabindex', '0')
    expect(screen.getByRole('tabpanel')).toHaveTextContent('Content 2')
  })
```

This is caused by the fact that allTabIds is used as a dependency on a useEffect, which in turn is a memo with a dependency on children. Children has a different reference on each render. 

Can be fixed by using `allTabIds.join(',')` instead of `allTabIds` as a dependency of the useEffect but there might be a better way.

arthur-amsterdam:
fixes #2008

### DIFF
--- a/packages/react/src/Tabs/Tabs.test.tsx
+++ b/packages/react/src/Tabs/Tabs.test.tsx
@@ -91,6 +91,45 @@ describe('Tabs', () => {
     expect(screen.getByRole('tabpanel')).toHaveTextContent('Content 2')
   })
 
+  it('should remember tab after rerender', async () => {
+    const user = userEvent.setup()
+
+    const { rerender } = render(
+      <Tabs>
+        <Tabs.List>
+          <Tabs.Button tab="one">Tab 1</Tabs.Button>
+          <Tabs.Button tab="two">Tab 2</Tabs.Button>
+        </Tabs.List>
+        <Tabs.Panel tab="one">Content 1</Tabs.Panel>
+        <Tabs.Panel tab="two">Content 2</Tabs.Panel>
+      </Tabs>,
+    )
+
+    const tabOne = screen.getByRole('tab', { name: 'Tab 1' })
+    const tabTwo = screen.getByRole('tab', { name: 'Tab 2' })
+
+    if (tabTwo) {
+      await user.click(tabTwo)
+    }
+
+    rerender(
+      <Tabs>
+        <Tabs.List>
+          <Tabs.Button tab="one">Tab 1</Tabs.Button>
+          <Tabs.Button tab="two">Tab 2</Tabs.Button>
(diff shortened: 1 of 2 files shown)


RubenSibon:
Hi @arthur-amsterdam, we're not exactly sure what the problem is that needs to be fixed. We see that the test succeeds, but I have personally not been able to reproduce this behaviour in [actual React code](https://codesandbox.io/p/sandbox/amsterdam-design-system-mys7n8).

The expected behaviour of Tabs is that the first tab is selected by default if the `activeTab` property is not set. By setting the `activeTab` you can make sure that the correct tab is selected on a rerender. Does this not work for your use-case?

My colleagues, Aram and Vincent, will come back on this soon.

arthur-amsterdam:
Hi @RubenSibon. Currently it is not possible to use the Tabs uncontrolled. I did not see a mention in the docs that it is required to make them controlled so I thought it was a bug. I can make them controlled but that just adds more logic to our codebase while that seems unnecessary.

alimpens:
This should be fixed in [this PR](https://github.com/Amsterdam/design-system/pull/2025). Please check if this resolves your problem and close this PR if it does.

arthur-amsterdam:
Thanks @alimpens!

PR #2020, opened 2025-04-24
fix(Date Input, Select, Time Input): Use correct colors for icons

VincentSmedinga:
# Describe the pull request

## What

Updates the color codes for the icons in the Date and Time Inputs in their disabled and hovered states. And uses availabe color tokens in our stylesheet for Storybook overrides.

## Why

These were probably forgotten when modeling the brand colour tokens. (Because they are hardcoded. A new ticket is available to improve the approach.)

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages-proprietary/tokens/src/components/ams/select.tokens.json
```diff
@@ -17,7 +17,7 @@
       "padding-inline": { "value": "{ams.space.m} calc(2 * {ams.space.m} + 1em)" },
```
alimpens, line 17: Not the right line, but: the chevron should change color on hover I think

VincentSmedinga: Cool, added – a bit to the bottom of the stylesheet to prevent descending specificity.

alimpens: It's a little confusing to me that we define a hover style on line 35 and another hover style on line 55. Those should be close together imo. Maybe we should not use nested selectors here, to make everything a bit more explicit and not create descending specificity issues.

VincentSmedinga: Wait, nesting selectors can make ordering to specificity more difficult if child selectors have different specificities.

I do agree we should improve this, but I think it affects many components so it would warrant a separate task.

alimpens: We could improve it in this PR first, seeing as we're running into these issues now. That will give us patterns we can implement everywhere else.

VincentSmedinga: I really don’t think we should include such a complicated aspect of CSS architecture into the original task.

I agree that it’s important to make it easy for people to grasp a stylesheet. I’ve grouped the invalid+hover styles back with the invalid ones, listed hover styles as the first group everywhere because they are the ones we repeat for various states, and disabled the Stylelint rule for descending specificity here.

VincentSmedinga: As a matter of fact, as I’m making a point not to slip refactoring in, I’ve reverted all reordering and just added a declaration (and a Stylelint exception) to let the chevron change color on hover, as requested.

VincentSmedinga: Ticket: https://gemeente-amsterdam.atlassian.net/browse/DES-1035

alimpens (approved):

PR #2028, opened 2025-05-02
fix(TextInput, PasswordInput): Restore `size` attribute functionality

alimpens:
# Describe the pull request

## What

It restores the `size` attribute functionality, allowing you to set the input's width in `ch`.

## Why

An input's width should be appropriate for the information to be entered.
You can use the `size` prop for this, but we broke that previously.

## How

Only set `inline-size: 100%` if the `size` attribute is not used. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/storybook/src/components/PasswordInput/PasswordInput.stories.tsx
+++ b/storybook/src/components/PasswordInput/PasswordInput.stories.tsx
@@ -21,7 +21,7 @@ const meta = {
     },
     size: {
       control: { min: 0, type: 'number' },
-      description: 'The size of the input.',
+      description: 'The width, expressed in the average number of characters.',
     },
   },
 } satisfies Meta<typeof PasswordInput>
--- a/storybook/src/components/PasswordInput/PasswordInput.stories.tsx
+++ b/storybook/src/components/PasswordInput/PasswordInput.stories.tsx
@@ -13,10 +13,15 @@ const meta = {
   args: {
     disabled: false,
     invalid: false,
+    size: 0,
   },
   argTypes: {
     disabled: {
       description: 'Prevents interaction. Avoid if possible.',
+    },
+    size: {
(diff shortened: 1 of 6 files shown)


dlnr (dismissed):

dlnr:
| It restores the size attribute functionality, allowing you to set the input's width in ch.

Don't think you set the css unit `ch`, if it did it probably fit 10 actual characters in the stories. The `size` attribute was there long before the css unit.

storybook/src/components/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -13,11 +13,16 @@ const meta = {
   args: {
     disabled: false,
     invalid: false,
+    size: 0,
```
VincentSmedinga, line 16: This shows the associated code as `<PasswordInput size={0} />`. We should probably remove this initial value to prevent people from inadvertently pasting it into their applications. The downside is that they must press the ‘Set Number’ button first, but this is how Storybook distinghuishes undefined and zero values.

```suggestion
```

storybook/src/components/TextInput/TextInput.stories.tsx
```diff
@@ -13,6 +13,7 @@ const meta = {
   args: {
     disabled: false,
     invalid: false,
+    size: 0,
```
VincentSmedinga, line 16: Same here.

```suggestion
```

alimpens:
> > It restores the size attribute functionality, allowing you to set the input's width in ch.
> 
> Don't think you set the css unit `ch`, if it did it probably fit 10 actual characters in the stories. The `size` attribute was there long before the css unit.

Good catch, it's in `em`s. I've updated the description.

VincentSmedinga (approved):

PR #2044, opened 2025-05-14
feat(Page Header): Make brand name larger in Compact Mode

dlnr:
# Describe the pull request

## What

The brandname in the Page Header on the compact theme was rather small, this changes the variable to header level 2.

## How

A new compact token file overriding just the font-size token for brandname.

## Checklist

- [ ] Add or update stories

## Additional notes


packages-proprietary/tokens/src/components/ams/page-header.compact.tokens.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ams": {
+    "page-header": {
+      "brand-name": {
+        "font-size": { "value": "{ams.typography.heading.2.font-size}" },
+        "line-height": { "value": "{ams.typography.heading.2.line-height}" }
```
VincentSmedinga, line 6: This introduces a build error:

> ams.page-header.brand-name.line-height.value tries to reference ams.typography.heading.2.line-height, which is not defined.

The compact stylesheet is built separately from the spacious one, and our compact typography tokens only override font size, not line height.

I also expected this reference to fall back to the spacious token, but Style Dictionary doesn’t know that we intend compact styles to be used as a layer on top of the spacious (default) ones. Can we express this in our build config, so that compact component tokens can use spacious brand and common tokens? Or does this separation serve a purpose and should we copy the common line height values? What do you think @alimpens?

VincentSmedinga: This also made me discover why we had `log: { verbosity: 'verbose' }` in the `modes.map` block of the build script – the current verbosity setting only applies to the default mode. You must add it to see the error message above. And we might improve the sharing of configs here.

dlnr: If the value isn't changed then it would make sense. It's weird that you can't even reference the value while it should be there though.

alimpens: Hm, tricky. The default mode and the other modes are built completely separately from each other, resulting in CSS files with custom properties that that only refererence other properties in that same file.

I'm not sure how we can let one mode reference a prop in another mode. It would need to check references, and if one doesn't exist, check another file? And then what? Add a reference to something that doesn't exist in the CSS file? That would kind of undermine the reference checks, I'm not sure that's a good idea. 

A solution to this is to render complete CSS files for all modes, so the compact CSS file also has all brand, common and component tokens. A downside to this is that you get an exponential growth of files if you want to combine modes, ie `compact-dark.css`, `compact-light.css`, `default-dark`, etc. I'm not sure that's the way to go either.

dlnr: After some consideration I would not bother to try. This way you're forced to not create obsolete lines in compact tokens and keep it clean.

dlnr: Also, this build error has been fixed so this comment could be resolved @VincentSmedinga ?

dlnr:
@VincentSmedinga  Please consider the Figma implementation because the working title now overlaps the menu button. Is Heading 2 too big? https://www.figma.com/design/9IGm6IdPUYizBNGsUnueBd/Amsterdam-Design-System?node-id=22197-14610&t=kDbcPKNIKljGzHuN-11

packages-proprietary/tokens/src/components/ams/page-header.compact.tokens.json
VincentSmedinga: I agree that we should try to keep the current approach of one default mode that sets everything, with separate modes that override aspects only (e.g. color, sizing, typography).

The issue at hand might be surfacing an oversight of ours. Compact mode overrides the values of our spacing and font-size tokens, but not the one for line-height. Even though their values are (incorrectly?) the same for spacious and compact, I would argue that line-height belongs in the spacing aspect.

Defining our line heights in our compact typography tokens allows them to be referenced in the compact styles for Page Header. This also resolves the issue where the brand name in compact mode now uses a compact font-size but a spacious line-height.

So I suggest we copy the spacious line heights into the compact typography tokens and use the correct line heights in the compact Page Header stylesheet.

VincentSmedinga:
> @VincentSmedinga Please consider the Figma implementation because the working title now overlaps the menu button. Is Heading 2 too big? https://www.figma.com/design/9IGm6IdPUYizBNGsUnueBd/Amsterdam-Design-System?node-id=22197-14610&t=kDbcPKNIKljGzHuN-11

The overlapping of the menu button is a bug in Figma; this doesn’t happen in the browser.

I see you point of Heading 2 being quite large: when I [toggle between Spacious and Compact mode](https://designsystem.amsterdam/demo-DES-1145_brandname-compact-h2/?path=/story/components-containers-page-header--default&args=brandName:Onderzoek+en+Statistiek&globals=viewport:phone;theme:Spacious), not much happens with the font size of the title, while the menu becomes considerably smaller. Let’s ask Bas if he thinks we should use Heading 3 instead.

PR #2096, opened 2025-06-10
fix: Use correct underline width for Standalone Link with icon

VincentSmedinga:
# Describe the pull request

## What

Changes the underline of a hovered Standalone Link with an icon from 2px to 3px.

## Why

It’s a bug, found my @dlnr. A Standalone Link with an icon is a ‘subtle’ link. This group has no underline initially, and a 2px underline on hover.

The variant of a Standalone Link with an icon inherited the hover underline border with for the variant without an icon, which is 3px because it must grow from the initial 2px.

## How

Add a token for the hover of a Standalone Link with an icon that overrides the hover of a Standalone Link without one.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/standalone-link.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/standalone-link.tokens.json
@@ -32,7 +32,8 @@
       "with-icon": {
         "text-decoration-line": { "value": "{ams.links.subtle.text-decoration-line}" },
         "hover": {
-          "text-decoration-line": { "value": "{ams.links.subtle.hover.text-decoration-line}" }
+          "text-decoration-line": { "value": "{ams.links.subtle.hover.text-decoration-line}" },
+          "text-decoration-thickness": { "value": "{ams.links.text-decoration-thickness}" }
         }
       }
     }
--- a/packages/css/src/components/standalone-link/standalone-link.scss
+++ b/packages/css/src/components/standalone-link/standalone-link.scss
@@ -49,5 +49,6 @@
 
   &:hover {
     text-decoration-line: var(--ams-standalone-link-with-icon-hover-text-decoration-line);
+    text-decoration-thickness: var(--ams-standalone-link-with-icon-hover-text-decoration-thickness);
   }
 }


packages-proprietary/tokens/src/components/ams/standalone-link.tokens.json
```diff
@@ -32,7 +32,8 @@
       "with-icon": {
         "text-decoration-line": { "value": "{ams.links.subtle.text-decoration-line}" },
         "hover": {
-          "text-decoration-line": { "value": "{ams.links.subtle.hover.text-decoration-line}" }
+          "text-decoration-line": { "value": "{ams.links.subtle.hover.text-decoration-line}" },
+          "text-decoration-thickness": { "value": "{ams.links.text-decoration-thickness}" }
```
VincentSmedinga, line 36: Now I see this change in the PR, I think this token’s value should be a new token `ams.links.subtle.hover.text-decoration-thickness`.

VincentSmedinga: I’ve moved the thickness declarion to the intial state as per [our conventions](https://github.com/Amsterdam/design-system/blob/develop/packages/css/documentation/coding-conventions.md#where-to-put-properties-for-various-states).

I don’t think we need a new token `ams.links.subtle.hover.text-decoration-thickness` as it would have the same value as `ams.links.text-decoration-thickness`. ‘Subtle’ is a variant of link that changes a few properties but uses the basic, shared value for text decoration thickness.

VincentSmedinga: Is it okay to use the regular links’ **initial** thickness to override the regular links’ **hover** thickness when hovering a variant of Standalone Link?

We must model these underline widths:

- regular links
  - default: 2px
  - hover: 3px
- subtle links
  - default: 0px
  - hover: 2px

When taking `text-decoration-line` into account, we want this resulting CSS:

```css
.regular-link {
  text-decoration: 2px underline;

  &:hover {
    text-decoration-thickness: 3px;
  }
}


.subtle-link {
  text-decoration: 2px none;

  &:hover {
    text-decoration-line: underline;
  }
}
```

Using tokens:

```css
.standalone-link {
  text-decoration-line: var(--ams-links-text-decoration-line); /* underline (UA default) */
  text-decoration-thickness: var(--ams-links-text-decoration-thickness); /* 2px (border-width-m) */

  &:hover {
    text-decoration-thickness: var(--ams-links-hover-text-decoration-thickness);  /* 3px (border-width-l) */
  }
}

.standaline-link-with-icon {
  text-decoration-line: var(--ams-links-subtle-text-decoration-line); /* none (token value) */
  text-decoration-thickness: var(--ams-links-text-decoration-thickness); /* this is the point */

  &:hover {
    text-decoration-line: var(--ams-links-subtle-hover-text-decoration-line); /* underline */
  }
}
```

I think we can reuse `ams.links.text-decoration-thickness` for the subtle variation. We define it on the initial state, it cascades to the hover selector, where the value of `text-decoration-line` gets overridden.

VincentSmedinga: I’ve compared all occurrences of

```
"text-decoration-line": { "value": "{ams.links.subtle.text-decoration-line}" }
```

and Breadcrumb, Card, Link List, Page Footer, Page Header, Pagination, and Table of Contents all already paired this with

```
"text-decoration-thickness": { "value": "{ams.links.text-decoration-thickness}" }
```

so this change to Standalone Link is consistent.

dlnr (approved):

PR #2129, opened 2025-07-11
feat: Menu component for application layout

dlnr:
# Describe the pull request

## What

A menu to be used in a responsive application layout menu. The desktop version will change to a vertical menu set to 112px as an example. In the future the layout container will determine the with of the menu.

## Why

Applications require a different layout compared to content sites. Users expect a vertical menu, which is the future might contain multiple levels of navigation, to rapidly switch between views on desktop. Mobile version still requires an extra click to open this menu.

## How

Being links the Menu resembles Link List a lot, including the color prop to be used on darker backgrounds. The only caveat compared to the design is the hover effect on darker backgrounds. If you hold on to the rules we've set for links the hover effect from the designs (lighter coloured backgrounds) to not apply or need to be reviewed.

## Checklist

- [X] Add or update stories

## Additional notes

- Figma scaffold: [ADS](https://www.figma.com/design/6jw9r2uTFbetk2XwiQN9Ym/DES-1320-%7C-Applicatie-menu?node-id=37619-12232)
- Figma: [DSIA](https://www.figma.com/design/Qwfx9FeXZ2NsfeRP0KIJJz/DSIA--Design-system-interne-applicaties-?node-id=1-33)
- Figma: [AIP](https://www.figma.com/design/4a6C5fMGJ8dd7ZQgdLVcnf/AIP_gemeenteAmsterdam?node-id=11011-10191&t=x5blhRO8CoJYxH79-11_)


packages/css/src/components/menu/menu.scss
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @media screen and (min-width: $ams-breakpoint-wide) {
+    align-items: center;
+    flex-direction: column;
+    max-inline-size: 112px;
+
+    // TODO: Discuss font size changes in breakpoint (scales to 24x24 on desktop)
+    .ams-menu__icon {
+      transform: scale(1.5);
```
dlnr, line 57: What would be a good way to do this? I preferred an aproach like this instead of changing props per breakpoint or tokens per breakpoint. This is pretty lightweight.

packages/css/src/components/menu/menu.scss
```diff
@@ -0,0 +1,76 @@
...
+  text-decoration-thickness: var(--ams-menu-link-text-decoration-thickness);
+  text-underline-offset: var(--ams-menu-link-text-underline-offset);
+
+  &:hover {
+    color: var(--ams-menu-link-hover-color);
+    text-decoration-line: var(--ams-menu-link-hover-text-decoration-line);
+  }
+
+  // TODO: Discuss className specificity in child components
+  .ams-menu__icon {
```
dlnr, line 46: This needed to be nested like this because the specificity of was the same as `.ams-icon` and got overruled.

VincentSmedinga (commented):
Good to see the start of this pattern entering our design system!

packages/css/src/components/menu/menu.scss
```diff
@@ -0,0 +1,73 @@
...
+  &:hover {
+    color: var(--ams-menu-link-hover-color);
+    text-decoration-line: var(--ams-menu-link-hover-text-decoration-line);
+  }
+
+  .ams-menu__icon {
+    align-self: initial;
+  }
+
+  @media screen and (min-width: $ams-breakpoint-wide) {
```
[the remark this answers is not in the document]
alimpens: I disagree here, using `window.matchMedia` would mean you would always have to use JS for this functionality. That's less performant and less robust.

packages/css/src/components/menu/menu.scss
dlnr: This would require javascript to change the prop in the icon. I wouldn't do that.

packages/css/src/components/menu/menu.scss
dlnr: Right but i'm not working in the icon.scss file right?

dlnr: To elaborate I do not want to target `.ams-icon` in this file. To target the className modifier on the `Icon` component inside `.ams-menu__link` seems like the way to go.

VincentSmedinga (dismissed):

VincentSmedinga (approved):
Yes! Our first real application-specific component 🎉

PR #2135, opened 2025-07-16
feat(Pagination): Rename `*visuallyHidden*` props to `*accessibleName*` to align with conventions

alimpens:
# Describe the pull request

## What

It renames `*visuallyHidden*` props to `*accessibleName*`.

## Why

This is more consistent with our conventions, see [Page Header](https://designsystem.amsterdam/?path=/docs/components-containers-page-header--docs).

## How

Added new props and deprecated the old ones. I've moved the inline defaults down, in order to be able to use both old and new props without issues. When we remove the deprecated props, we can move these to be inline again. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/checkbox.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/checkbox.tokens.json
@@ -32,7 +32,7 @@
       },
       "icon-container": {
         "block-size": { "value": "calc({ams.checkbox.font-size} * {ams.checkbox.line-height})" },
-        "inline-size": { "value": "1.5rem" }
+        "inline-size": { "value": "{ams.checkbox.font-size}" }
       },
       "indeterminate-indicator": {
         "stroke": { "value": "{ams.color.interactive.inverse}" }
--- a/packages-proprietary/tokens/src/components/ams/icon.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/icon.tokens.json
@@ -11,6 +11,18 @@
         "font-size": { "value": "{ams.typography.body-text.large.font-size}" },
         "line-height": { "value": "{ams.typography.body-text.large.line-height}" }
       },
+      "heading-0": {
+        "font-size": { "value": "{ams.typography.heading.0.font-size}" },
+        "line-height": { "value": "{ams.typography.heading.0.line-height}" }
+      },
+      "heading-1": {
+        "font-size": { "value": "{ams.typography.heading.1.font-size}" },
+        "line-height": { "value": "{ams.typography.heading.1.line-height}" }
+      },
+      "heading-2": {
+        "font-size": { "value": "{ams.typography.heading.2.font-size}" },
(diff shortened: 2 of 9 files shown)


packages/react/src/Pagination/Pagination.test.tsx
```diff
@@ -105,14 +105,22 @@ describe('Pagination', () => {
...
-  it('sets a custom id for the accessible label', () => {
+  it('sets a custom id for the accessible label - deprecated', () => {
     render(<Pagination linkTemplate={linkTemplate} totalPages={10} visuallyHiddenLabelId="custom-id" />)
 
     const component = screen.getByRole('navigation', { name: 'Paginering' })
 
     expect(component).toHaveAttribute('aria-labelledby', 'custom-id')
   })
 
+  it('sets a custom id for the accessible label', () => {
```
VincentSmedinga, line 116:
```suggestion
  it('sets a custom id for the accessible name', () => {
```

packages/react/src/Pagination/Pagination.test.tsx
```diff
@@ -145,15 +153,23 @@ describe('Pagination', () => {
...
+  it('renders an accessible label for the navigation - deprecated', () => {
     render(<Pagination linkTemplate={linkTemplate} totalPages={10} visuallyHiddenLabel="Pagination" />)
 
     const component = screen.getByRole('navigation', { name: 'Pagination' })
 
     expect(component).toBeInTheDocument()
   })
 
-  it('renders accessible labels for the ‘previous’ and ‘next’ links', () => {
+  it('renders an accessible label for the navigation', () => {
```
VincentSmedinga, line 164:
```suggestion
  it('renders an accessible name for the navigation', () => {
```

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -75,17 +100,22 @@ export const Pagination = forwardRef(
...
-        aria-labelledby={visuallyHiddenLabelId}
+        aria-labelledby={accessibleNameId || visuallyHiddenLabelId || 'ams-pagination-a11y-label'}
         className={clsx('ams-pagination', className)}
         ref={ref}
       >
-        <span className="ams-visually-hidden" id={visuallyHiddenLabelId}>
-          {visuallyHiddenLabel}
+        <span
+          className="ams-visually-hidden"
+          id={accessibleNameId || visuallyHiddenLabelId || 'ams-pagination-a11y-label'}
```
VincentSmedinga, line 109: Why change where we define default values?
(Same for `previousAccessibleName` and `nextAccessibleName`.)

VincentSmedinga: Never mind I should have read the description first.

alimpens:
> Doesn’t the PR title contain two too many asterisks?

No, I don't think so? You have both `nextAccessibleName` and `accessibleNameId` / `nextVisuallyHiddenLabel` and `visuallyHiddenLabelId`, so before and after for both prop groups.

PR #2150, opened 2025-07-25
chore: Upgrade to Storybook 9

VincentSmedinga:
# Describe the pull request

## What

Upgrades our Storybook to version 9.

## Why

To get all the latest and greatest.

## How

1. Ran their automated migration wizard and said yes to everything. Observed that there were no errors.
2. Adjusted some formatting manually and ran `npm run lint-fix` which caught some other things.
3. Moved the dependency on eslint-plugin-storybook to the root as our general ESLint config uses it
4. Replaced our storybook-addon-html plugin with a fork that works with Storybook 9, even though I had to remove the `getAbsolute` path function that the wizard added to see the HTML tab in action again. I chose to remove it everywhere, not just where calling the addon.

## Checklist

- [x] Add or update stories

## Additional notes

- This also resolves a missing peer dependency with vite that I accidentally introduced in #2118.


alimpens:
It seems they've partially implemented dark mode: 

image

Is there a way to explicitly set Storybook to light mode for the time being?

VincentSmedinga:
> Is there a way to explicitly set Storybook to light mode for the time being?

It was hard to find, but the cause is `color-scheme: light dark;` with which Storybook opts into both light and dark mode.

Our custom theme, which is based on the light Storybook theme, does a lot to keep everything light either way, but it seems not to address radio buttons. So they do become black on devices that use dark mode.

I’ve added an overridde to use only light mode on radio buttons. Just in case, I did the same for checkboxes and included the current selector that Storybook applies the color scheme to.

alimpens (approved):

PR #2161, opened 2025-08-04
feat(Inputs, Text Area, Select, Button): Improve line height of entered text

VincentSmedinga:
# Describe the pull request

## What

Decreases the line height of various form controls from 1.8 to 1.4.

## Why

While 1.8 is fine for body text, it is too large for text entered into a Text Area or a mult-line label of a Button. For single-line Inputs and Select, line height doesn’t appear between lines but only adds to their height.

The initial goal was to make buttons a bit less tall – they had become bulky after changing box shadows to borders in #1998. Decreasing line height seems to do enough here – although we could consider increasing the horizontal padding to adjust their ratio some more – and has the advantages above.

## How

Introduce a new common token for line heights in inputs and use and document it. I’ve added some mock content to our form example so that people actually see how entered data looks, and applied appropriate sizes to the inputs.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages-proprietary/tokens/src/components/ams/file-input.tokens.json
```diff
@@ -10,7 +10,7 @@
       "font-family": { "value": "{ams.typography.font-family}" },
       "font-size": { "value": "{ams.typography.body-text.font-size}" },
       "font-weight": { "value": "{ams.typography.body-text.font-weight}" },
-      "line-height": { "value": "{ams.typography.body-text.line-height}" },
+      "line-height": { "value": "{ams.inputs.line-height}" },
```
alimpens, line 13: Shouldn't this be on `file-selector-button` as well? The button is too small here I think

storybook/src/pages/amsterdam/FormPage/FormPage.tsx
```diff
@@ -56,14 +57,20 @@ export const FormPage = () => {
...
+                  autoComplete="family-name"
+                  id="familyName"
+                  name="familyName"
+                  size={32}
+                  value="van der Linden"
+                />
               </Field>
               <FieldSet legend="Woonplaats">
-                <Radio name="city" value="amsterdam">
+                <Radio checked name="city" value="amsterdam">
```
alimpens, line 73: I don't think we should show preselected radio buttons. That prevents a user from not answering, and it makes it easier to send an incorrect answer.

alimpens:
Nice work! Looks a lot better

packages-proprietary/tokens/src/components/ams/file-input.tokens.json
VincentSmedinga: Yes, good find, the difference was less than 2px 😅
The line hight on the file input itself doesn’t even do much.

storybook/src/pages/amsterdam/FormPage/FormPage.tsx
VincentSmedinga: Good idea. I reverted it and added this as documentation.

packages-proprietary/tokens/src/components/ams/file-input.tokens.json
alimpens: No, the button is always taller than a single line of text. But it's safer to keep it, in case the text wraps.

alimpens (approved):

PR #2184, opened 2025-08-22
chore: Use lint rules

alimpens:
# Describe the pull request

## What

Adds the recommended rules from the plugins we use to our ESLint config and cleans it up a bit. 

## Why

We load a lot ESLint plugins, but we didn't really use their rules, or only used them partly.
Using the recommended rules makes the code base a bit more consistent.

## How

Adding the rules to the config, remove rules that aren't necessary anymore, clean up.

## Checklist

- [x] Add or update stories

## Additional notes

There are still some choices we have to make on what rules to use, see the `TODO`s.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -186,7 +186,7 @@ export default tseslint.config(
           groups: ['title', 'component'],
         },
       ],
-      'perfectionist/sort-modules': 'off', // TODO: do we want this? In some cases, this means you use a type before it's defined (see Grid for example). That makes it harder to read imo.
+      'perfectionist/sort-modules': 'off', // We decide ourselves how we make our code readable
       'perfectionist/sort-union-types': 'off', // This causes more issues than it solves
 
       // React


eslint.config.mjs
```diff
@@ -135,138 +118,73 @@ export default tseslint.config(
...
+            ['internal', 'unknown'],
+            ['parent', 'sibling', 'index'],
+          ],
+          'newlines-between': 'never',
+          warnOnUnassignedImports: false,
         },
       ],
+
+      // Perfectionist
+      'perfectionist/sort-imports': 'off', // TODO: this actually works well, but conflicts with eslint-plugin-import. Do we want to go for the perfectionist format?
```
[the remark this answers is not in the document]
alimpens: We currently do not use `eslint-plugin-import` for module validity, safety, and policy, only for sorting and resolving. Do you mean to add rules? Typescript already checks if files and imports from files exist and stuff like that.

eslint.config.mjs
```diff
@@ -281,33 +184,15 @@ export default tseslint.config(
...
-          ...perfectionistCommonConfig.partition,
         },
       ],
-      'prefer-regex-literals': 'error',
-      radix: 'error',
-      // Start of React rules
-      'react/jsx-key': 'off',
-      'react/no-unknown-property': 'off',
-      'react/prop-types': 'off',
+      'perfectionist/sort-modules': 'off', // TODO: do we want this? In some cases, this means you use a type before it's defined (see Grid for example). That makes it harder to read imo.
```
alimpens, line 189: What do we want to do here?

VincentSmedinga: Hmm that’s too bad. I agree that using a type before it’s defined is confusing. I wonder why they sort exported members before local ones in their [default groups config](https://perfectionist.dev/rules/sort-modules.html#groups) – the other way around seems much more practical: we extract local types before using them in exported ones.

I’d advocate setting our own value for `groups`, but if you want to limit configuration here as well, let’s keep this rule turned off.

PR #2193, opened 2025-08-29
feat(Accordion): Display Headings in the size that matches their levels

alimpens:
# Describe the pull request

## What

It makes the Headings in an Accordion button smaller when you change the level. 
It also uses the Heading component instead of a custom one.

## Why

Accordions can be used in a lot of places in the document, so different sizes make sense.
It's easier and more consistent to reuse the Heading component for this, which already has this functionality.

## How

Replacing the custom heading component with Heading. Add tests and docs. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages/react/src/Accordion/AccordionSection.test.tsx
+++ b/packages/react/src/Accordion/AccordionSection.test.tsx
@@ -32,7 +32,7 @@ describe('Accordion section', () => {
     })
     const section = container.querySelector('.ams-accordion__panel')
 
-    expect(heading).toHaveClass('ams-heading')
+    expect(heading).toHaveClass('ams-heading ams-accordion__header')
     expect(button).toHaveClass('ams-accordion__button')
     expect(section).toBeInTheDocument()
   })


packages/react/src/Accordion/Accordion.tsx
```diff
@@ -15,15 +15,19 @@ export type AccordionProps = {
   /**
    * The hierarchical level of this Accordion’s Section Headings within the document.
    * There is no default value; determine the correct level for each instance.
-   * The size of the heading is fixed at level 3.
    */
-  headingLevel: HeadingProps['level']
+  headingLevel: Exclude<HeadingProps['level'], 1>
+  /** Uses larger or smaller text without changing its position in the heading hierarchy. */
+  headingSize?: Exclude<HeadingProps['size'], 'level-1' | 'level-6'>
```
VincentSmedinga, line 21: We could consider deprecating these values instead of changing them – we don’t have any other breaking changes yet. Let’s discuss on Wednesday.

PR #2207, opened 2025-09-17
Add issue templates

VincentSmedinga:
We have a template for our pull requests and the same would be useful for issues that people submit. As a simple good start, this adds the default ones that GitHub offers.


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
```
VincentSmedinga, line 1: We could also use [Discussions](https://github.com/Amsterdam/design-system/discussions) for this.

PR #2230, opened 2025-10-02
fix(Icon): Use capital I in name of icon for LinkedIn

VincentSmedinga:
# Describe the pull request

## What

This updates the lowercase i to an uppercase one in the file and component names of the icon for LinkedIn.

## Why

This is how they spell their name.

## How

Through `git mv`, among other things.

## Checklist

- [x] Add or update stories

## Additional notes

- We would normally consider this a breaking change, but we can’t deprecate the variation with the lowercase i. This is because the file names would only change in case, which is a problem in most operation systems. I suggest we release this in any next version and mention the rename explicitly in the release notes. The change is very easy to find and resolve, and it seems safe assume not many instances of this specific icon will be used in applications.


packages-proprietary/react-icons/src/LinkedIn.tsx
```diff
@@ -1,8 +1,13 @@
...
-const SvgLinkedin = (props: SVGProps<SVGSVGElement>) => (
+const SvgLinkedIn = (props: SVGProps<SVGSVGElement>) => (
   <svg aria-hidden="true" focusable="false" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" {...props}>
     <path d="M20.097 2.5H3.903c-.775 0-1.403.628-1.403 1.403v16.194c0 .775.628 1.403 1.403 1.403h16.194c.775 0 1.403-.628 1.403-1.403V3.903c0-.775-.628-1.403-1.403-1.403M8.163 18.687H5.306V9.612h2.857v9.074M6.735 8.354c-.909 0-1.646-.692-1.646-1.64 0-.947.737-1.64 1.646-1.64.91 0 1.647.693 1.647 1.64 0 .948-.737 1.64-1.647 1.64m11.958 10.34h-2.856v-4.956c0-1.462-.621-1.913-1.424-1.913-.847 0-1.678.637-1.678 1.95v4.92H9.878V9.618h2.747v1.258h.037c.276-.558 1.242-1.512 2.716-1.512 1.594 0 3.316.946 3.316 3.717z" />
   </svg>
 )
-export default SvgLinkedin
+
+/**
+ * @name This icon was previously spelled als ‘Linkedin’, with a lowercase ‘i’.
```
alimpens, line 10: [This is not what the name tag should be used for](https://jsdoc.app/tags-name)

packages-proprietary/react-icons/src/index.ts
```diff
@@ -177,7 +177,7 @@ export { default as LightningFillIcon } from './LightningFill'
 export { default as LineChartDownIcon } from './LineChartDown'
 export { default as LineChartUpIcon } from './LineChartUp'
 export { default as LinkIcon } from './Link'
-export { default as LinkedinIcon } from './Linkedin'
+export { default as LinkedInIcon } from './LinkedIn'
```
alimpens, line 180: This is currently a breaking change. Shouldn't we export both spellings, with a deprecation warning for `LinkedinIcon`?

Handling the SVG in `assets` is a little trickier though.

VincentSmedinga: Alright, let’s travel the royal road. I reverted the breaking change to the name of the SVG icon. The `generate` script doesn’t overwrite the file of the React icon – probably exactly because the file system is working case-insensitively. So we can keep the deprecation and the improved name of the React components. And I’ll create a ticket to rename the asset just before we publish our next major release. Is that acceptable?

PR #2232, opened 2025-10-03
chore: Add automated visual regression tests

Evi-2003:
# Describe the pull request

## What

Integrate [Chromatic](http://chromatic.com) tests with the Design System components (for now only Button and IconButton).

## Why

To prevent accidental changes on components and improve our workflow. 

## How

(How were these changes implemented? Provide a brief overview of the approach taken and any key considerations.)

We only include stories in Chromatic with the name "Test", so we test more variants within one story. This story is hidden from the Storybook-interface by using `["!dev", "!autodocs"]`. 

[Test Cases Button Component](https://designsystem.amsterdam/demo-chromatic-tests/?path=/story/components-buttons-button--test)

image

[Test Cases Icon Component](https://designsystem.amsterdam/demo-chromatic-tests/?path=/story/components-buttons-icon-button--test)

image

--

I've also added an GitHub Action in this PR that will run Chromatic on each pull-request. Within Github we can make this a requirement to succeed before merging.

## Checklist

- [x] Add or update documentation (separate [pull request](https://github.com/Amsterdam/design-system/pull/2250))
- [x] Add or update stories

## Additional notes

### DIFF
--- a/plop-templates/storybook.test.stories.tsx.hbs
+++ b/plop-templates/storybook.test.stories.tsx.hbs
@@ -18,6 +18,6 @@ export default meta
 type Story = StoryObj<typeof meta>
 
 export const Test: Story = {
-  // NOTE: When adding new variants, ensure they are included in the test story's render function for visual regression testing.
+  // NOTE: Add all variants of the component to this story for visual regression testing.
   tags: ['!dev', '!autodocs'],
 }


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: pnpm/action-setup@v2
+        with:
+          version: latest
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Run Chromatic
+        uses: chromaui/action@latest
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: storybook
```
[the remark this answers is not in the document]
Evi-2003: Thanks fixed it :-)

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,21 @@
+name: "Chromatic"
+
+on: push
+
+jobs:
+  chromatic:
+    name: Run Chromatic
```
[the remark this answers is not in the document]
Evi-2003: Fixed it, this leaves a new comment after each commit when you've not fixed it it yet

README.md
```diff
@@ -36,6 +36,12 @@ This project is part of a community initiative to use NL Design System component
 
 Join the `#nl-design-system` Slack via [praatmee.codefor.nl](https://praatmee.codefor.nl) for more information and to learn about things happening in our open-source community.
 
+## Thanks
+
+<a href="https://www.chromatic.com/"><img src="https://user-images.githubusercontent.com/321738/84662277-e3db4f80-af1b-11ea-88f5-91d67a5e59f6.png" width="153" height="30" alt="Chromatic" /></a>
+
+Thanks to [Chromatic](https://www.chromatic.com/) for providing the visual testing platform that helps us review UI changes and catch visual regressions.
+
```
Evi-2003, line 44: This is needed for the [Opensource Licensing of Chromatic](https://www.chromatic.com/pricing)

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,29 @@
...
+          node-version: 22.x
+      - uses: pnpm/action-setup@v2
+        with:
+          version: latest
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Run Chromatic
+        uses: chromaui/action@latest
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
Evi-2003, line 28: This token is setup within GitHub Secrets of the repo

storybook/src/components/IconButton/IconButton.stories.tsx
```diff
@@ -88,45 +96,46 @@ export const TestCases: Story = {
...
+                <IconButton
+                  {...args}
+                  className={variant === 'hovered' ? 'hover' : undefined}
+                  color={colorVariants.includes(variant) ? (variant as IconButtonProps['color']) : undefined}
+                  data-testid={`${variant}-${size}`}
+                  disabled={variant === 'disabled'}
+                  key={`${variant}-${size}`}
+                  // @ts-expect-error: We can pass arguments to the fn function from storybook/test. Only the typing gets wrong as we need to pass it via the args of the story.
+                  // The component onClick only accepts a MouseEventHandler without arguments.
+                  onClick={() => args.onClick?.({ variantSize: `${variant}-${size}` })}
```
Evi-2003, line 129: I don't think is the best way to do this. 
1. We pass the `fn`-function from storybook/test which accepts arguments via the story arguments. 
2. Within the story's render we need to pass variables to that function, that works, but there is a typing error from the typing of the component. Which says that it does not allow passing arguments to the function. 
3. I've tried different solutions but couldn't get the typing right. 

If there is a better way to solve this i would like to know about it.

Evi-2003: This is not relevant anymore, as we are not doing integration tests on buttons anymore.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,29 @@
+name: "Test"
+
+on: push
```
alimpens: I think we generally want to run this in the context of a pull request, not on every push, right? So `on: pull_request`?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22.x
+      - uses: pnpm/action-setup@v2
+        with:
+          version: latest
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
alimpens: You can copy these steps from [our `lint-test` workflow](https://github.com/Amsterdam/design-system/blob/develop/.github/workflows/lint-test.yml#L15), they do the same thing.

.github/workflows/lint-test.yml
```diff
@@ -54,4 +54,4 @@ jobs:
 
       - name: "Continuous Integration: test"
         run: |
-          pnpm run --if-present test
+          pnpm run --if-present --filter='!./storybook' test # Storybook is excluded as we have a separate action for the Storybook tests
```
alimpens: We can get rid of this change if we name the script to kick off the Chromatic tests `chromatic`, no?

plop-templates/storybook.test.stories.tsx.hbs
```diff
@@ -0,0 +1,25 @@
...
+  component: {{pascalCase name}},
+  args: {
+    children: 'Nieuw component',
+  },
+} satisfies Meta<typeof {{pascalCase name}}>
+
+type Story = StoryObj<typeof meta>
+
+export const Test: Story = {
+  render: () => <{{titleCase name}}/>,
```
alimpens: We don't need this render function for all components I think, we don't have to add it here imo.

storybook/src/components/Button/Button.test.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+const states = ['default', 'disabled', 'hover', 'icon']
+
+export const Test: Story = {
+  parameters: {
+    pseudo: {
+      hover: ['.hover'],
+    },
+  },
+  render: () => (
+    <table>
```
alimpens: This is a bit of a nitpick, but: you're using a `table` here for layout purposes, which you generally shouldn't do. A `table` is for tabular data. 

It's ok here because we only use it for visual tests, but be aware that you should generally avoid this.

plop-templates/storybook.test.stories.tsx.hbs
```diff
@@ -0,0 +1,25 @@
...
+
+const meta = {
+  title: '{{titleCase name}}',
+  component: {{pascalCase name}},
+  args: {
+    children: 'Nieuw component',
+  },
+} satisfies Meta<typeof {{pascalCase name}}>
+
+type Story = StoryObj<typeof meta>
```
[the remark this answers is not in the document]
Evi-2003: But it is used? 🤖

.github/workflows/lint-test.yml
Evi-2003: Yes, that's right. @VincentSmedinga should we only name the script `chromatic` without a `test` script? Then i can revert this change.

plop-templates/storybook.test.stories.tsx.hbs
Evi-2003: My idea here was that if a component is setup, all variants should be added in the test story.  This was to hint that it can be added there. 

Or would it be better to just remove that line, and add a comment that all variants should be added?

storybook/chromatic.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "buildCommand": "pnpm run build",
+  "outputDir": "dist/",
```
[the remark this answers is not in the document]
Evi-2003: Within the build command we change the output dir to `dist/`. Which is not the default, Chromatic needs to know were the build is saved, otherwise it gives an error. 
The dir here is not for the build, but for Chromatic to find the build.

.github/workflows/lint-test.yml
VincentSmedinga: Well, alright then 😅

storybook/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "build": "storybook build --output-dir dist/ --config-dir config/ --quiet && touch dist/.nojekyll",
+    "test": "npx chromatic",
```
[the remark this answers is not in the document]
alimpens: They're very different types of tests. `test` runs Jest unit tests, `chromatic` builds Storybook and publishes it to Chromatic. That isn't very fast, it would be quite annoying to always have that run when you run the unit tests.

Thinking about it, maybe we don't want to run Chromatic locally at all, only in the pipeline? What do you think @Evi-2003 ?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  chromatic:
+    name: Run Chromatic
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out branch
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
alimpens: Yeah, that makes sense

VincentSmedinga (dismissed):
Looks (very) good to me! One comment by @alimpens to go, I think?

plop-templates/storybook.test.stories.tsx.hbs
```diff
@@ -0,0 +1,23 @@
...
+  ...{{camelCase name}}Meta,
+  title: '{{titleCase name}}',
+} satisfies Meta<typeof {{pascalCase name}}>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Test: Story = {
+  // NOTE: When adding new variants, ensure they are included in the test story's render function for visual regression testing.
```
alimpens: This template is only used when you set up a new component. When you've done that, this comment will probably get deleted. So maybe we should adapt the time / phrasing of the comment to that point in time?
Something like:

```suggestion
  // NOTE: Add all variants of the component to this story for visual regression testing.
```

PR #2258, opened 2025-10-21
chore: Add linter rule to make sure story controls are inferred from the component

Evi-2003:
# Describe the pull request

## What

This make sure the intersection types are sorted object first.

## Why

Currently the controls are sometimes inferred from the component, and sometimes they are not.
Example:
CleanShot 2025-10-21 at 15 09 56@2x
The color control should show a select with the color options, like how it's done with Avatar. It should automatically be recognised that it should be a select.

Within the Avatar story (which works) we are not specifying it should be a `Select` control. It infers it from the component. 
For Badge it doesn't automatically infers it, because the props are not object-first there. 

I'm not completely sure why it only works object-first, maybe i will create a issue within the Storybook-repo for this. 

[Production](https://designsystem.amsterdam/?path=/docs/components-feedback-badge--docs)
[Feature build](https://designsystem.amsterdam/demo-sort-type-constituents/?path=/docs/components-feedback-badge--docs)

## Checklist

- [x] Add or update stories
- [x] Start the PR title with a Conventional Commit prefix, [as explained here](https://github.com/Amsterdam/design-system/blob/main/documentation/publishing.md?plain=1#

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,34 @@
+name: "Test"
+
+on: pull_request
+
+permissions:
+  contents: read
+
+jobs:
+  chromatic:
+    name: Run Chromatic
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out branch
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Install pnpm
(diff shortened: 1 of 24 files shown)


alimpens:
Nice work! I think I've figured out why this breaks. It seems to happen when we use a prop with the same name as a native HTML attribute. For Badge, Blockquote, DescriptionList and Heading for example that is `color`.

When we define `HTMLAttributes<HTMLElement>` before our own props, Storybook (or react-docgen-typescript) seems to pick the `color` defined in that generic type over our own definition, resulting in missing descriptions etc.

PR #2297, opened 2025-11-21
feat(Field, Field Set)!: Change spacing between elements

alimpens:
# Describe the pull request

This PR is a breaking change, because we remove the use of `ams-field-gap` in the Field CSS. Setting a custom value there will not work anymore.

While technically not breaking, we should also communicate to our users that they have to remove the utility classes that we previously recommended. 

## What

It changes the spacing between the elements of Field and Field Set. 

## Why

This was redesigned, in order to have a stronger visual coupling between Label / legend and descriptions, and Error Message and inputs. It also takes into account nested Fields and Field Sets. 

## How

Changed the `gap` property in Field to a margin between all elements, and used fancy CSS selectors to target some specific elements. Field Set already used a margin, updated that in a similar way. 

## Checklist

- [x] Add or update stories

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/page-header.compact.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/page-header.compact.tokens.json
@@ -4,6 +4,9 @@
       "brand-name": {
         "font-size": { "value": "{ams.typography.heading.2.font-size}" },
         "line-height": { "value": "{ams.typography.heading.2.line-height}" }
+      },
+      "mega-menu": {
+        "padding-block": { "value": "0" }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


packages/css/src/components/field/field.scss
```diff
@@ -8,7 +8,18 @@
   break-inside: avoid;
   display: flex;
   flex-direction: column;
-  gap: var(--ams-field-gap);
```
VincentSmedinga: This removal stops a possible custom value for the `gap` token from working. This PR deprecates the token, but we promised that deprecated features will continue to work for two more major releases.

Should we restore this line of CSS (and add a comment) and set our own value of the token to `initial`?

If so, we should also find a way to remember that we could or should change `display: flex` back to `block`, as we no longer use or need Flexbox and `block` is the default.

If not, we can revert to `display: block` now.

VincentSmedinga (approved):
Pretty cool to adopt this approach! Composition components applying things automatically so people get what they need without breaking a sweat 🪄

PR #2298, opened 2025-11-21
feat(Page Header)!: Add vertical space to Mega Menu in Spacious Mode

VincentSmedinga:
# Describe the pull request

## Breaking change

Although his change does not change the technical API, it may make the visual appearance of your website incorrect.

Take the following action when upgrading:

1. Remove any `padding*` properties (or equivalent custom styling) applied to the outer container of the Page Header’s collapsible menu, typically a Grid component. [See these examples.](https://designsystem.amsterdam/?path=/docs/components-containers-page-header--docs)

## What

Adds vertical space to Mega Menu in Spacious Mode.

Updates the row gap of the navigation part to the same size, so that it replaces the bottom padding of Page Header exactly if the collapsible menu is open. This prevents strange visual effects. It increases the space between the Logo and the Header Menu as well, if the latter wraps to a second line, but that doesn’t happen very often and it looks fine as well.

I experimented with various values. I think large space at both top and bottom looks best. Page Header itself has a large vertical padding itself as well, and a large gap between itts initial content and the mega menu. Makes the space around the Mega Menu double that of the other places. Looks very balanced to me, at all window widths, with various Mega Menu content, and for internal websites (with a Menu) as well.

## Why

So that the menu can be clearly distinguished from the rest of the content.

## How

Add a large vertical padding, override it in Compact Mode, remove it from markup in an example.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out branch
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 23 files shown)


packages-proprietary/tokens/src/components/ams/page-header.compact.tokens.json
```diff
@@ -4,6 +4,9 @@
       "brand-name": {
```
alimpens: Do we want to mark this as a breaking change? It technically isn't, but users will probably have to remove some padding from their mega menus.

alimpens (approved):

PR #2302, opened 2025-11-26
feat(ProgressList): Add statuses

Evi-2003:
# Describe the pull request

## What

Adds statuses to the Progress List.

## How

While developing the current-status, I noticed that the marker wasn’t perfectly round, especially when an outline was applied. This was caused by sub-pixel rendering issues. After a small refactor, removing the ::after element, the problem occurred much less frequently.

However, the marker still isn’t perfectly circular at certain resolutions. For example, in Safari on my MacBook it looked fine, but moving Safari to my 1440p monitor caused slight distortion. The issue is subtle and only noticeable when zoomed in.

Using round() fixes the problem, but it isn’t widely supported yet. The fallback uses the original value, which can still produce imperfect circles at some resolutions. Since the marker size is based on the font size, there doesn’t seem to be another reliable way to fix this.

--

I experimented with different ways of applying the border/outline. I initially used outline, but received feedback to be careful with it, as it can overwrite the focused state.

I then tried variations with `border` and `box-shadow`, `border-image`, and `background-image`. Using `border` with `box-shadow` seemed to work best, but it didn’t behave correctly in Safari. Safari handles box-shadows differently, causing the outer outline to be uneven.

I’ve now reverted to using `outline`, since the marker itself isn’t focusable, and it works across all browsers. 

--

For a next iteration of the marker we probably want to remove the checkmark for completed items, using only filled/unfilled circles. 

## Checklist

- [x] Add or update stories

## Additional notes

[Figma](https://www.figma.com/design/lTZ7XteXPym8AsNrrSoL2o/Finals--Amsterdam.nl?node-id=1-25616&p=f&t=NlWtlajNgKcARiMB-0)
[Preview](https://designsystem.amsterdam/?path=/story/components-containers-progress-list--default&args=headingLevel:2)

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/progress-list.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/progress-list.tokens.json
@@ -56,13 +56,16 @@
       },
       "shape": {
         "heading-2": {
-          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-2-font-size) * 0.8)" }
+          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-2-font-size) * 0.8)" },
+          "block-size": { "value": "calc(var(--ams-progress-list-marker-heading-2-font-size) * 0.8)" }
         },
         "heading-3": {
-          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-3-font-size) * 0.8)" }
+          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-3-font-size) * 0.8)" },
+          "block-size": { "value": "calc(var(--ams-progress-list-marker-heading-3-font-size) * 0.8)" }
         },
         "heading-4": {
-          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-4-font-size) * 0.8)" }
+          "inline-size": { "value": "calc(var(--ams-progress-list-marker-heading-4-font-size) * 0.9)" },
+          "block-size": { "value": "calc(var(--ams-progress-list-marker-heading-4-font-size) * 0.9)" }
         }
       },
       "upcoming": {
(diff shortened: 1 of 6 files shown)


packages/css/src/components/progress-list/progress-list.scss
```diff
@@ -17,6 +17,8 @@
   @include hyphenation;
   @include text-rendering;
   @include reset-ol;
+
+  list-style: none;
```
VincentSmedinga: This is already in the reset mixin.

```suggestion
```

packages/css/src/components/progress-list/progress-list.scss
```diff
@@ -45,62 +47,119 @@
...
+  align-self: center;
+  inline-size: 100%;
 }
 
 // Connector
 
 .ams-progress-list__connector {
   align-self: stretch;
-  background-color: var(--ams-progress-list-connector-background-color);
+  background-color: var(--ams-progress-list-step-connector-background-color);
```
VincentSmedinga: This connector has a background colour, others reset it and have a border. Can we use one of both in all situations? This will probably simplify it a bit further, and is more flexible for the future.

Also, we can’t be sure that the background will be white, so we cannot use that to reset a colour.

packages-proprietary/tokens/src/components/ams/progress-list.tokens.json
```diff
@@ -1,35 +1,99 @@
...
+          "marker": {
+            "background-color": { "value": "{ams.color.feedback.success}" },
+            "border": { "value": "none" }
+          }
+        },
+        "connector": {
+          "background-color": { "value": "{ams.color.background}" },
+          "border-color": { "value": "{ams.color.separator}" },
+          "border-style": { "value": "dashed" },
+          "border-width": { "value": "calc({ams.border.width.m} / 2)" }
```
VincentSmedinga: Can’t we apply the border to one side instead of halving a token? This also removes the border from the top and bottom, which is what we want.

VincentSmedinga (approved):
Sweet! 🚦

PR #2312, opened 2025-12-02
chore: Use pseudo addon only in dev/test environments

Evi-2003:
# Describe the pull request

## What

This makes sure the add-on for the pseudo-states is only loaded in the dev environment, or within Chromatic.

## Why

It causes issues with the usage of Skip Link, as it forces pseudo states on elements. 
The pseudo add on does not support the ability to be disabled/enabled when we need it, as it is [missing](https://github.com/storybookjs/storybook/issues/24539) a [param key](https://github.com/storybookjs/storybook/blob/next/code/addons/pseudo-states/src/manager.ts).

## How

By using `process.env` we can make sure that we are in the development environment.

Chromatic offers a [helper](https://www.chromatic.com/docs/ischromatic/) to check if we’re in the Chromatic environment. However, this doesn’t work because it’s intended for use in stories not config files.  On the same documentation page for isChromatic they explain that you can pass `IS_CHROMATIC` to the chromatic command. This works for tests but not for the Storybook build where we need to check if it’s Chromatic.

Therefore, we need to pass it via the Storybook build command. That’s why I’ve created a second Storybook build command.

## Checklist

- [X] Add or update stories

## Additional notes


storybook/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "scripts": {
     "build": "storybook build --output-dir dist/ --config-dir config/ --quiet && touch dist/.nojekyll",
+    "build:chromatic": "IS_CHROMATIC=true storybook build --output-dir dist/ --config-dir config/ --quiet && touch dist/.nojekyll",
```
alimpens: Can we reuse the `build` script here? So something like `"build:chromatic": "IS_CHROMATIC=true npm run build"`?

This would prevent the 2 scripts from drifting apart over time.

alimpens (approved):

PR #2538, opened 2026-04-09
feat(Progress List)!: Make collapsing and expanding steps optional

VincentSmedinga:
# Make collapsing and expanding steps optional

## Links

- [Story](https://designsystem.amsterdam/demo-DES-1769-progress-list-not-collapsible/?path=/story/components-containers-progress-list--not-collapsible)

## What

Adds a variation of Progress List that doesn’t collapse its steps – and restores it to be the default behaviour.

Steps render plain headings instead of buttons, which removes the chevron icons, blue interactive colour, and expand/collapse behaviour.

This is a (very small) breaking change because we made Progress List collapsible in the latest release, without an option for a static appearance. People who have implemented that must now add the `collapsible` prop to get the default behaviour they received until now.

## Why

Progress List can be used for smaller pieces of content, where collapsing is not necessary. The chevrons and colours bring visual overhead that can now be avoided.

Mijn Amsterdam already requested this.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


timvanoostrom (approved):

packages/react/src/ProgressList/ProgressList.test.tsx
```diff
@@ -126,6 +126,44 @@ describe('ProgressList', () => {
...
+        <ProgressList.Step heading="one">Content</ProgressList.Step>
+        <ProgressList.Step heading="two">Content</ProgressList.Step>
+      </ProgressList>,
+    )
+
+    const list = screen.getByRole('list')
+    const onKeyDown = vi.fn()
+
+    list.addEventListener('keydown', onKeyDown)
+    list.focus()
```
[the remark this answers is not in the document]
VincentSmedinga: Updated the test.

packages/react/src/ProgressList/ProgressListStep.test.tsx
```diff
@@ -423,6 +423,58 @@ describe('ProgressListStep', () => {
...
+  it('does not render the chevron icon when collapsible is false', () => {
+    const { container } = render(
+      <ProgressList collapsible={false} headingLevel={3}>
+        <ProgressList.Step heading="Test Step">Content</ProgressList.Step>
+      </ProgressList>,
+    )
+
+    const heading = container.querySelector('.ams-progress-list__heading')
+    const icon = heading?.querySelector('.ams-progress-list__icon')
+
```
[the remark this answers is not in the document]
VincentSmedinga: Updated the test.

RubenSibon (changes requested):
Everything seems good to me apart from the Chromatic tests. Text style is incorrect. Please, adjust that.

VincentSmedinga:
> Everything seems good to me apart from the Chromatic tests. Text style is incorrect. Please, adjust that.

Yes, updated to use `Paragraph` instead of `<p>`.

packages/react/src/ProgressList/ProgressList.tsx
```diff
@@ -19,6 +19,11 @@ export const progressListHeadingLevels = headingLevels.filter((level) => level !
 type ProgressListHeadingLevel = (typeof progressListHeadingLevels)[number]
 
 export type ProgressListProps = {
+  /**
+   * Whether the steps can be expanded and collapsed.
+   * @default true
```
VincentSmedinga: I think the feature to expand and collapse the steps of a Progress List should be an option, not the default. Hiding content shouldn’t be Plan A.

Proposed to toggle this boolean in our Teams channel, awaiting feedback.

RubenSibon: I gave my feedback: I agree that it's better to show the simpler and fully visible variant by default.

RubenSibon: After you've flipped the default variant I will give my approval (if you did the flipping correctly, that is 😉).

VincentSmedinga: Updated already after one supporting vote – this really seems like the natural approach.

VincentSmedinga: > After you've flipped the default variant I will give my approval (if you did the flipping correctly, that is 😉).

Done! Note that I’ve reordered all documentation, stories and tests to lead with the new default (not collapsed) everywhere.

VincentSmedinga: Everyone agreed, let’s do this.

packages/react/src/ProgressList/ProgressListStep.tsx
```diff
@@ -75,23 +79,30 @@ export const ProgressListStep = forwardRef(
...
-            </button>
+            {collapsible ? (
+              <button
+                aria-controls={panelId}
+                aria-expanded={!isCollapsed}
+                className="ams-progress-list__button"
+                onClick={() => {
+                  const next = !isCollapsed
+                  setIsCollapsed(next)
+                  onToggle?.(!next)
```
RubenSibon: Actually one nitpick (not a dealbreaker): despite quite some years of experience with programming I still find inversions/negations difficult to understand. Let me describe in pseudocode what happens her:

- Constant `next` is set to the inverse of `isCollapsed`:
  - `isCollapsed` == `true` => `next` == `false`;
  - `isCollapsed` == `false` or `undefined` (coercion) => `next` == `true`.
- The value of `next` is passed to `setIsCollapsed`.
- The value of `next` is inversed in-place and passed to an optional `onToggle` method.

If I think about it long enough I understand what's happening here, but it is not immediately obvious to me. It would help me and possibly other developers if there was a comment that explains what this does: why are we setting the internal state to the inverse of `isCollapsed` and why does the wrapping Progress List receive that value, but again negated?

RubenSibon (dismissed):
Approved, but with one nitpick about easy to understand code.

packages/react/src/ProgressList/ProgressListStep.tsx
VincentSmedinga: I think you’re right. We can remove the intermediate variable:

```ts
onClick={() => {
  setIsCollapsed(!isCollapsed)
  onToggle?.(isCollapsed)
}
```

Is that clear enough, or would you then still like comments?

And should we extract a function towards `onClick={handleClick}`?

RubenSibon: Yeah, this is clearer. And I'm in favour of extracting a function. That function can then have a comment explaining why the parent gets the value as is and the internal state needs to have the inversion.

VincentSmedinga: Cool, it is so.

packages/react/src/ProgressList/ProgressListStep.tsx
```diff
@@ -46,19 +50,25 @@ export const ProgressListStep = forwardRef(
...
+    const { collapsible, headingLevel } = useContext(ProgressListContext)
     const [isCollapsed, setIsCollapsed] = useState(defaultCollapsed ?? status === 'completed')
 
     const iconSize = `heading-${headingLevel}` as IconProps['size']
     const panelId = useId()
 
+    // Toggles the local collapsed state and emits the new state
+    const handleClick = () => {
+      setIsCollapsed(!isCollapsed)
+      onToggle?.(isCollapsed)
```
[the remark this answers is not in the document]
RubenSibon: Haha, this looks more complicated again, but I guess it’s less brittle as they said? Can you look into it @VincentSmedinga?

storybook/src/components/ProgressList/ProgressList.docs.mdx
```diff
@@ -32,6 +32,7 @@ Set `collapsible` to `true` to let users expand and collapse the content of each
 Current and upcoming steps then start expanded.
 Completed ones are initially collapsed.
 Use the `defaultCollapsed` prop on a step to override the default behaviour.
+The `onToggle` callback emits whether the Step is expanded.
```
RubenSibon: Can you rephrase this a bit? The word “whether” is a bit confusing because it does not suggest that the emitting happens every time the expanded state changes.

Maybe:

```suggestion
The `onToggle` callback emits the expanded state whenever it changes.
```

packages/react/src/ProgressList/ProgressListStep.tsx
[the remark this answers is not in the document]
VincentSmedinga: Back to what it was:

> The original isn't broken — it's correct by coincidence. Reading it cold, it looks like a bug: onToggle is called with isCollapsed while a "new state" is described in the comment. It works only because !nextCollapsed === currentCollapsed. So yes, it's brittle in the readability sense — a future maintainer will likely "fix" the apparent bug and break it.
> 
> But the proposed replacement trades one problem for another: it puts a side effect (onToggle) inside a state updater function. React docs explicitly warn against this — updater functions should be pure. In StrictMode, React intentionally calls updaters twice in development to surface impurity, which would fire onToggle twice per click. It also runs during render-phase work in some cases.
> 
> I'd suggest a third option that fixes the readability without the side-effect-in-updater pitfall:
> 
> ```
> // Toggles the local collapsed state and emits the new expanded state
> const handleClick = () => {
>   const nextIsExpanded = isCollapsed
>   setIsCollapsed(!isCollapsed)
>   onToggle?.(nextIsExpanded)
> }
> ```
> 
> This reads correctly at a glance, names the value being emitted, and keeps the setter call simple. Since handleClick only fires from a user click (not in rapid succession or from another effect), the updater-function form isn't buying you anything here — isCollapsed from the closure is already the right value.

RubenSibon (dismissed):
Approved!

You can decide if you want to apply the suggestion for that line in the docs.

VincentSmedinga:
> Approved!
> 
> You can decide if you want to apply the suggestion for that line in the docs.

Thanks, yes your line was better. The commit dismissed your approval, so you’ll have to do the merging yourself :)

RubenSibon (approved):

PR #2539, opened 2026-04-09
feat(Description List): Use container width instead of viewport to switch orientation

RubenSibon:
# Use container width instead of viewport to switch orientation of Description List

## What

Switches the Description List between vertical and horizontal orientations using a container query on the `ams-query-container-inline-size` query container. The component now responds to the inline size of its nearest query container (typically a Grid Cell) instead of the viewport.

## Why

When a Description List sits inside a narrow Grid Cell (e.g. 4 of 12 columns), the viewport can be wide enough to trigger a viewport-based horizontal layout while the available inline size is too small — causing cramped or broken content. A container query lets the component respond to the actual width of its ancestor.

## How

- **Description List CSS**: A `@container ams-query-container-inline-size (inline-size >= 32rem)` rule applies the horizontal grid layout. Below `32rem`, or outside any query container, the component renders vertically.
- **Sass variable**: CSS custom properties can't appear inside `@container` conditions, so the breakpoint is a Sass variable (`$ams-description-list-breakpoint: 32rem`).
- **Description List README**: Updated to describe the orientation behaviour relative to the nearest query container (typically a Grid Cell, but also a Dialog or Page), linking to the Query Container utility docs.
- **Stories**: Added an "Orientation" story showing side-by-side narrow and wide Grid Cells, placed first under "Examples". Updated the test story to include a narrow query container context for Chromatic.

## Checklist

- [x] Add or update stories

## Additional notes

- Related ticket: DES-1761
- [Storybook demo](https://designsystem.amsterdam/demo-DES-1761-dl-vertical-alignment-in-narrow-container/?path=/docs/components-text-description-list--docs)
- Container queries have [Baseline availability](https://caniuse.com/css-container-queries) across all modern browsers.
- The container query targets the named container `ams-query-container-inline-size` and uses `inline-size` (logical property) instead of `width`.


packages/css/src/components/description-list/README.md
```diff
@@ -6,8 +6,8 @@ A collection of terms and their descriptions.
 
 ## Design
 
-In a narrow window, descriptions appear below their term.
-After that, they display in a two-column layout.
+In a narrow container, descriptions appear below their term.
+Once the container reaches 32rem (or wider), they display in a two-column layout.
```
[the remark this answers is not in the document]
RubenSibon: Is it better to be specific about what container we have currently specified as the only possible containment context? Or is it better to keep it vague? 🤔

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
```diff
@@ -51,6 +51,13 @@ This ensures the colour of the text provides enough contrast.
 
 <Canvas of={DescriptionListStories.InverseColour} />
 
+### Narrow container
+
+The Description List uses container queries to switch between a vertical and horizontal layout.
+When its parent container is narrower than the breakpoint, terms and descriptions stack in a single column.
```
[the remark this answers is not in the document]
RubenSibon: Again, should we be specific or generic? Not sure yet.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -112,4 +125,19 @@ export const InverseColour: Story = {
...
+export const NarrowContainer: Story = {
+  parameters: {
+    layout: 'fullscreen',
+  },
+  render: (args) => (
+    <Grid>
+      <Grid.Cell span={{ narrow: 4, medium: 4, wide: 8 }}>
+        <DescriptionList {...args} />
+      </Grid.Cell>
+      <Grid.Cell span={{ narrow: 4, medium: 4, wide: 4 }} />
```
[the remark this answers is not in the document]
RubenSibon: I've adjusted the cell spans. Now the DL is in the column layout when the containment context's inline size is narrow or wide, but in the row orientation when its medium.

packages-proprietary/tokens/src/components/ams/grid.tokens.json
```diff
@@ -92,6 +92,12 @@
         "background-color": {
           "$value": "initial"
         },
+        "container-name": {
+          "$value": "grid-cell"
+        },
+        "container-type": {
+          "$value": "inline-size"
+        },
```
VincentSmedinga, line 136: The fact that Grid Cell now declares a containment context is a notable feature. It deserves its own mention in the change log and some documentation in the Grid docs. Let’s extact it into a separate PR.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -3,11 +3,15 @@
...
 
-@use "../../common/breakpoint" as *;
 @use "../../common/hyphenation" as *;
 @use "../../common/resets" as *;
 @use "../../common/text-rendering" as *;
 
+// Container query breakpoint
+// - Note: CSS Custom Properties can’t be used in `@container` queries.
+// - Keep aligned with Grid inline padding for the medium breakpoint: `37.5rem - (2 × 2.75rem) = 32rem`.
+$ams-description-list-cqi-medium: 32rem;
```
VincentSmedinga: We discussed using `cqi` in the name, but that was for the tokens. Now that we can’t use those, I’m wondering if this is still the best naming.

The ‘inline-size’ part isn’t too obvious (hidden in the `i` of `cqi`) and the ‘medium’ part doesn’t really apply here; we don’t have a ‘wide’ as well – DL has two appearances, one below this threshold and one above it.

So maybe `$ams-description-list-breakpoint` is all we need for now?

1. The ‘breakpoint’ term is well-known, even though it’s not in the spec and MDN doesn’t mention it, and people expect it to operate on the width / inline size.
2. We only need one turning point as we have only two appearances and don’t expect a third.
3. It matches our existing viewport breakpoints `$ams-breakpoint-*`. The difference in namespacing (none vs. description list) is a good indication, and the ones for media queries do deserve a specialisation to medium and wide.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -3,11 +3,15 @@
...
  */
 
-@use "../../common/breakpoint" as *;
 @use "../../common/hyphenation" as *;
 @use "../../common/resets" as *;
 @use "../../common/text-rendering" as *;
 
+// Container query breakpoint
+// - Note: CSS Custom Properties can’t be used in `@container` queries.
+// - Keep aligned with Grid inline padding for the medium breakpoint: `37.5rem - (2 × 2.75rem) = 32rem`.
```
VincentSmedinga: That relationship is a nice find – although not entirely accurate: at the medium breakpoint, the horizontal padding of Grid (`space-xl`) is

`1.8214rem + 2.1429vw = 1.8214 * 16 + 600 * 2.1429 / 100 = 42px`.

The perspective is different too. By subtracting the horizontal grid padding from the viewport width, you essentially reason about a Cell that covers all columns of the Grid. While the idea behind the container query is that the Description List changes its layout if there is enough space, say at around 400-500px, where we take the best reading length for a line of text into account, in combination with either one or two columns.

Adapted from the comments for our viewport breakpoints:

```suggestion
/**
 * CSS Custom properties cannot be used in container queries.
 * We resort to Sass variables for our breakpoint values.
 */

/** The container width at which Description List toggles its orientation. */
```

packages/css/src/components/description-list/description-list.scss
```diff
@@ -20,7 +24,7 @@
   @include text-rendering;
 }
 
-@media (min-width: $ams-breakpoint-medium) {
+@container grid-cell (min-inline-size: #{$ams-description-list-cqi-medium}) {
```
VincentSmedinga: Let’s turn back from our plan to use a named container from the start.

Container queries resolve against the nearest ancestor with a matching container-type (or container-name), walking up the tree.

Setting `container-type: inline-size` on the Grid Cell is the right call, since that’s the actual available space the component has to work with. But we don’t need the name. We already require all components to be laid out on the Grid, so this is the first containment context that components will find. It’s also premature: we only set one containment context, so no distincion is needed yet. Let’s update as we learn.

packages/css/src/components/description-list/README.md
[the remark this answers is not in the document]
VincentSmedinga: Let’s keep it practical for now – I think we help our users best by explaining what the specific container is.

```suggestion
Description Lists have a vertical layout when their parent Grid Cell is less than `32rem` wide (typically 512 pixels).
Above this size, they change to a horizontal layout, placing the descriptions to the right of their terms instead of below them.
```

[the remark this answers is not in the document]
VincentSmedinga: Let’s also use a list to explain the design choices: one item for the above, one for the terms column width, one for the terms being bold.

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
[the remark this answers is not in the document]
VincentSmedinga: I think we needn’t repeat the layout behaviour once more. Let’s call this story ‘Vertical orientation’ and make it the first example. Give it a decorator that wraps the story in `<div style={{container-type: inline-size; max-inline-size: '31.99rem' }}`.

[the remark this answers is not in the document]
VincentSmedinga: Or, maybe even nicer, use the ‘resize’ approach that Action Group has. Then call it just ‘Orientation’? Or ‘Automatic orientation’?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -3,12 +3,20 @@
...
 
-import type { Meta, StoryObj } from '@storybook/react-vite'
+import type { Meta, StoryFn, StoryObj } from '@storybook/react-vite'
 
-import { Link, Paragraph, UnorderedList } from '@amsterdam/design-system-react'
+import { Grid, Link, Paragraph, UnorderedList } from '@amsterdam/design-system-react'
 import { descriptionListTermsWidths } from '@amsterdam/design-system-react/dist/DescriptionList/DescriptionList'
 import { DescriptionList } from '@amsterdam/design-system-react/src'
 
+const inSingleCellGrid = (Story: StoryFn) => (
```
VincentSmedinga: The Grid is no longer necessary after removing the `grid-cell` name from the container query.

Just a decorator `<div style={{container-type: inline-size; max-inline-size: '31.99rem' }}` like above, and for all the stories?

And maybe add the `: Decorator` type.

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
```diff
@@ -49,9 +49,14 @@ export const Test: Story = {
       </DescriptionList.Section>,
     ],
   },
-  render: (args) =>
-    renderComponentVariants(DescriptionList, {
-      args,
-    }),
+  render: (args) => (
```
VincentSmedinga: Let‘s remove this and add one with a `31.99rem` div to the existing array.

VincentSmedinga (commented):
Yes, this looks really nice. The first time in 30 years web dev that I’ll be using container queries, exciting :)

As discussed offline, we’re adding this feature into our architecture for the first time. Let’s split up the PR, take the smallest steps possible, and document things well.

RubenSibon:
The PR's base was temporarily changed to `feat/DES-1761-grid-cell-as-containment-context`. This should be switched back to `develop` once that PR is merged.

packages/css/src/components/description-list/description-list.scss
RubenSibon: Yes, I agree. I will change the last part to `-breakpoint`.

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
RubenSibon: Yeah, so the current issue we have to think about: how to deal with a combination of container and media queries? Or: can we always depend on components that do container queries to always be nested in an element that has a containment context?

## Writing down structured thoughts

### Scope

- Description List's (DL) single breakpoint (whether that's media or container) for its orientation switch
- Grid Cells that now always provide containment contexts
- Media queries on the viewport
- Page that MAY also provide a containment context (debatable)

### Questions

Current situation (at time of writing): DL's only switch to the horizontal orientation when they sit in a Grid Cell.

- Do we bring back media queries? Will that conflict with the container queries? Can that be handled elegantly and without adding too many lines of CSS to the build output?
- Do we have to decide that a DL MUST be placed in a Grid Cell at all times? In practice this will always be the case if our documentation is followed. Our stories are an exception because we generally do not place components in Grids there to keep the documentation simple and pure.
- Do we add a catch-all containment context on Page that DL can use to switch its orientation on the same 32rem breakpoint? In the case of DL the coincidence is that this breakpoint is close to the medium viewport breakpoint. Page will act as some kind of viewport to do faux media queries on, but with the caveat that Page has a maximum inline size. We can also decide on a component-by-component basis if it needs media queries or container queries. In the case of DL no media queries makes sense because its orientation should always be decided by its containment context, not the window's width. Will that be true for many other components?

### AI advice

I will forward above thoughts to an LLM for advice.

RubenSibon: Here is a reply from AI on highest effort after a few follow-up prompts (I haven't tested or critically read it yet):

# DL orientation switching: media vs. container queries

## Q1 — Should we reintroduce media queries for DL?

No. Two reasons:

- **Cascade conflict.** A media query targeting the viewport can't know whether a given DL is sitting in a narrow Grid Cell on a wide screen. If MQ says "horizontal at 768px" and the cell is 24rem wide, DL goes horizontal in a space that can't fit it. Any MQ would need to be a conditional fallback — "only apply when no containment context exists" — and CSS has no clean way to express that.
- **Intent drift.** The reason `description-list.scss` uses `@container` is that a DL's orientation is a function of *its own available space*, not the window. Reintroducing MQ contradicts that principle for a payload saving that doesn't matter after gzip.

## Q2 — Should we mandate that DL always sits in a Grid Cell?

No, but the concern that motivated this question is narrower than I first thought.

Container queries resolve against the **nearest ancestor with a containment context**. So for in-tree wrappers that simply fill their parent — Accordion panels, Tab panels, most Cards — the Grid Cell's measurement is still accurate, because the wrapper is the same width as the cell. Those wrappers don't need to become containment contexts for DL to lay out correctly.

A Grid-Cell mandate fails in three specific situations:

1. **Portaled components.** Dialog, Modal, Popover, Drawer, Toast — these render outside the Grid tree. For a DL inside a Dialog, the nearest containment context is nothing, so DL stays vertical regardless of how much room the dialog has.
2. **Wrappers that constrain width independently of the cell.** A Card with `max-inline-size: 20rem` inside a 60rem cell: DL queries the cell (60rem ≥ 32rem) → goes horizontal → has only 20rem of real estate → cramped. DL needs to re-ask the question against the wrapper's actual width.
3. **Wrappers that *widen* relative to the cell.** Breakout is the canonical case — it deliberately escapes the parent's inline size, so children that query the cell get the wrong answer.

None of (1)–(3) are solvable by a documentation mandate. They need the wrapper itself to establish containment.

## Q3 — Should Page act as a catch-all containment context?

Yes, but with a twist: use **named containers**, not an unnamed one.

The risk of an unnamed containment context on Page is leakage. Every unnamed `@container` query inside the page tree would resolve against Page by default, and components expecting to measure their immediate wrapper could get surprising matches.

Notice that `grid.scss` already does:

```scss
container-name: var(--ams-grid-cell-container-name);
container-type: var(--ams-grid-cell-container-type);
```

The system is already set up for named containers driven by tokens. Lean into that:

1. Pick a shared name, e.g. `layout`.
2. Have Grid Cell set that name (it already does via the custom property — just make the default token value `layout`).
3. Have Page establish the same named container as a fallback: `container-type: inline-size; container-name: layout`.
4. Change DL's query to `@container layout (min-inline-size: #{$ams-description-list-breakpoint})`.
5. Any future wrapper that needs to act as a layout context opts in to the same name.

That gives you:

- DL in a Grid Cell → measures the cell.
- DL in a Dialog / directly in Page body → measures Page (fallback).
- DL in a story without either → stays vertical, which is honest: there is no layout context to measure against. Stories can wrap in a simple `<div style="container-type: inline-size; container-name: layout">` if they want to demo the horizontal variant.
- Other `@container` queries elsewhere are unaffected because they use different names or remain unnamed.

On Page's `max-inline-size`: that's a feature, not a caveat. It means DL's breakpoint responds to the actual readable content width, not a phantom 2560px monitor. The coincidence with the medium viewport breakpoint is only a coincidence — the right mental model is "Page width ≈ content width," which is exactly what DL should query.

## Which other wrappers should also provide the `layout` container?

The criterion is **two conditions, both must hold**:

1. The component can realistically contain another component that uses container queries (rich-content wrapper, not just a holder for buttons/links/text).
2. The component establishes an inline size that differs from its parent — by narrowing, by widening, or by being outside the normal flow.

Applying that to the Storybook groups:

### Required (minimal load-bearing set)

| Component | Group | Reason |
|---|---|---|
| Alert | Feedback | Rich-content region, semantic boundary |
| Card | Navigation | Often constrained narrower than its cell |
| Dialog | Containers | Portaled — no ancestor containment without this |
| Page | Containers | Global fallback (see Q3) |
| Breakout | Layout | Widens relative to its parent |
| Column | Layout | Narrows relative to its parent |

### Optional safety net

Accordion, Tabs, Page Header, Page Footer, Spotlight, Progress List. These usually fill their parent, so the cell's measurement is already accurate. Adding containment anyway is defensible as insurance against future layout changes inside these wrappers, but it isn't strictly required.

### Edge case

Table. Putting containment on the `<table>` element doesn't help content inside cells — you'd want it on `<td>`/`<th>`. Worth a separate decision.

### Excluded from Layout group

- **Grid** itself — containment already lives on Grid Cell; duplicating it on `.ams-grid` would be redundant and potentially confusing.
- **Action Group**, **Row**, **Overlap** — these don't wrap rich content; they hold buttons or arrange items at their parent's width. No realistic DL use case inside them, and they don't constrain width independently.

## TL;DR

- Keep DL on `@container`. Don't bring back media queries.
- Don't mandate DL-in-Grid-Cell; it can't fix the portaled or constraining-wrapper cases anyway.
- Introduce a named `layout` container. Grid Cell and Page both provide it. DL queries it by name.
- Add the same named container to the minimal load-bearing set: Alert, Card, Dialog, Breakout, Column.
- Treat the rest of the Containers group as optional safety net, and leave Action Group / Row / Overlap / Grid / Table out (Table pending a separate call).
- Rule of thumb for future components: container query if the question is "how much room do I have?"; media query if the question is "what kind of device am I on?" — and Grid's own responsive breakpoints stay on media queries because they're answering the second question.

VincentSmedinga: This a fine report and I think we can follow its suggestions.

Having a single, named query container makes the approach more targeted, less coincidental. A component will ask for containment, but doesn’t know which parent will provide it – and that’s a good thing. We assign the responsibility to a few of container components. Nicely decoupled. Teaching people how it works is easy.

Let’s pick a name and apply it to Grid Cell, Page, and Dialog. The other suggestions are less valid; Alert, Card, and Column are always in a Grid Cell and Breakout renders Grid Cells itself.

One addition: let’s add a utility class (and maybe a Sass mixin) that applies the same query containment. Both us and our users can use that to set a targeted reference point for container queries wherever it’s needed. We can set it on the right wrapper(s) in Storybook. 

I’m not fond of the name ‘layout’. I would have liked ‘grid’, but that doesn’t work well with Dialog or as a utility class. Thinking out loud: we could go with ‘container’ or ‘wrapper’, something with ‘inline-size’, ‘maximum’… The utiltiy class should support more names in the future, and an unnamed one, so something like `ams-container--layout`… Hmm.

RubenSibon: **Utility class**

Yes, I agree that adding a new CSS utility class with a mixin is a good idea. Do you then want to give developers the option to toggle that class via an attribute on Grid Cell, Page, and Dialog? Or should I still hardcode that?

**Name of inline-size layout containment context**

I agree that 'layout' might not be the best. Also because the containment context says nothing about what children should do after querying: any part of their appearance can be affected with a query, not just layout. The variable and class name should be about the container, not what a child potentially does with it.

'container' seems more to-the-point, but is also a very generic name that is already used in many places. It could be confusing if not all components in the Container category provide a containment context called 'container'. So I think that 'wrapper' is better in that case. 'constraint' is another option that basically means the same thing.

Searching online for the use of 'wrapper':

- One author on [CSS Tricks discusses the varying meanings](https://css-tricks.com/best-way-implement-wrapper-css/) of 'container' and 'wrapper' and says: 'in programming languages, the word container is generally used for structures that can contain more than one element. A wrapper, on the other hand, is something that wraps around a single object to provide more functionality and interface to it'. But in the end the author seems to adopt the convention that a wrapper is a kind of root div or a 'general page wrapper'.
- [Article on Geeks for Geeks](https://www.geeksforgeeks.org/css/wrapper-in-css/): a 'wrapper in CSS is an element often a <div> that is used to contain and manage the layout of other elements within a page'.

So two sources say a wrapper is a high-level element that acts as the root for many, often all, HTML elements that form a or the layout. This is basically the role that our Page component has for pages made with the design system.

I think that for our purposes we can understand 'wrapper' as any element that provides a containment context. Page is then the primary _and_ ultimate instance of a wrapper.

**Variable and utility class names**

Let me think out loud as well: we add this breakpoint variable and utility class so that some components can provide an inline-size constraint to a child that can query on that constraint to change its appearance. So I think `inline-size` should be part of the name. Maybe we should just use the `container` shorthand's prop name and values in the utils name? I.e.: `ams-container--inline-size` for inline-size containers that do not have a name and `ams-container--inline-size-wrapper` for containers of type `inline-size` that are named 'wrapper'.

RubenSibon: > Do you then want to give developers the option to toggle that class via an attribute on Grid Cell, Page, and Dialog? Or should I still hardcode that?

We agreed during a refinement that developers should not able to toggle because the containment context is a built-in feature of those components.

> Variable and utility class names

We now lean to being explicit about what the CSS in this class will do: `ams-containment-context--inline-size[-wrapper]`.

VincentSmedinga: Grid Cell, Page and Dialog get:

```css
  container-type: inline-size;
  container-name: something;
```

This way, Description List and any components that start using container queries can act on the width of a suitable ancestor.

We add a utility class `ams-containment-context` or `ams-containment-context--something` that sets the same properties.

We’re aware that there could be more containment contexts with other names, and they could have other types as well. Until we learn what exactly we need, we could go with the one above as the default, so no `--something` yet in the utility class.

We could extract a Sass mixin for the two lines of CSS, slightly overengineering though.

RubenSibon: I'll stick with `ams-containment-context` for now. Nothing more.

Page gets:

```css
  container-type: inline-size;
  container-name: ams-page-containment-context ams-layout-context;
```

As you see, the container name can get multiple values. So we can combine the component specific one and have a unifying one.

Grid Cell and Dialog get corresponding names, i.e. `ams-grid-cell-containment-context` and `ams-dialog-containment-context`.

For the unifying one I went for `ams-layout-context` in the end. Currently, that it what we add the context for: to affect the layout of children.

Agreed?

RubenSibon: My rationale for the two container names is:

- `ams-layout-context`: generic, unified and reusable containment context that 1:1 replaces the browser's media context which was used for media queries. When an `@container ams-layout-context` query is done it finds the closest containment context with that name. We use the phrase “layout context” because we use it as a replacement for `@media` queries in which we changed the layout of a child component.
- `ams-page-containment-context`: specific to this particular component. Handy for when you want to change the appearance of a child relative to this container. We use the phrase “containment context” because we do not want to specify what type this containment context has and how queries on it should be used (might not always affect the layout).

RubenSibon: > We add a utility class `ams-containment-context` or `ams-containment-context--something` that sets the same properties.

While working on it, I actually want to propose the following:

Let's call it `ams-inline-size-context`. Or go back to what I suggested earlier: `ams-container--inline-size`. Or: `ams-containment-context--inline-size`.

Why? Because with CSS utility classes it helps if the name tells you exactly, if feasible, what it does.

`ams-containment-context` does not tell you what type of container you will be querying on.

It is also in line with our other utility classes that are quite self-explanatory except `ams-body`. I.e.: `ams-aspect-ratio-3-4` => `aspect-ratio: 3-4` and `ams-mb-xl` => `margin-block-end: var(--ams-space-xl)`. If this literality is really important we could even go for something like `ams-container-layout-inline-size`.

What do you think?

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
RubenSibon: I’ve changed it, but it actually does not work well. If we set this `div` to that inline size all examples in this array will get that size. They are all in the same DL.

And setting the inline size to something below the breakpoint does not make sense because we want to test for the that the query works, right?

To test the media queries we should another DL (which has the container queries) and put in an element that’s wider than `32rem`.

Do you agree that I bring back another example, but in the same test story?

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
VincentSmedinga: Yes, I think `ams-inline-size-context` is great, it tells exactly what it does.

I knew a fitting name like this was hiding somewhere, we just had to find it, and I guess you did :)

RubenSibon: I’m going to split the utility class out of this PR. It does not belong here.

Also, I’m considering to combine it with #2548 (Page as containment context). Does that seem wise?

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
RubenSibon: In the end I managed with some AI help to combine the container query examples in the single Test story. The `container` prop in the style attribute needs to be removed once the `ams-inline-size-context` exists.

RubenSibon:
Converted to draft again because Page should be containment context (#2548) and the CSS utility class `ams-inline-size-context` (no branch or PR yet) should be added first.

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
```diff
@@ -19,6 +20,24 @@ export default meta
...
+ * Wrapper used by the single Chromatic "Test" story:
+ * - first render example is identical to the original Test story (container context will fallback to Page or none)
+ * - then render example in a narrow container-query context
+ */
+const DescriptionListWithContainerQueryExamples = (props: DescriptionListProps) => (
+  <>
+    <DescriptionList {...props} />
+
+    {/* Description list in a narrow container */}
+    <div className="ams-inline-size-context" style={{ inlineSize: '31.99rem' }}>
```
RubenSibon: As soon as the `ams-inline-size-context` utility is merged into develop and this branch, this example will display in the vertical orientation.

You can test it now by adding `container: ams-inline-size-context / inline-size;` in the style attribute.

RubenSibon:
Marking as ready for review again, ~because the DL now falls back to @media queries meaning @container queries are not required for the layout of the component to "work". Everything should now be as before when there is no containment context.~ [EDIT: fallback has been reverted]

If there’s still time it would be nice if @alimpens could take a look.

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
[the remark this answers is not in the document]
RubenSibon: You are correct. The class was added in anticipation of PR #2562.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -59,97 +148,54 @@
...
-@media (min-width: $ams-breakpoint-medium) {
-  .ams-description-list,
-  .ams-description-list > .ams-description-list__section:last-child {
-    > .ams-description-list__term:last-of-type {
-      margin-block-end: 0;
-    }
-  }
+// Respond to the width of an ancestor container
+@container ams-inline-size-context (min-inline-size: #{$ams-description-list-breakpoint}) {
+  @include ams-description-list--horizontal;
```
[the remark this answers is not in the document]
RubenSibon: Right, I forgot to update the PR’s description. Gonna do that!

storybook/src/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -108,6 +108,22 @@ export const RichDescription: Story = {
...
+  },
+  render: (args) => (
+    <Grid>
+      <Grid.Cell span={{ narrow: 4, medium: 4, wide: 4 }}>
+        <DescriptionList {...args} />
+      </Grid.Cell>
+      <Grid.Cell span={{ narrow: 4, medium: 8, wide: 8 }}>
+        <DescriptionList {...args} />
+      </Grid.Cell>
+    </Grid>
```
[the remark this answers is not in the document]
RubenSibon: All Stories are wrapped in Page. With the current spacing rules that we have a Grid Cell that spans 4 columns should always be below the container query breakpoint.

You are correct that it would be more robust to set an explicit max-inline-size though.

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
```diff
@@ -50,7 +69,7 @@ export const Test: Story = {
     ],
   },
   render: (args) =>
-    renderComponentVariants(DescriptionList, {
+    renderComponentVariants(DescriptionListWithContainerQueryExamples, {
       args,
     }),
```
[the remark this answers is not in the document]
RubenSibon: Alright, I don’t understand well enough how this works, but I can try your suggestion.

packages/css/src/components/description-list/README.md
```diff
@@ -6,12 +6,11 @@ A collection of terms and their descriptions.
...
-In a narrow window, descriptions appear below their term.
-After that, they display in a two-column layout.
-The column for the terms is as wide as the longest term, without wrapping.
-Its width can be adjusted to be ‘large’ (50%), ‘medium’ (33%), or ‘small’ (20%), which also allows the terms to wrap.
-
-The term is set in bold text.
+- The column for the terms is as wide as the longest term, without wrapping. Its width can be adjusted to be ‘large’ (50%), ‘medium’ (33%), or ‘small’ (20%), which also allows the terms to wrap.
+- The term is set in bold text.
+- Description Lists display in a vertical orientation by default. They display in a horizontal orientation (descriptions to the right of their terms) when enough space is available.
+  Inside a [container query context](/docs/utilities-css-inline-size-context--docs), this happens at a container width of `32rem` (typically 512 pixels).
```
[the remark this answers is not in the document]
RubenSibon: You are correct again. That utility is coming with another PR. We should absolutely check if the link works after the merge.

storybook/src/components/DescriptionList/DescriptionList.test.stories.tsx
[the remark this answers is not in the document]
RubenSibon: After asking advice on this in my IDE’s chat I got this back:

> The Copilot reviewer's concern is valid in principle: renderComponentVariants reads __docgenInfo from the component to generate prop combinations. Passing DescriptionListWithContainerQueryExamples (a plain function, not the React component) means it has no __docgenInfo, so renderComponentVariants won't discover props like termsWidth, color, or inverse.
> 
> But looking at line 72, it only passes args — it doesn't rely on auto-generated prop combinations. The Test story hardcodes its args with explicit children covering 1:1, 1:n, n:1, and n:m patterns. So renderComponentVariants here is just a layout wrapper, not a prop-combination generator.
> 
> Verdict: the Copilot reviewer is wrong in this case. The test story is fine as-is — it exercises the key structural variants via explicit args, and the wrapper correctly tests both container contexts. If you want Chromatic to also capture termsWidth variants in both contexts, you'd need a second story for that — but that's a scope expansion, not a bug.

My conclusion is that we can leave it as it is.

RubenSibon:
> changed the title feat(Description List): Switch between orientations based on container width [to] feat(Description List)!: Switch between orientations based on container width [2 minutes ago]

Changed the commit title to communicate that this is a breaking change. Reason: the DL does not render horizontally in isolation without our layout/container components or utility class that provide a containment context.

RubenSibon:
Converted to draft because changes from #2562 need to be applied here.

packages/css/src/components/description-list/description-list.scss
```diff
@@ -21,7 +28,7 @@
   @include text-rendering;
 }
 
-@media (min-width: $ams-breakpoint-medium) {
+@container ams-query-container-inline-size (inline-size >= #{$ams-description-list-breakpoint}) {
```
VincentSmedinga: Is this query repeated often enough to warrant a mixin? E.g.

```scss
@mixin horizontal-orientation {
  @container ams-query-container-inline-size (inline-size >= #{$ams-description-list-breakpoint}) {
    @content;
  }
}
```

packages/css/src/components/description-list/README.md
```diff
@@ -6,12 +6,9 @@ A collection of terms and their descriptions.
 
 ## Design
 
-In a narrow window, descriptions appear below their term.
-After that, they display in a two-column layout.
-The column for the terms is as wide as the longest term, without wrapping.
-Its width can be adjusted to be ‘large’ (50%), ‘medium’ (33%), or ‘small’ (20%), which also allows the terms to wrap.
-
-The term is set in bold text.
+- Description Lists have a vertical orientation when their nearest [query container](/docs/utilities-css-query-container--docs) — typically a Grid Cell, but also a Dialog or Page — is less than `32rem` wide (typically 512 pixels). Above this size, they change to a horizontal orientation, placing the descriptions to the right of their terms instead of below them.
```
VincentSmedinga: This is quite wordy and detailed. ‘Vertical orientation’, 3 other other components, rem and pixel values, both ‘less than’ and ‘above’, explaining what horizontal orientation means… 🙂 

How about

```suggestion
- The terms and descriptions stack vertically by default. They sit next to each other when their [[query container](https://claude.ai/docs/utilities-css-query-container--docs)](/docs/utilities-css-query-container--docs) is wide enough (at lest `32rem`).
```

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
```diff
@@ -18,6 +18,12 @@ import { DesignTokensTable } from "../../_components/DesignTokensTable/DesignTok
 
 ## Examples
 
+### Orientation
+
+The Description List uses container queries to switch from a vertical to a horizontal orientation when its nearest [query container](/docs/utilities-css-query-container--docs) is wider than a predefined breakpoint (see Design). If no query container is available, it displays in the default, vertical orientation.
```
VincentSmedinga: Again, a bit much in my opinion. The text repeats most of the guideline above and the example with two Description List is difficult to grasp and doesn’t work well on mobile.

Can we repeat the `_ams-resize-horizontal` trick once more, maybe with only the last three terms, and introduce it with something like ‘Resize the following example to see the Description List adapt its layout to the available width’? (Although this will not work on mobile as well, with a breakpoint at 500+ pixels.)

packages/css/src/components/description-list/README.md
RubenSibon: Yeah, much simpler. Thanks!

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
RubenSibon: Yeah, I could do that. It would probably work on most phones if people rotate their devices 90°, but then you’d have to inform them about it. Probably not worth the hassle.

I’ll see how I can simplify this story’s description.

packages/css/src/components/description-list/description-list.scss
RubenSibon: I applied this suggestion.

VincentSmedinga (approved):

PR #2540, opened 2026-04-10
fix(Storybook): work around Story rendering bug in Firefox 140 ESR

RubenSibon:
# Work around Storybook story rendering bug in Firefox 140 ESR

## What

Injects a small CSS override via Storybook's `managerHead` in [`storybook/config/main.ts`](storybook/config/main.ts) that sets `place-content: stretch` on the preview's `main > div`, restoring correct story rendering in Firefox 140 ESR.

## Why

Firefox 140 (ESR) triggers a layout bug in Storybook where stories fail to render correctly in the preview iframe. Tracked upstream in:

- storybookjs/storybook#33743
- storybookjs/storybook#33769

Since 140 is the current ESR line and many developers can only use this version of Firefox on their managed laptops a local workaround is needed until upstream ships a fix.

## How

Applied the community workaround from [storybookjs/storybook#33769 (comment)](https://github.com/storybookjs/storybook/issues/33769#issuecomment-3929210792): extend Storybook's `managerHead` with a `<style>` block targeting `main[aria-labelledby="main-preview-heading"] > div` and setting `place-content: stretch`. A `TODO` comment links the upstream issues so this can be removed once Storybook releases a fix.

This PR targets `main` directly as a hotfix (not `develop`).

## Checklist

- [x] Add or update unit tests — n/a, config-only workaround
- [x] Add or update documentation — n/a
- [x] Add or update stories — n/a
- [x] Add or update exports in index.\* files — n/a

## Additional notes

- Jira: DES-1798
- Remove the workaround once an upstream fix for storybookjs/storybook#33743 / storybookjs/storybook#33769 is released (I've subscribed to both issues).
- Worth a quick sanity check in both Firefox 140 ESR and a current stable Firefox to confirm the override is harmless where the bug isn't present.

### DIFF
--- a/storybook/config/main.ts
+++ b/storybook/config/main.ts
@@ -38,18 +38,6 @@ const config: StorybookConfig = {
     options: {},
   },
 
-  // TODO: remove this temporary workaround for Storybook's rendering of stories in Firefox 140 (ESR) when a fix has been released.
-  // Bug reports for Storybook: https://github.com/storybookjs/storybook/issues/33743, https://github.com/storybookjs/storybook/issues/33769
-  // Implemented workaround: https://github.com/storybookjs/storybook/issues/33769#issuecomment-3929210792
-  managerHead: (head) => `
-    ${head}
-    <style>
-      main[aria-labelledby="main-preview-heading"] > div {
-        place-content: stretch;
-      }
-    </style>
-  `,
-
   staticDirs: ['../../packages-proprietary/assets'],
   stories: ['../src/**/*.docs.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
 
--- a/storybook/config/main.ts
+++ b/storybook/config/main.ts
@@ -38,6 +38,18 @@ const config: StorybookConfig = {
     options: {},
   },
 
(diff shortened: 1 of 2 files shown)


storybook/config/main.ts
```diff
@@ -38,6 +38,18 @@ const config: StorybookConfig = {
     options: {},
   },
 
+  // TODO: remove this temporary workaround for Storybook's rendering of stories in Firefox 140 (ESR) when a fix has been released.
+  // Bug reports for Storybook: https://github.com/storybookjs/storybook/issues/33743, https://github.com/storybookjs/storybook/issues/33769
+  // Implemented workaround: https://github.com/storybookjs/storybook/issues/33769#issuecomment-3929210792
+  managerHead: (head) => `
```
VincentSmedinga: Can we add the CSS to our existing overrides stylesheet instead?

RubenSibon: I can try. Let's see.

RubenSibon: I've moved the CSS to the existing `storybook/config/manager-head.html` instead. This seems like the best place.

Evi-2003 (approved):

VincentSmedinga (approved):

PR #2541, opened 2026-04-10
chore(Storybook): Restore component variants rendering in test stories

Evi-2003:
## What

Restores the component variants within the test stories for visual regression testing.

## Why

`@storybook/react-vite` `10.3.4` upgraded `@joshwooding/vite-plugin-react-docgen-typescript` from `0.6.4` to `0.7.0`, which introduced two issues:

**`__docgenInfo` no longer on components**
The plugin now globs for component files using the Vite root (`storybook/`) as `cwd`, so it never finds components in `packages/react/src/`. `renderComponentVariants` breaks because it reads prop metadata directly from `__docgenInfo` at runtime.

**Enum prop values no longer extracted**
`0.7.0` now correctly passes the full compiler options from the resolved tsconfig to `react-docgen-typescript`. Without `tsconfigPath`, the plugin picks up `storybook/tsconfig.json`, which inherits `moduleResolution: "bundler"` from the root tsconfig. This causes `react-docgen-typescript` to fail to resolve component types correctly, breaking enum extraction. Pointing it at the new tsconfig.docgen.json (which uses `moduleResolution: "node"`) fixes this, but then `shouldExtractLiteralValuesFromEnum` needs to be set explicitly to restore the previous extraction behaviour.

## How

- **`tsconfigPath`**: points the plugin at `tsconfig.docgen.json` so the TypeScript program is built with `moduleResolution: "node"`, which `react-docgen-typescript` requires to correctly resolve component types.
- **`include`**: overrides the default glob so it reaches `packages/react/src/`, working around the plugin's incorrect `cwd`. Both this and `tsconfigPath` are required together.
- **`shouldExtractLiteralValuesFromEnum: true`**: restores enum value extraction, needed for `renderComponentVariants` to generate a variants


VincentSmedinga (approved):
Yes, great to have both worlds – the latest Storybook, and our funky component variants function.

PR #2543, opened 2026-04-10
feat(Grid): Allow components to adapt their appearance to the width of their Grid Cell

RubenSibon:
# Let Grid Cells act as a containment context for container queries

## What

Adds a container name and inline-size container type to the Grid Cell component so that descendants can respond to the cell's inline size via `@container` queries.

- New design tokens `ams.grid.cell.container-name` (default `ams-grid-cell`) and `ams.grid.cell.container-type` (default `inline-size`).
- `.ams-grid__cell` now sets `container-name` and `container-type` from those tokens.

## Why

Components placed inside a Grid Cell should be able to adapt their layout to the space the cell actually gives them, not the viewport. Container queries are the right primitive for this, but they require the cell to be declared as a containment context first. Exposing the container name and type as tokens keeps the behaviour configurable per theme.

## How

- Split out from the Description List container query work (DES-1761) so it can be reviewed and merged on its own, since other components will want to query the same context.
- Added `container-name` and `container-type` entries under `grid.cell` in `packages-proprietary/tokens/src/components/ams/grid.tokens.json`.
- Declared `container-name: var(--ams-grid-cell-container-name)` and `container-type: var(--ams-grid-cell-container-type)` on `.ams-grid__cell` in `packages/css/src/components/grid/grid.scss`.
- No React, markup, or API changes; purely additive CSS/tokens. Existing cells are unaffected because the defaults introduce a containment context without changing visual layout.

## Checklist

- [x] Add or update unit tests
- [x] Add or update documentation
- [x] Add or update stories (see Additional notes)
- [x] Add or update exports in index.* files
- [x] Verify visual regression tests pass (comment `/chromatic test` for a new test run)
- [x] Start the PR title with a Conventional Commit prefix, [as explained here](https://github.com/Amsterdam/design-system/blob/main/documentation/publishing.md?plain=1#L11).

## Additional notes

- Related ticket: DES-1761.
- Prerequisite for `feat/DES-1761-dl-vertical-alignment-in-narrow-container`, whose `@container ams-grid-cell (...)` queries rely on this containment context.
- No visual changes expected; please sanity-check Chromatic for the Grid stories anyway.
- No Story added at this time, because we do not have any components that do container queries yet and adding a custom component is too cumbersome.
- No unit or visual regression tests have been added. We should consider adding media and container query test capability to our Chromatic visual regression tests.

PR #2548, opened 2026-04-10
feat(Page): Allow components to query the width of Page to change their appearance

RubenSibon:
# Let Page act as a container query context

## What

Turns the Page component into a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) context, so components inside a page can adapt their layout to its inline size using `@container [ams-page] (…)` instead of `@media` queries against the viewport.

- Two new design tokens on `page`:
  - `container-name` (default `ams-page`)
  - `container-type` (default `inline-size`)
- `.ams-page` now sets `container-name` and `container-type` from those tokens.
- Documents the new capability in the Page README and tidies the Storybook docs page so the Home Page template link sits under *Examples* rather than above *Primary*.

## Why

Page is the root layout component of every City of Amsterdam website and, by definition, spans the full width of the viewport. By declaring it as a containment context, consumers get a single, stable query name (`ams-page`) to express "respond to the width of the browser window" without reaching for media queries. This keeps responsive behaviour consistent with the pattern already introduced on Grid Cell, and lets child components choose the right containment context for their situation (cell-level or page-level) without changing the querying mechanism.

Part of DES-1761, where Description List uses container queries to switch between stacked and column layouts based on the space it actually has.

## How

- Added `container-name` and `container-type` entries under `page` in `packages-proprietary/tokens/src/components/ams/page.tokens.json`.
- Declared `container-name: var(--ams-page-container-name)` and `container-type: var(--ams-page-container-type)` on `.ams-page` in `packages/css/src/components/page/page.scss`.
- Restructured the Page README: introduced a *Guidelines* section, kept the existing "root layout component" rule as a bullet, and added two bullets explaining container queries and the Page-as-viewport-proxy pattern.
- Moved the Home Page template link from the top of `Page.docs.mdx` into the *Examples* section, where it belongs.
- No React, markup, or API changes. The change is purely additive CSS and tokens; existing pages gain a containment context but their visual layout is unaffected.

## Checklist

- [x] Add or update unit tests (see Additional notes)
- [x] Add or update stories (see Additional notes)
- [ ] Verify visual regression tests pass (comment `/chromatic test` for a new test run)

## Additional notes

- Related ticket: DES-1761.
- Companion to `feat/DES-1761-grid-cell-as-containment-context`: Grid Cell provides the cell-scoped `ams-grid-cell` context, this PR provides the page-scoped `ams-page` context. Neither depends on the other — they can land in any order.
- Related to #2539 that allows a Description List to use this containment context.
- No visual changes are expected. It would still be good to sanity-check Chromatic for the Page stories.
- No Story added at this time, because we do not have any components that do container queries yet and adding a custom component is too cumbersome.
- No unit or visual regression tests have been added. We should consider adding media and container query test capability to our Chromatic visual regression tests.

### DIFF
--- a/packages-proprietary/tokens/src/components/ams/page.tokens.json
+++ b/packages-proprietary/tokens/src/components/ams/page.tokens.json
@@ -5,6 +5,12 @@
         "$value": "{ams.color.background.body}",
         "$type": "color"
       },
+      "container-name": {
+        "$value": "ams-page"
+      },
+      "container-type": {
+        "$value": "inline-size"
+      },
       "max-inline-size": {
         "$value": {
           "value": 90,
--- a/packages/css/src/components/page/README.md
+++ b/packages/css/src/components/page/README.md
@@ -6,6 +6,11 @@ Contains the entire website.
 
 The Page component wraps the [Page Header](https://designsystem.amsterdam/?path=/docs/components-containers-page-header--docs),
 [Page Footer](https://designsystem.amsterdam/?path=/docs/components-containers-page-footer--docs), and the main content in between.
-As a root layout component, it must be used for all websites for the City of Amsterdam.
 It is centered horizontally and sets a maximum width of 90 rems (usually 1.440 pixels).
 With a Menu, that becomes 120 rems (1.920 pixels).
+
+## Guidelines
+
+- As a root layout component, it must be used for all websites for the City of Amsterdam.
+- Components can adapt their appearance to the width of a container component and ultimately the Page component through a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries).
+- Because Page spans the entire width of the page, it effectively allows components to adapt their appearance to the width of the viewport by using container queries instead of [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Media_queries).
--- a/packages/css/src/components/page/README.md
+++ b/packages/css/src/components/page/README.md
@@ -13,4 +13,4 @@ With a Menu, that becomes 120 rems (1.920 pixels).
 
 - As a root layout component, it must be used for all websites for the City of Amsterdam.
 - Components can adapt their appearance to the width of a container component and ultimately the Page component through a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries).
-- Because Page spans the entire width of the page, it effectively allows components to adapt their appearance to the width of the viewport by using container queries instead of [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Media_queries).
+- Because Page is the top-level layout container, it allows components to adapt their appearance to the Page component’s available inline size by using container queries instead of [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Media_queries). This matches the viewport width only until the Page reaches its configured maximum width.
--- a/packages/css/src/components/page/page.scss
+++ b/packages/css/src/components/page/page.scss
@@ -5,6 +5,8 @@
 
 .ams-page {
   background-color: var(--ams-page-background-color);
+  container-name: var(--ams-page-container-name);
+  container-type: var(--ams-page-container-type);
   margin-inline: auto;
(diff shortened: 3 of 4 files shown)


alimpens:
Is the intention here to allow Page to be used as a container query context for consumers of the Design System? Or for internal use, i.e. Design System components? 

For that first use case, it feels a little odd to add this for all users, even those that won't use it. Doesn't a utility class make more sense there?

For the second use case, I think that would be breaking with a core pillar of the NLDS architecture: [CSS should be isolated](https://nldesignsystem.nl/handboek/developer/css-conventie/#ge%C3%AFsoleerde-theming). At least, if my assumption is correct that internal components will start relying on whether they are in a Page or Grid.Cell etc, and be less functional if they're not.

VincentSmedinga:
> Is the intention here to allow Page to be used as a container query context for consumers of the Design System? Or for internal use, i.e. Design System components?

The latter. [Our plan](https://github.com/Amsterdam/design-system/pull/2539#discussion_r3080627551) is to make Grid Cell, Page, and Dialog containment contexts. (Well, Grid Cell already is – #2543.) These are sensible defaults for Amsterdam websites that use the design system fully.

Components that use container queries rely on the width of their nearest containment ancestor. Typically, this will be Grid Cell. But we don’t want to require that as an ancestor, partly for the same reason you mention. To prevent that they will keep showing their layout for a narrow context even if more width is available, we make Page a fallback containment context.

We use tokens to set containment context, so people can make different choices there. And we will offer a utility class as well for people to use freely, both for Amsterdam and external use.

alimpens:
> To prevent that they will keep showing their layout for a narrow context even if more width is available, we make Page a fallback containment context.

This means that the functionality of a component is dependent on being in Page, no? So if a user cannot use Page for some reason, components break? This is a CSS side-effect, which decreases the robustness of the system. I would seriously reconsider this approach.

RubenSibon:
I very much appreciate you thinking along and challenging us. We need that!

**A response to some points**

We'll discuss this later today in call, but here are a few replies from me on some points.

> For the second use case, I think that would be breaking with a core pillar of the NLDS architecture: [CSS should be isolated](https://nldesignsystem.nl/handboek/developer/css-conventie/#ge%C3%AFsoleerde-theming). At least, if my assumption is correct that internal components will start relying on whether they are in a Page or Grid.Cell etc, and be less functional if they're not.

Good point. We already wanted to provide a utility class, but maybe we should use that class in all cases and not rely on internal characteristics of components. And then properly document the importance of that class for responsive layouts for developers.

> > To prevent that they will keep showing their layout for a narrow context even if more width is available, we make Page a fallback containment context.
> 
> This means that the functionality of a component is dependent on being in Page, no? So if a user cannot use Page for some reason, components break? This is a CSS side-effect, which decreases the robustness of the system. I would seriously reconsider this approach.

I wouldn’t say that components break; they just appear in their default, narrow state without queries of any kind applying.

And again, there is always the option to use the utility class on a parent component or use one of the mentioned components that provide the required containment context.

**General comment**

I also want to make a general comment about responsive layouts. For a very long time we've had media queries and only now we get wide support for container queries. But actually, we should've had container queries from the beginning. Because when elements are nested in other elements that have a constrained inline size, i.e. they’re smaller than the inline size of the viewport, it doesn't make sense that their dimensional appearance depends on the size of the screen/window.

Finally, we can easily and elegantly achieve that components such as Description List, which have a vertical and a horizontal orientation, can respond to how much size they actually get. And we can do that without complicated set-ups with a combination of media queries and properties.

I would call this progress. And yes, container queries require containment contexts that have to be applied somewhere higher in the DOM. Media queries are always available. That's a disadvantage, and you are correct that it does require us and downstream developers to do something to another component in the tree to affect another component. Let's discuss if we are willing to make that compromise.

RubenSibon:
> if my assumption is correct that internal components will start relying on whether they are in a Page or Grid.Cell etc, and be less functional if they're not.

If you haven't seen it, #2539, is the first example of a child component that will use this containment context. Maybe this will help you understand better what we're trying to achieve.

And as Vincent said, Grid Cell already has said containment context.

RubenSibon:
Closing in favour of #2562 that introduces the CSS utility class and adds or refactors said context for Page, Grid Cell and Dialog.

PR #2550, opened 2026-04-10
feat(Accordion): Rename `expanded` prop of Section to `defaultExpanded`

VincentSmedinga:
# Rename `expanded` prop to `defaultExpanded` for Accordion Section

## What

- Added `defaultExpanded`, following the React convention for uncontrolled initial-value props (`defaultValue`, `defaultChecked`, etc.).
- The old `expanded` prop is deprecated but still works. It emits a console warning when used. `defaultExpanded` takes priority if both are provided.

## Why

- The `expanded` prop name implied controlled behaviour that the component doesn't actually deliver — it only sets the initial state via `useState`.

## How

n/a

## Checklist

- [x] Add or update stories

## Test plan

- [x] Unit tests pass, including new tests for backwards compatibility, prop priority, and the deprecation warning
- [ ] Storybook stories render correctly with the new prop
- [ ] Verify console warning appears when using the old `expanded` prop

## Additional notes

- n/a

### DIFF
--- a/packages/react/src/Accordion/AccordionSection.test.tsx
+++ b/packages/react/src/Accordion/AccordionSection.test.tsx
@@ -5,7 +5,7 @@
 
 import { fireEvent, render, screen } from '@testing-library/react'
 import { createRef } from 'react'
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 
 import { Accordion } from './Accordion'
 
@@ -53,7 +53,19 @@ describe('AccordionSection', () => {
     expect(sectionContent).not.toHaveClass('ams-accordion__panel--expanded')
   })
 
-  it('adds --expanded class when expanded prop is true', () => {
+  it('adds --expanded class when defaultExpanded prop is true', () => {
+    const { getByText } = render(
+      <Accordion.Section defaultExpanded label={testLabel}>
+        {testContent}
+      </Accordion.Section>,
+    )
+
+    const sectionContent = getByText(testContent)
+
+    expect(sectionContent).toHaveClass('ams-accordion__panel--expanded')
+  })
+
+  it('still supports the deprecated expanded prop', () => {
     const { getByText } = render(
       <Accordion.Section expanded label={testLabel}>
         {testContent}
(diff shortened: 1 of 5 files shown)


packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -17,18 +17,27 @@ import { AccordionContext } from './AccordionContext'
...
 
 export const AccordionSection = forwardRef(
   (
-    { children, className, expanded = false, label, ...restProps }: AccordionSectionProps,
+    { children, className, defaultExpanded, expanded, label, ...restProps }: AccordionSectionProps,
     ref: ForwardedRef<HTMLDivElement>,
   ) => {
+    if (expanded !== undefined) {
+      console.warn('Accordion.Section: The `expanded` prop is deprecated. Use `defaultExpanded` instead.')
+    }
```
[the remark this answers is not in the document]
VincentSmedinga: Resolved.

packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -17,18 +17,29 @@ import { AccordionContext } from './AccordionContext'
 
 export type AccordionSectionProps = {
   /** Whether the content is displayed initially. */
+  defaultExpanded?: boolean
+  /**
+   * Whether the content is displayed initially.
+   * @deprecated Use the `defaultExpanded` prop instead.
```
VincentSmedinga:
```suggestion
   * @deprecated Use the `defaultExpanded` prop instead. Will be removed on or after 2026-10-17.
```

Evi-2003 (approved):

packages/react/src/Accordion/AccordionSection.tsx
```diff
@@ -17,18 +17,29 @@ import { AccordionContext } from './AccordionContext'
...
   (
-    { children, className, expanded = false, label, ...restProps }: AccordionSectionProps,
+    { children, className, defaultExpanded, expanded, label, ...restProps }: AccordionSectionProps,
     ref: ForwardedRef<HTMLDivElement>,
   ) => {
     const { headingLevel, sectionAs } = useContext(AccordionContext)
-    const [isExpanded, setIsExpanded] = useState(expanded)
+    const [isExpanded, setIsExpanded] = useState(defaultExpanded ?? expanded ?? false)
+
+    useEffect(() => {
```
RubenSibon, line 38: I think this `useEffect` and the `const` fallback above could include a TODO comment about removing and changing them when the deprecation window has expired. We can then CMD/CTRL-F on all the TODO comments. Maybe also include the deprecation date so that we can search on that too.

packages/react/src/Accordion/AccordionSection.test.tsx
```diff
@@ -65,6 +77,34 @@ describe('AccordionSection', () => {
...
+        {testContent}
+      </Accordion.Section>,
+    )
+
+    expect(spy).toHaveBeenCalledWith(
+      'Accordion.Section: The `expanded` prop is deprecated. Use `defaultExpanded` instead.',
+    )
+
+    spy.mockRestore()
+  })
```
RubenSibon, line 106: Interesting that you added tests for a deprecated prop. What is the rationale? Should we do that for more or all deprecated props? I haven’t found any such tests after a quick global search.

It seems to me that it just adds even more complexity and code that we have to remove in six months.

At the very least add TODO comments above these tests with the deprecation date so that we can search for them when it’s time to do the removal.

packages/react/src/Accordion/AccordionSection.tsx
VincentSmedinga: Not sure. As soon as we remove the `expanded` prop from the type, all occurrences become invalid and they get highlighted in the IDE. I would say that’s clear enough, and impossible to commit.

packages/react/src/Accordion/AccordionSection.test.tsx
VincentSmedinga: Same here. As soon as the `expanded` prop is gone, this test will clearly fail. Also, ‘@deprecated’ is more intentional about _what_ we must do about it than ‘TODO’, and it is less of a signal that something’s incomplete, which isn’t the case.

AI suggested adding this logging. It’s not too complex, so I’d be fine with doing this from now on. No need to update previously deprecated API, too much work for the probably low gain.

RubenSibon (approved):

PR #2551, opened 2026-04-11
feat(Progress List): Add controlled `collapsed` prop to Step

VincentSmedinga:
# Add controlled `collapsed` prop to Progres List Step

## What

Add a `collapsed` prop to `ProgressList.Step` that enables controlled mode for the collapsed/expanded state.

## Why

The component already had an `onToggle` callback but no way for a parent to control the collapsed state. This made it impossible to use `ProgressList.Step` as a fully controlled component.

## How

When `collapsed` is provided (not `undefined`), the component uses that prop directly instead of internal state.

The `onToggle` callback fires in both modes so the parent can respond to user clicks.

When `collapsed` is omitted, the existing uncontrolled behaviour with `defaultCollapsed` and internal `useState` is preserved.

## Checklist

- [x] Add or update stories
- [x] Start the PR title with a Conventional Commit prefix, [[as explained here](https://github.com/Amsterdam/design-system/blob/main/documentation/publishing.md?plain=1#L11)](https://github.com/Amsterdam/design-system/blob/main/documentation/publishing.md?plain=1#L11).

## Additional notes

- n/a


storybook/src/components/ProgressList/ProgressListStep.stories.tsx
```diff
@@ -44,3 +44,12 @@ export const Step: Story = {
...
 }
+
+export const Controlled: Story = {
+  args: {
+    children: <Paragraph>{exampleParagraph()}</Paragraph>,
+    collapsed: false,
+    heading: 'Aanpassing ontwerp fietspad Entreegebied',
+    status: 'current',
+  },
+}
```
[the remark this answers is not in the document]
VincentSmedinga: Resolved.

storybook/src/components/ProgressList/ProgressListStep.stories.tsx
```diff
@@ -39,8 +40,22 @@ type Story = StoryObj<typeof meta>
 export const Step: Story = {
   args: {
     children: <Paragraph>{exampleParagraph()}</Paragraph>,
-    defaultCollapsed: true,
```
RubenSibon, line 46: `defaultCollpsed` is visible in Storybook's Controls and toggling the boolean does nothing. This is also true for `develop`. Why does it show up there?

storybook/src/components/ProgressList/ProgressList.docs.mdx
```diff
@@ -24,6 +24,9 @@ Current and upcoming steps start expanded by default.
 Completed ones are initially collapsed.
 Use the `defaultCollapsed` prop to override this, e.g. to show a Progress List with all steps collapsed.
 
+Use the `collapsed` prop together with `onToggle` to control the collapsed state of a step from a parent component.
+This is useful when the expanded step should depend on other state in the application.
```
RubenSibon: It’s difficult for me and I assume other developers to understand how this works from the current single story. Is it possible to add another story that clearly demonstrates how this works? I.e. by adding the [Vitest test cases](https://github.com/Amsterdam/design-system/pull/2551/changes#diff-ce26cdde6f6df942b6c28b00702c79d752175cf4b9b0d47f8f01ee222bf0dfc3R426-R515) as stories?

packages/react/src/ProgressList/ProgressListStep.tsx
```diff
@@ -47,7 +50,9 @@ export const ProgressListStep = forwardRef(
     ref: ForwardedRef<HTMLLIElement>,
   ) => {
     const { headingLevel } = useContext(ProgressListContext)
-    const [isCollapsed, setIsCollapsed] = useState(defaultCollapsed ?? status === 'completed')
+    const isControlled = collapsed !== undefined
```
RubenSibon: In Storybook it is not possible to set the value of `collapsed` to `undefined` because it has a default value (`false`) in the Controls. So developers cannot test this specific case with these controls at the moment.

packages/react/src/ProgressList/ProgressListStep.tsx
```diff
@@ -81,7 +86,9 @@ export const ProgressListStep = forwardRef(
               className="ams-progress-list__button"
               onClick={() => {
                 const next = !isCollapsed
-                setIsCollapsed(next)
+                if (!isControlled) {
+                  setInternalCollapsed(next)
+                }
                 onToggle?.(!next)
```
RubenSibon: At the moment, it is not possible in the Step story to set or unset the `onToggle` control. Maybe nice to add that option?

storybook/src/components/ProgressList/ProgressListStep.stories.tsx
VincentSmedinga: The story is a controlled one; `collapsed` drives state, not `defaultCollapsed`. The control is therefore irrelevant and misleading. Fixed it by hiding `defaultCollapsed` in this story’s `argTypes`.

storybook/src/components/ProgressList/ProgressList.docs.mdx
VincentSmedinga: Yes, this was getting complicated. I’ve replaced the Step story with a Controlled story that works. The `defaultCollapsed` prop already got mentioned in the Collapsible example

packages/react/src/ProgressList/ProgressListStep.tsx
VincentSmedinga: ~~The Step story got removed – it wasn’t a very good example, and the second block of controls might be useful but noisy as well.~~

Also, `onToggle` expects a function, which isn’t easy to enter through the Storybook controls UI.

packages/react/src/ProgressList/ProgressListStep.tsx
VincentSmedinga: The Step story got removed.

storybook/src/components/ProgressList/ProgressList.docs.mdx
VincentSmedinga: Hmm, that was like throwing the baby out with the bathwater. Restored the story for the Step subcomponent.

VincentSmedinga (commented):
Can’t find some of your older comments in the GitHub UI. I’ve addressed your concerns about the Step story controls when restoring it.

storybook/src/components/ProgressList/ProgressList.stories.tsx
```diff
@@ -84,6 +85,77 @@ export const Collapsible: Story = {
...
+    docs: {
+      source: {
+        code: `const [expandedIndex, setExpandedIndex] = useState(1)
+
+<ProgressList collapsible headingLevel={3}>
+  <ProgressList.Step
+    collapsed={expandedIndex !== 0}
+    heading="Aanvraag ingediend"
+    onToggle={(expanded) => setExpandedIndex(expanded ? 0 : null)}
+    status="completed"
```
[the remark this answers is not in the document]
VincentSmedinga: I removed that duplicate source code, accepting that Show Code is slightly less pretty.

storybook/src/components/ProgressList/ProgressList.stories.tsx
```diff
@@ -84,6 +85,43 @@ export const Collapsible: Story = {
...
+          collapsed={expandedIndex !== 2}
+          heading="Besluit"
+          onToggle={(expanded) => setExpandedIndex(expanded ? 2 : null)}
+        >
+          <Paragraph>U ontvangt het besluit binnen 8 weken na uw aanvraag.</Paragraph>
+        </ProgressList.Step>
+      </ProgressList>
+    )
+  },
+}
```
RubenSibon, line 157: The [code in the Code tab](https://designsystem.amsterdam/demo-progress-list-controlled-collapsed/?path=/story/components-containers-progress-list--controlled) does not seem correct? It includes Storybook’s render and play functions. Not something that you want users to copy.

RubenSibon (commented):
I found one issue with the code in the Code tab.

RubenSibon (commented):
Lint error:

```sh
[2](https://github.com/Amsterdam/design-system/actions/runs/27305990369/job/80663898995?pr=2551#step:7:153)
. lint:js: /home/runner/work/design-system/design-system/packages/react/src/ProgressList/ProgressListStep.tsx
. lint:js:   25:3  error  Prop 'collapsed' should be read-only  react/prefer-read-only-props
```

storybook/src/components/ProgressList/ProgressList.stories.tsx
VincentSmedinga: Added a representation of the code through `parameters.docs.source.code`. This is the idiomatic fix here (Storybook can't auto-generate clean source for a stateful render, which is why the other controlled components hardcode source.code). Trade-off: the snippet is a hand-maintained copy of the JSX, so it can drift if the story changes; same as the existing DatePicker stories.

PR #2552, opened 2026-04-11
feat: Introduce Tab Navigation component

VincentSmedinga:
# Introduce Tab Navigation component

## What

Introduces a Tab Navigation component. It is a visual copy of Tabs, but with links instead of buttons.

## Why

We want to offer an alternative for the Tabs component that is more suitable for page navigation. Our teams now use Tabs with buttons to toggle between large amounts of content. This approach doesn’t provide a good user experience, as people can’t easily open a tab in a new window or share or bookmark it.

If most of the page changes when selecting a different tab, we’re in fact navigating to a different page. An that’s done with hyperlinks, not buttons.

## How

- Copied and adapted the Tabs component. The visual appearance of Tab Navigation is the same as for Tabs. We might change the appearance of Tabs in the future to highlight the difference in their usages.
- Added a vertical orientation, and the option for icons in the tabs. Both are being used in application designs.
- Updated documentation and examples. Most notably, removed the workarounds that indicate the downsides of using buttons for page navigation.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


packages/react/src/TabNavigation/TabNavigationLink.tsx
```diff
@@ -0,0 +1,45 @@
...
+  (
+    { children, className, icon, linkComponent, ...restProps }: TabNavigationLinkProps,
+    ref: ForwardedRef<HTMLAnchorElement>,
+  ) => {
+    const Link = linkComponent || 'a'
+
+    return (
+      <li className="ams-tab-navigation__item">
+        <Link {...restProps} className={clsx('ams-tab-navigation__link', className)} ref={ref}>
+          {icon && <Icon svg={icon} />}
```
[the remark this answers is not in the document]
RubenSibon: Can you address these concerns, @VincentSmedinga?

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  ],
+} satisfies Meta<typeof TabNavigation>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+const labels = ['Gegevens', 'Aanslagen', 'Documenten', 'Acties']
+
+export const Default: Story = {
```
RubenSibon, line 41: The user experience in narrow windows (especially mobile) is not great because:

- with the horizontal orientation a horizontal scrollbar will be visible in most cases (it’s true for all the horizontal stories);
- with vertical orientation there is no place for the content to the side (I assume it will wrap to a new line).

Maybe adjust the docs on how to deal with this. I’m not sure how... I guess best is to instruct developers to avoid Tab Navigation for narrow viewports. It’s simply not a suitable UI for mobile.

I know that this component was introduced because there was a demand from the developers of more complex internal apps of the City (in terms of data flows and UX). Those apps might have multiple levels of content and multiple navigation lists to get a user somewhere (menu’s in the header, sidebar, breadcrumbs, link lists and now tab navigation). But if those apps have to work well on mobile devices all app designers and developers should think hard about how they can reduce UI complexity and navigation within the app. How can we help developers to make keep the navigation as simple as possible. Maybe it’s time to include a more complex Internal app under Pages that combines many ways to navigate. You could use [Amopis in Prototypes](https://amsterdam.github.io/design-system-prototypes/amopis) as a starting point.

For the good old Tabs component all of the above is (no longer) a major issue because those are now much simpler: they will have less content and presumably fewer tabs (at least in the new story).

RubenSibon: Actually, a few possible solutions for the UX on narrow viewports:

- For the vertical orientation: developers should be instructed to render the tab’s “content” below the Tab Navigation component;
- For the horizontal orientation: don’t use it; always use the vertical orientation or automatically fallback to it. Or: make each tab wrap to its own row in narrow viewports, effectively making the tabs appear vertically.

I haven’t thought this through yet...

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+          ))}
+        </TabNavigation.List>
+      </TabNavigation>
+    )
+  },
+}
+
+const twelve = ['one', 'two', 'three', 'four', 'five', 'six', 'seven', 'eight', 'nine', 'ten', 'eleven', 'twelve']
+
+export const WithManyLinks: Story = {
```
RubenSibon, line 168: With this one the UX on narrow viewports is even worse. :-)

Horizontal scrolling should be avoided as much as possible.

packages/css/src/components/tabs/README.md
RubenSibon, line 1: I really like how Tabs are now much simpler in usage. 👍🏻

storybook/src/components/Tabs/Tabs.stories.tsx
RubenSibon, line 1: I really like how Tabs are now much simpler in usage. 👍🏻

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+              {label}
+            </TabNavigation.Link>
+          ))}
+        </TabNavigation.List>
+      </TabNavigation>
+    )
+  },
+}
+
+export const Vertical: Story = {
```
RubenSibon, line 79: Again (see comment at line 46), I would like to see what it looks like in context. Where does the tab’s “content” come exactly? How is the spacing and alignment?

Actually, there is nothing in the current documentation for developers how and where to place the tab’s “content”. I think it would be good to add that with a line and an example.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
RubenSibon: Aaah! https://github.com/Amsterdam/design-system/pull/2553

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
RubenSibon: Aaah! https://github.com/Amsterdam/design-system/pull/2553

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
RubenSibon: Good that I can see it in the context of a page:

Screen Shot 2026-04-14 at 17 00 10

I’m really not a fan of this UX on mobile. What do you think of it? I guess it’s an acceptable compromise when the apps are primarily used on desktop computers by a relatively small group of people and the mobile view is only used sparingly.

Another screenshot with the Menu open for dramatic effect:

Screen Shot 2026-04-14 at 17 04 54

packages/react/src/TabNavigation/TabNavigationLink.tsx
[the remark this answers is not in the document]
VincentSmedinga: Yes, 740af08c97a63b60d9964c3a693bcb24a6a2596f added `{...(!linkComponent && { ref })` and a clearer JSDoc comment.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: Sorry, didn’t link it up in the PR description.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: Ye olde Tabs already [scroll on production](https://designsystem.amsterdam/?path=/docs/components-containers-tabs--docs#with-wide-content), have been a long time. So I think the interaction itself is fine or at least good enough. We should add the feature that the active tab scrolls into view if it’s not.

Making the vertical orientation horizontal on narrow windows would surely help. Still not ideal, but a good first step that designers can work from.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: I don’t agree, scrolling is cheap. It’s fairly clear that more options are available (unless the cutoff is exactly after an item, but we can design a shade or arrow for that) and how to get to them.

[De Volkskrant](https://www.volkskrant.nl) and [NRC](https://www.nrc.nl) do the same on mobile – not the smallest publications.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: @RubenSibon I’ve updated the Tab Navigation component to use its vertical layout only from the medium breakpoint upwards, and rebased the branch for the example page in #2553 on top of it.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
RubenSibon: With both examples I actually disagree that it is good UX. I can’t visually see that you can scroll if the viewport has a width that causes the cutoff to be exactly after an item.

See the screenshots below off a simulated iPhone 12/13. The cutoff is not ideal in both cases. Yes, in the Volkskrant screenshot a light gray divider just visible, but I didn't notice it when the screen was the actual physical size of an iPhone 12/13.

Another issue is that to scroll these link lists you have to move your thumb all the way to the top of the screen. That requires quite a stretch!

Volkskrant on iPhone 12/13:
Screen Shot 2026-04-16 at 15 19 30

NRC on iPhone 12/13:
Screen Shot 2026-04-16 at 15 25 05

My guess is that both those publications consider the topics as secondary, uncritical navigation. I suspect that far majority of readers scroll down to discover the content highlighted by the publication for that moment. That is at least my behaviuor when I read the news (I rarely go to specific category).

packages/css/src/components/tab-navigation/README.md
```diff
@@ -8,6 +8,6 @@ A horizontal or vertical list of links, styled as tabs, to navigate between a se
 
 - Use Tab Navigation for navigating between pages that each have their own main content.
   Thanks to using links, every tab’s content has its own url that the user can open in a new window, share, or bookmark.
-- Use [Tabs](http://designsystem.amsterdam/?path=/docs/components-containers-tabs--docs) instead to switch between small pieces of secondary content without navigating away.
+- Use [Tabs](https://designsystem.amsterdam/?path=/docs/components-containers-tabs--docs) instead to switch between small pieces of secondary content without navigating away.
```
RubenSibon, line 11: Do we have to include the scheme in URLs because starting with `designsystem.amsterdam` does not resolve correctly?

If these URLs do resolve correctly without the schema, it might actually be better to leave them out:

```suggestion
- Use [Tabs](designsystem.amsterdam/?path=/docs/components-containers-tabs--docs) instead to switch between small pieces of secondary content without navigating away.
```

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
RubenSibon: At least our Tab Navigation has a horizontal scrollbar. The news publications don’t have that despite some recommending to always include them for usability and accessibility:

- https://adrianroselli.com/2019/01/baseline-rules-for-scrollbar-usability.html
- https://cerovac.com/a11y/2024/02/consider-accessibility-when-using-horizontally-scrollable-regions-in-webpages-and-apps/

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  component: TabNavigation,
+  argTypes: {
+    orientation: {
+      control: { type: 'inline-radio' },
+      options: ['horizontal', 'vertical'],
+    },
+  },
+  decorators: [
+    (Story, context) => (
+      <div style={{ maxWidth: context.args.orientation === 'vertical' ? '16rem' : undefined }}>{Story()}</div>
```
RubenSibon: The max width seems to cause an issue in both Chromium and Gecko (Firefox) browsers: there is more inline size space at the end when the value of `orientation` is `vertical`. See:

Image
Image

packages/css/src/components/tab-navigation/README.md
VincentSmedinga: Then it should be `[Tabs](//…`, but let’s not change this in one place only, and out of scope for this PR.

packages/react/src/TabNavigation/TabNavigationLink.tsx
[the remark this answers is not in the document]
VincentSmedinga: Does this resolve it?

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: I don’t understand. Your screenshots show horizontal Tab Navigations.

The max width only applies if the orientation is set to vertical, so it shouldn’t be active here. THe goal is to keep the indicator borders closer to the link labels.

I’m not sure where the excessive space is. I see a non-zero scroll position?

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: I agree that the UX should be finetuned or changed for mobile.

The current behaviour seems usable and acceptable for this iteration. Tabs has already ben doing this; teams are using them and we haven’t had any negative feedback on it.

I’ll add a ticket to progress the design for narrow windows.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: I agree that the UX should be finetuned or changed for mobile.

The current behaviour seems usable and acceptable for this iteration. Tabs has already ben doing this; teams are using them and we haven’t had any negative feedback on it.

I’ll add a ticket to progress the design for narrow windows.

storybook/src/components/TabNavigation/TabNavigation.stories.tsx
VincentSmedinga: Ah, saw what you meant offline. Removed the max-width decorator because of this side effect.

RubenSibon (approved):

PR #2562, opened 2026-04-17
feat(Dialog, Grid Cell, Page): Add containment context to allow container-based layouts

RubenSibon:
# Add containment context to Page, Grid Cell, and Dialog to allow container-based layouts

## What

- Add an `ams-inline-size-context` CSS utility class that turns any element into a container query context for its inline size.
- Page, Grid Cell, and Dialog now intrinsically provide this context, under the shared container name `ams-inline-size-context`.

## Why

Components need to adapt their layout to the inline size of their container rather than the viewport.
A single shared container name across Page, Grid Cell, and Dialog lets one `@container ams-inline-size-context (…)` rule work regardless of which ancestor provides the context.

## How

- New `ams-inline-size-context` utility class, with its own `container-name` and `container-type` design tokens.
- Added matching `container-name` and `container-type` tokens and CSS properties to Page, Grid Cell, and Dialog.
- Added a Storybook story and docs for the utility; updated the docs for Page, Grid, and Dialog.

## Checklist

- [x] Add or update unit tests (not relevant)
- [x] Add or update stories


RubenSibon:
@alimpens, this PR adds the `container-*` properties to a new CSS utility and to 3 container components that affect the page's layout.

Currently, not a single child component uses container queries so no side-effecting is taking place.

PR #2539 will add container queries, but I’m working on a system so that media queries do still work by default. Only when the container queries find a containment context up the tree will the media queries be overridden. Watch for those changes when they come.

With these two pieces of information I believe your concerns about isolated CSS and side effects are properly addressed/mitigated?

alimpens:
> @alimpens, this PR adds the `container-*` properties to a new CSS utility and to 3 container components that are affect the page's layout.
> 
> Currently, not a single child component uses container queries so no side-effecting is taking place.
> 
> PR #2539 will add container queries, but I’m working on a system so that media queries do still work by default. Only when the container queries find a containment context up the tree will the media queries be overridden. Watch for those changes when they come.
> 
> With these two pieces of information I believe your concerns about isolated CSS and side effects are properly addressed/mitigated?

Well, sort of. It seems container queries cannot really be used scoped to a component, dependencies on other components are (almost?) always necessary. This undermines the component isolation we currently do have, and makes the system less robust.

It's nice that we'll fall back on media queries when container queries do not work, but that's still patching an issue we're creating by using container queries in the first place. Imo, if a problem is solvable without using container queries, that should always have preference.

RubenSibon:
> Imo, if a problem is solvable without using container queries, that should always have preference.

I agree, for DL we could look at a way so that it collapses (wraps) from a default horizontal orientation to a vertical orientation when it gets too little inline size. That solution would avoid both media and container queries altogether.

storybook/src/utils/InlineSizeContext/InlineSizeContext.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+)
+
+const meta = {
+  title: 'Utilities/CSS/Inline Size Context',
+  component: InlineSizeContext,
+  args: {
+    children: (
+      <Paragraph>
+        This Paragraph sits inside an Inline Size Context, which establishes a containment context of type
+        `inline-size`. Descendants can query its inline size:
```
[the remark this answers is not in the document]
VincentSmedinga: Not needed.

packages/css/src/components/inline-size-context/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Class name
+
+`ams-inline-size-context`
+
+## Guidelines
+
+- Apply this class to an element whose width should drive the layout of its descendants.
+- Descendants can then adapt responsively using `@container ams-inline-size-context (min-inline-size: …)`, instead of `@media (min-width: …)`.
+- [Page](/docs/components-containers-page--docs), [Grid Cell](/docs/components-layout-grid--docs), and [Dialog](/docs/components-containers-dialog--docs) already establish this context, so you don’t need the utility class on those components.
+- Override the default container name or type through the `--ams-inline-size-context-name` and `--ams-inline-size-context-type` custom properties if you need a component-specific container name.
```
RubenSibon:
```suggestion
- Override the default container name or type with the `--ams-inline-size-context-name` and `--ams-inline-size-context-type` custom properties if you need a component-specific container name.
```

packages/css/src/components/inline-size-context/README.md
RubenSibon, line 1: @VincentSmedinga, maybe we should make more clear that in most use-cases you don't need to set this class because in most cases, when the docs are followed, elements will be placed in a Grid Cell or Dialog that already provide the narrowest possible inline size containment context. What do you think? Where to add this?

RubenSibon:
> > @alimpens, this PR adds the `container-*` properties to a new CSS utility and to 3 container components that are affect the page's layout.
> > Currently, not a single child component uses container queries so no side-effecting is taking place.
> > PR #2539 will add container queries, but I’m working on a system so that media queries do still work by default. Only when the container queries find a containment context up the tree will the media queries be overridden. Watch for those changes when they come.
> > With these two pieces of information I believe your concerns about isolated CSS and side effects are properly addressed/mitigated?
> 
> Well, sort of. It seems container queries cannot really be used scoped to a component, dependencies on other components are (almost?) always necessary. This undermines the component isolation we currently do have, and makes the system less robust.
> 
> It's nice that we'll fall back on media queries when container queries do not work, but that's still patching an issue we're creating by using container queries in the first place. Imo, if a problem is solvable without using container queries, that should always have preference.

Hi @alimpens, an update on where this landed.

We went back and forth on this and took a different path than the one I floated above. In #2539 we initially built the `@media` fallback I described, but we've since reverted it: the Description List now uses a container query exclusively, as a 1:1 replacement of the old viewport breakpoint. The result is simpler CSS than the dual-mechanism approach, and (after weighing it against the alternatives) we're keeping it. A few reasons we're comfortable with that:

- **A containment context is virtually always present.** When our usage docs are followed, content lives inside Page or Dialog — both provide `ams-inline-size-context`. In most cases, it's also nested in a Grid Cell, which provides a narrower one.
- **Nothing breaks without one.** If a component ends up outside any containment context, it renders in its mobile-first vertical form. Content stays readable; the layout is just not optimised for the available width.
- **There's an explicit escape hatch.** The new `ams-inline-size-context` utility class lets consumers establish a context on any element when they want one.
- **Art-direction props are on the radar.** For cases where consumers want to override the default, we're considering orientation props on components — a more explicit, component-scoped form of art direction.
- **Cross-team alignment.** We've discussed this approach with the NLDS community and received no disapproving or blocking feedback.

We do hear you on component isolation, and we don't take that lightly. In this case the dependency is on a single, well-known container name (`ams-inline-size-context`) that we control across three layout components, and the alternatives we considered (e.g. flex-wrap-only solutions) didn't clearly solve the underlying problem without trade-offs of their own.

Happy to dig into any specific scenario if there's one you're worried about.

> Note: above comment has been edited by an LLM for tone and convincingness. 🤖😅

alimpens:
> > > @alimpens, this PR adds the `container-*` properties to a new CSS utility and to 3 container components that are affect the page's layout.
> > > Currently, not a single child component uses container queries so no side-effecting is taking place.
> > > PR #2539 will add container queries, but I’m working on a system so that media queries do still work by default. Only when the container queries find a containment context up the tree will the media queries be overridden. Watch for those changes when they come.
> > > With these two pieces of information I believe your concerns about isolated CSS and side effects are properly addressed/mitigated?
> > 
> > 
> > Well, sort of. It seems container queries cannot really be used scoped to a component, dependencies on other components are (almost?) always necessary. This undermines the component isolation we currently do have, and makes the system less robust.
> > It's nice that we'll fall back on media queries when container queries do not work, but that's still patching an issue we're creating by using container queries in the first place. Imo, if a problem is solvable without using container queries, that should always have preference.
> 
> Hi @alimpens, an update on where this landed.
> 
> We went back and forth on this and took a different path than the one I floated above. In #2539 we initially built the `@media` fallback I described, but we've since reverted it: the Description List now uses a container query exclusively, as a 1:1 replacement of the old viewport breakpoint. The result is simpler CSS than the dual-mechanism approach, and (after weighing it against the alternatives) we're keeping it. A few reasons we're comfortable with that:
> 
> * **A containment context is virtually always present.** When our usage docs are followed, content lives inside Page or Dialog — both provide `ams-inline-size-context`. In most cases, it's also nested in a Grid Cell, which provides a narrower one.
> * **Nothing breaks without one.** If a component ends up outside any containment context, it renders in its mobile-first vertical form. Content stays readable; the layout is just not optimised for the available width.
> * **There's an explicit escape hatch.** The new `ams-inline-size-context` utility class lets consumers establish a context on any element when they want one.
> * **Art-direction props are on the radar.** For cases where consumers want to override the default, we're considering orientation props on components — a more explicit, component-scoped form of art direction.
> * **Cross-team alignment.** We've discussed this approach with the NLDS community and received no disapproving or blocking feedback.
> 
> We do hear you on component isolation, and we don't take that lightly. In this case the dependency is on a single, well-known container name (`ams-inline-size-context`) that we control across three layout components, and the alternatives we considered (e.g. flex-wrap-only solutions) didn't clearly solve the underlying problem without trade-offs of their own.
> 
> Happy to dig into any specific scenario if there's one you're worried about.
> 
> > Note: above comment has been edited by an LLM for tone and convincingness. 🤖😅

This reads to me like you've mitigated the issues I mentioned by...not mitigating them? That's a little odd, imo. I think these issues still stand:

- _A containment context is virtually always present:_ This isn't true for users outside of Amsterdam
- _Nothing breaks without one:_ Depends on your definition of breaking I guess, but imo: the responsive behaviour breaks, which is part of the functionality of a component
- _The new `ams-inline-size-context` utility class lets consumers establish a context on any element when they want one:_ Which means components using a container query are dependent on a side-effect. 

As you can probably tell, I don't think this is the right decision. I'm a little surprised to hear that a flex-wrap-only solution didn't solve the underlying problem for a Description List. We've implemented that solution without issues in Meldingen.

RubenSibon:
> This reads to me like you've mitigated the issues I mentioned by...not mitigating them? That's a little odd, imo.

No, we were trying to mitigate your concerns before by adding a fallback to media queries. However, after discussing we’ve decided to abandon that because it adds a lot of SCSS that won’t be used if our docs are followed. Those commits have been reverted. The 1:1 replacement of media queries with container queries is elegant and simple.

> I think these issues still stand:
> 
> * _A containment context is virtually always present:_ This isn't true for users outside of Amsterdam

Users outside of Amsterdam who want to use our components that do container queries may also use our `ams-inline-size-context` utility class or our `container-*` tokens. Or they can provide their own containment context with 1 or 2 lines of CSS.

The need for a container query context (or containment context) is documented for every component that uses container queries.

> * _Nothing breaks without one:_ Depends on your definition of breaking I guess, but imo: the responsive behaviour breaks, which is part of the functionality of a component

With media queries the layout can also look weird (broken?) if the child component sits inside a narrow container. It actually never made sense to use the width of the viewport to affect the layout of elements nested in containers that are much narrower than the viewport.
 
> * _The new `ams-inline-size-context` utility class lets consumers establish a context on any element when they want one:_ Which means components using a container query are dependent on a side-effect.

When composing the layout of a website or app you are dealing with (side) effects by definition, because elements have to work together to form a coherent whole. We find it acceptable to have intentional and well-documented side effects here.

> As you can probably tell, I don't think this is the right decision. I'm a little surprised to hear that a flex-wrap-only solution didn't solve the underlying problem for a Description List. We've implemented that solution without issues in Meldingen.

I agree that flex-wrap-only solution would be preferable to both media and container queries. That would be the most elegant solution. Yet, for various reasons, we’ve decided to use media queries to affect the layout of several components in the past. Now we’re replacing those media queries with container queries to more precisely affect the layout of components based on the inline space they actually get.


I’ll not comment further on this thread here. @VincentSmedinga is going to finish the review and will probably take a good, deep look at the underlying concerns and possibilities (including the flex-wrap-only solution).

packages-proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -37,6 +37,12 @@
           "nl.amsterdam.type": "shadow"
         }
       },
+      "container-name": {
+        "$value": "ams-dialog-containment-context ams-inline-size-context"
```
VincentSmedinga: The three local `ams-*-containment-context` container names aren’t used or documented anywhere. I’d suggest leaving them out for now; we can add them later if a concrete need comes up.

packages-proprietary/tokens/src/components/ams/inline-size-context.tokens.json
```diff
@@ -0,0 +1,12 @@
+{
+  "ams": {
+    "inline-size-context": {
+      "container-name": {
+        "$value": "ams-inline-size-context"
```
VincentSmedinga: One more thought on the name: `ams-context-inline-size` might be a better ordering. It scales more naturally if we add other containment contexts later, e.g. `ams-context-scroll-state`. (In case you’re wondering: we don’t use BEM `--` in utility classes.)

The docs page could then be ‘Context’, with the `inline-size` variation as the single start.

packages/css/src/components/dialog/README.md
```diff
@@ -12,6 +12,7 @@ A popup window in which the user must perform an action to proceed.
 - Wrap multiple buttons in an [Action Group](https://designsystem.amsterdam/?path=/docs/components-layout-action-group--docs).
 - To open the Dialog, use `Dialog.open(dialogId)` from the React package.
 - To close it, either call the `Dialog.close` function or add a `<form>` as in the ‘confirmation’ example.
+- Dialog provides a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) context for its content.
```
VincentSmedinga: This may be easier to understand for our audience:

```suggestion
- Components placed inside Dialog can adapt their layout to the width of the Dialog itself, rather than the width of the browser window.
  This uses a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries).
```

Can you update the other components as well?

packages/css/src/components/inline-size-context/README.md
VincentSmedinga: We can make that the first guideline, or a sentence before the list like we do for Accordion.

packages/css/src/components/inline-size-context/README.md
```diff
@@ -0,0 +1,16 @@
+<!-- @license CC0-1.0 -->
+
+# Inline Size Context
+
+Turns an element into a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) context that descendants can query against its inline size.
```
VincentSmedinga: I’d rather keep links out of the rationale — let’s move them to a ‘See also’ section. In general, our users trust us to pick the right web features and aren’t looking for implementation details.

storybook/src/utils/InlineSizeContext/InlineSizeContext.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+    {children}
+  </div>
+)
+
+const meta = {
+  title: 'Utilities/CSS/Inline Size Context',
+  component: InlineSizeContext,
+  args: {
+    children: (
+      <Paragraph>
```
VincentSmedinga: Maybe we can use the `resize` approach that we have for Action Group, and switch the background colour at a certain width? And then simple content like ‘Below a width of x, the background is …, otherwise it is …’. Should be a fairly narrow width to make it work on mobile devices?

storybook/src/utils/InlineSizeContext/InlineSizeContext.docs.mdx
```diff
@@ -0,0 +1,13 @@
...
+import * as InlineSizeContextStories from "./InlineSizeContext.stories.tsx";
+import README from "../../../../packages/css/src/components/inline-size-context/README.md?raw";
+
+<Meta of={InlineSizeContextStories} />
+
+<Markdown>{README}</Markdown>
+
+<Primary />
+
+<Controls />
```
VincentSmedinga: These can be removed, unless we want people to choose a width for the container query?

```suggestion
```

packages/css/src/components/inline-size-context/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Class name
+
+`ams-inline-size-context`
+
+## Guidelines
+
+- Apply this class to an element whose width should drive the layout of its descendants.
+- Descendants can then adapt responsively using `@container ams-inline-size-context (min-inline-size: …)`, instead of `@media (min-width: …)`.
+- [Page](/docs/components-containers-page--docs), [Grid Cell](/docs/components-layout-grid--docs), and [Dialog](/docs/components-containers-dialog--docs) already establish this context, so you don’t need the utility class on those components.
+- Override the default container name or type with the `--ams-inline-size-context-name` and `--ams-inline-size-context-type` custom properties if you need a component-specific container name.
```
VincentSmedinga: If we extract a code block below, we can make these guidelines easier to read.

packages-proprietary/tokens/src/components/ams/dialog.tokens.json
RubenSibon: Correct, they are not used or documented at the moment.

My thinking to have these names was: such specific names allow consumers (or ourselves) to target these specific containers when they write their own `@container` queries. But without documentation users will not discover that ability. And it is indeed the question if this feature will be used much or at all.

I will remove them for now.

packages-proprietary/tokens/src/components/ams/inline-size-context.tokens.json
RubenSibon: Do you then also want to change the utility class’ name?

If we swap the parts around would `ams-container-context-inline-size` or `ams-containment-context-inline-size` then be better?

Note that [MDN consistenly mentions](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries#using_container_size_queries) the need to declare a “containment context”, not a “container context”.

VincentSmedinga: I’ve done some more considering:

- ams-inline-size-context
- ams-context-inline-size
- ams-container-inline-size
- ams-context-container-inline-size
- ams-containment-context-inline-size
- ams-query-container-inline-size

and I think the last one is best. Only ‘Container’ may be misleading, ‘Context’ generic, the other two with two words complicated, but ‘query container’ is more precise, reads well, not too complex and exactly what the `container-type` and `-name` properties create. Slightly long, but that’s acceptable as it will be seldomly used. And MDN literally writes “establishes a query container for …” for all [type values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/container-type#values), so it’s really what it is.

So let’s go with ‘Query container’ as the name of the page etc.

VincentSmedinga: @RubenSibon I’ve applied this, extracted a common token for the container name, and spent some more time on the documentation page. Can you review and let me know if you agree?

VincentSmedinga (dismissed):

VincentSmedinga:
@alimpens Thanks for your feedback and pushback. We understand that adopting container queries introduces a new paradigm. And you’re not wrong in stating that some components now rely on external context for their optimal appearance.

We have taken our time to consider this approach carefully. Our conclusion is that we do want to take this step forward.

Container queries have long been a highly anticipated feature of the web. Experts have put much thought into its design. It recently became ‘Baseline widely available’, having been stable in major browsers for 30 months. They are a real part of the platform.

By nature, they require a relationship between elements. Your main concern is that (the appearance and behaviour of) components will be less encapsulated. We think this is acceptable. The chance of failure is slim, the amount of damage is small, the resolution is both simple and a common approach in many other cases.

Specifically for Description List, we have researched other ways to have it adapt its orientation intrinsically, given the requirements of correct layout with idiomatic HTML (`dl`, `dt`, `dd` only, with `div` only to wrap multiple `dt`) and support for 1:n, m:1 and n:m terms and descriptions. We haven’t found one.

We can see other components benefit from container queries in the future. We’d even not be surprised if we end up replacing all media queries, as a well-chosen container is arguably a better context than the viewport to let components base their layout decisions on.

When using Amsterdam Design System as intended, components never lack a containment context; they will almost always be wrapped by a Grid Cell, otherwise by a Page or Dialog. No regression here. For custom or onforeseen patterns, the utility class that adds inline size containment context is very easy to add. The same goes for external consumers who don’t use our Grid, Page or Dialog. We have been in contact with NL Design System, both community members and the lead developer. They agree that while the side effect may be real, our approach is also convenient, prefixed, well-documented, and cheap to adopt.

Again, a Description List without containment context is perfectly usable; its base appearance is fine. This will be the same for any other components using container queries. In rare cases, they just won’t improve their appearance for the available space. On the other side, using media queries can equally be seen as ‘broken’: they give Description List horizontal layout in a narrow container on a wide window, which seems even less usable than the exception we have with container queries.

Having to add a marker somewhere is not new. All our utility classes add determinate and documented effects – a couple of text elements without a margin class between them would also be ‘broken’ or not-encapsulated. CSS positioning, grid and flex layout require a relationship between elements and descendants. Many compositions require adding markup for accessibility.

So. We’re taking responsibility, both curious and comfortable with what it will bring. It’s called software for a reason – if all hell breaks loose, we can just change it once more. But we’re pretty sure it won’t.

packages/css/src/components/page/README.md
```diff
@@ -12,5 +12,4 @@ With a Menu, that becomes 120 rems (1.920 pixels).
 ## Guidelines
 
 - As a root layout component, it must be used for all websites for the City of Amsterdam.
-- Components placed inside Page can adapt their layout to the width of the Page itself, rather than the width of the browser window.
-  This uses a [CSS container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries).
+- Page is a [query container](/docs/utilities-css-query-container--docs) for inline size, so that elements in it can adapt their appearance to the width of the page rather than that of the browser window.
```
RubenSibon: Maybe we could add that `Page`’s containment context/query container is a last-resort one. Ideally and in most cases a Grid Cell or Dialog will provide the appropriate context to query against.

VincentSmedinga: Good point, updated.

It’s even not a last resort. The Page container is the _only_ containment context for things that sit outside a Grid Cell or Dialog: Page Header, Page Footer, skip links, anything in the main flow that isn’t in a grid.

Also removed the unnecessary complexity of “rather than the width of the browser window”. No need to refer to media queries here as well. We could add a focused page about responsive design in the developer guide later.

VincentSmedinga (approved):
Hello container queries! 🎉

PR #2581, opened 2026-04-22
fix(Field Set): Don’t grow beyond container width for wide inputs or textareas

### LINKED ISSUE #2580: FieldSet overflow issue
Kind of a mistery here, if you add Field and TextInput components to a FieldSet the `size` property will cause overflow.

Image

If you change the fieldset to a div everything is ok, both are block elements. I've set up a branch with a test in the internal navigation page story so the overflow is easily visible.

VincentSmedinga:
# Don’t grow beyond container width for wide inputs or textareas

## What

Adds `min-inline-size: 0` to `.ams-field-set` so it no longer grows to accommodate children with an intrinsic width (inputs with a `size` attribute, textareas with `cols`).

Also clarifies the Size documentation for Text Input and Password Input to note that the input won’t grow beyond the width of its container.

This was reported in #2580.

## Why

The native `<fieldset>` element has a UA style that effectively sets `min-inline-size: min-content`, so it refuses to shrink below the intrinsic size of its children. When a `TextInput` with a `size` attribute is placed inside a `FieldSet`, the fieldset expands to fit the input and overflows its containing grid cell. Swapping the fieldset for a `<div>` sidesteps the quirk but drops the semantics.

`max-inline-size: 100%` on the input can’t help here: the containing block (the fieldset) has already grown, so 100% equals the expanded width.

## How

One-line CSS fix in `field-set.scss` with an inline comment explaining the UA quirk. The input’s existing `max-inline-size: 100%` then clamps correctly, because the fieldset now tracks its container's width.

Clarified documentation on the `size` attribute for Text Input and Password Input.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/storybook/src/components/TextInput/TextInput.docs.mdx
+++ b/storybook/src/components/TextInput/TextInput.docs.mdx
@@ -100,13 +100,14 @@ We follow the advice against using `type="number"` by [NL Design System](https:/
 
 ### Size
 
-Make the width of the input appropriate for the information to be entered.
+Match the width of the input to the expected length of its content.
 Use the `size` attribute for this.
+The input won’t grow beyond the width of its container.
 
-Avoid making it too narrow – ensure the user has enough space to operate comfortably.
-This also gives browsers or extensions an opportunity to add an icon or button at the end.
+Avoid sizes that feel cramped: the user needs room to type and review comfortably.
+Extra space also leaves room for icons or buttons that browsers and extensions may add.
 
-We suggest a size of 15 for a phone number, 30 for an e-mail address, 40 for a web address, and 7 for a postal code.
+As a rule of thumb: 7 for a postal code, 15 for a phone number, 30 for an email address, and 40 for a web address.
 
 <Canvas of={TextInputStories.Size} />
 
--- a/storybook/src/pages/internal/NavigationPage/NavigationPage.stories.tsx
+++ b/storybook/src/pages/internal/NavigationPage/NavigationPage.stories.tsx
@@ -6,7 +6,17 @@
 import type { Meta, StoryObj } from '@storybook/react-vite'
 import type { MouseEvent } from 'react'
 
-import { Breadcrumb, Grid, Heading, TabNavigation } from '@amsterdam/design-system-react'
+import {
+  Breadcrumb,
+  Field,
+  FieldSet,
+  Grid,
+  Heading,
+  Label,
+  Select,
+  TabNavigation,
+  TextInput,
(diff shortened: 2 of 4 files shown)


storybook/src/components/TextInput/TextInput.docs.mdx
```diff
@@ -100,13 +100,14 @@ We follow the advice against using `type="number"` by [NL Design System](https:/
...
 Use the `size` attribute for this.
+The input won’t grow beyond the width of its container.
 
-Avoid making it too narrow – ensure the user has enough space to operate comfortably.
-This also gives browsers or extensions an opportunity to add an icon or button at the end.
+Avoid sizes that feel cramped: the user needs room to type and review comfortably.
+Extra space also leaves room for icons or buttons that browsers and extensions may add.
 
-We suggest a size of 15 for a phone number, 30 for an e-mail address, 40 for a web address, and 7 for a postal code.
+As a rule of thumb: 7 for a postal code, 15 for a phone number, 30 for an email address, and 40 for a web address.
```
dlnr, line 110: What should be the length for first and sir name? I always struggle with these because they can obviously vary.

storybook/src/pages/internal/NavigationPage/NavigationPage.stories.tsx
```diff
@@ -97,6 +107,24 @@ const meta = {
           <Heading className="ams-mb-m" level={3}>
             {currentSubMenu.label}
           </Heading>
+          <FieldSet legend="Test Inline size">
```
dlnr: Should you leave the demo of the outside in the docs? I would think there could be form examples for internal apps.

storybook/src/components/TextInput/TextInput.docs.mdx
VincentSmedinga: I’m not sure. We had a ticket for it on a couple of sprints, but it never got done.

GOV.UK says:
> Fields must be long enough to accommodate the names of your users. You should use population data or data about your existing users to do this.

I’ll check with design this week.

storybook/src/pages/internal/NavigationPage/NavigationPage.stories.tsx
VincentSmedinga: Let’s scope this PR to the pure bugfix. Form examples in application pages would be great – I’ll add a ticket and you’re welcome to share some of the forms you encountered in your projects.

dlnr (approved):

PR #2584, opened 2026-04-23
feat(Table of Contents): Make items with nested lists optionally collapsible

VincentSmedinga:
# Make the Table of Contents collapsible

## Links

- [Docs](https://designsystem.amsterdam/demo-DES-1819-collapsible-toc/?path=/docs/components-navigation-table-of-contents--docs#collapsible)
- [Story](https://designsystem.amsterdam/demo-DES-1819-collapsible-toc/?path=/story/components-navigation-table-of-contents--collapsible)

## What

- Adds a `collapsible` prop to `TableOfContents`. When set, every item that contains a nested list gets its own toggle button, separate from the link, to expand and collapse that list.
- Adds supporting props: `defaultExpanded` and `onToggle` on `TableOfContents.Link`, and `showAccessibleLabel` / `hideAccessibleLabel` on the root to customise the toggle button’s accessible name (defaults `Toon submenu van` / `Verberg submenu van`).
- Highlights the current page’s link through `aria-current="page"`.
- Preserves consumer-provided nested list ids for `aria-controls`, falling back to generated ids only when one is missing.
- Restricts arrow-key navigation to visible toggle buttons, and only when the keyboard event originates from a toggle.
- Adds a `Collapsible` story and component documentation.

## Why

A long, fully expanded Table of Contents is hard to scan when it has many nested sections.
Making items collapsible lets readers fold away the sections they are not using while keeping the current page’s branch open — the behaviour a documentation or handbook layout needs.
It mirrors the established pattern in Accordion and Progress List rather than introducing a new interaction model.

## How

- **Tokens → CSS:** Adds toggle, current-link, and item-gap tokens, and tightens the nested-list padding tokens. In `collapsible` mode items use a two-column grid so labels of items without a toggle align with those that have one. Collapsed subtrees are hidden but stay visible in `@media print`, and the chevron rotates with a reduced-motion-aware transition. The existing gap-based spacing is kept, so the `gap` and `list.gap` tokens are unchanged and nothing is deprecated.
- **React:** Introduces an internal `TableOfContentsContext` to pass `collapsible` and the accessible labels to each link without prop drilling. `TableOfContents.Link` detects a direct `TableOfContents.List` child, renders an `IconButton` toggle (`aria-expanded` + `aria-controls`), preserves explicit nested list ids, and manages its own expanded state.
- **Accessibility:** When a subtree that holds focus is collapsed, focus moves back to its toggle button so it is not lost on a hidden element.

## Checklist

- [x] Add or update stories

## Additional notes

- Implements DES-1819.
- The Handbook Page example that demonstrates this component has moved to a separate, stacked PR (branch `feat/DES-1819-handbook-page`) so this PR stays focused on the component.
- `TableOfContentsContext` is internal to the component, so no public `index.*` export was added.
- All new props are optional and backwards compatible: with `collapsible` unset, no toggle buttons render and existing markup is unchanged.
- Follow-up for SPA-specific control (a `collapsed` prop on `TableOfContents.List`) is tracked separately in Jira.


storybook/src/components/TableOfContents/TableOfContents.stories.tsx
RubenSibon, line 41: Sanity check: the default story has no nested items. When you toggle the `collapsible` boolean to `true` all the items will get indentation which looks a bit odd. Is this the intended and desired behaviour, @VincentSmedinga? Maybe yes, not to complicate the logic?

storybook/src/components/TableOfContents/TableOfContents.docs.mdx
```diff
@@ -22,6 +22,13 @@ import { DesignTokensTable } from "../../_components/DesignTokensTable/DesignTok
 
 <Canvas of={TableOfContentsStories.MultipleLevels} />
 
+### Collapsible
+
+Set the `collapsible` prop on the root to let each item with a nested list expand and collapse via a separate toggle button.
+Use `defaultExpanded` on a Link to pre-open the current page's ancestors, and `aria-current="page"` on the link itself to highlight it.
```
RubenSibon: @VincentSmedinga, have you and the designers considered opening the descendants by default or via another prop as well?

In the [demo](https://designsystem.amsterdam/demo-DES-1819-collapsible-toc/?path=/story/pages-public-handbook-page--default) I find that it’s not immediately obvious that an item has children. The only indication of that is the down-facing chevron.

Maybe users would expect immediate children of the active link to be automatically expanded?

storybook/src/components/TableOfContents/TableOfContents.stories.tsx
VincentSmedinga: That would be using the component incorrectly. So not desired, but correct output given the input. Our policy is not to programatically guard against this kind of errors.

storybook/src/components/TableOfContents/TableOfContents.docs.mdx
VincentSmedinga: Yes, that’s what it should do indeed. Sorry I forgot to specify that. See e.g. amsterdam.nl/pga for reference.

A server-side rendered website can achieve this by not adding the `ams-table-of-contents__item--collapsed` class to the link that represents the current page.

For dynamic (single page) rendering, we should probably add a `collapsed` prop to the `List` subcomponent. Fine to add that in a separate PR.

RubenSibon (changes requested):
1. Make active TOC Link descendants expanded by default
2. Address GitHub Copilot’s comments
3. We have a [unique landmark violation](https://68db9df886b46f139748c074-xwcwwofybs.chromatic.com/?path=/story/components-tableofcontents--test&a11ySelection=).

RubenSibon (dismissed):
Approving to reset my earlier request for changes.

@VincentSmedinga, can you continue your review?

storybook/src/components/TableOfContents/TableOfContents.docs.mdx
RubenSibon: The change for multipage set-ups has been implemented.

A Jira story has been made for the latter point on SPAs.

RubenSibon (dismissed):

packages-proprietary/tokens/src/components/ams/table-of-contents.tokens.json
```diff
@@ -20,7 +20,8 @@
         }
       },
       "gap": {
-        "$value": "{ams.space.s}",
+        "$value": "initial",
```
VincentSmedinga: This introduces the same problem as Aram mentioned for a earlier change in Page Header. This deprecation approach may work fine within our own CSS, but it doesn’t for people using this token directly. Reverting.

storybook/src/components/TableOfContents/TableOfContents.docs.mdx
```diff
@@ -17,12 +17,46 @@ import * as TableOfContentsStories from "./TableOfContents.stories.tsx";
 
 <Controls />
 
+## Usage guidelines
+
+### When to use
+
+Use a Table of Contents for any list of navigation links that correspond to named sections of content: anchor links to sections on the current page, links to a set of related pages, or a deep multi-level tree for a handbook or documentation site.
+For the last case, use `collapsible` mode so branches can be expanded and collapsed.
+
```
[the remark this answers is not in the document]
VincentSmedinga: This removes a required empty line, has nothing to do with “one sentence per line”.

packages/react/src/TableOfContents/TableOfContents.tsx
```diff
@@ -3,49 +3,107 @@
...
   ) => {
+    const innerRef = useRef<HTMLElement>(null)
+
+    useImperativeHandle(ref, () => innerRef.current as HTMLElement)
+
+    const { keyDown } = useKeyboardFocus(innerRef, {
+      focusableElements: [
+        '.ams-table-of-contents__toggle:not([disabled]):not(.ams-table-of-contents__item--collapsed > .ams-table-of-contents__list .ams-table-of-contents__toggle)',
+      ],
+      rotating: true,
```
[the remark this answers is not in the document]
VincentSmedinga: That exclusion (`.item--collapsed > .list .toggle`) is the clearest way to express “toggles inside a collapsed subtree, but not the collapsed item's own toggle.” It’s a Baseline-supported selector in 2026, and the only alternative – replicating it by walking ancestors in JS inside the shared useKeyboardFocus (used by Accordion, ProgressList, Tabs) – is more error-prone and complicated by jsdom having no layout in tests. Not worth the trade.

Did add a comment to explain the selector.

packages/react/src/TableOfContents/TableOfContentsLink.tsx
```diff
@@ -3,30 +3,118 @@
...
+    const [isExpanded, setIsExpanded] = useState(defaultExpanded ?? false)
+
+    const panelId = useId()
+    const buttonRef = useRef<HTMLButtonElement>(null)
+    const itemRef = useRef<HTMLLIElement>(null)
+
+    const listChild = findListChild(children)
+    const isExpandable = collapsible && !!listChild
+    // Reuse a provided nested list id to keep aria-controls references stable.
+    const nestedListId = (listChild?.props as { id?: string } | undefined)?.id ?? panelId
```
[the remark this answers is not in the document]
VincentSmedinga: Updated.

packages/react/src/TableOfContents/TableOfContents.test.tsx
```diff
@@ -75,4 +76,393 @@ describe('TableOfContents', () => {
...
+      expect(nestedList).toBeInTheDocument()
+      expect(nestedList).toHaveClass('ams-table-of-contents__list')
+    })
+
+    it('ignores a blank nested list id and falls back to a generated id', () => {
+      render(
+        <TableOfContents collapsible>
+          <TableOfContents.List>
+            <TableOfContents.Link defaultExpanded href="#a" label="A">
+              <TableOfContents.List id="  ">
```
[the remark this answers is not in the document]
VincentSmedinga: Switched the test to `id=""` so the markup stays valid. The component still trims whitespace defensively, so a blank id in either form falls back to a generated id.

packages/react/src/TableOfContents/TableOfContents.test.tsx
```diff
@@ -75,4 +76,393 @@ describe('TableOfContents', () => {
...
+      expect(nestedList).toBeInTheDocument()
+      expect(nestedList).toHaveClass('ams-table-of-contents__list')
+    })
+
+    it('ignores a blank nested list id and falls back to a generated id', () => {
+      render(
+        <TableOfContents collapsible>
+          <TableOfContents.List>
+            <TableOfContents.Link defaultExpanded href="#a" label="A">
+              <TableOfContents.List id="">
```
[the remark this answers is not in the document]
VincentSmedinga: This rule and the earlier one (`r3442982094`) cancel out: an `id` can't be both non-empty and whitespace-free *and* represent a blank value, so there’s no literal that satisfies both — the test was deliberately passing an invalid id to check the component ignores it. Rather than keep flip-flopping, I’ve removed the test: the fallback to a generated id is already covered by the case that renders a nested list with no `id`, so coverage stays at 100%.

PR #2604, opened 2026-05-08
chore(docs): Add cells to default Grid example

VincentSmedinga:
# Describe the pull request

## What

Adds three cells to the [default Grid example](https://designsystem.amsterdam/demo-default-grid-example/?path=/docs/components-layout-grid--docs#design).

## Why

We didn’t add cells to the first example to let it focus on the Grid, the actual component, rather than on Cells. But @dlnr pointed out that this makes the controls below the grid appear broken, because they have no visual effect. That’s a fair point.

We could remove the controls, but that would break the convention. Adding cells in the first example anyway is defensible because the point of a grid is positioning things on it.

I’ve chosen two rows to show the effect of `verticalGap`, and spans of `all`, `8` and `4` because they form the minimal hint at the general ‘header, body, sidebar’ layout. It did require using three values for `span` immediately, but I guess that’s acceptable as well.

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/storybook/src/components/Grid/Grid.stories.tsx
+++ b/storybook/src/components/Grid/Grid.stories.tsx
@@ -69,6 +69,14 @@ const CellStoryTemplate: CellStory = {
 
 export const Default: Story = {
   ...StoryTemplate,
+  args: {
+    ...StoryTemplate.args,
+    children: [
+      <Grid.Cell className="_ams-item" key={1} span="all" />,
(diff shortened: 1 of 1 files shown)


RubenSibon (approved):
Yeah, this is certainly an improvement!

Having an empty box with only the grid columns visible and no working controls was more confusing than helpful.

PR #2605, opened 2026-05-08
chore(docs): Clarify favicon and web app installation steps

RubenSibon:
## What

Refines the Favicon and Web App developer guides:

- Reframes the install step as a check that `@amsterdam/design-system-assets` is present, and uses `npm install` instead of `npm i`.
- Splits the “Link assets” section into clearer subsections: _Copy or symlink assets_, _Edit Web Manifest_ (Web App only), and _Linking in HTML_.
- States explicitly that you can copy the icon assets (often into `public`), with symlinking presented as an alternative for macOS and Linux shells.
- Replaces `rm -r` with `unlink` for removing symlinks, and notes the argument must be a name rather than a path.
- Adds a note that linking static assets in HTML depends on the project’s framework or build tool.
- Updates “Submit changes” to mention copied files as well as symlinks.

## Why

The previous wording implied symlinking was the recommended (or only) path and used commands that could affect the original files. The new structure better matches how teams actually integrate the assets and avoids the footgun of `rm -r` on a symlink target.

Resolves [DES-1822](https://amsterdam.atlassian.net/browse/DES-1822).

## How

Editorial changes only to:

- `storybook/src/brand/assets/favicon.docs.mdx`
- `storybook/src/docs/developer-guide/web-app.docs.mdx`

No code, components, or exports affected.

## Checklist

- [x] Add or update stories

## Additional notes

- Docs-only change; no visual regressions expected.
- Please review the wording around copy-vs-symlink — happy to tighten further if either path should be recommended over the other.

[DES-1822]: https://gemeente-amsterdam.atlassian.net/browse/DES-1822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storybook/src/brand/assets/favicon.docs.mdx
+++ b/storybook/src/brand/assets/favicon.docs.mdx
@@ -18,10 +18,10 @@ This set covers all common browsers (Chrome, Safari, Edge, and Firefox).
 
 ### Install assets
 
-First, install these assets as they come with the [Amsterdam Design System Assets](https://www.npmjs.com/package/@amsterdam/design-system-assets):
+First, ensure that you have the [Amsterdam Design System Assets](https://www.npmjs.com/package/@amsterdam/design-system-assets) installed:
 
 ```sh
-npm i @amsterdam/design-system-assets
+npm install @amsterdam/design-system-assets
 ```
 
 ### Link assets
--- a/storybook/src/brand/assets/favicon.docs.mdx
+++ b/storybook/src/brand/assets/favicon.docs.mdx
@@ -12,7 +12,7 @@ Represent the website in bookmarks, on home screens, in syndication and other to
 Here’s how to display a website icon (a favicon) in the web browser.
 This set covers all common browsers (Chrome, Safari, Edge, and Firefox).
 
-> There is separate documentation on [how to publish a web app with the proper app names and icons](/docs/docs-developer-guide-web-app--docs).
+**Note**: There is separate documentation on [how to publish a web app with the proper app names and icons](/docs/docs-developer-guide-web-app--docs).
 
 ## How to use
 
@@ -42,14 +42,16 @@ cd public
 ln -s ../node_modules/@amsterdam/design-system-assets/favicon favicon
 ```
 
-> Important: the target path (the first argument) is resolved relative to the directory containing the symlink, not your current shell directory. The simplest way to avoid surprises is to `cd` to that directory first, as in the example above.
+**Important**: the target path (the first argument) is resolved relative to the directory containing the symlink, not your current shell directory. The simplest way to avoid surprises is to `cd` to that directory first, as in the example above.
 
-> Remove the symlink without affecting the original files:
->
-> ```sh
-> # Note: do not include a trailing slash — that follows the symlink and unlink will refuse to remove a directory.
-> unlink favicon
-> ```
+#### Remove symlink
+
+Remove the symlink without affecting the original files:
+
+```sh
+# Note: do not include a trailing slash — that follows the symlink and unlink will refuse to remove a directory.
+unlink favicon
+```
 
 #### Link in HTML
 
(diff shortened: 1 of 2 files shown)


storybook/src/brand/assets/favicon.docs.mdx
```diff
@@ -42,13 +45,16 @@ ln -s ../node_modules/@amsterdam/design-system-assets/favicon favicon
...
+> Remove the symlink without affecting the original files by unlinking:
 >
 > ```sh
-> rm -r favicon
+> # May not be a path, but only the file or directory name without a trailing slash.
+> unlink favicon
 > ```
 
-Then link all these assets by placing the following tags in the `<head>`:
+#### Linking in HTML
```
VincentSmedinga: The other headings use the imperative mood (Install, link, copy) – would be good to be consistent here.

storybook/src/brand/assets/favicon.docs.mdx
```diff
@@ -60,9 +66,11 @@ Then link all these assets by placing the following tags in the `<head>`:
 </head>
 ```
 
+How this is done exactly depends on your project structure and build process. Follow the documentation of your framework or build tool for the best way to link static assets in the HTML.
+
 ### Submit Changes
```
VincentSmedinga:
```suggestion
### Submit changes
```

We use sentence case in headings, not title case. There are a few more: Edit Web Manifest, and SVG Icon and Apple Touch Icon (I don’t think they’re trademarks).

storybook/src/brand/assets/favicon.docs.mdx
```diff
@@ -18,21 +18,24 @@ This set covers all common browsers (Chrome, Safari, Edge, and Firefox).
...
+npm install @amsterdam/design-system-assets
 ```
 
 ### Link assets
 
-You can manually [symlink (symbolic link)](https://en.wikipedia.org/wiki/Symbolic_link#POSIX_and_Unix-like_operating_systems) and copy the files to the root or a `public` folder next to where the `index.html` is located in your project.
-An advantage of symlinking is that it tracks changes to the assets when there are package updates. Symlinks are basically shortcuts to files or directories.
+#### Copy or symlink assets
+
+Copy the icon assets found in `node_modules/@amsterdam/design-system-assets/favicon` to your project. Often the best place is the `public` folder, but the proper location might also be the root directory or any other directory where static files are stored. This depends on your project structure and build process.
```
VincentSmedinga: > happy to tighten further if either path should be recommended over the other

The links seem preferable because they accept updates. If there are no significant downsides, I would present this as the default approach, then mention copying as an alternative. Or more specifically for Windows, if that doesn’t support linking?

VincentSmedinga (commented):
Nice update.

Can we take this opportunity to remove the usage of `>` blockquotes to highlight sections? They look nice and fitting visually, but they’re not quotes as implied. These are the only files that have this. Elsewhere, we mostly use a bold prefix, like `**Note**:`. We could add a Note component later.

storybook/src/brand/assets/favicon.docs.mdx
RubenSibon: This has already been changed in the latest commit with AI suggestions.

storybook/src/brand/assets/favicon.docs.mdx
RubenSibon: That is a consideration, but my thinking was that copying works on all platforms. Therefore, it is the most reliable solution and should be offered as the default method.

But maybe switching it around makes more sense. I’ll think about it.

RubenSibon: I’ve thought about it and I think it’s better to leave it as I wrote it. Mainly for the reason I gave 29 minutes ago.

I think Aram would have agreed because he had issues with writing documentation that excludes or ignores a subset of developers.

VincentSmedinga: Fair enough!

VincentSmedinga (approved):

PR #2608, opened 2026-05-08
chore: Workflows to clean up stale demo environments and gh-pages previews

RubenSibon:
# Workflows to clean up stale demo environments and gh-pages previews

Closes DES-1831.

## What

Two new GitHub Actions workflows, plus a supporting Python script:

- **`.github/workflows/cleanup-environments.yml`** — deletes obsolete `demo-*` GitHub Environments via the REST API.
- **`.github/workflows/cleanup-gh-pages-demos.yml`** — deletes obsolete `demo-*` directories from the `gh-pages` branch and pushes the resulting commit.
- **`.github/scripts/cleanup_environments.py`** — the API-side cleanup logic, extracted so the workflow YAML stays thin and the script can be tested locally.

Both target the same set: a `demo-X` environment / directory is obsolete when no remote branch matches `X` after stripping the leading `<prefix>/` segment — the same transform `feature-branch-deploy.yml` applies when creating them.

## Why

`feature-branch-deploy.yml` provisions a `demo-*` GitHub Environment and a `demo-*/` directory on `gh-pages` per feature branch. These accumulate indefinitely: stale entries clutter the Environments list, inflate the gh-pages branch, and aren't covered by any existing automation. Cleaning them up has been ad-hoc and manual until now.

## How

**Triggers and dry-run defaults:**

| Workflow | Schedule trigger | Manual trigger | Real deletion requires |
|---|---|---|---|
| `cleanup-environments.yml` | Weekly, Sun 04:00 UTC (dry-run only) | Yes (defaults to dry-run) | Manual dispatch with `dry_run=false` |
| `cleanup-gh-pages-demos.yml` | On `release: published` where the tag ends `.0.0` (real deletion) | Yes (defaults to dry-run) | Either a major release, or manual dispatch with `dry_run=false` |

The gh-pages cleanup is tied to major releases (release-please tags ending `.0.0`, e.g. `design-system-react-v4.0.0`) rather than a fixed schedule, so demo previews remain available throughout a release cycle and are pruned at the natural cadence (~every 3 months).

**Safety guards (both workflows):**

- **Empty-branch-list abort.** If the source of truth (API or `git ls-remote`) returns zero branches, the run aborts with a clear error rather than marking every demo entry obsolete. The bash variant captures `git ls-remote` into a tempfile first because `mapfile -t < <(...)` does not propagate failures from process substitution under `set -e`.
- **Default to dry-run** for manual invocations; scheduled environment runs are always dry-run.
- **Max-delete cap of 25** in the Python script, with an `--i-really-mean-it` override flag.
- **Request timeouts** on every API call; URL-encoded environment name in the DELETE.
- **Improved error reporting** on API failures — the response body is included so org-level rejections (e.g. classic-PAT bans, SSO requirements) are visible in the log instead of a bare `403 Forbidden`.

**Token handling for `cleanup-environments.yml`:**

Deleting a GitHub Environment requires repo `Administration: write`, which is **not** an exposable scope on `GITHUB_TOKEN` via the workflow `permissions` block. The workflow resolves the token as `${{ secrets.ENV_ADMIN_TOKEN || secrets.GITHUB_TOKEN }}`, so listing and dry-runs work out of the box on `GITHUB_TOKEN`; only real deletion requires an `ENV_ADMIN_TOKEN` secret (fine-grained PAT or GitHub App installation token with `Administration: Read and write`). The header comment in the workflow explains this in full.

## Checklist

- [x] Add or update documentation — header comments in both workflow files document triggers, dry-run defaults, and token requirements.
- [x] ~~Add or update unit tests~~ — N/A (workflow infrastructure).
- [x] ~~Add or update stories~~ — N/A.
- [x] ~~Add or update exports in index.\* files~~ — N/A.
- [x] ~~Comment `/chromatic test` and verify visual regression tests pass~~ — N/A.

## Additional notes

- **Local end-to-end testing is pending.** Amsterdam org forbids classic PATs, and `gh auth token` issues a classic PAT, so I cannot run the Python script against the real repo locally yet. A fine-grained PAT has been requested; once approved I'll do a `--dry-run` against `Amsterdam/design-system` and confirm the output before flipping to a real run via the Actions UI. Reviewers comfortable merging without local validation should know this; reviewers who want a live demo should hold the merge.
- **`ENV_ADMIN_TOKEN` secret is not required for this PR to land.** Without it, scheduled and dry-run invocations work fine. To enable real environment deletion, an org admin needs to provision the secret in repo settings — that's a follow-up.
- The `cleanup-gh-pages-demos.yml` `if:` guard currently matches major-version tags from any release-please-managed package (`design-system-react-v*.0.0`, `design-system-tokens-v*.0.0`). If both packages bump majors in the same release-please run, the workflow fires twice — the second run is a harmless no-op. Worth flagging in case we want to narrow it later.


.github/workflows/cleanup-environments.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: Install dependencies
+        run: pip install 'requests==2.32.3'
+
+      - name: Cleanup environments
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          DRY_RUN: ${{ github.event.inputs.dry_run || 'true' }}
+          STALE_DAYS: ${{ github.event.inputs.stale_days || '0' }}
+        run: |
```
VincentSmedinga, line 67: We should probably write the script to a file and invoke it with `python script.py`.

.github/workflows/cleanup-environments.yml
```diff
@@ -0,0 +1,139 @@
...
+        uses: actions/setup-python@v5 # SHA pinned by Dependabot on first run
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install 'requests==2.32.3'
+
+      - name: Cleanup environments
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
VincentSmedinga: We'll need either:
  - `permissions: administration: write` (often disallowed for `GITHUB_TOKEN` depending on org settings), or
  - a PAT / GitHub App token with repo admin scope stored as a secret.

RubenSibon: You're right that admin scope is needed. Worth noting administration isn't a valid GITHUB_TOKEN permission scope at all, so option 1 isn't available to us — went with option 2: read ENV_ADMIN_TOKEN with a fallback to GITHUB_TOKEN so listing/dry-runs work unchanged.

.github/workflows/cleanup-environments.yml
RubenSibon: I made that change. Also added some guards. A dry-run is now the default. We’ve to manually trigger the actual run.

The trigger for a dry-run now is every major release (i.e. ending in `.0.0`). So we only have to do this clean-up once every 3 months or so.

RubenSibon:
@VincentSmedinga, I’ve set this to ready for review, but note the Additional notes.

In other words: currently the only feasible way to test these workflows is by merging this PR. It _should_ be safe because the scripts should only run dryly. But please, review carefully.

.github/scripts/cleanup_environments.py
```diff
@@ -0,0 +1,163 @@
...
+
+def paginate(session: requests.Session, url: str) -> Iterator[list | dict]:
+    page = 1
+    sep = "&" if "?" in url else "?"
+    while True:
+        r = _check(session.get(f"{url}{sep}per_page=100&page={page}", timeout=REQUEST_TIMEOUT))
+        yield r.json()
+        if "next" not in r.links:
+            return
+        page += 1
```
VincentSmedinga: `paginate()` mixes two pagination strategies: it builds the URL manually with an incrementing `page` counter, but uses `r.links["next"]` to decide when to stop. Pick one. Following the link header directly is simpler and removes the sep calculation:

```py
def paginate(session: requests.Session, url: str) -> Iterator[list | dict]:
    next_url: str | None = f"{url}?per_page=100"
    while next_url:
        r = _check(session.get(next_url, timeout=REQUEST_TIMEOUT))
        yield r.json()
        next_url = r.links.get("next", {}).get("url")
```

.github/workflows/cleanup-environments.yml
```diff
@@ -0,0 +1,67 @@
...
+          sparse-checkout: .github/scripts
+          sparse-checkout-cone-mode: false
+
+      - name: Set up Python
+        uses: actions/setup-python@42375524d0a3c83d1a8f3d1779a827123c5c2a43 # v5.4.0
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install 'requests==2.32.3'
```
VincentSmedinga: Version is pinned but not hash-verified. Consider a `requirements.txt` with `--generate-hashes` and `pip install --require-hashes -r ...` for stronger supply-chain assurance. Low priority for a cleanup script, but worth noting.

.github/workflows/cleanup-environments.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Cleanup environments
+        env:
+          # Falls through to GITHUB_TOKEN when the admin secret isn't set so
+          # dry-runs and listing still work; see permissions note above.
+          GITHUB_TOKEN: ${{ secrets.ENV_ADMIN_TOKEN || secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+        run: |
+          python .github/scripts/cleanup_environments.py \
+            ${{ (github.event.inputs.dry_run == 'false') && '--no-dry-run' || '--dry-run' }} \
+            --stale-days "${{ github.event.inputs.stale_days || '0' }}"
```
VincentSmedinga: When triggered by schedule, `github.event.inputs.dry_run` is empty, so this ternary correctly resolves to `--dry-run`. That behaviour is intentional but non-obvious — a short comment here would help a future editor who might add `default: "false"` to the schedule trigger thinking it controls this path.

.github/workflows/cleanup-gh-pages-demos.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Configure git
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Remove stale demo directories
+        env:
+          # Release-triggered runs always perform the deletion; manual runs
+          # honour the workflow input (defaults to dry-run).
+          DRY_RUN: ${{ github.event_name == 'release' && 'false' || github.event.inputs.dry_run || 'true' }}
```
VincentSmedinga: `DRY_RUN` is set to the same expression in both this step and the "Commit and push" step (line 102). Moving it to a `env:` block at the job level would remove the duplication and the risk of the two definitions drifting.

.github/workflows/cleanup-gh-pages-demos.yml
```diff
@@ -0,0 +1,113 @@
...
+            awk '{print $2}' "$ls_remote_out" \
+              | sed -E 's|^refs/heads/||; s|^[^/]*/||'
+          )
+
+          if [ "${#active_short_names[@]}" -eq 0 ]; then
+            echo "ERROR: remote branch list is empty — refusing to proceed." >&2
+            exit 2
+          fi
+
+          for dir in demo-*/; do
```
VincentSmedinga, line 132: The Python script supports `--stale-days` to skip recently-updated environments. The bash cleanup has no equivalent — it deletes all stale directories regardless of age. Worth either adding parity or noting the asymmetry in the workflow header comment.

VincentSmedinga (commented):
The "local end-to-end testing is pending" note in the PR description should be a blocker before merge. Would like to see at least a --dry-run pass against Amsterdam/design-system confirming the branch → short-name transform produces the expected output and the environment/directory lists look correct before this lands.

Also, let’s discuss whether we’re really sure to introduce a Python script while we both aren’t very comfortable with that language. To be honest, all my comments are AI-assisted.

RubenSibon:
> Also, let’s discuss whether we’re really sure to introduce a Python script while we both aren’t very comfortable with that language. To be honest, all my comments are AI-assisted.

True! We can ask a Pythonista from Engineering Enablement to review this code. Or any other experiences Python developer in the organisation.

RubenSibon:
Closing this in favour of #2640 which actually connected to an open Jira issue.

@VincentSmedinga’s comments in this PR have been addressed. I also pushed those and more changes to THIS branch/PR so that you can see the changes with your comments.

PR #2609, opened 2026-05-11
fix: Remove obsolete font format files (.eot and .woff)

VincentSmedinga:
# Describe the pull request

## What

- Removes the `.eot` and `.woff` font files from our build.
- Moves the remaining `.woff2` files into the main directory.
- This is not breaking; the `index.css` file is the API. Also, see ‘why’.

## Why

- Internet Explorer 11 is dead; modern browsers don't need EOT.
- All modern browsers support WOFF2; WOFF is redundant.
- Makes our build output significantly smaller.
  - (even though compressed-size-action doesn’t correctly indicate that below)

## How

n/a

## Checklist

- [x] Add or update stories

## Additional notes

- n/a


RubenSibon (approved):
Seems good to me!

One question about usage of Light and Extra Bold. I did a quick search on these terms with and without spaces and in their numeric form (i.e. `300` and `800`) and it seems that we do not use those variants*? Can they be removed from the project if we never use them and do not suggest users to use them?

(*) I [did find one instance](https://github.com/Amsterdam/design-system/blob/e564df39610d285f2575d62771788c7da7501454/packages-proprietary/tokens/src/brand/ams/typography.tokens.json#L36) where we use a font weight of `800` in the Typography tokens, but the token is called `bold` and not `extra-bold`. Is that correct?

VincentSmedinga:
> One question about usage of Light and Extra Bold. I did a quick search on these terms with and without spaces and in their numeric form (i.e. `300` and `800`) and it seems that we do not use those variants*? Can they be removed from the project if we never use them and do not suggest users to use them?

We don’t use Light and Bold. We use the Extra Bold weight for bold text. But this PR was to remove unused formats, not to filter font weights as well. I think we better leave them in for completeness. No one is downloading them unless a consumer decides to use it. And we’re on track to use regular Bold for bold text.

> (*) I [did find one instance](https://github.com/Amsterdam/design-system/blob/e564df39610d285f2575d62771788c7da7501454/packages-proprietary/tokens/src/brand/ams/typography.tokens.json#L36) where we use a font weight of `800` in the Typography tokens, but the token is called `bold` and not `extra-bold`. Is that correct?

Yes, the public API is `regular` and `bold` for easy understanding. Under the hood, they use the regular (CSS 400) and extra bold (800) font weights of Amsterdam Sans.

PR #2620, opened 2026-05-13
chore(docs): Clarify Field Set guidelines and improve example stories

VincentSmedinga:
# Clarify Field Set guidelines and improve example stories

## What

- Rewrite the Field Set README so the guidelines explicitly cover grouping Radios and Checkboxes, and clarify that `optional` only applies to such lists.
- Drop the redundant `optional: false` default from the FieldSet story args.
- Replace the clipped placeholder descriptions in the Radio “In a Field Set” stories with complete sentences.
- Update the DateInput `MemorableDateWithValidation` story to demonstrate partial validation (one missing field) instead of marking every field invalid.

## Why

- The previous Field Set guidelines didn’t make it clear when `optional` is meaningful, and the example stories used fragments where full sentences are more representative of real usage.
- The DateInput validation example also didn’t reflect a realistic partial-validation scenario.

## How

Documentation and Storybook-only changes; no component behavior or API was touched.

## Checklist

- [x] Add or update stories

## Additional notes

- n/a

### DIFF
--- a/packages/css/src/components/field-set/README.md
+++ b/packages/css/src/components/field-set/README.md
@@ -10,7 +10,7 @@ Groups a set of Radio buttons or Checkboxes, or other related form fields.
   This allows them to share a label, description, and error message, while each option still has its own label.
 - A Field Set can also express the relation between a group of inputs or other controls.
   For instance, you might want to group several text inputs into one Field Set when requesting an address or various parts of a name.
-- Mark a Field Set as `optional` only for a list of Radios or Checkboxes, and if checking any of them is not necessary.
+- Mark a Field Set as `optional` only for a list of Radios or Checkboxes, and if selecting any option is not necessary.
 - There’s no need to add white space between the children of a Field Set.
   The component does this automatically.
 
--- a/packages/css/src/components/field-set/README.md
+++ b/packages/css/src/components/field-set/README.md
@@ -2,14 +2,17 @@
 
 # Field Set
 
-A component to group related form inputs.
+Groups a set of Radio buttons or Checkboxes, or other related form fields.
 
 ## Guidelines
 
-- Use Field Set when you need to show a relationship between multiple form inputs. For example, you may need to group a set of text inputs into a single Field Set when asking for an address.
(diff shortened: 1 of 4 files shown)


packages/css/src/components/field-set/README.md
```diff
@@ -2,14 +2,17 @@
...
 
-- Use Field Set when you need to show a relationship between multiple form inputs. For example, you may need to group a set of text inputs into a single Field Set when asking for an address.
-- Set `optional` to `true` if all inputs in a Field Set are not required.
-- Use `hint` to show a custom hint text.
-- There’s no need to add white space between the children of a Field Set. The component does this automatically.
+- Group a collection of [Radio](/docs/components-forms-radio--docs) buttons or [Checkboxes](/docs/components-forms-checkbox--docs) into a Field Set.
+  This allows them to share a label, description, and error message, while each option still has its own label.
+- A Field Set can also express the relation between a group of inputs or other controls.
+  For instance, you might want to group several text inputs into one Field Set when requesting an address or various parts of a name.
+- Mark a Field Set as `optional` only for a list of Radios or Checkboxes, and if selecting any option is not necessary.
```
alimpens, line 13: I don't think this is quite true. For example, the memorable date example can be optional in its entirety. You would / should use `optional` there.

storybook/src/components/DateInput/DateInput.stories.tsx
```diff
@@ -107,19 +107,26 @@ export const MemorableDateWithValidation: Story = {
   render: (args) => (
     <FieldSet aria-describedby="description-b error-b" invalid legend="Wanneer ben je geboren?">
       <Paragraph id="description-b">Bijvoorbeeld 1 1 2000.</Paragraph>
-      <ErrorMessage id="error-b">De datum moet in het verleden liggen.</ErrorMessage>
+      <ErrorMessage id="error-b">Vul alle drie de velden in.</ErrorMessage>
```
alimpens, line 110: I think this error message can be a bit clearer. How about 'Vul in in welk jaar je geboren bent.'?

PR #2630, opened 2026-05-19
feat: Sort table with links

RubenSibon:
# Sort table with links

## What

Adds a `SortingWithLinks` story to the internal Table Page, demonstrating how to sort a table by clicking links — a one-click alternative to the existing `SortingWithSelect` story, suited to tables with a small fixed set of sort options.

## Why

Picks up DES-1762 / the `sorteren-via-links` prototype. When there are only a few sort options, links let a user pick a sort with a single click instead of the select-then-confirm flow.

## How

- Mirrors the structure of `SortingWithSelect` (same `Grid` / `Heading` / `Table` scaffold and the same `?sort=…` URL-param flow via `sortAddresses`).
- A `Row` holds the label and three sort options. The currently applied sort renders as `<strong aria-current="true">` (bold, no link) so we don't link the user to where they already are — mirroring how `Pagination` treats the current page.
- Registers the new story in `stories/index.ts` and `TablePage.stories.tsx`.

## Visual review

Inspect the current visual output in the [Storybook deploy](https://designsystem.amsterdam/demo-DES-1762-table-sorting-via-links/?path=/story/pages-internal-table-page--sorting-with-links).

## Checklist

- [x] Add or update stories

## Additional notes

- Ticket: DES-1762.
- Open for review — the right component setup for the row of sort options isn't obvious to me. I'm unsure about nesting `Link`s in a `Row` inside a `Paragraph`. `StandaloneLink` isn't an ideal alternative either, since it requires an icon — a row of icons next to the label looks messy and distracting. Feedback on the intended pattern is welcome.


RubenSibon:
I’m going to look into using Action Group instead of Links inside a Row inside a Paragraph.

RubenSibon:
Hi @bas-kroese, with the components that we currently have it was quite difficult to get a satisfying look and HTML structure for sorting a Table’s columns with links.

Have a look in the [Storybook demo](https://designsystem.amsterdam/demo-DES-1762-table-sorting-via-links/?path=/story/pages-internal-table-page--sorting-with-links).

Using Standalone Links is not a good-looking solution because they require icons. So we resorted to using inline Links nested inside a Paragraph which is again nested in a Row.

We need the Paragraph so that the text and links get the proper styling and font family.

We need the Row to align the Paragraph to right end of the Grid Cell.

> I’m going to look into using Action Group instead of Links inside a Row inside a Paragraph.

Using Action Group does not anything meaningful and is superfluous when we also use Row to align to the right (Action Group can’t do that).

With @VincentSmedinga we came to the conclusion that this is the moment that we might need some kind of wrapper component for a line of text and link elements. We also might need such a thing for [Article](https://designsystem.amsterdam/?path=/story/pages-public-article-page--default) and Card bylines. Think of a publish date, last edit date, the author, etcetera. But in this case we’re talking more about a navigation list. Certainly something to think more about in a refinement!

RubenSibon:
Closing this as we’ve decided not to pursue this pattern further.

PR #2632, opened 2026-05-19
chore(lint): Scope React lint rules to packages that use React

VincentSmedinga:
# Scope React lint rules to packages that use React

## What

Split the root ESLint config so the `eslint-plugin-react` plugin, settings, and rules are no longer applied to every JS/TS file in the repo. General JS/TS rules still apply repo-wide; React-specific rules now apply only to `packages/react`, `packages-proprietary/react-icons`, and `storybook`.

## Why

The root config set `react: { version: 'detect' }` globally, but React isn’t installed at the root. Every ESLint run emitted:

> Warning: React version was set to "detect" in eslint-plugin-react settings, but the "react" package is not installed. Assuming latest React version for linting.

Scoping the React block to packages that actually depend on React lets version detection resolve from those packages’ own `node_modules`, silencing the warning. It also stops the React rules from being applied to non-React code (token build scripts, SVGO configs).

## How

Extracted the React plugin, `react: { version: 'detect' }` setting, recommended rules, and `react/react-in-jsx-scope: 'off'` override into a new config block in [eslint.config.mjs](eslint.config.mjs) with a `files` glob limited to the three React-using packages.

The remaining general block keeps the same `files: ['**/*.{js,jsx,ts,tsx}']` and the parser/JSX/import-resolver settings, which are still needed wherever `.jsx`/`.tsx` files appear.

## Checklist

- [x] Add or update stories

## Additional notes

Verified locally: `pnpm run lint:js` no longer emits the React version warning, and the remaining errors/warnings match `develop`.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -71,9 +71,9 @@ export default tseslint.config(
     ...eslintConfigPrettier,
   },
 
-  // JavaScript, TypeScript and React
+  // JavaScript and TypeScript
   {
-    name: 'amsterdam-design-system/javascript-typescript-react',
+    name: 'amsterdam-design-system/javascript-typescript',
     files: ['**/*.js', '**/*.jsx', '**/*.ts', '**/*.tsx'],
     plugins: {
       '@typescript-eslint': tsPlugin,
@@ -81,7 +81,6 @@ export default tseslint.config(
       import: importPlugin,
       vitest,
       perfectionist,
-      react,
     },
     languageOptions: {
       parser: tsParser,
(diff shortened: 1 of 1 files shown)


RubenSibon (commented):
Good catch! This would certainly be an improvement.

I want to test this out locally before I approve.

Also:

- My fingers are itching to refactor the ESLint config all together: I did the migration from *rc files to the flat file config, but since then we haven’t really looked at updating the config with new insights and developments. Many plugins have surely migrated to the flat file config by now. There will also be many instances of better ways to structure the config for a monorepo like ours^. One thing that I think would be better is to actually have React specific config in the React packages and then import that partial from there.
- There’s the [upgrade to ESLint 10](https://gemeente-amsterdam.atlassian.net/browse/DES-1693) that we could pick up sometime.

^ See also this discussion on the flat file config and monorepos: https://github.com/eslint/eslint/discussions/16960

RubenSibon (approved):
It works well!

I’ve made a ticket to address (my) other issues with our current ESLint config.

PR #2637, opened 2026-05-21
chore(ESLint): Scope configs per package & modernize, clean up rules

RubenSibon:
# ESLint: Scope configs per package and modernize, clean up rules

## What

Overhaul of the ESLint setup for DES-1852, organised around three themes:

### 1. Per-package config split

Each workspace package now owns the rules that apply to it.

- `packages/react/eslint.config.mjs` owns the React preset (plugin, settings, recommended rules) and exports `reactPreset` for other React-using packages to extend.
- `storybook/eslint.config.mjs` owns the Storybook preset and the MDX configuration (every `.mdx` file in the repo is a Storybook doc).
- `packages-proprietary/react-icons/eslint.config.mjs` is new, scoped to that package and extending `reactPreset`.
- Root `eslint.config.mjs` is now purely structural: globs, plugins, extends, settings — no rule lists.
- Plugin devDependencies moved alongside the configs that use them (`eslint-plugin-react` → React package, `eslint-plugin-storybook` + `eslint-plugin-mdx` → Storybook).

### 2. Address `tseslint.config()` deprecation

- Switched the root config wrapper from `tseslint.config(...)` to `defineConfig(...)` from `eslint/config` (ESLint ≥ 9.22). Drop-in replacement; the typescript-eslint docs recommend this migration.
- Removed the now-unused `typescript-eslint` meta package from root devDependencies.

### 3. Further improvements

- **`defineConfig` `extends` field** instead of spreading preset rules inline (where possible).
- **`--report-unused-disable-directives` moved from CLI to config** via `linterOptions.reportUnusedDisableDirectives: 'error'`. The flag is gone from `lint:js`.
- **Working TypeScript path resolution.** `import/resolver` extensions were missing leading dots and never matched `.ts`/`.tsx`. Switched to `eslint-import-resolver-typescript`; `import/no-cycle` and `import/no-named-as-default` now actually evaluate.
- **`eslint-plugin-import` → `eslint-plugin-import-x`** — the original is unmaintained; `import-x` is the active fork with the same rule set and faster TS resolver.
- **Vitest preset scoped to test files** (`**/*.{test,spec}.{js,jsx,ts,tsx}`) instead of leaking across the repo.
- **`globals.node` added** to language options so Node scripts (plopfile, rollup configs, etc.) no longer require `/* eslint-disable no-undef */`. Two such file-level directives were removed as a side effect.
- **JSX parser option scoped to React fragment** instead of every JS/TS file.
- **TS-aware twins for core rules**: `no-loop-func` → `@typescript-eslint/no-loop-func`, `no-invalid-this` → `@typescript-eslint/no-invalid-this`, `no-unused-expressions` → `@typescript-eslint/no-unused-expressions`.
- **Pruned deprecated rules**: `no-new-symbol` (removed from ESLint 9 recommended; replaced by `no-new-native-nonconstructor`), `no-return-await` (deprecated in 8.46).
- **`react/prop-types: 'off'`** added explicitly to `reactPreset` and the legacy `storybook/.eslintrc.json` deleted — that file had been dead since the flat-config migration.
- **`prettier` config placed last** in the config array (per its own docs).
- **`eslint.rules.mjs`** extracted from the root config: rules grouped by concern (`tsRules`, `coreRules`, `baselineRules`, `importRules`, `perfectionistRules`). Root config dropped from 262 to ~110 lines, all structural.
- **Storybook `flat/recommended` properly spread** instead of accidentally relying on `defineConfig`'s nested-array flattening.
- **JSON block ignores `pnpm-lock.yaml`** to skip the lockfile.

## Why

The ESLint config had grown organically since the flat-config migration and had drifted in several directions:

- Plugin configuration was centralised in the root config, even for plugins only used by one package — making the root file long and obscuring which package depends on what.
- `tseslint.config()` was emitting a deprecation warning at every editor load.
- The TypeScript path resolver was silently broken (`import/no-cycle` etc. never evaluated against `.ts`/`.tsx` imports), so the rules looked enforced but weren't.
- `eslint-plugin-import` is unmaintained.
- Several rules were redundant (TS-aware versions exist), deprecated (`no-new-symbol`, `no-return-await`), or scoped too broadly (`vitest` recommended applied repo-wide).
- A legacy `storybook/.eslintrc.json` had been dead since the flat-config migration but stayed in the tree as a confusing artefact.

The ticket scoped this work to: (1) place config alongside the code it lints, (2) clear the `tseslint.config` deprecation, (3) hunt for further improvements. This PR delivers all three.

## How

The work landed across small, reviewable commits — one per concern — so the history can be bisected if a regression surfaces. Notable approach decisions:

- **Per-package `eslint.config.mjs` files imported by the root.** ESLint's experimental `unstable_config_lookup_from_file` would allow each directory to use its nearest config, but it remains experimental. JS imports of per-package fragments give the same locality without leaving the supported path. Globs in fragments stay root-relative because ESLint resolves them relative to the importing config — documented in a comment at the top of each fragment.
- **`reactPreset` exported from the React package** so any package shipping React code (Storybook, React-icons) can extend the same baseline without re-declaring plugin/settings/rules. Each consumer applies its own `files` glob.
- **`defineConfig`'s `extends` field used for plugin presets that don't themselves use `extends`.** `@typescript-eslint/eslint-plugin`'s `configs.recommended` already uses `extends` internally, and `defineConfig` rejects nested `extends`, so it stays as a `...rules` spread with a one-line comment explaining why.
- **`@typescript-eslint/only-throw-error` evaluated but skipped** — it requires type-aware linting, which would be a much bigger investment. Left in the follow-up section instead.
- **Lint output preserved at baseline (10 warnings, 0 errors)** throughout the PR; each commit verified individually.

## Checklist

- [x] Add or update unit tests — not applicable (build-system change).
- [x] Add or update stories — not applicable.
- [x] Comment `/chromatic test` and verify visual regression tests pass — not applicable (no visual changes).

## Additional notes

- Ticket: DES-1852.
- Verification: `pnpm lint:js` → 10 warnings, 0 errors (identical to the pre-PR baseline: 7× `no-explicit-any`, 3× `vitest/no-disabled-tests`). `pnpm -r run lint` (typecheck across all packages) passes. All tests in the Vitest suite also pass.
- Follow-ups intentionally out of scope:
    - `eslint-plugin-react-hooks` adoption — DES-1854, landed separately because the v7 preset surfaced real React-hook bugs to fix.
    - `react/prefer-read-only-props` adoption — patch ready, separate ticket because it auto-fixes ~90 component files.
    - Type-aware linting (`parserOptions.projectService`) — would unlock `only-throw-error`, `no-floating-promises`, `no-unsafe-*`. Deserves its own ticket; slower lint and requires every file in a tsconfig include.
- Areas where extra review attention is welcome: the per-package fragment pattern (`packages/react/eslint.config.mjs`, `storybook/eslint.config.mjs`, `packages-proprietary/react-icons/eslint.config.mjs`) — first time we're splitting the config this way, so feedback on whether the ownership model holds up is appreciated.


packages-proprietary/react-icons/eslint.config-partial.mjs
VincentSmedinga, line 1: We shouldn’t name these extracted files `eslint.config.mjs` – that's ESLint’s reserved auto-discovery filename if it searches from the current working directory upward.

For example, the scripts that generate icons or logos run `eslint` with their working directory inside a package, so they will pick up the local `eslint.config.mjs` and treat it as the complete config, which it isn’t. These scripts now fail.

IDEs have the same issue.

Let’s use a non-magic name like `eslint.fragment.mjs` or `eslint.config.fragment.mjs`.

eslint.config.mjs
```diff
@@ -3,251 +3,92 @@ import json from '@eslint/json'
...
-  groups: ['x-small', 'small', 'medium', 'large', 'x-large', '2x-large'],
-  useConfigurationIf: {
-    allNamesMatchPattern:
-      '^(x-small|xs|small|s|narrow|phone|min|start|medium|m|tablet|large|l|wide|desktop|max|x-large|xl|2x-large|2xl|end)$',
-  },
-}
+import { baselineRules, coreRules, importRules, perfectionistRules, tsRules } from './eslint.rules.mjs'
+import reactIconsConfig from './packages-proprietary/react-icons/eslint.config.mjs'
+import reactConfig from './packages/react/eslint.config.mjs'
+import storybookConfig from './storybook/eslint.config.mjs'
```
VincentSmedinga: These `*Config` parts should be updated to match their new file names.

packages/react/eslint.config.mjs
```diff
@@ -0,0 +1,38 @@
+import react from 'eslint-plugin-react'
+
+/* `reactPreset` is the un-scoped React rule set. Other packages that ship React code
+ * (Storybook, the React-icons package) import it and apply it to their own globs.
+ */
```
VincentSmedinga:
```suggestion
/* `The unscoped React ruleset. Other packages that ship React code apply it to their own globs. */
```

storybook/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -4,6 +4,7 @@
  */
 
 import type { Meta, StoryObj } from '@storybook/react-vite'
+import type { MouseEvent as ReactMouseEvent } from 'react'
```
VincentSmedinga: Does the rename serve a purpose? We don’t usually do this. Isn’t `MouseEvent<HTMLInputElement>` below just fine? Multiple occurrences.

storybook/src/components/SearchField/SearchField.stories.tsx
```diff
@@ -5,6 +5,7 @@
 
 import type { SearchFieldProps } from '@amsterdam/design-system-react/src'
 import type { Meta, StoryObj } from '@storybook/react-vite'
+import type { ChangeEvent as ReactChangeEvent } from 'react'
```
VincentSmedinga: Another unnecessary rename, I think.

VincentSmedinga (commented):
Yes, that’s a sweet cleanup! One sneaky regression though, simple to fix, and a few nitpicks.

storybook/src/components/Checkbox/Checkbox.stories.tsx
RubenSibon: Well, it’s to distinguish these events imported from the `react` package from HTML events such as [MouseEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent).

But not renaming will work fine too. What do you prefer?

storybook/src/components/SearchField/SearchField.stories.tsx
RubenSibon: Same consideration as above.

storybook/src/components/Checkbox/Checkbox.stories.tsx
VincentSmedinga: Let’s drop the alias and use plain `MouseEvent<HTMLInputElement>`.

React’s synthetic event types intentionally share the DOM global names, and shadowing the native `MouseEvent` within the module is the idiomatic pattern — the distinction only matters when a file uses both, which these stories don’t.

It also keeps us consistent with the React package itself (`TabsButton`, `Dialog` both import `MouseEvent` from `react` unaliased).

packages-proprietary/react-icons/eslint.config-partial.mjs
VincentSmedinga: I think I understand why you chose `eslint.react.config.mjs` etc. Does feel a bit awkward to repeat the package name in the file name – it’s already there in the path.

./packages-proprietary/**react-icons**/eslint.**react-icons**.config.mjs

We don’t do it anywhere else and we slightly lose the information that it’s partial config.

But I could live with it – up to you whether you want to use `fragment` or `partial` or not.

RubenSibon: Alright, I’ll change it to something that does not repeat the package name. So, `fragment`, `partial` or something else?

RubenSibon: Adjusted! Hopefully this can be merged quickly so that we can move on with the other 3 follow-up PRs.

VincentSmedinga (approved):

PR #2639, opened 2026-05-21
chore(pnpm): post-v11 cleanup and explicit build step

RubenSibon:
# pnpm: cleanup and explicit build step after upgrade to pnpm v11 

## What

- Drop the root `postinstall` script and add a `Build workspace packages` step (running `pnpm build`) to the five CI workflows that previously relied on it: [main-branch-deploy.yml](.github/workflows/main-branch-deploy.yml), [feature-branch-deploy.yml](.github/workflows/feature-branch-deploy.yml), [lint-test.yml](.github/workflows/lint-test.yml), [chromatic.yml](.github/workflows/chromatic.yml), [publish.yml](.github/workflows/publish.yml).
- Update [CONTRIBUTING.md](CONTRIBUTING.md) so first-time setup is `pnpm install && pnpm build`, with a note explaining why `pnpm install` alone no longer rebuilds.
- Align [pnpm-workspace.yaml](pnpm-workspace.yaml) with pnpm v11 supply-chain defaults: drop `minimumReleaseAge: 1440` (already the v11 default) and the `allowBuilds: { …: false }` map (a no-op under v11's default `strictDepBuilds: true`). Add `blockExoticSubdeps: true` to refuse transitive deps from git URLs, tarballs, or other non-registry sources.
- Remove the npm/yarn-style `workspaces` array from [package.json](package.json); pnpm reads workspace globs from `pnpm-workspace.yaml` and the two lists could drift silently.
- Replace `pnx lint-staged` with `pnpm exec lint-staged` in [.husky/pre-commit](.husky/pre-commit) — `lint-staged` is a local devDep so `pnpm exec` skips the registry fetch that `pnx` (= `pnpm dlx`) does.
- Document the preference for `pnx` over `npx` for one-off tools in [documentation/continuous-integration.md](documentation/continuous-integration.md).

## Why

After the pnpm 10 → 11 upgrade, `pnpm install` short-circuits to a no-op when the lockfile and store are already in sync, and **skips lifecycle scripts including `postinstall`** in that case. The root `postinstall` was our build trigger for workspace `dist/` directories, so the common post-pull flow (`git pull` → `pnpm install`) silently left builds stale. Demonstrated locally: delete `packages/css/dist/`, run `pnpm install`, get "Already up to date" in ~50 ms with no rebuild — even `pnpm install --force` doesn't reinstate it.

The rest is incidental cleanup that piled up around the v11 migration: settings that duplicate v11 defaults, a dead workspaces array that npm/yarn would use but pnpm ignores, and an opportunity to add `blockExoticSubdeps` while we were touching the file.

## How

- Reproduced the postinstall regression by wiping `packages/*/dist`, `packages-proprietary/*/dist`, and `storybook/dist`, then running `pnpm install` and confirming the dirs stayed missing. `pnpm build` after that restores all five expected artifacts.
- For the CI workflows: `pnpm build` (root script = `pnpm -r run build`) is added between install and any consumer step. Two existing workflows (`pr-bundle-size.yml`, `pr-coverage.yml`) already use `--ignore-scripts` plus an explicit build and were not touched.
- In [publish.yml](.github/workflows/publish.yml), the new build step sits before the existing `git reset --hard HEAD && git clean -fd`. Verified `dist/` is gitignored (line 10), so `clean -fd` leaves the just-built artifacts in place.
- For `pnpm-workspace.yaml`: verified via [pnpm.io/settings](https://pnpm.io/settings) that `minimumReleaseAge` defaults to 1440 in v11 and that `strictDepBuilds: true` blocks unlisted packages by default, making the `false` entries redundant. Ran `pnpm install` after the changes — clean, no warnings about unrecognised settings, no `blockExoticSubdeps` rejections against the current tree.
- Verified the overrides (`@tootallnate/once@<3.0.1` and `tar: ">=7.5.11"`) are still load-bearing with `pnpm why` — they hold transitive deps at safe versions, so they stayed.

## Checklist

- [x] Add or update unit tests — n/a, no runtime code changed
- [x] Add or update documentation — [CONTRIBUTING.md](CONTRIBUTING.md) updated; [documentation/continuous-integration.md](documentation/continuous-integration.md) updated
- [x] Add or update stories — n/a
- [x] Add or update exports in `index.*` files — n/a
- [x] Comment `/chromatic test` and verify visual regression tests pass — not expected to change visuals, but worth running once to confirm the new CI flow produces a green Chromatic build

## Additional notes

- Resolves DES-1850.
- After merge, contributors pulling `develop` for the first time will need to run `pnpm install && pnpm build` rather than just `pnpm install`. Documented in CONTRIBUTING.md.
- The fastest place to spot regressions is the first run of [lint-test.yml](.github/workflows/lint-test.yml) on this PR — if React tests pass there, the explicit-build path is wired correctly.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -63,8 +63,8 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Build workspace packages
-        run: pnpm build
+      - name: Build packages
+        run: pnpm --filter=!storybook -r run build
 
       - name: Run Chromatic
         uses: chromaui/action@1bbb5819252d9782cb7d77122d9becae467def18 # v16.10.1
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
+        run: pnpm build
+
       - name: Run Chromatic
         uses: chromaui/action@1bbb5819252d9782cb7d77122d9becae467def18 # v16.10.1
         with:
--- a/.github/workflows/feature-branch-deploy.yml
+++ b/.github/workflows/feature-branch-deploy.yml
@@ -48,10 +48,10 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Build Storybook
+      - name: Build workspace packages
         env:
           STORYBOOK_BUILD_PATH: demo-${{ env.BRANCH_NAME }}
-        run: pnpm --filter=storybook build
+        run: pnpm build
 
       - name: Create a version mark
         run: |
--- a/.github/workflows/feature-branch-deploy.yml
+++ b/.github/workflows/feature-branch-deploy.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Build workspace packages
+      - name: Build packages and Storybook
         env:
           STORYBOOK_BUILD_PATH: demo-${{ env.BRANCH_NAME }}
         run: pnpm build
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -48,8 +48,8 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Build workspace packages
-        run: pnpm build
+      - name: Build packages
+        run: pnpm --filter=!storybook -r run build
 
       - name: "Continuous Integration: lint"
         run: |
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -48,6 +48,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
+        run: pnpm build
+
       - name: "Continuous Integration: lint"
         run: |
           pnpm run --if-present lint
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,6 +49,10 @@ jobs:
         if: ${{ steps.release.outputs.releases_created == 'true' }}
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        run: pnpm build
+
       # Reset any tree modifications from earlier steps so pnpm's git checks
       # pass without us having to disable them with --no-git-checks.
       - name: Reset working tree before publish
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,9 +49,9 @@ jobs:
         if: ${{ steps.release.outputs.releases_created == 'true' }}
         run: pnpm install --frozen-lockfile
 
-      - name: Build workspace packages
+      - name: Build packages
         if: ${{ steps.release.outputs.releases_created == 'true' }}
-        run: pnpm build
+        run: pnpm --filter=!storybook -r run build
 
       # Reset any tree modifications from earlier steps so pnpm's git checks
       # pass without us having to disable them with --no-git-checks.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,7 +172,7 @@ To enable correct validation and to fix lint/style errors on save, add this to y
 - Build workspace packages: `pnpm build`
 - Run storybook: `pnpm start`
 
-Re-run `pnpm build` after pulling source changes — `pnpm install` is a no-op when the lockfile is unchanged, so it will not rebuild `dist/` for you.
+After pulling updates that include source changes, run `pnpm build` again to rebuild the packages.
 
 ### Run unit tests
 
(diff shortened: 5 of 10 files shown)


documentation/continuous-integration.md
```diff
@@ -29,7 +29,7 @@ We have used [pin-github-action](https://www.npmjs.com/package/pin-github-action
 Use this command to get the correct format:
 
 ```sh
-npx pin-github-action -c " {ref}" /path/to/workflow.yaml
+pnx pin-github-action -c " {ref}" /path/to/workflow.yaml
```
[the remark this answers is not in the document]
RubenSibon: Not correct since pnpm 11: https://pnpm.io/cli/pnx

.github/workflows/chromatic.yml
```diff
@@ -63,6 +63,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
```
VincentSmedinga: Do we need the word ‘workspace’? The names of our actions and jobs always puzzle me a bit, so I’d say the clearer / shorter the better.

CONTRIBUTING.md
```diff
@@ -169,8 +169,11 @@ To enable correct validation and to fix lint/style errors on save, add this to y
 ### Run storybook
 
 - Install dependencies: `pnpm install`
+- Build workspace packages: `pnpm build`
 - Run storybook: `pnpm start`
 
+Re-run `pnpm build` after pulling source changes — `pnpm install` is a no-op when the lockfile is unchanged, so it will not rebuild `dist/` for you.
```
VincentSmedinga: This sentence is heavy on jargon, can we make it easier to understand?

pnpm-workspace.yaml
```diff
@@ -8,9 +7,15 @@ overrides:
   tar: ">=7.5.11"
 engineStrict: true
 savePrefix: ""
+# Refuse to resolve transitive deps from git URLs, tarballs, or other non-registry sources.
+# Supply-chain hardening — see https://pnpm.io/supply-chain-security
+blockExoticSubdeps: true
```
VincentSmedinga: This doesn’t seem necessary: `true` is already [the default value](https://pnpm.io/settings#blockexoticsubdeps) for this setting. Removing `minimumReleaseAge` relies on that default, too.

.github/workflows/lint-test.yml
```diff
@@ -48,6 +48,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
+        run: pnpm build
```
VincentSmedinga: This workflow doesn’t need to build Storybook, I think.

.github/workflows/chromatic.yml
```diff
@@ -63,6 +63,9 @@ jobs:
...
+        run: pnpm build
```
VincentSmedinga: The Chromatic action below will build Storybook, so we can filter it out here.

.github/workflows/feature-branch-deploy.yml
```diff
@@ -48,10 +48,10 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      - name: Build Storybook
+      - name: Build workspace packages
```
VincentSmedinga: Would you say Storybook is a package? If not, it’s clearer to say ‘Build packages and Storybook’ here, to further highlight the difference with the workflows that exclude building Storybook.

.github/workflows/publish.yml
```diff
@@ -49,6 +49,10 @@ jobs:
         if: ${{ steps.release.outputs.releases_created == 'true' }}
         run: pnpm install --frozen-lockfile
 
+      - name: Build workspace packages
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        run: pnpm build
```
VincentSmedinga: Storybook is not published, so we can filter it from building here.

VincentSmedinga (commented):
On a roll! 🧹

.github/workflows/chromatic.yml
RubenSibon: I guess we don’t need it. I will remove all instances.

VincentSmedinga (approved):

RubenSibon:
> /chromatic test

I’m looking into why the [Chromatic build is failing](https://github.com/Amsterdam/design-system/actions/runs/26287178011/job/77377636080#step:8:793).

RubenSibon:
> I’m looking into why the [Chromatic build is failing](https://github.com/Amsterdam/design-system/actions/runs/26287178011/job/77377636080#step:8:793).

What is happening is a workflow-context mismatch.

The real failure is: Storybook cannot resolve `@amsterdam/design-system-tokens/dist/index.css` from `preview.tsx`.
That CSS file is generated during the tokens build in `build.js`.

Why it fails in your Chromatic run:

1. Your PR branch workflow includes an explicit build step before Chromatic in `chromatic.yml`.
2. But the failing run was triggered via issue_comment (/chromatic test).
3. For issue_comment, GitHub Actions uses the workflow definition from the default branch (develop), not from your PR branch.
4. On develop, this workflow still runs install then Chromatic without the new build step.
5. Since `postinstall` was removed in this PR, `pnpm install` no longer creates dist artifacts, so the tokens CSS import is missing at Storybook build time.

So the failure is expected until the default-branch workflow also has the explicit build step (or you trigger Chromatic through a PR event that uses the PR workflow version).

Practical options:

1. Re-trigger via a pull_request event path (for example a new commit with a trigger that runs on synchronize, if enabled).
2. Merge/rebase so develop’s Chromatic workflow includes the build step.
3. Temporarily keep a build-producing mechanism available for this transition period.


~I’m choosing option 1.~ EDIT: I ended up marking this PR as a Draft and then as Ready for review again.

PR #2641, opened 2026-05-22
feat(React): Mark all component props as read-only

RubenSibon:
# Enforce read-only props

## What

Enables `react/prefer-read-only-props` in the shared React preset and applies the resulting `readonly` updates across React components and Storybook samples.
`96 files changed, 354 insertions, 366 deletions` (vs `develop`).
This is a type-safety and consistency change; no runtime behaviour is intended to change.

Two layers of `readonly` are applied:

- **Per-field `readonly` modifiers** on component-owned props.
- **`Readonly<...>` wrappers** on intersected React/base prop types (`HTMLAttributes<...>`, `PropsWithChildren<...>`, `Omit<InputHTMLAttributes<...>, ...>`, `SVGProps<...>`, etc.), so inherited props are also immutable in component implementations.

This is thread 1 of the **DES-1860 Strict Types** story.

## Why

The components in this repository do not intentionally mutate props, but this was convention-driven.
With read-only props enforced by types and linting, accidental mutation now fails earlier.

Benefits:

- Bugs are caught at compile/lint time instead of at runtime.
- Prop types communicate an explicit immutable contract for component internals.

No consumer-facing API behaviour is expected to change.

## How

- Enabled `react/prefer-read-only-props: 'error'` in the shared preset at [packages/react/eslint.config-partial.mjs](packages/react/eslint.config-partial.mjs).
- Applied repo-wide auto-fix for per-field readonly updates.
- Performed a manual pass for intersections/unions where auto-fix does not add `Readonly<...>` wrappers.
- Updated the component scaffolding template at [plop-templates/react.tsx.hbs](plop-templates/react.tsx.hbs) so new components default to:
  - `Readonly<PropsWithChildren<...>>` props
  - `forwardRef<Element, Props>(...)` typing instead of `ForwardedRef` callback parameter typing

## Verification

- `pnpm lint:js` -> passes with the existing baseline warnings unrelated to this PR’s runtime behaviour.
- `pnpm -r run lint` -> passes (typecheck across `packages/react` and `storybook`).

## Checklist

- [x] Add or update unit tests — not applicable (type-level change, no runtime behaviour).
- [x] Add or update stories — not applicable (Storybook samples received the same read-only treatment; behaviour unchanged).
- [x] Comment `/chromatic test` and verify visual regression tests pass — recommended as defence in depth, but no visual changes expected.

## Additional notes

- Diff is intentionally mostly mechanical: either `foo: T` -> `readonly foo: T`, or `& X<...>` -> `& Readonly<X<...>>`.
- Ticket: DES-1860.


packages/react/src/Breakout/BreakoutCell.tsx
```diff
@@ -16,37 +16,37 @@ import { breakoutCellClasses } from './breakoutCellClasses'
...
 export type BreakoutCellProps = {
   /** The HTML element to use. */
-  as?: BreakoutCellTag
-} & BreakoutCellRowSpanAndStartProps &
-  PropsWithChildren<HTMLAttributes<HTMLElement>> &
-  (BreakoutCellSpanAllProp | BreakoutCellSpanAndStartProps)
+  readonly as?: BreakoutCellTag
+} & Readonly<BreakoutCellRowSpanAndStartProps> &
+  Readonly<BreakoutCellSpanAllProps | BreakoutCellSpanAndStartProps> &
+  Readonly<PropsWithChildren<HTMLAttributes<HTMLElement>>>
```
VincentSmedinga, line 49: The PR body contradicts this diff. The “Additional notes” says `BreakoutCellRowSpanAndStartProps` and the discriminated union `BreakoutCellSpanAllProps | BreakoutCellSpanAndStartProps` were “left as-is” to avoid breaking type narrowing. But both of them get wrapped in `Readonly<>`.

packages/react/eslint.config-partial.mjs
```diff
@@ -10,6 +10,7 @@ export const reactPreset = {
   rules: {
     ...react.configs.recommended.rules,
     ...reactHooks.configs.recommended.rules,
+    'react/prefer-read-only-props': 'warn',
```
VincentSmedinga: Warnings don’t fail CI; mew components that forget `readonly` will silently slip through. If the goal is enforcement, this rules should emit errors.

```suggestion
    'react/prefer-read-only-props': 'error',
```

packages/react/src/Accordion/Accordion.tsx
VincentSmedinga, line 1: (Can’t attach the comment to the correct file, but at least this blocks CI.)

We should update our scaffolding template in `plop-templates/react.tsx.hbs` accordingly:

1. wrap `PropsWithChildren` in `Readonly<>`
2. use `forwardRef<>` instead of `ForwardedRef` (should have been part of #2643)

VincentSmedinga (commented):
React already treats props as immutable by spec. Mutating them is undefined behavior, and React’s docs explicitly forbid it. But TypeScript doesn’t know that. Adding `readonly` lifts React’s runtime convention into the type system, so the compiler enforces what was previously enforced only by code review.

While this does bring a benefit, it’s small in practice. The “accidental mutation” bug it prevents is rare in a library this disciplined — we already destructure, spread `…restProps`, and never reassign. So this catches a class of bug that mostly isn’t happening.

The visual noise (`readonly` on every field, `Readonly<>` wrapping every intersected React type) is unfortunate. I haven’t seen this anywhere else in applications or libraries.

Are you sure we should take this on?

packages/react/eslint.config-partial.mjs
RubenSibon: I agree. It felt a bit too forceful to immediately start with `error`, but if we move ahead with this we should.

Note, that this only applies to per-field props, not to intersections (the `Readonly<>` wrappers).

RubenSibon:
> React already treats props as immutable by spec. Mutating them is undefined behavior, and React’s docs explicitly forbid it. But TypeScript doesn’t know that. Adding `readonly` lifts React’s runtime convention into the type system, so the compiler enforces what was previously enforced only by code review.
> 
> While this does bring a benefit, it’s small in practice. The “accidental mutation” bug it prevents is rare in a library this disciplined — we already destructure, spread `…restProps`, and never reassign. So this catches a class of bug that mostly isn’t happening.
> 
> The visual noise (`readonly` on every field, `Readonly<>` wrapping every intersected React type) is unfortunate. I haven’t seen this anywhere else in applications or libraries.
> 
> Are you sure we should take this on?

Indeed, as you say this rule enforces an explicit safety type that is in practice (almost?) never violated. The intent of this rule is to make it explicit in TypeScript-land that this rule exists in the React runtime.

Another benefit I see: making this explicit might help less experienced React contributors understand how React works. That’s one of the reasons I have always believed it is best if budding developers start learning JavaScript or React programming with TypeScript in strict mode out of the gate.

So we have to decide if the minor benefit of being more explicit in our typing is a good trade-off with more ‘visual noise’. Personally, I don’t mind that noise so much. I like the explicitness a bit more. But if you don’t see the point and prefer to keep it less verbose, I’m fine with not merging this PR too.

packages/react/src/Breakout/BreakoutCell.tsx
RubenSibon: You’re right: the Additional notes text is stale. The current implementation does wrap both `BreakoutCellRowSpanAndStartProps` and the discriminated union in `Readonly`, and narrowing still works in the current version.

I’ll update the PR description to match the actual diff.

VincentSmedinga (approved):
Subtle, but autofixable with a linter, so let’s go.

PR #2643, opened 2026-05-22
fix(React): Type forwarded refs as HTML elements

RubenSibon:
# Type forwarded refs as HTMLElement

## What

Replaces `ForwardedRef<any>` with `forwardRef<HTMLElement, …>` (or `ForwardedRef<HTMLElement>` for Menu) in the seven polymorphic-`as` layout components that were previously accepting any ref type:

- `Row`, `Column`, `Grid`, `GridCell`, `Spotlight`, `BreakoutCell`, `Menu`

Also consolidates two duplicate `@amsterdam/design-system-react/src` imports in the Grid stories into a single line (separate commit).

Part of the **DES-1860 Strict Types** story. The read-only props thread shipped separately ([feat/DES-1860-strict-typing-readonly-props](https://github.com/Amsterdam/design-system/tree/feat/DES-1860-strict-typing-readonly-props)). This branch closes the remaining `@typescript-eslint/no-explicit-any` warnings.

## Why

The seven layout components had `ForwardedRef<any>` because their `as` prop could render different HTML elements (`<div>`, `<article>`, `<section>`, etc.), each with its own ref type. `any` was the original way to side-step the variance problem: it accepted any ref type the consumer passed in.

Two real downsides:

- `any` silenced the `@typescript-eslint/no-explicit-any` rule but provided no type safety inside the component implementations (`ref.current` had no autocomplete, no method checking).
- It signalled "we couldn't be bothered to type this" rather than the actual design constraint (the rendered element is `<div>` by default but configurable).

`HTMLElement` is the lowest common ancestor of every value in the `as` unions for these components. Switching to it:

- Removes seven `no-explicit-any` warnings (lint is back to the 3-warning baseline of unrelated `vitest/no-disabled-tests` findings).
- Restores meaningful IntelliSense on `ref.current` (DOM API autocomplete works).
- Keeps consumer ergonomics stable: `useRef<HTMLElement>(null)` works for any allowed `as` value.

## How

For each component, the change is mechanical:

    - import type { ForwardedRef, … } from 'react'
    + import type { ElementType, … } from 'react'

    - export const Row = forwardRef(
    -   ({ … }: RowProps, ref: ForwardedRef<any>) => (
    + export const Row = forwardRef<HTMLElement, RowProps>(
    +   ({ … }, ref) => {
    +     const Tag = (as ?? 'div') as ElementType
        …

The implementation uses `as: ElementType` to type the dynamic JSX tag — needed because TypeScript would otherwise narrow `<Tag>` to a specific intrinsic element based on the literal `'div' | 'article' | …` union and reject the wider `Ref<HTMLElement>`.

**Considered and rejected**: a fully polymorphic generic shape (`<T extends RowTag = 'div'>(...) => …` with `Ref<HTMLElementTagNameMap[T]>`). It compiled cleanly and gave technically more accurate ref typing per `as` value, but the VS Code hover output became unreadable:

    <T extends "article" | "div" | "section" = "div">(props: {
      ref?: …
    } & { as?: T | undefined } & {
      [P in Exclude<keyof React.PropsWithoutRef<React.ComponentProps<T>>, keyof RowOwnProps | "as">]: …
    } & { align?: MainAlign; alignVertical?: CrossAlign; gap?: RowGap; wrap?: boolean }): …

TypeScript can compute that correctly; tooling can't simplify it. Contributors and consumers preferred the cleaner `HTMLElement` hint output, and the practical safety gain of per-tag narrowing was small (refs onto layout containers are rare in consumer code).

## Checklist

- [x] Add or update unit tests — not applicable; existing `supports ForwardRef in React` tests for each component already cover the contract and still pass.
- [x] Add or update stories — only the Grid story import consolidation (cosmetic).
- [x] Comment `/chromatic test` and verify visual regression tests pass — recommended as a sanity check; no visual changes expected (TypeScript-only change).

## Additional notes

- Ticket: DES-1860.
- Verification:
    - `pnpm lint:js` → 3 warnings, 0 errors (matches the new baseline: 3× `vitest/no-disabled-tests`, all unrelated).
    - `pnpm -r run lint` (typecheck across `packages/react` and `storybook`): passes.
- Grid's public prop type widens from `HTMLAttributes<HTMLDivElement>` to `HTMLAttributes<HTMLElement>` — a minor surface shift to align the type with the component's polymorphic `as` prop (which accepts 8 tags, not just `<div>`). Consumers typing `useRef<HTMLDivElement>` against `<Grid>` will need to widen to `useRef<HTMLElement>`. Other components already used `HTMLAttributes<HTMLElement>`; only Grid was narrowed.
- No consumer prop-validation trade-off: `HTMLAttributes<T>` is React's base type and only includes attributes common to every HTML element. Element-specific attributes like `href` (on `AnchorHTMLAttributes`) were not accepted before and are not accepted now.
- The two-commit shape (type refactor + import consolidation) is unrelated content split for review clarity.


package.json
```diff
@@ -26,7 +26,7 @@
     "clean:dist": "pnpm -r run clean:dist",
     "lint": "pnpm run --aggregate-output '/^lint:.+$/' && pnpm -r --no-bail run lint",
     "lint:css": "stylelint --allow-empty-input --report-needless-disables '**/*.{css,scss}'",
-    "lint:js": "eslint . --report-unused-disable-directives",
+    "lint:js": "eslint .",
     "lint:package-json": "npmPkgJsonLint **/package.json",
```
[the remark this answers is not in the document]
RubenSibon: You’re correct. This is a branch of another PR that should be merged first. The diff will be much smaller after.

packages/react/src/Breakout/BreakoutCell.tsx
```diff
@@ -41,41 +41,34 @@ type BreakoutCellRowSpanAndStartProps = {
...
 }
 
+type BreakoutCellSpanKeys = 'colSpan' | 'colStart' | 'has' | 'rowSpan' | 'rowStart'
+
 export type BreakoutCellProps = {
   /** The HTML element to use. */
   as?: BreakoutCellTag
 } & BreakoutCellRowSpanAndStartProps &
-  PropsWithChildren<HTMLAttributes<HTMLElement>> &
+  PropsWithChildren<Omit<HTMLAttributes<HTMLElement>, BreakoutCellSpanKeys>> &
```
VincentSmedinga: The `Omit<>` wrappers in `BreakoutCell`, `GridCell`, and `Grid` aren’t doing anything. None of those keys (`colSpan`, `rowSpan`, `span`, `start`, `has`, `paddingBottom/Top/Vertical`) exist on `HTMLAttributes<HTMLElement>`. They look like they’re preventing a clash, but there’s nothing to clash with. We can drop them and use `PropsWithChildren<HTMLAttributes<HTMLElement>>` directly.

packages/react/src/Menu/Menu.tsx
```diff
@@ -24,11 +24,11 @@ export type MenuProps = {
 export const MenuRoot = forwardRef(
```
VincentSmedinga: Can this component use `forwardRef<HTMLElement, MenuProps>` instead of `ForwardedRef` like all others?

storybook/src/components/Grid/Grid.stories.tsx
```diff
@@ -6,8 +6,7 @@
 import type { GridCellProps } from '@amsterdam/design-system-react/src'
 import type { Meta, StoryObj } from '@storybook/react-vite'
 
-import { Paragraph } from '@amsterdam/design-system-react/src'
-import { Grid } from '@amsterdam/design-system-react/src'
+import { Grid, Paragraph } from '@amsterdam/design-system-react/src'
```
VincentSmedinga: Stories import their own component from source to get the type annotations. This isn’t necessary for other components, so `Paragraph` can be imported from `@amsterdam/design-system-react` here.

We could add a line to our AGENTS.md for this. A custom linter might be overkill?

VincentSmedinga (commented):
Nice push to get rid of these warnings!

The "href" example in the PR description is wrong. The body says `<Row as="div" href="…">` will no longer fail at compile time because `HTMLAttributes<HTMLElement>` “accepts the union of all HTML attributes.” It doesn’t; `HTMLAttributes<HTMLElement>` is the base type and excludes `href` (that’s on `AnchorHTMLAttributes`). So `<Row as="a" href="…">` fails both before and after.

Grid widens its public prop type from `HTMLAttributes<HTMLDivElement>` to `HTMLAttributes<HTMLElement>`. Tiny semver-minor surface shift. Fine, but worth a mention in the PR description since it’s the one component that actually narrowed before.

RubenSibon:
> The "href" example in the PR description is wrong.
> 
> Grid widens its public prop type from `HTMLAttributes<HTMLDivElement>` to `HTMLAttributes<HTMLElement>`. Tiny semver-minor surface shift. Fine, but worth a mention in the PR description since it’s the one component that actually narrowed before.

I’ve remove the former and added the latter in the PR description.

VincentSmedinga (approved):

PR #2645, opened 2026-05-23
feat(Dialog): Improve inner and outer borders in forced colors mode

Robbert:
# Describe the pull request

## What

- prevent `border-width: 0` for the dialog in forced colors mode
- add a 1px border around the dialog body in forced colors mode

Screenshot 2026-05-23 at 18 00 37

## Why

Improved accessibility

I was reminded of the necessity when browsing GitHub with forced colors mode

Screenshot 2026-05-23 at 18 01 48

## How

Similar approach to Utrecht components: use a media query to force a certain design in forced colors mode, and override design tokens where necessary. In this case the `border-width` design token is respected, but to avoid the risk of `--ams-dialog-border-style: none` this `border-style` is hardcoded.

## Checklist

- [ ] Add or update stories

## Additional notes

- Discussed this in NL Design System Developer Open Hour
- For an ideal solution more changes are needed, but I didn't see a way without breaking changes. `gap` should be removed in favour of a `padding-bottom` for the dialog header and `padding-block-start` for the dialog footer. Then the scrollable area of the dialog body aligns with the borders of the header and footer. It would be nice to have a larger scrollable area anyway, the current small size is pretty terrible on small viewports

Screenshot 2026-05-23 at 18 00 01


packages/css/src/components/dialog/dialog.scss
```diff
@@ -25,6 +25,14 @@ so do not apply these styles without an `open` attribute. */
   max-block-size: var(--ams-dialog-max-block-size);
   max-inline-size: var(--ams-dialog-max-inline-size);
 
+  /* Ensure there always is a border in forced colors mode */
+  @media (forced-colors: active) {
+    background-color: canvascolor;
```
VincentSmedinga: This has more effect, I think :)

```suggestion
    background-color: Canvas;
```

VincentSmedinga (approved):

VincentSmedinga:
Thanks @Robbert! We had added the same-colur border for forced colours – but certain token values can indeed render the forced colours styles useless. This is more robust; we’ll apply the `max(1px, …)` guard to our Select as well.

I’ll add your further improvements for the component to our list. We’ve already planned diving back into Dialog, Modal Dialog, Alert Dialog.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -86,6 +101,13 @@ so do not apply these styles without an `open` attribute. */
   padding-block: var(--ams-dialog-footer-padding-block);
   padding-inline: var(--ams-dialog-footer-padding-inline);
 
+  /* Use a border in forced colors mode for visual hierarchy and demarkating the scrollable body */
+  @media (forced-colors: active) {
+    border-block-start-color: currentColor;
+    border-block-start-style: solid;
```
RubenSibon, line 107: We could already apply the rule of 3 for the common props and maybe also the query with `@contents` in this file?

We’ll be reusing that mixin in other components as well and when that happens we could move the mixin from this file to `src/common/`.

RubenSibon (commented):
One comment about rule of 3 in this file, @VincentSmedinga.

packages/css/src/components/dialog/dialog.scss
VincentSmedinga: The three are different – one sets all borders, one only the bottom, one only the top. Moving only the media query into a mixin would obfuscate things in my opinion.

RubenSibon: Oh, right! Sorry!

RubenSibon (approved):

PR #2646, opened 2026-05-26
chore(Table, Table Page): Document caption requirement & add inline sorting controls demo

RubenSibon:
# Document caption requirement and add a new story for inline sorting controls placement

## What

- Make `SortingWithSelect` the single sorting demo on the Internal Table Page: the heading and the sort `Select` share one row above the table, and the table carries a visually hidden `<caption>` for its accessible name.
- Add a "Sort and filter controls" section to `TablePage.docs.mdx` explaining that controls belong outside the `<caption>`, and documenting two supported layouts (visible caption with controls above; recommended inline layout with a visually hidden caption).
- Add a Table guideline in the component README: don't replace the caption by naming the table another way, because the caption provides both the accessible name and the data-table semantics.
- Add a supporting reference (Scott O'Hara) inline in the README and under "See also" in `Table.docs.mdx`.

## Why

User story DES-1837 asked whether the sort `Select` could sit top-right next to the table heading, ideally inside the `<caption>`. Desk research concluded:

1. **No interactive controls in `<caption>`.** A caption's full subtree forms the table's accessible name ([accname 1.2, steps 2C/2F](https://www.w3.org/TR/accname-1.2/)), so a `<select>`'s current value would be concatenated into the table name on every announcement. `aria-describedby` does not override the name.
2. **Don't promote caption-less tables.** Naming a table via `aria-labelledby`/`aria-label` instead of a caption causes some browsers to expose it as a layout table, stripping row/column semantics ([Scott O'Hara](https://www.scottohara.me/blog/2019/04/19/tabled-tables.html)). The caption is the only naming method that preserves semantics.

So the controls stay outside the table, the table always keeps a caption, and the docs capture both the rule and the recommended layout.

## How

- The story keeps the controls in a `<form>` (sort state via URL, consistent with the page's existing guidance) inside a `<Row align="between">` alongside the heading; the table uses `<Table.Caption className="ams-visually-hidden">`.
- The README guideline links the "layout table" claim to the source; the same source is listed under the component docs' "See also".
- No change to the `Table` component API — `Table.Caption` already accepts `className`, so `ams-visually-hidden` is opt-in.

## Checklist

- [x] ~Add or update unit tests~ (not relevant)
- [x] Add or update stories

## Additional notes

- Jira: DES-1837.
- Manual screen-reader verification still needed before merge: confirm in NVDA/Firefox and VoiceOver/Safari that the table is announced as a data table named "Gegevens per adres", not a "group" or layout table.
- Story deploy links: [Table](https://designsystem.amsterdam/demo-DES-1837-table-toolbar-in-table-header/?path=/docs/components-containers-table--docs) and [Table Page](https://designsystem.amsterdam/demo-DES-1837-table-toolbar-in-table-header/?path=/docs/pages-internal-table-page--docs).


VincentSmedinga (commented):
I think the ‘inline’ example works better, mainly because the DOM order is more logical. And consequently, the order in which items are read out by screen readers, and how they stack in narrow windows. Although duplicating the header text is unfortunate.

Should we make this the single example? Not sure it really adds something to advise both, mainly if the non-inline was seems inferior.

RubenSibon:
> I think the ‘inline’ example works better, mainly because the DOM order is more logical. And consequently, the order in which items are read out by screen readers, and how they stack in narrow windows. Although duplicating the header text is unfortunate.
> 
> Should we make this the single example? Not sure it really adds something to advise both, mainly if the non-inline was seems inferior.

I’ve made the ‘inline’ example the recommended and only example. In the docs I do suggest that users could also choose to have the controls above the table and leave the caption visible within.

Do you agree with that?

P.S. I’ve also documented more explicitly in Table that the Caption is required.

packages/css/src/components/table/README.md
```diff
@@ -2,7 +2,7 @@
 
 # Table
 
-Structures data in rows and columns to help users work with large amounts of information.
+Presents related values in rows and columns, from a compact reference table to a large data set users need to scan, sort, or filter.
```
RubenSibon: > from a compact reference table

This reads a bit confusing because these kinds of simple tables are primarily not meant for the compact theme.

Maybe:

```suggestion
Presents related values in rows and columns: whether it is a simple reference table or a large data set to inspect, sort and filter.
```

RubenSibon (commented):
I think it would be better to split the Table documentation stuff into a new branch/PR and scope this PR to only the Table Page refactor and enhancements.

RubenSibon (commented):
I approve your last batch of commits @VincentSmedinga!

In the end I guess we can keep the enhancements to the Table and Table Page docs in 1 PR.

I guess you have to force a merge because neither of us can approve this PR at this point (you are the last committer and I opened the PR).

PR #2660, opened 2026-06-04
feat(Date Picker): Add Date Picker component

VincentSmedinga:
# Add Date Picker component

## Links

- [Date Picker docs](https://designsystem.amsterdam/demo-DES-1856-date-picker/?path=/docs/components-forms-date-picker--docs)

## What

Adds `DatePicker`, a form control for selecting a single date or a date range, in *Components / Forms*.

## Why

It completes the Calendar / Date Picker split.

Calendar covers browsing content grouped by date; the Date Picker is the form control for *choosing* a date or range — a grid of day buttons with full keyboard support — which is a distinct concern with different markup and interaction.

## How

- Controlled component with a `mode` discriminated union: single mode takes a `Date | null` value and calls `onChange` with a `Date`; range mode (`mode="range"`) takes a `DateRange` (`{ start, end }`) and calls `onChange` with the updated range.
  The first pick sets the start; a second pick on a later date or the same date completes the range (one-day ranges are valid). Picking before the current start discards the selection and begins a new range.
- `minDate` and `maxDate` bound both selection and month navigation; `isDateDisabled` marks individual dates unavailable while keeping them reachable by keyboard.
- Follows the [[WAI-ARIA grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/)](https://www.w3.org/WAI/ARIA/apg/patterns/grid/): `role="grid"` with a roving `tabindex`, arrow / Home / End / Page navigation, and Enter or Space to select. Focus follows keyboard navigation and clicks across month boundaries. The grid takes its accessible name from the visible month caption; each day button is labelled with its full date including weekday.
- Range mode sets `aria-multiselectable="true"` on the grid and marks every date within the selection as `aria-selected`. The start and end dates append `rangeStartAccessibleName` / `rangeEndAccessibleName` to their accessible names (default: `'startdatum'` / `'einddatum'`); when start and end fall on the same day, both suffixes are announced. All navigation button labels and range endpoint suffixes are overridable props for localisation.
- Forced-colours support: selected and hovered dates use `Highlight` / `HighlightText` so the selection remains visible when the OS overrides colours.
- Design tokens alias the Calendar tokens so the two components stay visually in step. Reuses the `useMonthNavigation` hook and date helpers shared with Calendar.

## Checklist

- [x] Add or update stories

## Additional notes

- Implements DES-1856.
- Story to review: *Components / Forms / Date Picker* (link above once the feature-branch preview deploys).
- The day hover background reuses the Icon Button's `rgb(0 70 153 / 12.5%)` value — there is no neutral-surface token yet, so this is a temporary workaround.
- Page Up / Page Down clamp the focused day to the last valid day of the target month (e.g. 31 Jan → 28 Feb) rather than overflowing.


packages/css/src/components/date-picker/date-picker.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.ams-date-picker__grid {
+  display: grid;
+  grid-template-columns: repeat(7, 1fr);
+}
+
+.ams-date-picker__row {
+  display: contents;
+}
```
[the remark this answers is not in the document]
VincentSmedinga: Done.

packages/react/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,276 @@
...
+  readonly rangeStartAccessibleName?: string
+} & Readonly<Omit<HTMLAttributes<HTMLDivElement>, 'onChange'>>
+
+type DatePickerSingleProps = {
+  /** Selects a single date. This is the default mode. */
+  readonly mode?: 'single'
+  /** Called when the user selects or clears the date. */
+  readonly onChange: (value: Date | null) => void
+  /** The currently selected date, or `null`. */
+  readonly value: Date | null
```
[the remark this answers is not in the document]
VincentSmedinga: Done.

packages/react/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    const nextFocusDate = getNextFocusDate(event.key, event.shiftKey, focusedDate)
+    if (nextFocusDate === undefined || isOutOfBounds(nextFocusDate, minDate, maxDate)) {
+      return
+    }
+
+    event.preventDefault()
+    moveFocusTo(nextFocusDate)
+  }
```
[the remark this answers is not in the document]
VincentSmedinga: Done.

packages/react/src/DatePicker/DatePickerBody.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const weeks = Array.from({ length: cells.length / 7 }, (_, index) => cells.slice(index * 7, index * 7 + 7))
+
+  return (
+    <div aria-labelledby={captionId} className="ams-date-picker__grid" onKeyDown={onKeyDown} role="grid">
+      <div className="ams-date-picker__row" role="row">
+        {Array.from({ length: 7 }).map((_, index) => (
+          <span className="ams-date-picker__weekday" key={`weekday-${index}`} role="columnheader">
+            {weekdayFormatter.format(new Date(2026, 5, 1 + index))}
+          </span>
+        ))}
```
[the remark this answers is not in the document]
VincentSmedinga: Done.

packages/react/src/DatePicker/DatePickerBody.tsx
```diff
@@ -0,0 +1,107 @@
...
+      className="ams-date-picker__body"
+      onKeyDown={onKeyDown}
+      role="grid"
+    >
+      <div className="ams-date-picker__row" role="row">
+        {Array.from({ length: 7 }).map((_, index) => (
+          <span className="ams-date-picker__weekday" key={`weekday-${index}`} role="columnheader">
+            {weekdayFormatter.format(new Date(2026, 5, 1 + index))}
+          </span>
+        ))}
```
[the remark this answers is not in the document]
VincentSmedinga: Done.

packages-proprietary/tokens/src/components/ams/date-picker.tokens.json
```diff
@@ -0,0 +1,103 @@
...
+          "text-decoration-line": {
+            "$value": "line-through",
+            "$extensions": {
+              "nl.amsterdam.type": "textDecorationLine"
+            }
+          }
+        },
+        "hover": {
+          "background-color": {
+            "$value": "rgb(0 70 153 / 12.5%)",
```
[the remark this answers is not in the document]
VincentSmedinga: No, not yet.

storybook/src/components/DatePicker/DatePicker.docs.mdx
```diff
@@ -0,0 +1,82 @@
...
+## Usage guidelines
+
+### When to use
+
+Use a Date Picker when someone needs to choose a date or date range from a visible calendar, for example to book an appointment.
+
+### When not to use
+
+Use a [Date Input](/docs/components-forms-date-input--docs) instead when the user knows the date and can type it into a form field.
+Use a [Calendar](/docs/components-navigation-calendar--docs) instead to browse content grouped by date, such as an events calendar.
```
RubenSibon:
```suggestion
- Use a [Date Input](/docs/components-forms-date-input--docs) when the user knows the date and can type it into a form field.
- Use a [Calendar](/docs/components-navigation-calendar--docs) to browse content grouped by date, such as an events calendar.
```

“Instead” is not needed because the heading ‘When not to use’ already implies that these are alternative use-cases.

I think that making this a list adds clarity.

I’ve considered switching the use-case and the imperative form:

```
- When the user knows the date and can type it into a form field use a [Date Input](/docs/components-forms-date-input--docs).
```

But I guess it’s better the way it was.

storybook/src/components/DatePicker/DatePicker.docs.mdx
```diff
@@ -0,0 +1,82 @@
...
+### Range selection
+
+<Canvas of={DatePickerStories.Range} />
+
+### Limiting the selectable period
+
+Use `minDate` and `maxDate` to bound the selectable period.
+Dates outside the range cannot be selected, and the month and year navigation buttons stop at the edges of the range.
+
+<Canvas of={DatePickerStories.LimitedToOneMonth} />
```
RubenSibon: Maybe it would be nice if this example or another example showed what the Date Picker looks like when the `minDate` or `maxDate` are inside the current month. What do the not-selectable dates look like?

It’s not possible to play around with the `min`- and `maxDate` in the controls. Same for many of the other props. Why is this so?

EDIT: I’ve [learned why](https://github.com/Amsterdam/design-system/pull/2660#discussion_r3395078212).

storybook/src/components/DatePicker/DatePicker.docs.mdx
RubenSibon, line 1: Clicking on the `Show code` buttons in the [demo](https://designsystem.amsterdam/demo-DES-1856-date-picker/?path=/docs/components-forms-date-picker--docs) takes me to an empty Storybook view. What’s going on?

storybook/src/components/DatePicker/DatePicker.docs.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+<Meta of={DatePickerStories} />
+
+<Title />
+
+<Description of={DatePickerStories} />
+
+<Primary />
+
+<Controls />
```
RubenSibon, line 18: Why are many of the controls not editable?

EDIT: I’ve [learned why](https://github.com/Amsterdam/design-system/pull/2660#discussion_r3395078212).

storybook/src/components/DatePicker/DatePicker.docs.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+In the default single mode, `value` is a `Date` or `null` and `onChange` receives the picked date.
+In range mode, set `mode="range"`; `value` is a `DateRange` (`{ start, end }`) and `onChange` receives the updated range.
+The first pick sets the start, the second sets the end, and picking before the start begins a new range.
+
+## Examples
+
+### Range selection
+
+<Canvas of={DatePickerStories.Range} />
```
RubenSibon, line 62: I think this example could use a bit more guidelines. How will users know that this a date range selection? Developers or content editors need to add that context with text. The component in its bare state does not communicate “I’m a date range picker!”.

storybook/src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+  argTypes: {
+    defaultMonth: { control: false }, // A Date object has no usable controls panel widget and can break the story.
+    isDateDisabled: { control: false }, // A function prop has no usable controls panel widget.
+    locale: { control: false }, // Shown for reference only; localisation examples to follow.
+    maxDate: { control: false }, // A Date object has no usable controls panel widget and can break the story.
+    minDate: { control: false }, // A Date object has no usable controls panel widget and can break the story.
+    mode: { control: false }, // The story wrapper owns this prop.
+    onChange: { control: false }, // The story wrapper owns this prop.
+    value: { control: false }, // The story wrapper owns this prop.
+  },
```
RubenSibon, line 43: Okay, so now I understand why the controls are not editable. But users will not see these comments, so it might be confusing why they cannot play around with them.

And if we could have some kind of workaround to get these controls working, that would be a very nice feature. Disabling controls because of some internal technical reason is not optimal.

RubenSibon (commented):
Review in progress...

packages/react/src/DatePicker/utils.ts
RubenSibon, line 1: Nice! Our own date utils instead of a third-party lib.

packages/react/src/DatePicker/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+import type { DateRange } from './DatePicker'
+
+import { getDaysInMonth, startOfDay } from '../common/dates'
+
+/** Whether the date is before `minDate` or after `maxDate` (comparing by day). */
+export const isOutOfBounds = (date: Date, minDate?: Date, maxDate?: Date) =>
+  (minDate !== undefined && startOfDay(date) < startOfDay(minDate)) ||
+  (maxDate !== undefined && startOfDay(date) > startOfDay(maxDate))
```
RubenSibon: I guess our linter allows this, but I don’t like arrow functions with an implicit return were the return block is on the next (multiple) lines (see also [arrow-body-style#as-needed](https://eslint.org/docs/latest/rules/arrow-body-style#as-needed)). I would prefer:

```suggestion
export const isOutOfBounds = (date: Date, minDate?: Date, maxDate?: Date) => {
  return (minDate !== undefined && startOfDay(date) < startOfDay(minDate)) ||
  (maxDate !== undefined && startOfDay(date) > startOfDay(maxDate))
}
```

Same for the other functions with this pattern.

packages/react/src/DatePicker/DatePickerHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+  readonly disableNextYear?: boolean
+  /** Disables the button that navigates to the previous month. */
+  readonly disablePreviousMonth?: boolean
+  /** Disables the button that navigates to the previous year. */
+  readonly disablePreviousYear?: boolean
+  readonly goToNextMonth: () => void
+  readonly goToNextYear: () => void
+  readonly goToPreviousMonth: () => void
+  readonly goToPreviousYear: () => void
+  readonly month: Date
```
RubenSibon, line 37: Why do these props not have comments?

packages/react/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,276 @@
...
+   * @default startdatum
+   */
+  readonly rangeStartAccessibleName?: string
+} & Readonly<Omit<HTMLAttributes<HTMLDivElement>, 'onChange'>>
+
+type DatePickerSingleProps = {
+  /** Selects a single date. This is the default mode. */
+  readonly mode?: 'single'
+  /** Called when the user selects a date. */
+  readonly onChange: (value: Date | null) => void
```
RubenSibon: `DatePickerSingleProps.onChange` is typed `(value: Date | null) => void`, but implementation only ever calls it with a `Date`.

**Why it matters:** public API type/documentation mismatch; consumers may expect `null` events that never happen.

**Suggestion:** either narrow to `(value: Date) => void` or explicitly document future/intentional nullable callback behaviour.

Credits: GitHub Copilot.

packages/react/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+    const { start, end } = selection.value
+
+    if (start !== null && isSameDay(date, start)) {
+      return rangeStartAccessibleName
+    }
+
+    if (end !== null && isSameDay(date, end)) {
+      return rangeEndAccessibleName
+    }
```
RubenSibon: In range mode, if user picks the same date twice (`start === end`), `nextRange` creates a complete range, but the accessible name only gets `rangeStartAccessibleName` (not end), because `getBoundaryLabel` returns early on start match.

- `packages/react/src/DatePicker/utils.ts` (`nextRange`)
- `packages/react/src/DatePicker/DatePicker.tsx` (`getBoundaryLabel`)

**Why it matters:** this makes a completed one-day range indistinguishable (for screen readers) from an incomplete range with only `start` set.

**Suggestion:** when `start` and `end` are both set and same day, append both labels (e.g. `"startdatum, einddatum"`), and add a test for this case.

> At the moment, in these stories, a user cannot input the same date for both the start and end with mouse pointer, touch or keyboard. But maybe we still should make the underlying logic even more robust?

Good question — and yes, I’d still make it robust.

I do think this case is currently possible in the component itself:

- In range mode, first pick sets `{ start: date, end: null }`.
- Picking the **same** date again goes through `nextRange` and returns `{ start: date, end: date }`.
- That can happen via click/tap or pressing `Enter`/`Space` twice on the focused day.

Even if your product flow *usually* avoids that, `DatePicker` is controlled, so a parent can also pass `{ start, end }` with equal dates directly.

So I’d recommend a defensive fix anyway. Two valid approaches:

1. **Allow one-day ranges (recommended)**  
   Keep current selection logic, but if `start === end`, announce both boundaries in the accessible name (and add a test).

2. **Disallow one-day ranges**  
   Normalize in `nextRange` so a second pick on the same date does not set `end` (or restarts range), and document that rule.

Credits: GitHub Copilot

packages-proprietary/tokens/src/components/ams/date-picker.tokens.json
RubenSibon, line 1: Right now sharing is mostly `date-picker -> calendar` aliasing. That works, but conceptually it couples DatePicker to Calendar naming.

A cleaner long-term model is an internal foundation layer, e.g. `ams.date-grid.*`:

- `ams.date-grid.font-*`
- `ams.date-grid.gap`
- `ams.date-grid.inline-size`
- `ams.date-grid.day.current.font-weight`
- `ams.date-grid.day.outline-offset`

Then:
- `ams.calendar.*` aliases `ams.date-grid.*`
- `ams.date-picker.*` aliases `ams.date-grid.*`
- component-specific tokens stay local (`calendar.day.link.*`, `date-picker.day.selected.*`, etc.)

This gives shared styling without implying Calendar is the “source of truth” for DatePicker.

VincentSmedinga: It’s a fair point, but I’m not against making Calendar the source of truth for Date Picker.

The foundation is also not internal as `ams.date-grid.*` tokens will all be exposed in the browser unless we add extra configuration.

packages/react/src/DatePicker/DatePicker.tsx
VincentSmedinga: Yes, removed `null`.

packages/react/src/DatePicker/DatePicker.tsx
VincentSmedinga: Ah, this is an interesting UX flow.

I initially dismissed single date ranges for being useless, but then realised it can be frustrating for a user if they want to focus on a single date for a particular feature, but can’t because the Date Picker is configured to only accept ranges.

So the more permissive approach is the way to go. I’ve updated the accessibility labels and will pick this up with UX.

packages/react/src/DatePicker/DatePickerHeader.tsx
VincentSmedinga: Added them.

packages/react/src/DatePicker/utils.ts
VincentSmedinga: Sure. Even introduced intermediate variables to end up with single, self-contained lines.

storybook/src/components/DatePicker/DatePicker.docs.mdx
VincentSmedinga: Not sure, they look fine for me:

Image

storybook/src/components/DatePicker/DatePicker.docs.mdx
VincentSmedinga: Applied your suggestion separately.

storybook/src/components/DatePicker/DatePicker.docs.mdx
VincentSmedinga: Yes, added a good intro.

storybook/src/components/DatePicker/DatePicker.docs.mdx
VincentSmedinga: > Maybe it would be nice if this example or another example showed what the Date Picker looks like when the `minDate` or `maxDate` are inside the current month. What do the not-selectable dates look like?

Good idea, I’ve added a ‘Limiting the selectable period’ example.

storybook/src/components/DatePicker/DatePicker.stories.tsx
VincentSmedinga: I’ve updated the descriptions in code and added a note below the controls.

storybook/src/components/DatePicker/DatePicker.docs.mdx
RubenSibon: Now they do for me too. 🤷🏻

RubenSibon (approved):
Code looks good and the stories work as expected in the demo.

PR #2666, opened 2026-06-08
chore: Adopt oxlint and oxfmt, and modernise the lint/format toolchain

VincentSmedinga:
# Adopt oxlint and oxfmt, and modernise the lint/format toolchain

### What

Moves the design system's linting and formatting onto the [oxc](https://oxc.rs) toolchain, with related tooling cleanups:

- **oxfmt fully replaces Prettier** for all formatting (JS/TS/JSX, JSON, CSS/SCSS, YAML, HTML).
- **oxlint is the sole linter for all JavaScript/TypeScript**, running its native rules plus `eslint-plugin-perfectionist`, `eslint-plugin-baseline-js`, and `eslint-plugin-storybook` through oxlint's `jsPlugins` loader.
- **`perfectionist/sort-imports` is enforced** across all TS, JS, and MDX files with explicit import groups (types → external → internal → relative → side-effects). `#storybook/` and `#tokens/` are registered as internal patterns, so aliased Storybook imports sort with project code rather than node_modules.
- **ESLint is reduced to the languages oxlint can't lint** — JSON, Markdown, and MDX.
- **`lint-staged` → `nano-staged`** (zero-dependency, faster) with the same pre-commit behaviour.
- **Markdown content linting now actually runs.** It was a silent no-op (`markdown.configs.recommended.rules` is `undefined` in `@eslint/markdown` 8); it now uses the GFM language with the recommended rules, so `require-alt-text`, `no-empty-links`, etc. are enforced.
- **Root-level ESLint plugins are declared explicitly** (`eslint-plugin-storybook`, `eslint-plugin-mdx`, `eslint-plugin-react`) instead of relying on pnpm hoisting from sub-packages.
- **`AGENTS.md` gains efficient linter/formatter guidance** (tool ownership, fast targeted commands, "don't hand-format/sort").

Removed: `prettier`, `eslint-config-prettier`, `eslint-plugin-oxlint`, `lint-staged`, and the now-unused ESLint JS/TS plugins (`@typescript-eslint/*`, `@vitest/eslint-plugin`, `eslint-plugin-import-x`, `eslint-import-resolver-typescript`, `globals`, `@eslint/js`).

Added: `oxlint`, `oxfmt`, `nano-staged`. The published packages (CSS, tokens, React, assets) are unchanged — this only affects the developer toolchain.

### Why

ESLint and Prettier were the slowest parts of the local and CI feedback loop; oxlint and oxfmt are Rust-based and much faster, and consolidating on one toolchain removes the ESLint/Prettier coordination. Along the way this surfaced and fixed real gaps: Markdown linting was doing nothing, and the root tooling depended on plugins declared only in sub-packages.

Like many others, we haven't been able to upgrade to ESLint 10 for a while due to a blocking issue in [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react/issues/3977). This led us to research alternatives, where [this article](https://evilmartians.com/chronicles/stop-writing-rules-in-agents-md-use-agent-hooks-and-nano-staged-instead) pointed us at the oxc toolchain, mentioning "oxfmt now has 1:1 compatibility with Prettier […] and oxlint is also production-ready."

### How

Seven reviewable commits, each independently revertible:

1. `replace ESLint/Prettier tooling with oxlint and oxfmt` — generates `.oxlintrc.json` (`@oxlint/migrate`), adds `.oxfmtrc.json` mirroring the old Prettier style, converts stale `eslint-disable` → `oxlint-disable`.
2. `style: format with oxfmt` — the one-off oxfmt pass (a single embedded-script semicolon; the codebase already matched).
3. `lint all JS/TS with oxlint via jsPlugins, slim ESLint to JSON/Markdown/MDX`.
4. `replace lint-staged with nano-staged`.
5. `document efficient linter and formatter use in AGENTS.md`.
6. `declare the root-level ESLint plugins explicitly`.
7. `enable Markdown content linting` — the GFM/recommended fix, plus the two issues it surfaced (a CSS block missing its language; a broken `#guidelines` anchor in the Page Header README).

A merge commit from `develop` brought in two additional changes that landed there in the interim — **`perfectionist/sort-imports` enforcement** (with ESLint auto-fixing the whole codebase) and the **`#storybook`/`#tokens` path aliases**. Both are reflected in this branch: `.oxlintrc.json` already carries the full `sort-imports` config including the `internalPattern` for the aliases, and `storybook/eslint.config-partial.mjs` carries the same rule for MDX files.

Parity for the JS/TS rules was verified by linting a deliberately-broken file with both tools (identical findings), and the new Markdown rules were verified to flag `require-alt-text` and `no-empty-links`.

`oxfmt` honours `<!-- prettier-ignore -->` (its documented Markdown ignore syntax), so those comments were kept.

CI needs no workflow change — a new `lint:format` script is picked up by the existing `pnpm run lint` aggregator.

### Checklist

- [x] Add or update unit tests — _not applicable; no source behaviour changes. Full suite (1085 + 78) passes._
- [x] Add or update documentation — _AGENTS.md (incl. the React package note), `definition-of-done.md`, `CONTRIBUTING.md` (VS Code extension → Oxc), and two doc fixes surfaced by Markdown linting._
- [x] Add or update stories — _not applicable._
- [x] Add or update exports in index.\* files — _not applicable._
- [x] Comment `/chromatic test` and verify visual regression tests pass — _no visual changes expected (only lint-directive comments in story files); worth running to confirm._

### Additional notes

**Performance (median of 3, Apple M1 Max):**

| Command | Before (ESLint + Prettier) | After (oxlint + oxfmt) |
| --- | --- | --- |
| Formatter, whole repo | `prettier --check .` 4.39s | `oxfmt --check .` **0.54s** |
| `lint:js` (the command devs run) | `eslint .` 10.36s | `oxlint && eslint .` **4.68s** |
| `pnpm run lint` (full) | 16.54s | **12.59s** |

**For reviewers to weigh in on:**

- **oxlint's `jsPlugins` loader is alpha** (its own schema notes "not subject to semver") and is now the only thing linting our JS/TS. It's isolated in commit 3, so `git revert <commit-3>` falls back to a fully-stable setup (oxlint native rules + a reduced ESLint) without touching the Prettier → oxfmt work.
- **Rules dropped** in the jsPlugins phase (no native/jsPlugins equivalent): `consistent-return`, `react/prefer-read-only-props`, `padding-line-between-statements` (generated logos), `@typescript-eslint/no-invalid-this`.
- **MDX content linting is still off** — `.mdx` is syntax-validated but its content isn't linted (no `remark-lint` config). Intentionally out of scope here; tracked as a separate follow-up.
- **MDX is not formatted by oxfmt** (Prettier never did either; ~108 files of churn avoided). Config-flip to enable later.


storybook/src/components/LinkList/LinkList.stories.tsx
```diff
@@ -21,7 +21,6 @@ const meta = {
 
 export default meta
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const linkMeta = {
   component: LinkList.Link,
 } satisfies Meta<typeof LinkList.Link>
```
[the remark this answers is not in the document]
VincentSmedinga: Checked this empirically: oxlint’s `no-unused-vars` rule does count a `typeof` reference in a type position as a usage, so `linkMeta` is not flagged. The file lints clean without a disable directive. No change needed here.

AGENTS.md
```diff
@@ -118,9 +118,9 @@ For details, rely on the official documentation and per-package instructions:
 Key agent expectations:
 
 - Before submitting work, cross-check the full definition-of-done checklist in [documentation/definition-of-done.md](documentation/definition-of-done.md).
-- Run the most specific relevant lint/test commands for the package you touched before relying on full `pnpm run lint` / `pnpm run test`.
+- Run the most specific relevant lint/test command — e.g. `oxlint` for JS/TS or `oxfmt --check` for formatting — before relying on the full `pnpm run lint` / `pnpm run test`.
```
RubenSibon, line 121: This tells me you haven’t included the [agent hooks](https://evilmartians.com/chronicles/stop-writing-rules-in-agents-md-use-agent-hooks-and-nano-staged-instead#agent-hooks) in this PR yet.

But now that I’ve read that part of the article again (see link above) I do see that it is agent vendor specific. Can you do some research if there’s a more generic option that at the very least works with GitHub Copilot?

RubenSibon (commented):
Great work!

I have yet to test and review this properly, but I’m sure the build process will be much faster and the config will be even more consolidated and consistent. And it’s great that Markdown linting now actually works, even though some work on MDX linting/formatting remains.

But a few things from the description I find unfortunate:

- “ESLint is reduced to the languages oxlint can’t lint — JSON, Markdown, and MDX.” -- _it would be better if we could get rid of ESLint altogether, now we still have similar tooling for similar tasks. How does this affect the full code quality checking pipeline?_
- “Root-level ESLint plugins are declared explicitly (eslint-plugin-storybook, eslint-plugin-mdx, eslint-plugin-react) instead of relying on pnpm hoisting from sub-packages.” -- _I see that oxlint uses the JSON format and that cannot import partials. They do not support (M)JS? Because I actually like the separation of concerns we had with package particular rules scoped to that package. Another downside of basic JSON is that there can be no clarifying comments or use comments and newlines to improve structure and readability. The latter downsides could maybe be avoided [with JSONC or JSON5](https://go-tools.org/blog/json5-jsonc-formatting-guide)._
- “MDX content linting is still off — .mdx is syntax-validated, but its content isn't linted (no remark-lint config). Intentionally out of scope here; tracked as a separate follow-up.” -- _Is that issue on the backlog already?_
- Rules dropped in the jsPlugins phase (no native/jsPlugins equivalent): consistent-return, react/prefer-read-only-props, padding-line-between-statements (generated logos), @typescript-eslint/no-invalid-this. -- _Can you elaborate on this? Why can’t we have these rules any more? We recently added `prefer-read-only-props`._

I’m going to test this with Visual Studio Code extensions for ESLint & Prettier that I currently have set up. I think that contributors should ideally be able to continue developing without any or too many workflow disruptions.

Furthermore, I’ll also add other regular contributors (@alimpens, @dlnr, @Evi-2003) as reviewers so that they can test the linting and formatting with their IDEs and set-ups.

RubenSibon (commented):
I’ve done some testing:

- Linting and formatting from the command-line works flawlessly!
- But I couldn’t rely on the ESLint and Prettier extensions in VS Code any more.
- The [Oxc extension](https://marketplace.visualstudio.com/items?itemName=oxc.oxc-vscode) _does_ work as expected. But I had to manually look for it; I did not see a prompt to install it.

So, in conclusion:

- I think I can accept this change after you’ve addressed my concerns in the earlier remark.
- It might be wise to warn contributors about that they have to look at their tooling.
- Reviews from regular contributors with other IDEs and set-ups might still be valuable.

RubenSibon (commented):
I’ve done a bit more testing by intentionally adding code and formatting mistakes to `TableOfContents.tsx`.

Most issues were found and automatically fixed by either the Oxc linter or formatter, but the import order rules were not applied:

```tsx
/**
 * @license EUPL-1.2+
 */

import type { HeadingProps } from '../Heading'

import { Heading } from '../Heading'

import { TableOfContentsLink } from './TableOfContentsLink'
import { clsx } from 'clsx'
import { TableOfContentsList } from './TableOfContentsList'
```

Above import order is incorrect.

This is because `eslint-plugin-import-x` is no longer included in the project. Isn’t that a major limitation? Maybe add it to the ESLint config again for now if Oxc does not support it?

packages/react/package.json
```diff
@@ -27,7 +27,7 @@
     "build": "tsc --project ./tsconfig.build.json && rollup -c",
     "build:watch": "rollup -c --watch",
     "clean:dist": "rimraf dist/",
-    "generate-logos": "svgr --config-file src/Logo/svgr.config.mjs ../../packages-proprietary/assets/logo/*-logo.svg --out-dir src/Logo/brands && prettier --write src/Logo/brands && eslint src/Logo/brands --fix",
+    "generate-logos": "svgr --config-file src/Logo/svgr.config.mjs ../../packages-proprietary/assets/logo/*-logo.svg --out-dir src/Logo/brands && oxfmt --write src/Logo/brands && oxlint --fix src/Logo/brands",
```
Evi-2003, line 30: After running this command the result is different, there are changes in the generated files and errors in the terminal.

Evi-2003 (changes requested):
Nice implementation! 🚀

I've had a bit of trouble setting it up in Neovim, but once I figured it out it was pretty easy to set up.

All the rules seem to be working for me, also the Perfectionist rules. These rules do not seem to be working in VS Code. I've installed the extension but all the JS-based rules do not seem to be working. Which seems related to Ruben's comment about the import order.

This seems to be an issue related to the VS Code extension, however I [can't find](https://github.com/oxc-project/oxc-vscode/issues) a related issue for it. 

--

The implementation seems to work, but for new developers it might be a hassle to get started if they did not work with Oxlint before. We can expect everyone to have Prettier installed, but we can't for Oxfmt.

The extensions for Oxlint seems to be released not that long ago and might have more issues, which probably will get fixed over time. 

I'm not sure if the 4-second performance gain is worth the hassle of probably more maintenance, and the difficulty of getting started in the design system code if your dev environment is not prepared for Oxlint.

Less dependencies are great, but I think it would be more important to stay with the more traditional linting packages for now, at least until everything works for most editors/ more developers have adopted to it.

VincentSmedinga:
> I'm not sure if the 4-second performance gain is worth the hassle of probably more maintenance, and the difficulty of getting started in the design system code if your dev environment is not prepared for Oxlint.
> 
> Less dependencies are great, but I think it would be more important to stay with the more traditional linting packages for now, at least until everything works for most editors/ more developers have adopted to it.

Agree with you, @Evi-2003 – it may be a bit early to bet everything on Oxc while they’re below 1.0, we need to keep ESLint partly anyway, and editor support is a bit rough.

Thanks for the effort @RubenSibon, good groundwork regardless. Closing this for now; we can easily reopen once things have matured.

PR #2673, opened 2026-06-11
fix(Progress List): Show dashed connector for collapsed current step with substeps

Evi-2003:
## What
Fixes the connector for a `Step` that is both `current` and `collapsed` with substeps, it now correctly shows as dashed instead of solid.

Before: 
CleanShot 2026-06-11 at 12 06 33

After:
CleanShot 2026-06-11 at 12 07 34

Also updates a test story to cover this case.

## Why
The line style for current steps with substeps was always applied, even when the substeps were collapsed. When collapsed, the dashed line (from the default style) should remain, as the substeps are hidden. The solid line should only appear when the substeps are visible.

## How
Added `:not(.ams-progress-list__step--collapsed)` to the CSS selector that sets `border-inline-style: solid`, so the solid connector only applies when the step is expanded.

## Checklist
- [x] Add or update stories

## Additional notes
- Manually collapse the current step to see the change
[Develop](https://designsystem.amsterdam/?path=/story/components-containers-progress-list--with-substeps)
[Branch build](https://designsystem.amsterdam/?path=/story/components-containers-progress-list--with-substeps)

### DIFF
--- a/packages/css/src/components/progress-list/progress-list.scss
+++ b/packages/css/src/components/progress-list/progress-list.scss
@@ -124,7 +124,7 @@
     border-inline-style: var(--ams-progress-list-step-completed-connector-border-inline-style);
   }
 
-  .ams-progress-list__step--current.ams-progress-list__step--has-substeps & {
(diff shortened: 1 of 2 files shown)


RubenSibon (approved):
LGTM! 🥳

PR #2674, opened 2026-06-11
feat(icons): Add new icons

dlnr:
# Describe the pull request

## What

New Icons from the UX department:

- ChevronDoubleBackward
- ChevronDoubleForward
- Box
- TheatricalMasks
- SpeechBalloon
- Refresh (functional)
- RotateCounterclockwise (functional)
- RotateClockwise (functional)
- ZoomIn (functional)
- ZoomOut (functional)
- BirdHouse
- Bridge
- WasteContainer
- ArchiveBox
- MarketStall

Config changes to SVGO to make the process of adding icons a lot easier. Figma adds fill colors, width and height to the svg, there's no way around this. There was a lot of manual stuff that I did to remove this stuff before generating the icon components. This is no longer needed because the generate command now does this for you. One implications is that the svg files in the assets now still have these attributes. Let me know if this is not wanted and i'll change or add a script to do the source files.

## How

Added to assets and batch processed through svgo to add to react-icons as component.

## Checklist

- [ ] Add or update stories

## Additional notes


RubenSibon:
Why are the width and height of all other icons (not in scope) reduced to 24px from 300px?

packages-proprietary/react-icons/src/Bell.tsx
```diff
@@ -9,9 +9,4 @@ const SvgBell = (props: SVGProps<SVGSVGElement>) => (
     />
   </svg>
 )
-
-/**
- * @deprecated The ‘Bell’ icon will be removed on or after 2026-07-09.
- * Use ‘Notification’ instead.
- */
```
VincentSmedinga, line 16: I understand that the `generate` command rewrites every React icon from scratch, but we should not delete these deprecation annotations. Revert manually for now?

I’ve created a [ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-1889) to configure deprecations in the generator.

packages-proprietary/react-icons/src/index.ts
```diff
@@ -186,10 +189,6 @@ export { default as LightningFillIcon } from './LightningFill'
 export { default as LineChartDownIcon } from './LineChartDown'
 export { default as LineChartUpIcon } from './LineChartUp'
 export { default as LinkIcon } from './Link'
-export {
-  /** @deprecated This icon misspells the brand as ‘Linkedin’, with a lowercase ‘i’. Use ‘LinkedInIcon’ instead. Will be removed on or after 2026-05-01. */
-  default as LinkedinIcon,
-} from './LinkedIn'
```
VincentSmedinga, line 192: This is a breaking change; we must wait to remove this export until the next major release.

packages-proprietary/react-icons/src/index.ts
```diff
@@ -1,11 +1,8 @@
-/**
- * @license EUPL-1.2+
- * Copyright Gemeente Amsterdam
- */
-
```
VincentSmedinga, line 5: The license block should not be removed.

packages-proprietary/assets/icons/Apple.svg
```diff
@@ -1 +1,3 @@
...
\ No newline at end of file
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
VincentSmedinga, line 1: I understand that Figma adds `fill`, `width` and `height`.

Width and height seem fine – they give the browser an intrinsic size when the file is used as  or embedded as <object>, preventing layout shift.

But `fill="none"` seems odd, this makes the paths invisible if they don’t have a fill, and Apple for instance doesn’t.

Should we run the svg source files through SVGO as well to add `fill="black"` to `<svg>`, and to strip fills from paths? For now, our icons don’t have explicit colours.

VincentSmedinga (commented):
Thanks for adding the new icons. Updating the SVGO config to ease the process seems logical, although it also introduces scope creep.

If it’s not too much work, I think the best approach would be to strip the new icons so that improving the configuration becomes the goal of this PR. Effects on the svgs and React icons are then fine.

Except for the removal of deprecation comments and license banners; so I guess generating these should become part of the PR. This is then a `fix` PR.

A second `feat` PR can then add the new icons, which profit from the improved configuration right away.

packages-proprietary/react-icons/src/Bell.tsx
VincentSmedinga: In the meantime, @Evi-2003 has implemented this in #2695 🚀

packages-proprietary/assets/icons/Apple.svg
dlnr: Width and height make the scaling impossible but i'll keep it in mind when rebuilding this PR in two parts.

dlnr:
Splitting the config and new icons changes into two pull requests.

PR #2679, opened 2026-06-11
chore: Sort imports in MDX files with perfectionist

VincentSmedinga:
# Sort imports in MDX files with perfectionist

## What

Added `perfectionist/sort-imports` to the MDX ESLint config and auto-fixed all 93 existing Storybook docs pages to comply. Also updated the plop template so newly generated component docs start out in the correct order.

## Why

We already enforce sorted imports on all TypeScript and JavaScript files via `eslint-plugin-perfectionist`. MDX files were excluded, leading to inconsistent import ordering across Storybook docs.

## How

1. Added a new `amsterdam-design-system/mdx-imports` config block in `storybook/eslint.config-partial.mjs`, targeting `**/*.mdx` with `perfectionist/sort-imports` using the same `natural` / `asc` settings as the TS config.
1. Ran `eslint --fix` to apply the rule across all existing docs pages.
1. Updated `plop-templates/storybook.docs.mdx.hbs` to produce correctly ordered imports for new components.

## Checklist

- [x] Add or update stories

## Additional notes

- All 93 changed MDX files were auto-fixed by ESLint.


RubenSibon (commented):
> We already enforce sorted imports on all TypeScript and JavaScript files via eslint-plugin-perfectionist.

This is not true.

We use [`eslint-plugin-import-x`](https://github.com/Amsterdam/design-system/blob/develop/eslint.config.mjs#L9) (which is a fork of `eslint-plugin-import`) to sort imports in (J/T)S(X) files.

Check [our ESLint rules](https://github.com/Amsterdam/design-system/blob/develop/eslint.rules.mjs#L86C1-L92C2).

It would actually be an improvement if we could let Perfectionist handle that sorting as well. Then we could get rid of the ESLint import plugin. That would make migrating to Oxc easier too.

RubenSibon (approved):

VincentSmedinga:
Superseding this with a broader PR that adds `perfectionist/sort-imports` to TypeScript files as well, with `newlinesBetween: 1` and explicit import groups so the blank-line separation is enforced everywhere. That PR covers MDX with the same config, so this one is no longer needed. Closing to avoid the 93 docs files being touched twice with different settings. Thanks for the review, Ruben!

Issue #2618, opened 2026-05-13
Logo SVG cut off (viewbox issue?)


svoenie-ams:
Hi all,

I noticed all the logos on Chromium have a problem, where it cuts off the last letter a bit. Happens on Chrome/Edge but not on Firefox. Seems like the SVG element is cutting it off. See the images provided.

<img width="191" height="95" alt="Image" src="https://github.com/user-attachments/assets/ee7901bf-e799-4360-80fe-43d81e491235" />

<img width="224" height="76" alt="Image" src="https://github.com/user-attachments/assets/c40f4093-6b95-4efd-bf14-78a855cf3514" />

https://designsystem.amsterdam/?path=/docs/components-media-logo--docs

VincentSmedinga:
Hello Sven – you have quite the keen eye! We’ll look into it.

VincentSmedinga:
We’ve tweaked some pixels @svoenie-ams – all of them should be visible after updating to our [latest release](https://github.com/Amsterdam/design-system/releases/tag/design-system-react-v4.2.0).

PR #2683, opened 2026-06-12
fix(Logo): Keep logo clear of the clipping edge to prevent cut-off

Evi-2003:
# Describe the pull request

## What

Adds a margin to the `viewBox` of all logos so the logo no longer touches the right edge:

- Updates all 16 logo SVGs in `packages-proprietary/assets/logo/`.
- Regenerates the 8 React components from those sources, the `.tsx` diffs contain nothing but the new `viewBox` values.
- Restores the `class="ams-logo__…"` attributes in `ggd-amsterdam-inspectie-logo.svg`. They existed in the generated component but not in the source, so the next regeneration would have silently stripped the classname.

| Logo | Artwork right edge | viewBox width before | viewBox width after (+1, and rounded up after 3 decimals) |
|---|---|---|---|
| Amsterdam | 114.854 | 115 | 115.855 |
| Amsterdam English | 115.002 | 115 | 116.003 |
| GGD Amsterdam | 114.854 | 115 | 115.855 |
| GGD Amsterdam Inspectie | 89.1965 | 89.1966 | 90.197 |
| Museum Weesp | 112.2558 | 112.5581 | 113.256 |
| Stadsarchief | 97.631 | 98 | 98.632 |
| Stadsbank van Lening | 150.838 | 151.4286 | 151.839 |
| VGA Verzekeringen | 134.809 | 134.8837 | 135.81 |

- Measured with [getBBox()](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox) in viewBox units.

## Why

The last part of most logos were cut off in Chrome, Edge and Safari at some viewport sizes. 

The logo's height comes from a viewport-relative token, so the rendered element is a fractional pixel size at almost every viewport width. The width is derived from that height via the viewBox aspect ratio and is fractional too. 

Layout and painting handle this fine, both are subpixel-precise. The problem is the clip: WebKit/Blink(Chromium) browsers snap the SVG's clipping window to whole device pixels at paint time (see [BlinkNG](https://developer.chrome.com/docs/chromium/blinkng), [WebKit LayoutUnit](https://trac.webkit.org/wiki/LayoutUnit)), and that snapped edge causes a part of the logo to disappear, even though the viewBox isn't wrong. 

CleanShot 2026-06-12 at 16 00 42

The old viewBoxes were tight with 0 to 0.59 units of accidental slack, so the artwork was exactly to that edge and was being cut off. Firefox clips with subpixel precision, so there was no issue there.

Sizing workarounds do not help: rounding the height (CSS `round()`) still results a fractional width, and whole CSS pixels still become fractional device pixels under browser zoom or 125%/150% display scaling. The artwork itself has to stay clear of the edge.

## How

Measured the exact right edge of each logo's artwork with `getBBox()` in a browser, then applied one rule to every viewBox width:  roundUp(artwork right edge + 1, 3 decimals)

- **+1 unit** 1px at the minimum logo height (40px), which covers the worst case: 0.5px clip-snap error plus ~0.5px of anti-aliased ink fringe.
- **Rounded *up* at 3 decimals**, never to nearest: e.g. GGD Inspectie's artwork ends at 89.19649…, so +1 = 90.19649…; rounding to nearest (90.196) would leave 0.9995 units. Rounding up gives 90.197.
- **3 decimals** because that is the svgo emits when generating the React components; keeping the source on the same grid to be identical at both places.

The change is invisible: the rendered scale of the artwork is driven by the height (viewBox height stays 40), so widening the viewBox only adds ~1px of empty space on the right of the element. Top and bottom edges are unchanged.

## Checklist

- [x] Add or update stories

## Additional notes

[Related issue](https://github.com/Amsterdam/design-system/issues/2618).
[Prod](https://designsystem.amsterdam/?path=/docs/components-media-logo--docs)
[Branch deploy](https://designsystem.amsterdam/?path=/docs/components-media-logo--docs).


packages-proprietary/assets/logo/amsterdam-english.svg
RubenSibon, line 1: The _Gemeente Amsterdam_ and _City of Amsterdam_ logos used to have the same view box width and height*. Now there is a difference. The logos are hot-swapped on the main site when the language is switched. Could this difference cause further issues?

(*) In the GitHub rich diff viewer the displayed height differs 1px despite the view box height being 40px in both cases. I am not sure if that matters, but it’s worth a mention.

RubenSibon (commented):
I have one remark about the Dutch and English logos of the City.

I don’t know enough about SVG view box and sub-pixel rendering to judge whether this +1px and rounding to 3 decimals is an adequate solution. I’ll leave that judgement to Vincent.

packages-proprietary/assets/logo/amsterdam-english.svg
VincentSmedinga: > The _Gemeente Amsterdam_ and _City of Amsterdam_ logos used to have the same view box width and height*. Now there is a difference. The logos are hot-swapped on the main site when the language is switched. Could this difference cause further issues?

That hot-swapping on the website is a workaround that we don’t support. We don’t have to take that into account.
 
> (*) In the GitHub rich diff viewer the displayed height differs 1px despite the view box height being 40px in both cases. I am not sure if that matters, but it’s worth a mention.

This is because GitHub renders svgs that don’t have intrinsic dimensions at 300 pixels wide. On our websites, we render logos with a fixed height and varying width, so no problem here.

VincentSmedinga:
> I don’t know enough about SVG view box and sub-pixel rendering to judge whether this +1px and rounding to 3 decimals is an adequate solution. I’ll leave that judgement to Vincent.

@RubenSibon The viewbox is an internal coordinate system for the svg. In the context of CSS, it only affects the aspect ratio of the image. This is why the updated logos might show slightly more whitespace on the right side. We prefer this to the cropping issue it resolves.

I will soon discuss the management of our logo files with Communications; perhaps we can achieve more standardization in the future.

VincentSmedinga (approved):

PR #2690, opened 2026-06-17
feat(Pagination): Add tokens and elements for theming, active state, ellipsis, and relative-link

VincentSmedinga:
# Add Pagination tokens and elements for theming, active state, ellipsis, and relative-link

## What

Expands the Pagination token surface and adds new elements so consumers (including NL Design System theme authors) can restyle Pagination by overriding tokens.

- Adds 32 new tokens covering states (hover, active, current, current-hover) and elements (root, list, link, ellipsis, relative-link).
- Adds the `:active` state for link.
- Renames the internal `Spacer` component to `Ellipsis` and adds an `__ellipsis` class so the placeholder can be themed and aligned with the page numbers.
- Adds an `__relative-link` class for previous/next buttons, with a full token set that mirrors link.

## Why

The Pagination token surface was minimal — most states and properties had no override slot, forcing downstream consumers to redeclare CSS for any change.

This PR opens up the full surface so themes can override without touching CSS, introduces visible interactions (the `:active` state), and gives previous/next buttons and the ellipsis their own targetable element classes.

## How

- Every new token holds the concrete value we actually want for the Amsterdam brand, expressed as a regular token: a reference where one fits (e.g. `link.active.color` → `link.color`), or a plain value where the property is simply unset in our brand (`0`, `transparent`, `auto`).
- No CSS-wide keyword theming hooks (`initial`/`inherit`) and no separate `external` extension — these are ordinary tokens that consumers can override like any other. We express our design choices directly: “between these items, the gap is zero.”
- Sorted tokens consistently: own CSS properties alphabetically, then variants (`hover`, `active`), then children (`list`, `link`, `relative-link`, `ellipsis`, `current`) in natural hierarchical order.

## Tokens added

Values are the Amsterdam defaults. References use shorthand: `{link.X}` is `{ams.pagination.link.X}`.

### Pagination root and list

| Token | Amsterdam value |
| --- | --- |
| `pagination.gap` | `0` |
| `pagination.list.gap` | `0` |

### Link

| Token | Amsterdam value |
| --- | --- |
| `link.background-color` | `transparent` |
| `link.font-weight` | `{ams.typography.body-text.font-weight}` |
| `link.min-block-size` | `auto` |
| `link.min-inline-size` | `auto` |
| `link.padding-block` | `0` |

### Link variants

| Token | Amsterdam value |
| --- | --- |
| `link.hover.background-color` | `transparent` |
| `link.active.background-color` | `transparent` |
| `link.active.color` | `{link.color}` |
| `link.current.background-color` | `transparent` |
| `link.current.color` | `{link.color}` |
| `link.current.hover.background-color` | `transparent` |
| `link.current.hover.color` | `{link.hover.color}` |

### Ellipsis

| Token | Amsterdam value |
| --- | --- |
| `ellipsis.color` | `{ams.color.text.default}` |
| `ellipsis.padding-block` | `{link.padding-block}` |
| `ellipsis.padding-inline` | `0` |

### Relative-link

All 15 tokens reference the matching `link.*` token.

| Token | References |
| --- | --- |
| `relative-link.background-color` | `{link.background-color}` |
| `relative-link.color` | `{link.color}` |
| `relative-link.font-weight` | `{link.font-weight}` |
| `relative-link.gap` | `{link.gap}` |
| `relative-link.outline-offset` | `{link.outline-offset}` |
| `relative-link.padding-block` | `{link.padding-block}` |
| `relative-link.padding-inline` | `{link.padding-inline}` |
| `relative-link.text-decoration-line` | `{link.text-decoration-line}` |
| `relative-link.text-decoration-thickness` | `{link.text-decoration-thickness}` |
| `relative-link.text-underline-offset` | `{link.text-underline-offset}` |
| `relative-link.hover.background-color` | `{link.hover.background-color}` |
| `relative-link.hover.color` | `{link.hover.color}` |
| `relative-link.hover.text-decoration-line` | `{link.hover.text-decoration-line}` |
| `relative-link.active.background-color` | `{link.active.background-color}` |
| `relative-link.active.color` | `{link.active.color}` |

## Checklist

- [x] Add or update unit tests — added assertions for the new `__relative-link` and `__ellipsis` element classes
- [x] Add or update documentation — component docs auto-render the new tokens
- [x] Add or update stories — Default args bumped so Chromatic captures every new state
- [x] Add or update exports in index.* files — no public exports affected by the rename

## Additional notes

- Closes DES-1888.
- Zero visual change from the pre-PR baseline: every new token resolves to the same computed CSS the component already produced (`transparent` background, `0` gap, base link colour, etc.).


storybook/src/components/Pagination/Pagination.stories.tsx
RubenSibon, line 1: Out of scope for this PR I guess, but it would actually be nice if the links in the [story demo](https://designsystem.amsterdam/demo-DES-1888-pagination-link-tokens/?path=/story/components-navigation-pagination--default) actually changed the render state of Pagination so that the “active” page is actually highlighted as the current one.

Didn’t we have that feature once? I probably misremember.

packages-proprietary/tokens/README.md
```diff
@@ -310,6 +345,28 @@ We add a subtype for certain types to further specify their purpose and preview.
...
+  "color": {
+    "$value": "{ams.pagination.link.color}",
+    "$extensions": {
+      "nl.amsterdam.external": true,
+      "nl.amsterdam.type": "color"
+    }
+  }
+}
+```
+
```
RubenSibon: Seems like a sensible extension that clearly distinguishes our own brand tokens. 👍🏻

RubenSibon (dismissed):

storybook/src/components/Pagination/Pagination.stories.tsx
VincentSmedinga: Yes, we had that when they were buttons :)

Making links do what they’re supposed to do, but within Storybook, and both in a div on a docs page and in a story iframe, is a broader problem. We’ve done some trickery in the Table Pages for example; would indeed be nice if we could make it straightforward and reusable. I’ve added a [ticket](https://gemeente-amsterdam.atlassian.net/browse/DES-1895) for it.

packages-proprietary/tokens/README.md
VincentSmedinga: Thanks. I’ve been thinking a bit more about it.

While the ‘external’ property on its own does seem clear, I’m not sure about the `initial` and `inherit` values. Sad that we need two different ones, each in specific cases. And there’s more complexity that they introduce: should we implement them in our own components in other stacks (React Native, Mendix, CMSes)? What if we ‘promote’ them and give them an actual value? How much work does this administrative divide introduce for us? Aren’t ‘initial’ and ‘inherit’ prone to unintended side effects, or even bugs if the context slightly changes?

We could just give them the values we actually want, without depending on these keywords – so `0`, `none`, `auto` etc. Treat them as any other token, express our design choices: “Between these things, the gap is zero.” Bundle size doesn’t matter – these tokens already increased it anyway, and compression and caching go a long way with CSS.

VincentSmedinga: Replaced initial/inherit theming hooks with concrete token values.

packages-proprietary/tokens/src/components/ams/pagination.tokens.json
```diff
@@ -19,34 +19,81 @@
           "nl.amsterdam.type": "fontWeight"
         }
       },
+      "gap": {
+        "$value": "0",
+        "$extensions": {
+          "nl.amsterdam.subtype": "space",
+          "nl.amsterdam.type": "dimension"
+        }
+      },
```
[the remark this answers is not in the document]
VincentSmedinga: Went with option (2). The tokens now hold concrete values (`0`, `transparent`, `auto`, and same-component references) as regular tokens, and the `initial`/`inherit` hooks and the `nl.amsterdam.external` extension have been dropped — see the discussion above. I’ve updated the PR description to match.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -86,7 +86,7 @@ export const Pagination = forwardRef(
           {accessibleName || 'Paginering'}
         </span>
         {page !== 1 && (
-          <Link className="ams-pagination__link" href={linkTemplate(page - 1)} rel="prev">
+          <Link className="ams-pagination__link ams-pagination__relative-link" href={linkTemplate(page - 1)} rel="prev">
             <Icon svg={ChevronBackwardIcon} />
```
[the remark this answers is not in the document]
VincentSmedinga: Good call. Added unit tests asserting that the previous/next links carry `ams-pagination__relative-link` and that the ellipsis renders an `ams-pagination__ellipsis` element.

PR #2692, opened 2026-06-17
fix(Pagination): Generate a unique fallback id for the accessible name

VincentSmedinga:
# Generate a unique fallback id for the accessible name of Pagination

## What

`Pagination`’s `accessibleNameId` falls back to a `useId()`-generated id instead of the hardcoded `'ams-pagination-a11y-label'` string. Consumers who don’t pass `accessibleNameId` get a guaranteed-unique id automatically.

## Why

Two `Pagination` instances on the same page previously produced duplicate HTML ids and shared an accessible name, which is invalid HTML and trips axe’s `landmark-unique` and duplicate-id checks. Consumers shouldn’t have to set `accessibleNameId` manually just to render two paginations on a page.

## How

- Calls `useId()` at the top of the render function — before the `totalPages <= 1` early return, to satisfy the rules-of-hooks.
- The consumer-provided `accessibleNameId` still wins; the generated id only kicks in when none is provided.
- Added a unit test that renders two `Pagination` instances and asserts their `aria-labelledby` ids differ.

## Checklist

- [x] Add or update stories

## Additional notes

- Companion to `chore(Storybook): Give variant-matrix landmarks unique accessible names`, which addresses the same family of axe violations from the test-harness side.

### DIFF
--- a/packages/react/src/Pagination/Pagination.test.tsx
+++ b/packages/react/src/Pagination/Pagination.test.tsx
@@ -115,6 +115,21 @@ describe('Pagination', () => {
     expect(component).toHaveAttribute('aria-labelledby', 'custom-id')
   })
 
+  it('generates a unique fallback id when none is provided', () => {
+    render(
+      <>
+        <Pagination linkTemplate={linkTemplate} totalPages={10} />
+        <Pagination linkTemplate={linkTemplate} totalPages={10} />
+      </>,
+    )
+
+    const [first, second] = screen.getAllByRole('navigation', { name: 'Paginering' })
+
+    expect(first).toHaveAttribute('aria-labelledby')
+    expect(second).toHaveAttribute('aria-labelledby')
+    expect(first.getAttribute('aria-labelledby')).not.toBe(second.getAttribute('aria-labelledby'))
+  })
+
   it('should not render when totalPages is 1 or less', () => {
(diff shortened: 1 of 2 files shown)


RubenSibon (approved):
Sensible safeguard, but this shouldn’t be necessary because having more than 2 pagination components on 1 page should not occur. Only one can be attached to the current state in the URL search params.

Or is this for situations that there are 2 or more that use programmatic rendering?

We haven’t explicitly documented that there may only be 1 of these on the page, but maybe we should?

VincentSmedinga:
> Sensible safeguard, but this shouldn’t be necessary because having more than 2 pagination components on 1 page should not occur. Only one can be attached to the current state in the URL search params.
> 
> Or is this for situations that there are 2 or more that use programmatic rendering?
> 
> We haven’t explicitly documented that there may only be 1 of these on the page, but maybe we should?

Our policy has usually been that we can’t be sure that certain patterns don’t make sense.

Multiple paginations on a single page can seem unintuitive, but I could still imagine two related sets of items next to each other, that can both need pagination. Two params in the url would be perfectly fine for this.

In general, we offer a box of components and invite people to get creative with it. We only advise against real accessibility or usability problems, and I don’t think two paginations would qualify as that.

This change to generate unique ids instead of one equally ‘made up’ internal id is small, internal and used in several other components.

PR #2695, opened 2026-06-18
chore(react-icons): Generate icon deprecation comments

Evi-2003:
# Describe the pull request

## What

Makes the icon generator add the `@deprecated` JSDoc comments itself, configured in a single config file.

- `deprecations.mjs` listing the deprecated icon components (removal date + replacement) and the deprecated alias re-export (`LinkedinIcon`).
- The SVGR `template` adds the `@deprecated` comment to a component's default export for every icon in the config.
- `indexTemplate.mjs` re-exports the deprecated aliases with their comment, and now also emits the `@license` header so it is no longer added by hand after each run.

## Why

Deprecation comments were added to the generated icon files by hand, so every `pnpm run generate` overwrote them and they had to be re-added manually (the same was true for the `@license` header on `index.ts`). Moving this into configuration means the comments generation and there is one place to manage which icons are deprecated.

## How

A custom SVGR component `template` reads the config and attaches the comment to the export. To keep the generated code identical to before, the comment is attached as a trailing block comment of the component, so the generator keeps the blank line above it. The same config feeds `indexTemplate` for the deprecated alias exports.

Verified by running `pnpm run generate`: the 11 deprecated component files come back byte-for-byte identical (comment + blank line intact), and the only change is the corrected export sort order in `index.ts`. `eslint` and `pnpm run build` pass, and the `@deprecated` comments land in the published `dist/*.d.ts` types.

## Checklist

- [x] Add or update stories _(n/a — no component changes)_

## Additional notes

- To confirm the behaviour: delete a `@deprecated` block from a generated icon file, run `pnpm run generate`, and it is restored.

### DIFF
--- a/packages-proprietary/react-icons/deprecations.mjs
+++ b/packages-proprietary/react-icons/deprecations.mjs
@@ -43,17 +43,21 @@ for (const [iconName, { removeOnOrAfter }] of Object.entries(deprecatedIcons)) {
   }
 }
 
-// The `@deprecated` JSDoc comment for each deprecated icon, keyed by file
-// basename.
-/** @type {Record<string, string>} */
-export const deprecatedIconComments = Object.fromEntries(
-  Object.entries(deprecatedIcons).map(([iconName, { removeOnOrAfter, replacement }]) => [
-    iconName,
-    [
-      '/**',
-      ` * @deprecated The ‘${iconName}’ icon will be removed on or after ${removeOnOrAfter}.`,
-      ` * Use ‘${replacement}’ instead.`,
-      ' */',
-    ].join('\n'),
-  ]),
-)
+/**
+ * Builds the `@deprecated` JSDoc comment for an icon, or `undefined` if it is
+ * not deprecated.
+ *
+ * @param {string} iconName File basename of the icon, e.g. `TrashBin`.
+ */
+export const deprecatedIconComment = (iconName) => {
+  const deprecation = deprecatedIcons[iconName]
+
+  if (!deprecation) return undefined
+
+  return [
+    '/**',
+    ` * @deprecated The ‘${iconName}’ icon will be removed on or after ${deprecation.removeOnOrAfter}.`,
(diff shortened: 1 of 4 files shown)


packages-proprietary/react-icons/deprecations.mjs
```diff
@@ -0,0 +1,39 @@
+// Icons whose component is deprecated, by file basename (e.g. `TrashBin`,
```
VincentSmedinga, line 8: The license banner is missing.

packages-proprietary/react-icons/deprecations.mjs
```diff
@@ -0,0 +1,39 @@
...
+// exported as `TrashBinIcon`). Adds an `@deprecated` comment to the component.
+export const deprecatedIcons = {
```
VincentSmedinga, line 11: Can we type this variable to be sure? I would think something like `Record<keyof typeof Icons, {r: string, r: string}>`? We could even type the date string with a regex-like thing :)

packages-proprietary/react-icons/deprecations.mjs
```diff
@@ -0,0 +1,39 @@
...
+  LinkedinIcon: {
+    comment:
+      'This icon misspells the brand as ‘Linkedin’, with a lowercase ‘i’. Use ‘LinkedInIcon’ instead. Will be removed on or after 2026-05-01.',
+    icon: 'LinkedIn',
+  },
+}
+
+// Builds the `@deprecated` JSDoc comment for an icon, or `undefined` if it is
+// not deprecated. `iconName` is the file basename, e.g. `TrashBin`.
+export function deprecatedIconComment(iconName) {
```
VincentSmedinga: Does this need to be a function while the others are consts?

VincentSmedinga (commented):
Thanks for picking this up, could mean that we’ll be releasing new icons tomorrow!

VincentSmedinga (approved):

Issue #2585, opened 2026-04-23
Improvements to scrolling containers


dlnr:
While working with the Tabs and Table components i've noticed it would help to add these rules:

```css
overflow: auto hidden;
overscroll-behavior-x: contain;
```

`overflow` Shorthand to make sure overflow-y is set to hidden and nog auto to prevent the accidental popping up of scrollbars. 
`overscroll-behavior-x` to disable to back navigation on touch screens and prevent the scroll chaining to the body.

There is much more we could do to containers that scroll like `scrollbar-behavior` or styling them with `scrollbar-color`, `scrollbar-gutter` and `scrollbar-width` to prevent the ugly windows fixed scrollbars.  (Edit: never mind the baseline settings would not allow this)

Another welcome addition to Tabs would be if the selected Tab would scroll in view using something like `scroll-snap-align: start;` because on a small screen, if the active tab is outside the viewport, the tab would not be visible forcing you to place another title to show the user where they are.

PR #2714, opened 2026-06-23
feat(Dialog, Table, Tabs, Tab Navigation): Improve scrolling behaviour

VincentSmedinga:
## What

Hardens the scrolling containers in four components:
- **Tabs** & **Table** — explicit `overflow-y: hidden` + `overscroll-behavior-x: contain`
- **Tabs** — `scroll-snap-type: x mandatory` on the list and `scroll-snap-align: start` on the buttons
- **Dialog** — explicit `overflow-x: hidden` on the body (mirror axis)
- **Tab Navigation** — explicit `overflow-y: hidden` on the list (vertical variant resets overflow)

## Why

Addresses #2585. `overflow-x: auto` silently makes the cross-axis compute to `auto`, which can pop a stray scrollbar; `overscroll-behavior-x: contain` prevents scroll-chaining and touch back-navigation. Tabs and Table were the only horizontal scrollers missing `overscroll-behavior` — Tab Navigation and Image Slider already had it, so this brings them in line. The Tabs snap matches Tab Navigation so tabs rest on their edges while scrolling.

## How

Pure SCSS, no React changes. `overflow` is written as longhands to match house style. Each `overscroll-behavior-*` carries the existing `plugin/use-baseline` “Progressive enhancement” disable comment, consistent with the other scroll containers. Tab Navigation’s vertical variant resets `overflow-y: visible` so the column layout keeps growing freely.

## Checklist

- [x] Add or update unit tests — n/a, CSS-only; existing tests pass
- [x] Add or update documentation — n/a
- [x] Add or update stories — n/a (an optional narrow-overflow demo could be added)
- [ ] Comment `/chromatic test` and verify visual regression tests pass — no diffs expected

## Additional notes

- Addresses #2585.
- CSS snap improves scroll feel but doesn’t reveal an already-selected off-screen tab on load — that needs a small `scrollIntoView`, deferred.
- Follow-up planned: a stylelint guard (defensive-css `scroll-chaining`) to enforce `overscroll-behavior` on future scrollers.

### DIFF
--- a/packages/css/src/components/dialog/dialog.scss
+++ b/packages/css/src/components/dialog/dialog.scss
@@ -82,6 +82,7 @@ so do not apply these styles without an `open` attribute. */
 }
 
 .ams-dialog__body {
+  overflow-x: hidden;
   overflow-y: auto;
   /* stylelint-disable-next-line plugin/use-baseline -- Progressive enhancement. */
   overscroll-behavior-y: contain;
--- a/packages/css/src/components/tab-navigation/tab-navigation.scss
+++ b/packages/css/src/components/tab-navigation/tab-navigation.scss
@@ -13,6 +13,7 @@
   box-shadow: var(--ams-tab-navigation-list-box-shadow);
   display: flex;
   overflow-x: auto;
+  overflow-y: hidden;
   /* stylelint-disable-next-line plugin/use-baseline -- Progressive enhancement. */
   overscroll-behavior-x: contain;
   scroll-snap-type: x mandatory;
@@ -26,6 +27,7 @@
       box-shadow: var(--ams-tab-navigation-list-vertical-box-shadow);
       flex-direction: column;
       overflow-x: visible;
+      overflow-y: visible;
     }
   }
 }
--- a/packages/css/src/components/table/table.scss
+++ b/packages/css/src/components/table/table.scss
@@ -5,6 +5,9 @@
 
 .ams-table {
   overflow-x: auto;
+  overflow-y: hidden;
+  /* stylelint-disable-next-line plugin/use-baseline -- Progressive enhancement. */
+  overscroll-behavior-x: contain;
 }
 
 .ams-table__table {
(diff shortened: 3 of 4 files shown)


packages/css/src/components/dialog/dialog.scss
```diff
@@ -82,6 +82,7 @@ so do not apply these styles without an `open` attribute. */
 }
 
 .ams-dialog__body {
+  overflow-x: hidden;
```
dlnr, line 85: This is cool, also if you don't want to use the shorthand, but perhaps they should also be tokens?

dlnr (approved):
Nice job. Next step progressively enhanced `scrollbar-color`? The non-apple scrollbar in these nested scrollable areas are super-ugly.

packages/css/src/components/dialog/dialog.scss
VincentSmedinga: As discussed elsewhere: this PR addresses a few (possible) bugs and inconsistencies.

Some of these could use tokens (scroll snapping yes, overflow and overscoll no, I guess?). That’s a separate task as it would affect other components than these.

Same for using shorthand or not; looking for a linting solution that makes sure we don’t forget to set both overflow directions instead of only one.

VincentSmedinga:
> Next step progressively enhanced `scrollbar-color`? The non-apple scrollbar in these nested scrollable areas are super-ugly.

Could be a next step. First: decide if we want to override the appearance of the operating system. If so: make a visual design first.

dlnr:
> > Next step progressively enhanced `scrollbar-color`? The non-apple scrollbar in these nested scrollable areas are super-ugly.
> 
> Could be a next step. First: decide if we want to override the appearance of the operating system. If so: make a visual design first.

Wouldn't style the body scrollbar (ever), just the scrolling containers. Mainly because of the lack of consistency between mac and windows scrollbars. The apple overlay scrollbar only appears when scrolling, causing it to overlap the tabs for example. The windows scrollbar show a gutter not matching the design of the components and basically look weird in the middle of your page.

Using this line will cause the scrollbar to always be visible, but not be so ugly.

```css
scrollbar-color: var(--ams-color-interactive) transparent;
```

This still matches the design of the OS to some level.

You could also set it to a thin size, but turns out many people actually try to grab the scrollbar.

PR #2842, opened 2026-07-25
fix(Storybook): Put the Skip Links in a navigation landmark

VincentSmedinga:
## Links

- [Skip Link](https://designsystem.amsterdam/?path=/docs/components-navigation-skip-link--docs)
- [Page](https://designsystem.amsterdam/?path=/docs/components-containers-page--docs)
- Deploy links follow once the preview is up.

## What

A Skip Link now sits in a `nav` with an accessible name, instead of loose in the Page container:

```tsx
<nav aria-label="Snelkoppelingen">
  <SkipLink href="#inhoud">Direct naar inhoud</SkipLink>
</nav>
```

This changes the two Page Layout decorators behind the public and internal templates, the `WithMenu` story of Page, and the snippets in both introductions, on the Page and Skip Link pages, and in the Home Page layout sketch. No component in `packages` changes, so nothing here needs a release.

## Why

A Skip Link comes before the Page Header, which put it ahead of every landmark on the page. axe reported `region` — "All page content should be contained by landmarks" — on every page template because of it. That rule is tagged best-practice rather than WCAG, so this was never a conformance failure, but it was the only finding on our page templates, and it also maps to RGAA 9.2.1, which an auditor working from a checklist may well raise.

Wrapping the link is enough to resolve it, and it gives the group a name of its own alongside the Hoofdmenu, the Kruimelpad and a page's own Table of Contents.

## How

All of a page's Skip Links share one `nav`, rather than each getting its own — the Handbook Page declares two, and two landmarks for two hidden links would be noise.

On the internal layout the `nav` carries `ams-page__area--skip-link`. The grid that `withMenu` sets up selects direct children of the Page, so the class has to move to whichever element is that child. The comment in the decorator said that none of those children may be wrapped; it now says wrapping is fine as long as the wrapper takes the class.

The Skip Link stories are unchanged. They render the component on its own, with no Page around it, so a `nav` there would be scaffolding that demonstrates nothing.

Verified with axe-core against three public and three internal page stories: every one of them reported `region` before and none of them reports anything now. The only remaining `incomplete` result is `color-contrast` on the Page Header's `aria-hidden` brand name, which is unrelated and untouched. I also forced focus on the Skip Link in both layouts and compared the rendering, since the focused state is the only thing that could regress — the band still spans the full Page width and pushes the page down, including on the internal grid where the area class moved.

## Checklist

- [x] Add or update stories

## Additional notes

Skip Links are hidden until focused, so I do not expect visual regressions, but every story under `Pages/**` is snapshotted and the internal templates changed a grid area, so the Chromatic run is worth a look rather than a wave-through.

The naming is the one thing worth a second opinion: `aria-label="Snelkoppelingen"` was chosen to sit alongside Hoofdmenu and Kruimelpad as a plain noun. It is one string in three places if you would rather have something else.

### DIFF
--- a/storybook/src/components/Page/Page.docs.mdx
+++ b/storybook/src/components/Page/Page.docs.mdx
@@ -29,7 +29,7 @@ As a root layout component, it must be used for all websites for the City of Ams
 #### Public websites
 
 Use the Page as it is.
-Its children – a [Skip Link](/docs/components-navigation-skip-link--docs), the Page Header, the sections of the page, and the Page Footer – follow one another in document order.
+Its children – a `nav` holding the [Skip Links](/docs/components-navigation-skip-link--docs), the Page Header, the sections of the page, and the Page Footer – follow one another in document order.
 See the [introduction to the public templates](/docs/pages-public-introduction--docs) for the structure of a page.
 
 #### Internal websites
@@ -70,7 +70,7 @@ An element placed directly in the Page, outside a [Grid](/docs/components-layout
 
 ## See also
 
-- [Skip Link](/docs/components-navigation-skip-link--docs) – the first element in the Page, above the Page Header.
+- [Skip Link](/docs/components-navigation-skip-link--docs) – sits in the first element of the Page, above the Page Header.
 - [Page Header](/docs/components-containers-page-header--docs) – sits at the top of the Page.
 - [Menu](/docs/components-navigation-menu--docs) – fills the menu column that `withMenu` adds, on an internal website.
 - [Page Footer](/docs/components-containers-page-footer--docs) – sits at the bottom of the Page.
--- a/storybook/src/components/Page/Page.stories.tsx
+++ b/storybook/src/components/Page/Page.stories.tsx
@@ -69,9 +69,9 @@ export const WithMenu: Story = {
   },
   render: ({ children, ...args }) => (
     <Page {...args}>
-      <SkipLink className="ams-page__area--skip-link" href="#inhoud">
-        Direct naar inhoud
-      </SkipLink>
+      <nav aria-label="Snelkoppelingen" className="ams-page__area--skip-link">
+        <SkipLink href="#inhoud">Direct naar inhoud</SkipLink>
+      </nav>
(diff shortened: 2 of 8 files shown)


VincentSmedinga:
Closing this: the wrapper is the wrong remedy, and checking it against the rest of the ecosystem is what settled it.

[NL Design System's page on 2.4.1](https://www.nldesignsystem.nl/wcag/2.4.1/) is explicit that there is nothing to fix here:

> Het plaatsen van een skiplink binnen de header-landmark is *best practice*. Maar, staat de skiplink bovenaan de pagina als eerste focusable element en buiten een landmark, dan is dit **geen** overtreding van dit WCAG-succescriterium.

So the axe result that prompted this — `region`, which is one of its best practices rather than a success criterion — was never a conformance problem.

And a `nav` is specifically the thing not to do. The [Utrecht README](https://github.com/nl-design-system/utrecht/blob/main/components/skip-link/README.md) says so in as many words:

> Niet: `<nav class="skip-links">…</nav>` Wel: `<div class="skip-links">…</div>`

Their `SkipLink` renders a `<p>` around the link, with no ARIA and no landmark, and they place the skip link *before* the landmarks rather than inside one. The NL Design System component page says nothing about a wrapper at all.

There was a second signal I should have taken more seriously while building this: the markup kept growing. A landmark needs a name, our naming convention rules out `aria-label`, so the name needed a hidden text node, and then the same block had to be copied into both Page Layouts, both introductions, the Home Page sketch and the Page story. That much ceremony around a link that stays hidden until it is focused was the tell.

Thanks to Copilot for the two review comments — both were right, and both are moot now.

What lands instead is a note in the Skip Link documentation recording that the placement is deliberate, that axe flags it, and that a `nav` is not the answer, with both sources cited. The point is that the next person to run an audit finds the reasoning instead of reopening this. PR follows.

Leaving one thing on the table for whoever picks it up: NL Design System calls placing the skip link inside the header landmark a best practice. That would clear the axe result without adding a landmark, but it needs a slot on Page Header and a decision about the Skip Link's design, which is a band above the header today rather than inside it. Worth a ticket, not worth bundling here.

PR #2869, opened 2026-07-31
feat(Storybook): Add internal Data Page example

VincentSmedinga:
# Describe the pull request

## Links

- Deploy links follow once the first preview is up.
- [Grid](https://designsystem.amsterdam/?path=/docs/components-layout-grid--docs), the Subgrid section in particular.
- [News Overview Page](https://designsystem.amsterdam/?path=/docs/pages-public-news-overview-page--docs) and [Search Results Page](https://designsystem.amsterdam/?path=/docs/pages-public-search-results-page--docs), which introduced this layout.

## What

Adds a Data Page to the internal page templates: a detail page showing all data of a single object, here an inspection report for a bridge. A sidebar holds the object's metadata and in-page navigation, and the sections of the report sit next to it — status assessments, condition reports, documents, and empty states.

The internal page layout gains a `menuItems` story parameter, so a page can show the sections of its own application in the Menu instead of the default set. The other internal templates are unaffected and keep the default items.

## Why

Internal applications keep asking for this shape: a narrow column of context beside a long stack of content that has to line up with the rest of the page. We had examples of a filter column beside a list of results, but none of a metadata column beside a set of report sections, and none where the sidebar itself is more than one block.

It also serves as a second, non-public exercise of the Subgrid, which is still new.

## How

Each side of the page is a `Grid.Subgrid` rather than a Cell:

- the sidebar spans 3 columns and holds two Cells, one for the metadata and the button, one for the table of contents;
- the sections span the remaining 9 and hold one Cell each.

Because each side is a single grid item, neither column has to span rows to stay beside the other, and every Cell still sits on the columns of the page. The alternative — Cells placed straight in the Grid — drops the second sidebar block below the sections, and a Cell nested in a Cell has no columns to align to.

Some smaller decisions:

- The Description Lists use plain Term and Description elements. A Section only groups several terms that share one description, which does not occur here; removing them changes the markup, not the rendering.
- Table captions follow both documented patterns: wrapped around the heading where nothing sits between it and the table, and visually hidden where a row of buttons does.
- Empty tables keep a row saying no data is available, rather than the informative Alert the Search Results Page uses for an empty result set. Happy to switch if you prefer one pattern across templates.

Opened as a draft — a few iterations on the content are likely, and it keeps Chromatic from running until it is worth reviewing.

## Checklist

- [x] Add or update stories

## Additional notes

- Page templates have no test stories: Chromatic snapshots every story under `Pages/` directly.
- The table of contents links are the first real in-page anchors in a page template; they target the section Cells.


VincentSmedinga:
Has been reviewed and approved by @dlnr outside of GitHub – especially the Subgrid approach within an internal website.

At the same time, this specific template is not yet generic enough to serve as an example in Storybook, so we will not merge the PR. Other templates for internal websites will follow.

PR #2944, opened 2026-08-20
feat(Known Date Input): Introduce as separate component

RubenSibon:
# Split the memorable date pattern from Date Input

## Links

- [DES-1947](https://gemeente-amsterdam.atlassian.net/browse/DES-1947)
- Feature branch deploy: [Known Date Input docs](https://designsystem.amsterdam/demo-DES-1947-date-input-group/?path=/docs/components-forms-known-date-input--docs), its [translated story](https://designsystem.amsterdam/demo-DES-1947-date-input-group/?path=/story/components-forms-known-date-input--translated) and its [test story](https://designsystem.amsterdam/demo-DES-1947-date-input-group/?path=/story/components-forms-known-date-input--test).
- [Date Input Group](https://nldesignsystem.nl/date-input-group/) and [Date Input](https://nldesignsystem.nl/date-input/) at NL Design System, which document the two patterns separately.
- [Date input](https://design-system.service.gov.uk/components/date-input/) at the gov.uk Design System, the reference for this pattern’s anatomy and validation.
- [Known Date](https://webawesome.com/docs/components/known-date/) at Web Awesome, where the name comes from.

## What

Introduces Known Date Input: separate text inputs for the parts of a date that the user knows by heart.

- Tokens: `ams.known-date-input.*`, with the gaps, the space between a label and its input, and two input widths.
- CSS: `.ams-known-date-input`, with its `__field`, `__input` and `__input--year` elements. Field Set puts the same space before a Known Date Input as before a Field.
- React: `KnownDateInput` with an `autoComplete` prop, and its parts `KnownDateInput.Day`, `.Month` and `.Year`, with unit tests.
- Storybook: a documentation page, stories for the component on its own, in a Field Set, with validation, for a month and year, and in another language, plus a test story.

Date Input loses its two ‘Memorable date’ stories and their documentation, which now live on the new page.

Date Input and Known Date Input refer to each other, and Text Input, Time Input, Calendar, Date Picker, Field Set and the shared Inputs token page link to the new component.

The Localisation guide explains how the two date components differ in the order they show, and how to wrap a Known Date Input for another language.

## Why

Date Input documented two patterns on one page: the native date control, and separate text inputs for a date the user knows by heart. The better option for a date of birth was easy to miss at the bottom of that page. NL Design System separates the two, and so do we now.

The name says when to use the component rather than what it looks like, which helps readers choose between our three date components. It deviates from NL Design System’s Date Input Group, which is still in the Help Wanted stage.

Before, every consumer assembled each field by hand: a Label, a Text Input, an id to link them, a numeric keyboard, an autocomplete value and a width. Now the component does that.

## How

- Each part renders a field: a Label and a Text Input. It sets what never changes:
  - `inFieldSet` on the Label
  - `inputMode="numeric"`
  - the width
  - a generated id, unless you pass one
  - the autocomplete value that follows from `autoComplete` on the Known Date Input
- `autoComplete` takes `birth-date` or `credit-card-expiry-date`, which the field maps to the `bday-*` and `cc-exp-*` tokens. Leave out the Day for a payment card; the component doesn’t guard against it.
- Labels default to Dag, Maand and Jaar, and `label` overrides them.
- Other props, `className`, `style` and the ref go to the input, as in Search Field Input, so a spread of React Hook Form’s `register()` works. The types leave out `autoComplete`, `inputMode`, `size` and `type`, because the field sets those itself.
- Each field sets the native `size` attribute: 2 for a day or a month, 4 for a year. This keeps the fields narrow when CSS fails to load. It also stops Text Input’s full-width rule for inputs without `size` from matching, so a plain class carries the width from the tokens.
- The widths are `calc(3ch + …)` and `calc(5ch + …)`: two or four digits, a character to spare for the caret, plus the padding and border from `ams.inputs.*`.
- The fields wrap onto a second row when they no longer fit, at narrow widths and at large text sizes.
- Day, Month and Year are thin wrappers around one internal field. Each declares its own `label`, so every props table shows its own default.
- Known Date Input has no `disabled` prop, because a disabled Field Set already disables every input inside it.
- The component never reads the locale. The fields appear in the order you place them, so a form in American English puts the Month first, as the translated story shows. A Date Input can’t do that: the browser picks the order of its segments.
- Safari on iOS only autofills a birth date when each field’s `id` or `name` contains the English word for its part, whatever language the page or the phone is in. Tests on an iPhone showed this. The generated ids therefore end in `-day`, `-month` and `-year`, and the docs ask consumers who set their own `id` or `name` to include that word.

## Checklist

- [x] Add or update stories

## Additional notes

- The name and the API follow the discussion in [this comment](https://github.com/Amsterdam/design-system/pull/2944#issuecomment-5696538103) and [this review](https://github.com/Amsterdam/design-system/pull/2944#pullrequestreview-5222253178).
- Few browsers autofill a birth date. Safari does. Chrome recognises the `bday-*` tokens but doesn’t fill them, and Firefox doesn’t support them. We keep the attributes, because they’re how we meet WCAG 1.3.5, Identify Input Purpose, and password managers can use them.
- The translated story uses English example content, where our stories are normally in Dutch. That’s the point of the story, and Character Count’s translated story does the same.
- In the sidebar, Known Date Input sorts between Invalid Form Alert and Label, not next to Date Input. Our `storySort` only orders categories, and the cross-links from Date Input help readers find it.
- The component count on the Storybook home page moved from 66 to 68. Counting per `documentation/introduction-statistics.md` gives 68, so the figure was already one behind before this branch.
- The design team has not reviewed the new tokens. The widths started from what the earlier stories produced with `size={2}` and `size={4}`, and the gaps were adjusted in review.

[DES-1947]: https://gemeente-amsterdam.atlassian.net/browse/DES-1947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/css/src/components/date-input-group/date-input-group.scss
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-date-input-group {
+  column-gap: var(--ams-date-input-group-column-gap);
+  display: flex;
+  flex-wrap: wrap;
```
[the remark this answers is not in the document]
RubenSibon: Added

storybook/src/components/DateInputGroup/DateInputGroup.docs.mdx
```diff
@@ -0,0 +1,105 @@
...
+Three separate fields let someone type the date the way they would say it, without first working out which format the browser expects.
+
+This is the pattern the [NL Design System](https://nldesignsystem.nl/richtlijnen/formulieren/wanneer-welk-form-element/iedereen-kan-invullen) and [gov.uk](https://design-system.service.gov.uk/components/date-input/) recommend for these dates.
+
+### When not to use
+
+Use a [Date Input](/docs/components-forms-date-input--docs) for a date the user does not know by heart and has to look up, such as an appointment.
+Its native control opens a calendar to browse, which a group of text fields cannot offer.
+
+Use a [Calendar](/docs/components-navigation-calendar--docs) when picking a date is the task of the page rather than one answer in a form.
```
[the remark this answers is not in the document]
RubenSibon: Good catch. Adjusted!

packages/css/src/components/date-input-group/date-input-group.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.ams-date-input-group__field {
+  display: flex;
+  flex-direction: column;
+  row-gap: var(--ams-date-input-group-field-row-gap);
+}
+
+// A Text Input without a `size` attribute fills its container; match that selector so an author
+// who does set one keeps the width the attribute gives them.
+.ams-date-input-group__field > .ams-text-input:not([size]) {
```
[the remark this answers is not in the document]
RubenSibon: Added!

RubenSibon (commented):
Vincent suggested naming this Known Date [as Web Awesome does](https://webawesome.com/docs/components/known-date/).

I think that name might be an improvement over Date Input Group. Known Date is more explicit about the key reason for using this over Date Input. 

Counterarguments:
- it is less descriptive of what you get: a group of 3 text inputs;
- it deviates from [NLDS’ Date Input Group](https://nldesignsystem.nl/date-input-group/).

Still, I lean to Known Date. The NLDS component is still in the Help Wanted-stage and names describing purpose to implementing developers are preferable over names describing structure.

VincentSmedinga:
> Vincent suggested naming this Known Date [as Web Awesome does](https://webawesome.com/docs/components/known-date/).
> 
> I think that name might be an improvement over Date Input Group. Known Date is more explicit about the key reason for using this over Date Input.

Actually my suggestion was Known Date Input, including the ‘Input‘ to match the other components.

> Counterarguments:
> 
> * it is less descriptive of what you get: a group of 3 text inputs;

I’d say this is no problem; it’s not really a goal for a component name to explain its appearance. E.g. Progress List doesn’t say anything about lines and circles.

> * it deviates from [NLDS’ Date Input Group](https://nldesignsystem.nl/date-input-group/).

True, but it also works the other way around. NLDS is curious about decisions made in the community so that they can learn from them and get inspired.

NLDS and ADS could even indefinitely maintain different names for a component, it’s no requirement that they are the same.

> Still, I lean to Known Date. The NLDS component is still in the Help Wanted-stage and names describing purpose to implementing developers are preferable over names describing structure.

Let’s do it. The name really helps in explaining our users when to use which of the three date input patterns that we offer.

VincentSmedinga (changes requested):
Every field in a Known Date Input is built the same way. It pairs a Label with `inFieldSet` and a Text Input with `inputMode="numeric"`, links the two with `id` and `htmlFor`, and the year gets `wide`. The stories repeat all of that for every field, and every consumer will too. I’d like the component to do it for them.

I suggest giving each part of the date its own subcomponent, which sets what never changes and passes everything else to the input:

```tsx
<FieldSet legend="Wanneer ben je geboren?">
  <KnownDateInput autoComplete="bday">
    <KnownDateInput.Day name="dag" />
    <KnownDateInput.Month name="maand" />
    <KnownDateInput.Year invalid name="jaar" />
  </KnownDateInput>
</FieldSet>
```

Each part renders its own Label and Text Input. It sets `inFieldSet`, `inputMode`, the width, a generated id and the autocomplete value, such as `bday-day`. Its label defaults to Dag, Maand or Jaar. For month and year, you leave out `Day`. The Field subcomponent and its `wide` prop go away. Search Field Input, Checkbox and Radio already generate an id and link their label the same way.

The types would look like this:

```tsx
export type KnownDateInputProps = {
  /**
   * Lets the browser fill in a date it already knows.
   * Use `bday` for a date of birth, and `cc-exp` for the expiry date of a payment card.
   */
  readonly autoComplete?: 'bday' | 'cc-exp'
} & Readonly<PropsWithChildren<HTMLAttributes<HTMLDivElement>>>

/** What every part shares: the props of a Text Input, minus the ones the part sets itself. */
export type KnownDateInputPartProps = Omit<TextInputProps, 'autoComplete' | 'inputMode' | 'size' | 'type'>

export type KnownDateInputDayProps = {
  /**
   * The text of the label.
   * @default 'Dag'
   */
  readonly label?: string
} & KnownDateInputPartProps

// KnownDateInputMonthProps and KnownDateInputYearProps are the same, with 'Maand' and 'Jaar' as their defaults.
```

A few notes on these types:

- The parts set `autoComplete`, `inputMode`, `size` and `type` themselves, so TypeScript rejects those props. Everything else a Text Input takes still works, including the ref and a spread of React Hook Form’s `register()`.
- `id` stays open, so Invalid Form Alert can link to a field. Without one, the part generates an id.
- `label` is declared on each part separately, so each props table shows its own default.
- `className` and `style` go to the input, as they do in Search Field Input.
- `autoComplete` only takes `bday` and `cc-exp`, because those are the only autofill values that come in parts. `cc-exp` has no day, so a Day part gets no autocomplete value then.
- The group has no `disabled` prop. A disabled Field Set already disables every input inside it.

Do you agree this would be an improvement?

RubenSibon:
> Every field in a Known Date Input is built the same way. It pairs a Label with `inFieldSet` and a Text Input with `inputMode="numeric"`, links the two with `id` and `htmlFor`, and the year gets `wide`. The stories repeat all of that for every field, and every consumer will too. I’d like the component to do it for them.
> 
> I suggest giving each part of the date its own subcomponent, which sets what never changes and passes everything else to the input:
> 
> [...]
> 
> Do you agree this would be an improvement?

Yes!

I’ve implemented the feature as you requested: Known Date Input now has Day, Month and Year parts.

Each part renders its own Label and Text Input, links them with a generated id (or the `id` you pass), and sets `inFieldSet`, `inputMode` and the autocomplete value derived from `autoComplete` on the Known Date Input (`bday` or `cc-exp`; a Day gets none with `cc-exp`).

Labels default to Dag, Maand and Jaar, and `className`, `style` and the ref go to the input, as in Search Field Input.

The Field subcomponent and `wide` are gone.

One addition: each part also sets the native `size` attribute itself, 2 for a day or month and 4 for a year. A Text Input without `size` gets `inline-size: 100%` from `.ams-text-input:not([size])`, which comes later in the bundle, so the old Field needed a `:not([size])` selector trick to win. With `size` set, that rule no longer matches, a plain `.ams-known-date-input__input` class carries the width from the tokens, and the fields stay narrow when CSS fails to load. `size` is still left out of the part props, so consumers can't override it.

Day, Month and Year are thin wrappers around one shared internal part. Each declares its own `label` prop, so every props table shows its own default.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+The labels default to Dutch, and the parts appear in the order you place them.
+For a form in another language, set `label` on each part and place the parts in the order its readers expect.
+This example asks for a date of birth in American English, which puts the month first.
+
+A Known Date Input does not read `lang`, so the order is always the one in your markup.
+Still set `lang` on the Field Set, so screen readers pronounce its text in the right language.
+See the [Localisation guide](/docs/docs-guidelines-localisation--docs) for more.
+
+<Canvas of={KnownDateInputStories.Translated} />
```
RubenSibon, line 96: @VincentSmedinga, note the Translated story that I added. As well as a few lines in the Localisation docs about the how Known Date Input differs from the more native Date Input.

storybook/src/components/KnownDateInput/KnownDateInput.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+    'KnownDateInput.Month': KnownDateInput.Month,
+    'KnownDateInput.Year': KnownDateInput.Year,
+  },
+} satisfies Meta<typeof KnownDateInput>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
```
VincentSmedinga, line 25: The default example should be an example of the component alone – not in a Field Set already, as in this case. All / most other form components have separate ‘In a Field’ and ‘In a Field with validation’ stories, let’s do that here as well.

packages-proprietary/tokens/src/components/ams/known-date-input.tokens.json
```diff
@@ -0,0 +1,55 @@
...
+        "$value": "{ams.space.m}",
+        "$description": "The space between the parts of the date.",
+        "$extensions": {
+          "nl.amsterdam.hint": "Must be a single value: it sets a longhand property, which takes only one.",
+          "nl.amsterdam.subtype": "space",
+          "nl.amsterdam.type": "dimension"
+        }
+      },
+      "row-gap": {
+        "$value": "{ams.space.s}",
```
VincentSmedinga: The space between label and input is small already – so the groups need a step larger between them to help visual group the levels. I think it also improves the appearance if an Error Message is present.

```suggestion
        "$value": "{ams.space.m}",
```

packages/css/src/components/known-date-input/known-date-input.scss
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+.ams-known-date-input {
```
VincentSmedinga, line 6: Not this line, but I think the Field Set description and the Known Date Input can use more space between them; I’d say `space-m`.

Field Set already has a couple of `:has(+` selectors, including `.ams-field-set > :has(+ .ams-error-message) {` which already targets another component.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+Without one, the part generates its own.
+
+Accept what the user types in the format they type it.
+Both `1` and `01` are a valid month, and rejecting one of the two turns a formatting preference into an error the user has to solve.
+
+## Examples
+
+### With validation
+
+To report an error, add `invalid` and an [Error Message](/docs/components-forms-error-message--docs) to the Field Set, and connect the Error Message to the Field Set with `aria-describedby`.
```
VincentSmedinga, line 75: Should we make explicit that our controls don’t do validation? I think it would answer a question people will have if we start with ‘to report an error’ like we do now.

If yes, I suggest including it in Features as well, and for all components (as a separate task). And we could use a ‘Form validation’ guideline that explains how we’d want that to work, what our component do and don’t do in that respect, and refer to the ee-rhf-ads repo.

```suggestion
The component does not validate its values itself.
To report an error, add `invalid` and an [Error Message](/docs/components-forms-error-message--docs) to the Field Set, and connect the Error Message to the Field Set with `aria-describedby`.
```

packages-proprietary/tokens/src/components/ams/known-date-input.tokens.json
```diff
@@ -0,0 +1,55 @@
...
+            "$value": "calc(5ch + (2 * {ams.inputs.padding-inline}) + (2 * {ams.inputs.border-width}))",
+            "$description": "Fits a four-digit year, with a character to spare for the caret. Includes the padding and border of the input, which sizes itself with `border-box`.",
+            "$extensions": {
+              "nl.amsterdam.hint": "Must be a single value: it sets a longhand property, which takes only one.",
+              "nl.amsterdam.type": "dimension"
+            }
+          }
+        }
+      },
+      "part": {
```
VincentSmedinga: ‘Part’ isn’t a bad name, but ‘field’ might be better. This `div` plays the same role as the root element of Field, wrapping label and input. It explains the ‘field set’ container and spares us a new concept. (Replace all occurrences, e.g. I see one in the `$description` of the `column-gap` token.)

packages/react/src/KnownDateInput/KnownDateInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { KnownDateInputDay } from './KnownDateInputDay'
+import { KnownDateInputMonth } from './KnownDateInputMonth'
+import { KnownDateInputYear } from './KnownDateInputYear'
+
+export type KnownDateInputProps = {
+  /**
+   * Lets the browser fill in a date it already knows.
+   * Use `bday` for a date of birth, and `cc-exp` for the expiry date of a payment card.
+   */
+  readonly autoComplete?: 'bday' | 'cc-exp'
```
VincentSmedinga: Let’s use human readable values here – `birth-date` and `credit-card-expiry-date`.

(Not `birthday`; that’s annually, and: “A **birth date**, as a full date.” – [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#bday).)

(Or `date="birth"` and `date="credit-card-expiry"`?)

packages/react/src/KnownDateInput/KnownDateInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+      </div>
+    </KnownDateInputContext.Provider>
+  ),
+)
+
+KnownDateInputRoot.displayName = 'KnownDateInput'
+
+/**
+ * Lines up the parts of a date that a user knows by heart, each in a field of its own.
+ * Typing a day, month, and year separately asks less of someone than filling in a single date control.
```
VincentSmedinga:
```suggestion
 * A set of three text inputs for the day, month and year of a date.
 * Typing these separately is easier for known dates than browsing through a date control.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+<Description of={KnownDateInputStories} />
+
+<Primary />
+
+<Controls />
+
+## Subcomponents
+
+Compose a Known Date Input from its parts: Known Date Input Day, Month, and Year.
```
VincentSmedinga:
```suggestion
Compose a Known Date Input from its Day, Month, and Year parts.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+Each part renders a [Label](/docs/components-forms-label--docs) and a [Text Input](/docs/components-forms-text-input--docs), links the two, and sets what every part of a date needs: a numeric keyboard, a width that fits its digits, and the autocomplete value that matches the `autoComplete` of the Known Date Input.
```
VincentSmedinga: Mand! 🧺 😁 

```suggestion
Each part renders a [Label](/docs/components-forms-label--docs) and a [Text Input](/docs/components-forms-text-input--docs), links the two, sets a numeric keyboard, a fitting width, and a chosen autocomplete value.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+Everything else you pass to a part, such as `name`, `invalid`, a `ref`, or a `className`, goes to its input.
+
+The labels default to Dag, Maand, and Jaar.
+Set `label` on each part for a form in another language.
```
VincentSmedinga: This connection with `label` is another reason to rename ‘part’ to ‘field’.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+### When to use
+
+Use a Known Date Input for a date the user already knows or can copy from a document in front of them: a date of birth, the expiry date of a passport, the day something happened.
+Separate fields let someone type the date the way they would say it, without first working out which format the browser expects.
+
+This is the pattern the [NL Design System](https://nldesignsystem.nl/richtlijnen/formulieren/wanneer-welk-form-element/iedereen-kan-invullen) and [gov.uk](https://design-system.service.gov.uk/components/date-input/) recommend for these dates.
+
+### When not to use
+
+Use a [Date Input](/docs/components-forms-date-input--docs) for a date the user does not know by heart and has to look up, such as an appointment.
```
VincentSmedinga: This contraction seems odd to me: ‘a date […] that […] has to look up’.

Maybe ‘they have’ instead of ‘has’? Or even simpler, leave the duplication out:

```suggestion
Use a [Date Input](/docs/components-forms-date-input--docs) for a date that the user has to look up, such as for an appointment.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+Set `label` on each part for a form in another language.
+
+## Usage guidelines
+
+### When to use
+
+Use a Known Date Input for a date the user already knows or can copy from a document in front of them: a date of birth, the expiry date of a passport, the day something happened.
+Separate fields let someone type the date the way they would say it, without first working out which format the browser expects.
+
+This is the pattern the [NL Design System](https://nldesignsystem.nl/richtlijnen/formulieren/wanneer-welk-form-element/iedereen-kan-invullen) and [gov.uk](https://design-system.service.gov.uk/components/date-input/) recommend for these dates.
```
VincentSmedinga: More direct:

```suggestion
Both [NL Design System](https://nldesignsystem.nl/richtlijnen/formulieren/wanneer-welk-form-element/iedereen-kan-invullen) and [gov.uk](https://design-system.service.gov.uk/components/date-input/) recommend this pattern for these dates.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+Use a [Date Input](/docs/components-forms-date-input--docs) for a date the user does not know by heart and has to look up, such as an appointment.
+Its native control opens a calendar to browse, which a group of text fields cannot offer.
+
+Use a [Date Picker](/docs/components-forms-date-picker--docs) when the user has to choose a date or a date range from a calendar they can see, such as when booking an appointment.
+
+### How to use
+
+Wrap a Known Date Input in a [Field Set](/docs/components-forms-field-set--docs) whose legend asks for the date as a question.
+The legend is what ties the parts together for a screen reader; without it, three labels read as three unrelated questions.
```
VincentSmedinga:
```suggestion
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+Put a Paragraph with an example date above the Known Date Input and connect it to the Field Set with `aria-describedby`.
+Write the example in the order the parts appear, so it doubles as a format hint.
+
+Set `autoComplete` to `bday` for a date of birth, or to `cc-exp` for the expiry date of a payment card, so the browser can fill in a date it already knows.
+Leave it out for any other date.
```
VincentSmedinga:
```suggestion
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+Give each part an `id` when an [Invalid Form Alert](/docs/components-forms-invalid-form-alert--docs) has to link to it.
+Without one, the part generates its own.
```
VincentSmedinga:
```suggestion
Give each part an `id` so that an [Invalid Form Alert](/docs/components-forms-invalid-form-alert--docs) can link to it.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+Accept what the user types in the format they type it.
+Both `1` and `01` are a valid month, and rejecting one of the two turns a formatting preference into an error the user has to solve.
```
VincentSmedinga:
```suggestion
Both `1` and `01` are a valid month, and rejecting one of the two gives the user an unnecessary task.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+## Examples
+
+### With validation
+
+To report an error, add `invalid` and an [Error Message](/docs/components-forms-error-message--docs) to the Field Set, and connect the Error Message to the Field Set with `aria-describedby`.
+Set `invalid` on the parts that are at fault, and only on those: mark the year alone if the year is what is missing.
```
VincentSmedinga:
```suggestion
Set `invalid` only on the fields that are at fault.
```

storybook/src/components/KnownDateInput/KnownDateInput.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <KnownDateInput.Year invalid name="jaar" />
+      </KnownDateInput>
+    </FieldSet>
+  ),
+}
+
+export const MonthAndYear: Story = {
+  render: (args) => (
+    <FieldSet aria-describedby="description-c" legend="Sinds wanneer woon je op dit adres?">
+      <Paragraph id="description-c">Bijvoorbeeld 3 2019.</Paragraph>
```
VincentSmedinga: I understand the example, but I’m not entirely sure that we wouldn’t use a `<Select>` for the month in this case. Maybe leave it for now, but extend the instruction?

```suggestion
      <Paragraph id="description-c">Bijvoorbeeld 3 2019 voor maart 2019.</Paragraph>
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+### Translated
+
+The labels default to Dutch, and the parts appear in the order you place them.
+For a form in another language, set `label` on each part and place the parts in the order its readers expect.
+This example asks for a date of birth in American English, which puts the month first.
+
+A Known Date Input does not read `lang`, so the order is always the one in your markup.
+Still set `lang` on the Field Set, so screen readers pronounce its text in the right language.
+See the [Localisation guide](/docs/docs-guidelines-localisation--docs) for more.
```
VincentSmedinga:
```suggestion
Make sure the `lang` attribute is in effect; see the [Localisation guide](/docs/docs-guidelines-localisation--docs) for more.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+<Canvas of={KnownDateInputStories.Translated} />
+
+## Design
+
+Each part is as wide as the answer it takes: two digits for a day or a month, four for a year.
+That difference is a hint in itself, and it keeps the parts from reading as three interchangeable boxes.
+The widths hold without CSS too, because each input also carries a `size` attribute for its digits.
+
+The parts wrap onto a second line when they no longer fit beside each other, which happens at narrow widths and at large text sizes rather than at a fixed breakpoint.
```
VincentSmedinga: Beware for AI’s ‘rather than’! :)

```suggestion
The parts wrap onto a second line when they no longer fit beside each other, which happens at narrow widths and at large text sizes.
```

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
VincentSmedinga: Good idea.

Please add Known Date Input to its ‘The same pattern applies to the Date Picker…’ section.

And maybe even a TSX example with `const AmericanKnownDateInput`?

storybook/src/components/DateInput/DateInput.docs.mdx
```diff
@@ -133,15 +114,16 @@ That behaviour is the reason for using the native control rather than rebuilding
 
 The format the field shows and expects comes from the browser and the operating system, not from the language of the page.
 A visitor whose browser is set to another locale sees the order of day, month and year they are used to, which is helpful — but it also means the page cannot promise which one they will see.
+A [Known Date Input](/docs/components-forms-known-date-input--docs) shows the same order to everyone, which is what makes an example date above the fields possible there.
```
VincentSmedinga: Complicated language: “, which is what makes an example date above the fields possible there.”

packages/react/src/KnownDateInput/KnownDateInputPart.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+/** The Label and Text Input that Day, Month and Year each render. */
+export const KnownDateInputPart = forwardRef(
+  ({ className, id, label, part, ...restProps }: PartProps, ref: ForwardedRef<HTMLInputElement>) => {
+    const generatedId = useId()
+    const inputId = id || generatedId
+    const { autoComplete } = useContext(KnownDateInputContext)
+
+    // A payment card has no day to fill in.
+    const hasAutoComplete = autoComplete && !(autoComplete === 'cc-exp' && part === 'day')
```
VincentSmedinga: But it also doesn’t even render a part / field for it, right? Seems like this condition is for when people use `cc-exp` with a `KnownDateInput.Day`, which is their mistake. We don’t generally guard incorrect implementations.

packages/react/src/KnownDateInput/KnownDateInput.tsx
VincentSmedinga: Autocomplete doesn’t work correctly on my iPhone. This happens after tapping the bar on top of the virtual keyboard containing my birth date:

Image

VincentSmedinga: Can we add tests for this?

VincentSmedinga: Can you test on Android?

packages/react/src/KnownDateInput/KnownDateInputField.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Label } from '../Label/Label'
+import { TextInput } from '../TextInput/TextInput'
+import { KnownDateInputContext } from './KnownDateInputContext'
+
+/** What every part shares: the props of a Text Input, minus the ones the part sets itself. */
+export type KnownDateInputPartProps = Omit<TextInputProps, 'autoComplete' | 'inputMode' | 'size' | 'type'>
+
+type PartProps = {
+  readonly label: string
+  readonly part: 'day' | 'month' | 'year'
```
VincentSmedinga, line 29: Just to be complete, can we add TSDoc comments to these props as well?

storybook/src/components/KnownDateInput/KnownDateInput.stories.tsx
VincentSmedinga: Separate task: all stories named ‘In a field with validation’ could just be named ‘With validation’ (and sit right after the ‘In a field’ one).

VincentSmedinga (changes requested):
Great to see this as a separate, full-fledged component!

Remarks are mostly about docs and a class name or prop value here and there – but also the question why autocomplete doesn’t seem to work on iPhone.

packages/react/src/KnownDateInput/KnownDateInput.tsx
RubenSibon: Web reports show Safari's split-birthday autofill is unreliable ([MUI #11063](https://github.com/mui/material-ui/issues/11063), [Webflow forum](https://discourse.webflow.com/t/cant-get-my-form-to-be-compatible-with-autofill-date/226027), [12 Days of Web](https://12daysofweb.dev/2021/form-autocomplete)); no authoritative WebKit source found. Unit tests can't cover real AutoFill (jsdom has none). Keep attributes regardless: they satisfy WCAG 1.3.5.

I’ve tried to test this on my Android in two browsers (Chromium and Firefox), but I see no way to autofill a birthdate. I can autofill addresses and payment card details. I tried the latter, but that also filled in the month and year into the day field when that was selected.

What you can still do: test [gov.uk's date-of-birth example](https://components.publishing.service.gov.uk/component-guide/date_input/with_autocomplete_for_date_of_birth_fields) on same iPhone and see if that does work.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
RubenSibon: Added Known Date Input to that section, with an `AmericanKnownDateInput` example.

storybook/src/components/KnownDateInput/KnownDateInput.stories.tsx
RubenSibon: Not going to do the second request as this is out of the scope of this story. Could be a refactor accross many components, but not too high priority imo.

RubenSibon: One the first request: default now shows the component on its own, followed by ‘In a Field Set’ and ‘In a Field Set with validation’, as for Checkbox and Radio.

packages/css/src/components/known-date-input/known-date-input.scss
RubenSibon: Added `.ams-known-date-input` to the Field Set rule that already puts `space-m` before a Field or Field Set, so a description or Error Message above it gets that space too.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
RubenSibon: Applied the suggestion here.

A Features entry for all form components and a ‘Form validation’ guideline sound good as a separate task.

packages-proprietary/tokens/src/components/ams/known-date-input.tokens.json
RubenSibon: Renamed: the element is now `.ams-known-date-input__field`, the token `ams.known-date-input.field.row-gap`, and the internal component `KnownDateInputField`.

The docs call the label and input a field throughout; Day, Month and Year are still ‘parts’, as in your suggestion for the Subcomponents section.

packages/react/src/KnownDateInput/KnownDateInput.tsx
RubenSibon: Now `birth-date` and `credit-card-expiry-date`, still on `autoComplete` since that’s what it controls.

The component maps them to [bday](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#bday) and [cc-exp](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#cc-exp).

RubenSibon: After we did some troubleshooting with various test cases hopefully we have no isolated the issue and fixed it: for iOS/Safari’s birthdate autofill to work they require English terms in either the `id` or the `name`. We have added it to the `id`. See the [docs and comments](https://github.com/Amsterdam/design-system/pull/2944/commits/7e3d759bb45264508e459e5b1a52e4f8d8dcae66) in the code.

Can you test [our Dutch story](https://designsystem.amsterdam/demo-DES-1947-date-input-group/iframe.html?id=components-forms-known-date-input--in-a-field-set&viewMode=story) again on your iPhone?

packages/react/src/KnownDateInput/KnownDateInputField.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const { autoComplete } = useContext(KnownDateInputContext)
+
+    return (
+      <div className="ams-known-date-input__field">
+        <Label htmlFor={inputId} inFieldSet>
+          {label}
+        </Label>
+        <TextInput
+          {...restProps}
+          autoComplete={autoComplete && `${autoCompleteTokens[autoComplete]}-${part}`}
```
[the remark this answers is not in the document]
RubenSibon: Developers shouldn’t use `KnownDateInput.Day` if that is not applicable for the use-case such as with creditcard expiry dates.

As Vincent said before: [[w]e don’t generally guard incorrect implementations](https://github.com/Amsterdam/design-system/pull/2944#discussion_r4035723387).

On this point, maybe we should add a story for creditcards. What do you think, @VincentSmedinga?

storybook/src/components/KnownDateInput/KnownDateInput.test.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </FieldSet>
+      <FieldSet invalid legend="Wanneer ben je geboren?">
+        <KnownDateInput>
+          <KnownDateInput.Day defaultValue={16} />
+          <KnownDateInput.Month defaultValue={8} />
+          <KnownDateInput.Year invalid />
+        </KnownDateInput>
+      </FieldSet>
+      {/* Too narrow for three fields side by side, so the year drops onto a second row. */}
+      <div style={{ maxInlineSize: '14rem' }}>
```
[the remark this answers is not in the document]
RubenSibon: I’ll check what can be done.

packages/react/src/KnownDateInput/KnownDateInput.tsx
```diff
@@ -0,0 +1,46 @@
...
+        {children}
+      </div>
+    </KnownDateInputContext.Provider>
+  ),
+)
+
+KnownDateInputRoot.displayName = 'KnownDateInput'
+
+/**
+ * A set of three text inputs for the day, month and year of a date.
```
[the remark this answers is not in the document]
RubenSibon: Good point. It now reads: ‘Separate text inputs for the parts of a date, such as the day, month and year.’

storybook/src/components/KnownDateInput/KnownDateInput.test.stories.tsx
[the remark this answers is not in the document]
RubenSibon: Changed: the constraint now sits on the Field Set, so there’s no extra `div`, and it’s calculated from the Known Date Input tokens.

It leaves room for the day and month side by side, but not the year, so the wrap still shows and keeps doing so if the widths or gap change.

storybook/src/components/KnownDateInput/KnownDateInput.docs.mdx
VincentSmedinga: I don’t see the sentence, but that’s fine – it will be added with https://gemeente-amsterdam.atlassian.net/browse/DES-2037.

packages/react/src/KnownDateInput/KnownDateInput.tsx
VincentSmedinga: > Can you test [our Dutch story](https://designsystem.amsterdam/demo-DES-1947-date-input-group/iframe.html?id=components-forms-known-date-input--in-a-field-set&viewMode=story) again on your iPhone?

Works correctly with both English and Dutch language settings 🎉  Nice browser quirk!

packages/react/src/KnownDateInput/KnownDateInputField.tsx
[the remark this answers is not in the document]
VincentSmedinga: Not necessery, I think. We won’t be be performing creditcard payments ourselves, and we already have an example for a month and a year.

VincentSmedinga (approved):
This will spare countless individuals the hassle of hunting through a date picker just to locate their birth date! 🎉

PR #2952, opened 2026-08-25
feat(File Card, File List)!: Present one file, in a list or on its own

RubenSibon:
# feat(File Card, File List)!: Present one file, in a list or on its own

## Links

- [DES-1959](https://gemeente-amsterdam.atlassian.net/browse/DES-1959)
- [File Card documentation](https://designsystem.amsterdam/demo-DES-1959-file-component/?path=/docs/components-forms-file-card--docs)
- [File List documentation](https://designsystem.amsterdam/demo-DES-1959-file-component/?path=/docs/components-forms-file-list--docs)
- [Using a file input](https://designsystem.amsterdam/demo-DES-1959-file-component/?path=/story/components-forms-file-list--with-input)
- [File at NL Design System](https://www.nldesignsystem.nl/file/)

## What

Introduces File Card, which presents one file with its type and size and the actions available for it. It works on its own as well as inside a File List, so an attachment on a case page and an upload in a form are the same component.

File List Item now renders whatever is put inside it. Its `file` and `onDelete` props are deprecated until 2027-04-01 and warn in the development build.

Breaking: the deprecated props no longer display a thumbnail for an image file.

## Why

We had no way to show a file outside a list. Everything known about a file was locked inside a list item, while the rest of government uses the same display on its own. NL Design System calls this component File, and [Discussion 347](https://github.com/orgs/nl-design-system/discussions/347) shows nine organisations building it under five names, so a shared model makes ours recognisable to anyone arriving from another municipal system.

It is also the occasion to settle six defects at once, all of which lived in that list item:

- The visible text could not be translated, and `formatFileSize` produced English decimal points with trailing zeros. A zip file was called a Document, and a file of a terabyte read `1.00 undefined`.
- The thumbnail used the file name as its alternative text, so screen readers announced that name twice.
- A long name was cut off with an ellipsis, losing exactly the date or reference that tells two files apart.
- Only one action was possible, and it was fixed inside the list item.
- `URL.createObjectURL` ran in the render and `URL.revokeObjectURL` appeared nowhere in the source, so every render leaked a blob URL and the file behind it.
- After a delete, the focused element disappeared and focus fell back to `document.body`.

## How

File Card takes `name`, `size`, `type` and `previewUrl` rather than a `File` object. That keeps the global `File` name out of consumers' modules, where importing a component of that name would shadow the constructor, and it makes the component work for files that already sit on a server. It also leaves the object URL with whoever creates it, so no path remains in which the component makes a URL that nobody revokes.

`formatDetailsText` returns the line below the name, in Dutch by default, following the formatter-property pattern of Character Count. `Intl.NumberFormat` replaces `toPrecision`, units reach terabytes, and a missing value drops its part instead of printing `undefined`. Dutch and English formatters are exported.

The deprecated `file` prop renders a File Card internally rather than the old markup, so consumers who have not migrated get the fixes without acting. That also stops React emitting `ams-file-list__item-preview`, `ams-file-list__item-info` and the misnamed `ams-file-input__item-details`; the three stay in the stylesheet for consumers writing the markup by hand. The `ams.file-list.file.*` tokens point at their `ams.file-card.*` replacements and carry a `$deprecated` message, per the procedure in the tokens README.

```diff
- <FileList.Item file={file} onDelete={remove} />
+ <FileList.Item>
+   <FileCard name={file.name} onDelete={remove} size={file.size} type={file.type} />
+ </FileList.Item>
```

## Checklist

- [x] Add or update stories
- [x] Tested with NVDA
- [ ] Tested with JAWS
- [ ] Tested with VoiceOver

## Additional notes

**Three decisions taken against the acceptance criteria**, each for a stated reason:

- The delete button does carry the file name, in an `ams-visually-hidden` span. The criterion's reasoning is about screen clutter and button width, and hidden text costs neither; without it, someone reviewing the buttons of a page out of context hears the same word for each. The visible label stays the first part of the accessible name, so WCAG 2.5.3 holds.
- The File Input example creates its object URLs in the change handler rather than in a `useEffect`. The repo's `react-hooks/set-state-in-effect` rule makes the literal pattern an error. `URL.revokeObjectURL` is still in an effect clean-up.
- The migration paragraph is not in the release notes. `documentation/release-notes.md` rules out upgrade instructions and deprecation notices on that page and asks a major to name what changed and leave migration to the changelogs. The before-and-after fragment is on the File List page instead.

**A Firefox bug surfaced during screen reader testing.** Firefox does not refresh what a screen reader reads beside the browse button when the selection is changed in code, so NVDA announced a file after it had been removed. Chromium is unaffected. Reproduced on Firefox ESR 140.14.0, which matters because that is what many government desktops are held at. The workaround is its own commit and reverts on its own if this is fixed upstream.

**The removal date, 2027-04-01, is an assumption.** It is at least six months after a release landing in September or October 2026. If the 5.0 date is known, it should be checked against the two token files, the stylesheet and the `FileListItem` props.

**Still outstanding:**

- Verification of the post-delete focus behaviour with a second screen reader. NVDA covers both the next/previous move and the empty list.
- Feeding our approach back into [NL Design System Discussion 347](https://github.com/orgs/nl-design-system/discussions/347).
- The UX decisions on the visible defaults: `Verwijder` or `Verwijderen`, the wording for file types, whether `Document` returns as a fallback, how many digits a size gets, and whether `previewUrl` ships at all. These are default values in one file, but they are visible to everyone using the component.
- Release notes, drafted and waiting on a date and version numbers.

[DES-1959]: https://gemeente-amsterdam.atlassian.net/browse/DES-1959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/css/src/components/file-card/file-card.scss
```diff
@@ -0,0 +1,55 @@
...
+
+.ams-file-card {
+  display: flex;
+  flex: 1;
+  flex-wrap: nowrap;
+  font-family: var(--ams-file-list-file-font-family /* @deprecated */, var(--ams-file-card-font-family));
+  font-size: var(--ams-file-list-file-font-size /* @deprecated */, var(--ams-file-card-font-size));
+  font-weight: var(--ams-file-list-file-font-weight /* @deprecated */, var(--ams-file-card-font-weight));
+  gap: var(--ams-file-list-file-gap /* @deprecated */, var(--ams-file-card-gap));
+  line-height: var(--ams-file-list-file-line-height /* @deprecated */, var(--ams-file-card-line-height));
```
dlnr: Why would you need deprecations in a new component?

packages/react/src/FileCard/FileCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {details && <span className="ams-file-card__details">{details}</span>}
+        </div>
+        {onDelete && (
+          <div className="ams-file-card__actions">
+            <Button onClick={deleteFile} variant="tertiary">
+              {deleteButtonLabel}
+              <span className="ams-visually-hidden">{` ${name}`}</span>
+            </Button>
+          </div>
+        )}
```
dlnr, line 86: Having only the option to delete really narrows its usage. Actions, like the Dialog component, could be a slot to add your own buttons. Stuff like download, publish, remove (with a dialog) or edit to name a view I've been building in apps dealing with assets.

storybook/src/components/FileList/FileList.stories.tsx
```diff
@@ -14,16 +14,12 @@ const meta = {
...
-        file={new File(['sample2'], 'sample2.txt', { lastModified: Date.now(), type: 'text/plain' })}
-        key="2"
-        onDelete={() => {}}
-      />,
+      <FileList.Item key="1">
+        <FileCard name="paspoort.pdf" onDelete={() => {}} size={1536000} type="application/pdf" />
+      </FileList.Item>,
+      <FileList.Item key="2">
+        <FileCard name="loonstrook.pdf" onDelete={() => {}} size={248000} type="application/pdf" />
+      </FileList.Item>,
```
dlnr, line 22: Is filelist really necessary? I've originally created the list so that we could "force" users to add only filelist.item components as a child component. But your spec for this new feature is that FileCard (or just call it File..) should function as such on itself, like isolated. This new approach would allow you to just use UnorderedList without markers or grid as ul [like search results](https://designsystem.amsterdam/demo-DES-1959-file-component/?path=/story/pages-public-search-results-page--default) as it is just a column with a gap.

packages/react/src/FileCard/FileCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Presents one file with its type and size, and the actions available for it.
+ *
+ * @see {@link https://designsystem.amsterdam/?path=/docs/components-forms-file-card--docs File Card docs at Amsterdam Design System}
+ */
+export const FileCard = forwardRef(
+  (
+    {
+      className,
+      deleteButtonLabel = 'Verwijder',
```
dlnr, line 52: This is a pitfall to have static dutch as default label. It's not apparent you need to supply a translated label to have a proper deletebutton.

packages/css/src/components/file-card/file-card.scss
```diff
@@ -0,0 +1,55 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@use "../../common/text-rendering" as *;
+
+.ams-file-card {
```
dlnr, line 8: It's not really a card if you display it inline like a list item.

I would've built this with a container query to allow this to look like an actual card if there is enough space. Something like this?

Image

storybook/src/components/FileList/FileList.stories.tsx
dlnr: Or just suggest a story using `<Column>` or `<Row>`, this would make it very flexible. 

The UL (in file list) was more for screenreaders though, but in the case of a11y and selected files of a file input this might be reading it double. So the screenreader would read out the selected files when the focus is on the input en then again when focussing the list, or probably only the delete button, so not ideal.

US Design system has some testing done for the file input: https://designsystem.digital.gov/components/file-input/accessibility-tests/ , using the devtools a11y tree did not work well for me, selected files are not visible in the tree. This might be a devtools bug though.

packages/react/src/FileCard/FileCard.tsx
RubenSibon: Yes, we certainly want this component to have the flexibility to add more actions in the future. For now, we start by only implementing the one action `FileList.Item` already had: deletion.

In a future iteration we can think about the best way to have more actions: providing a slot might very well be the best solution, but currently that is beyond the scope of this PR.

packages/css/src/components/file-card/file-card.scss
RubenSibon: This is a pattern [we introduced a while ago](https://github.com/Amsterdam/design-system/pull/2707) to graciously fallback from deprecated tokens to the tokens that replace them in CSS with the usage of [custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties#custom_property_fallback_values).

In most cases, the deprecated tokens will not be present, and the fallback will kick-in.

If developers still define those deprecated tokens in their project manually, those tokens will apply.

But for a new component it does indeed not make sense to have those deprecations. I’ll remove them here.

storybook/src/components/FileList/FileList.stories.tsx
RubenSibon: Good point! File List has been reduced to a wrapper with little added value. We could keep it for legacy reasons, but maybe a better solution could be found. Let me think about it.

On the point of screen readers reading the list twice: this is not the case for NVDA in the current demo.

packages/react/src/FileCard/FileCard.tsx
RubenSibon: After giving it some more thought, it would be good to get the PO’s opinion on this. He originally wrote the requirement that File Card should for now have one built-in action: deletion, but he also wrote that his envisioned API is not yet set in stone. We’ll discuss next week.

storybook/src/components/FileList/FileList.stories.tsx
RubenSibon: After reading through the ticket again, the PO was clear about seeing no reason to get rid of File List. He said:

> FileList blijft de container: de naam klopt, de ul-markup klopt en er zijn geen gebreken in gevonden.

I agree there is no pressing reason to remove (deprecate) this component at this time.

packages/css/src/components/file-card/file-card.scss
RubenSibon: This is also a discussion to be had with the PO. It goes above the scope of this PR and acceptance requirements.

packages/react/src/FileCard/FileCard.tsx
RubenSibon: This is the same convention `CharacterCount`, `FormFieldStatus`, and `Pagination` already use throughout this repo: a Dutch default, overridable via a prop, documented in the localisation guide.

I agree that we should make as obvious as possible that a translated label should be provided. But that goes for all components that use this convention.

packages/css/src/components/file-card/file-card.scss
RubenSibon: Removed the deprecated tokens from the new component: https://github.com/Amsterdam/design-system/pull/2952/commits/0d1b0af5ab3082705c4722b00c715d1eeb078463

packages/css/src/components/file-card/file-card.scss
dlnr: You've said that you're naming it FileCard to align with NLDS but they call it [File](https://nldesignsystem.nl/file/)  just like the implementation from Den Haag. This might make more sense.

RubenSibon: No, I did not say that. We align with the model, but not necessarily with the complete specification or naming conventions of other community implementations (such as Den Haag):

> Waarom we aansluiten bij NLDS. File staat op [NL Design System](http://nldesignsystem.nl/)  als component met een gedeelde naam en een gedeeld model: meta-informatie over een bestand plus één of meer acties. Discussie [#347](https://github.com/orgs/nl-design-system/discussions/347) laat zien dat Den Haag, DUO, Amsterdam, Open Gemeenten, Rotterdam, Open Formulieren, Open Inwoner, DSO en Groningen ditzelfde ding onder vijf verschillende namen bouwen. Aansluiten bij één model maakt onze component herkenbaar voor wie uit een ander gemeentelijk systeem komt, en het is de aanleiding om de zes tekortkomingen hierboven in één keer op te lossen in plaats van los van elkaar.

> De NLDS-naam is File, maar File is een globale naam in de browser en in TypeScript. Een const File op moduleniveau overschaduwt de ingebouwde File-constructor in de hele module, ook in de regels erboven, en dat raakt niet alleen ons — FileList.test.stories.tsx en FileListItem.stories.tsx maken allebei een new File([...]) aan — maar ook elke consument die de component importeert in een module waar bestanden worden geconstrueerd.

> De component heet daarom File Card, in code FileCard, met het CSS-blok ams-file-card volgens onze conventie dat de React-naam en de blokklasse gelijk lopen. Die naam is niet verzonnen: Atlassian gebruikt in de Forge UI Kit FileCard, met fileName en fileSize als gegevens en onDownload en onDelete als acties die verschijnen zodra je ze meegeeft — functioneel het dichtst bij wat wij willen — en Ant Design X hanteert dezelfde naam. Ter vergelijking: Salesforce Lightning noemt het File, met .slds-file, .slds-file__icon, .slds-file__figure en .slds-file__text, maar bij hen bestaat de naam alleen in CSS, waar het probleem niet speelt; IBM Carbon noemt het FileUploaderItem en koppelt het daarmee vast aan uploaden, wat wij juist niet willen. Dat wij ook een Card hebben en deze weergave daar geen variant van is, nemen we voor lief. In de documentatie noemen we File als NLDS-naam, zodat de component vindbaar is voor wie die term kent.

Above are citations of the Jira item which you could read for the proper context.

Again, when Vincent is back, we can discuss naming and how we want to align with the rest. Indeed, Card might not be the best option, but calling it just File while it is metadata about a file and there are clashes with reserved keywords does not seem ideal either.

storybook/src/components/FileList/FileList.stories.tsx
dlnr: Having a third file component might be confusing to users, especially since FileList.Item and FileCard look exactly the same. FileList.Item will probably be the one the IDE autocompletes or ai suggests, so you're going to get a component that'll be used very randomly.

RubenSibon: That’s a fair point. Let’s discuss that next week as well.

PR #2953, opened 2026-08-27
feat(Tokens): Split two-value spacing tokens into start and end

RubenSibon:
# Split two-value spacing tokens into start and end

## Links

- [DES-1970](https://gemeente-amsterdam.atlassian.net/browse/DES-1970)
- Follows [#2883](https://github.com/Amsterdam/design-system/pull/2883), which recorded the constraint as a hint
- [Tokens that must hold a single value](https://github.com/Amsterdam/design-system/blob/feat/DES-1970-split-tokens-with-multiple-independent-values/packages-proprietary/tokens/AGENTS.md#tokens-that-must-hold-a-single-value) in the tokens `AGENTS.md`
- [`ams/require-single-value-token`](https://github.com/Amsterdam/design-system/blob/feat/DES-1970-split-tokens-with-multiple-independent-values/stylelint/README.md#amsrequire-single-value-token) in the Stylelint plugin README
- Dialog docs: https://designsystem.amsterdam/demo-DES-1970-split-tokens-with-multiple-independent-values/?path=/docs/components-containers-dialog--docs
- Menu docs: https://designsystem.amsterdam/demo-DES-1970-split-tokens-with-multiple-independent-values/?path=/docs/components-navigation-menu--docs
- Modal Dialog docs: https://designsystem.amsterdam/demo-DES-1970-split-tokens-with-multiple-independent-values/?path=/docs/components-containers-modal-dialog--docs
- Select docs: https://designsystem.amsterdam/demo-DES-1970-split-tokens-with-multiple-independent-values/?path=/docs/components-forms-select--docs

## What

Makes the single-value constraint on dimension tokens structural instead of advisory, in three parts:

1. Splits the ten tokens that hold two values — four on Dialog, one on Menu, four on Modal Dialog, one on Select — into a `-start` and `-end` token each, and deprecates the originals.
2. Fails the token build when a token typed as a dimension resolves to more than one value.
3. Adds an `ams/require-single-value-token` Stylelint rule that reports a `var()` resolving to more than one value inside `calc()`, `clamp()`, `max()` or `min()`, or on a longhand that takes a single value.

## Why

A token read through a math function or assigned to a longhand takes one value.
Given two, the declaration is invalid at computed-value time and the property falls back to its initial value rather than to the token, so the spacing disappears rather than degrading.
Nothing distinguished the two kinds of token beyond a `nl.amsterdam.hint` someone has to read.
Making Grid, Page Header, Page Footer and Menu clear a display cutout wraps spacing tokens in `max()`, which turns this from a latent risk into an active one.

## How

Each deprecated token now references its two halves, e.g. `ams.dialog.header.padding-block` is `{ams.dialog.header.padding-block-start} {ams.dialog.header.padding-block-end}`, so it resolves to the same value as before.
Three older aliases that forwarded to one of the ten (`ams.dialog.header.medium.padding-block`, `ams.dialog.footer.medium.padding-block` and `ams.menu.wide.padding-block`) point at the halves too, so their migration message names a token that is not itself deprecated.
The deprecations run until 2027-02-25, six months from when they were added.

The component stylesheets are unchanged: they keep reading the deprecated shorthand tokens through the `padding-block` and `padding-inline` shorthands.
That keeps an application that overrides one of those tokens working during the deprecation window, since no CSS can read one half of a two-value custom property.

The build check runs on the default, Compact and Lo-fi dictionaries before any output is written, so an offending token never reaches `dist/`.
It reads both `nl.amsterdam.type` and the DTCG `$type`, and skips deprecated tokens.

The Stylelint rule builds on the plugin's existing token resolution.
It judges each operand of a math function on its own, so an operand it cannot resolve does not hide a problem in the one beside it.
It treats a `var()` nested in another `var()` as a fallback rather than judging it twice.
Shorthands such as `padding-block` and `gap` are not checked, since they take a value per side on purpose.

## Checklist

- [x] Add or update stories

## Additional notes

- `feat` rather than `fix`: the twenty `-start` and `-end` tokens are new public API, which warrants a minor release of the tokens package.
- No stories or index exports needed: the change adds tokens and tooling, not components, and the rendered components do not change.
- No visual change is expected. Built from `develop` and from this branch, every existing custom property resolves to the same value; the only difference is the twenty new properties. `/chromatic test` has been triggered to confirm.
- The fourteen `nl.amsterdam.hint` notes added in #2883 stay as documentation of why a token needs one value; the checks make them advisory rather than load-bearing.
- Out of scope, per the story: this guards the tokens and stylesheets in this repository, not an application that overrides a custom property with two values.

[DES-1970]: https://gemeente-amsterdam.atlassian.net/browse/DES-1970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stylelint/require-single-value-token.mjs
```diff
@@ -0,0 +1,234 @@
...
+      /* Interpolation only resolves once the stylesheet is compiled. A value without a reference
+       * holds whatever it literally says, which needs no token resolution to read.
+       */
+      if (hasInterpolation(declaration.value) || !declaration.value.includes('var(')) {
+        return
+      }
+
+      for (const { functionName, reference } of findMathReferences(declaration.value)) {
+        const call = declaration.value.slice(reference.start, reference.end)
+        const values = resolveToValues(call)
```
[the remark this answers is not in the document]
RubenSibon: Thanks for flagging this — traced it through and it doesn't reproduce in the actual pipeline.

postcss strips CSS comments from `decl.value` at parse time, before the rule ever sees it. `var(--foo /* @deprecated */, fallback)` becomes `var(--foo , fallback)` by the time `require-single-value-token.mjs` reads `declaration.value`, so the lookup key `resolveValue` extracts is already clean — no comment embedded.

This is the same pattern the other five rules in this plugin already rely on (`no-fixed-sizes`, `require-grid-minmax`, etc.), and the test at `rules.test.mjs:350-354` covers exactly this case: a deprecated alias (`var(--ams-alias-two-value-padding-block /* @deprecated */, ...)`) resolving to two values does get reported.

`resolveValue` would indeed mis-resolve if fed a raw string with the comment still embedded, but that's not a path the rule (or any caller) exercises — `custom-properties.mjs` is always called with postcss-parsed values. Not changing this one.

RubenSibon:
I change the commit type to `fix` and expanded the scope because these changes are published to packages. This way Release Please will pick them up.

packages-proprietary/tokens/lint/README.md
```diff
@@ -10,7 +10,7 @@ Reports design tokens that the build defines but nothing consumes.
 Until this check, nothing enforced that, and review missed at least one case for over a year: `--ams-search-field-input-cancel-button-color` has never been referenced by [search-field.scss](../../../packages/css/src/components/search-field/search-field.scss) in the entire history of that file.
 
 Every other check in the repository runs from source to the token dictionary, never the other way.
-In [.stylelintrc.json](../../../.stylelintrc.json), `csstools/value-no-unknown-custom-properties` and the five `ams/*` rules all take the built token CSS as an `importFrom` dictionary and validate declarations against it.
+In [.stylelintrc.json](../../../.stylelintrc.json), `csstools/value-no-unknown-custom-properties` and the six `ams/*` rules all take the built token CSS as an `importFrom` dictionary and validate declarations against it.
```
VincentSmedinga: Let’s stop updating the count.

```suggestion
In [.stylelintrc.json](../../../.stylelintrc.json), `csstools/value-no-unknown-custom-properties` and the `ams/*` rules all take the built token CSS as an `importFrom` dictionary and validate declarations against it.
```

packages-proprietary/tokens/AGENTS.md
```diff
@@ -44,19 +44,27 @@ Token files use the `.tokens.json` extension and follow the DTCG format:
...
-Nothing enforces this yet, so every such token carries a `nl.amsterdam.hint` that says so, in one of two wordings:
+
+Two checks enforce this, so it does not have to be remembered:
+
+- `pnpm build` fails and names every offending token, through [style-dictionary/multiValueDimensions.js](style-dictionary/multiValueDimensions.js).
+- `ams/require-single-value-token` reports a stylesheet reading such a token where one value fits, through [the Stylelint plugin](../../stylelint/README.md).
+
+The build check exempts a deprecated token, since it is frozen at the value it shipped with until it is removed.
+The Stylelint rule does not: it judges the declaration in front of it, deprecated or not, so a deprecated alias that still resolves to two values is still reported where one value fits.
+Where a value per side is wanted, use a type that takes several values, or split the token in two and set the longhands — as `ams.dialog.header.padding-block` was split into `ams.dialog.header.padding-block-start` and `ams.dialog.header.padding-block-end`.
```
VincentSmedinga: This offers “use a type that takes several values” as an alternative to splitting the token. None of our other types is meant for spacing, and the build check only applies to tokens typed as a dimension, so this reads as a way to skip the check. The tokens README only mentions splitting the token. Could this say the same?

```suggestion
Where a value per side is wanted, split the token in two and set the longhands, as `ams.dialog.header.padding-block` was split into `ams.dialog.header.padding-block-start` and `ams.dialog.header.padding-block-end`.
```

stylelint/require-single-value-token.mjs
```diff
@@ -0,0 +1,236 @@
...
+
+const ruleName = 'ams/require-single-value-token'
+
+/* The functions that compute one value from an expression, so each of their operands is one value. */
+const MATH_FUNCTIONS = ['calc', 'clamp', 'max', 'min']
+
+/* The properties that take exactly one value. Shorthands are deliberately absent: `padding-block`
+ * and `gap` take two on purpose, and the design system ships tokens that hold two for them.
+ * So are `border-block-width` and the corner radii, which take a second value of their own.
+ */
```
VincentSmedinga, line 28: After this PR, the only two-value tokens left for `padding-block` and `gap` are deprecated, and the build rejects new ones. Leaving shorthands out of the list still makes sense because they take two values, but the comment doesn't need the part about the tokens we ship.

```suggestion
/* The properties that take exactly one value. Shorthands are deliberately absent: `padding-block`
 * and `gap` take two on purpose. So are `border-block-width` and the corner radii, which take a
 * second value of their own.
 */
```

VincentSmedinga (commented):
Great! This could prevent quite an array of bugs. Just some remarks to get the docs right.

packages/css/src/components/dialog/dialog.scss
```diff
@@ -70,7 +70,8 @@ so do not apply these styles without an `open` attribute. */
   flex-wrap: nowrap;
   gap: var(--ams-dialog-header-gap);
   justify-content: space-between;
-  padding-block: var(--ams-dialog-header-padding-block);
+  padding-block-end: var(--ams-dialog-header-padding-block-end);
+  padding-block-start: var(--ams-dialog-header-padding-block-start);
```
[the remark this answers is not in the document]
RubenSibon: Good catch — I reverted the header/footer `padding-block` changes in `dialog.scss` so the deprecated `--ams-dialog-*-padding-block` tokens are consumed again (including the medium fallback chain). This keeps downstream overrides working through their deprecation window.

packages/css/src/components/menu/menu.scss
```diff
@@ -57,7 +57,8 @@
     display: none;
     margin-inline: 0; // Override negative margin above.
     max-inline-size: var(--ams-menu-wide-max-inline-size /* @deprecated */, var(--ams-menu-vi-wide-max-inline-size));
-    padding-block: var(--ams-menu-wide-padding-block /* @deprecated */, var(--ams-menu-vi-wide-padding-block));
+    padding-block-end: var(--ams-menu-vi-wide-padding-block-end);
+    padding-block-start: var(--ams-menu-vi-wide-padding-block-start);
```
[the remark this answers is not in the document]
RubenSibon: Fixed. `menu.scss` now uses `padding-block: var(--ams-menu-wide-padding-block /* @deprecated */, var(--ams-menu-vi-wide-padding-block));` again, so both deprecated wide tokens remain effective until removal.

packages/css/src/components/modal-dialog/modal-dialog.scss
```diff
@@ -117,19 +117,22 @@ Content elements – e.g. a heading and a subtitle – stack without needing a w
   overflow-y: auto;
   /* stylelint-disable-next-line plugin/use-baseline -- Progressive enhancement. */
   overscroll-behavior: contain;
-  padding-block: var(--ams-modal-dialog-body-padding-block);
+  padding-block-end: var(--ams-modal-dialog-body-padding-block-end);
+  padding-block-start: var(--ams-modal-dialog-body-padding-block-start);
```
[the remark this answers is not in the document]
RubenSibon: Fixed — I restored `padding-block` usage for modal-dialog body/footer at both base and medium breakpoint. That reactivates the deprecated `--ams-modal-dialog-*-padding-block` tokens and preserves existing downstream overrides during deprecation.

packages/css/src/components/select/select.scss
```diff
@@ -21,7 +21,8 @@
   max-inline-size: 100%;
   outline-offset: var(--ams-select-outline-offset);
   padding-block: var(--ams-select-padding-block);
-  padding-inline: var(--ams-select-padding-inline);
+  padding-inline-end: var(--ams-select-padding-inline-end);
+  padding-inline-start: var(--ams-select-padding-inline-start);
```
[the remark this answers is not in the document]
RubenSibon: Addressed. `select.scss` now consumes `--ams-select-padding-inline` again via `padding-inline`, so deprecated-token overrides keep working until the documented removal date.

packages-proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -126,8 +141,8 @@
         },
         "medium": {
           "padding-block": {
-            "$deprecated": "Use `ams.dialog.header.vi-medium.padding-block` instead. Will be removed on or after 2026-10-20.",
-            "$value": "{ams.dialog.header.vi-medium.padding-block}",
+            "$deprecated": "Use `ams.dialog.header.vi-medium.padding-block-start` and `ams.dialog.header.vi-medium.padding-block-end` instead. Will be removed on or after 2026-10-20.",
+            "$value": "{ams.dialog.header.vi-medium.padding-block-start} {ams.dialog.header.vi-medium.padding-block-end}",
```
VincentSmedinga: The stylesheets read the deprecated tokens again, but this alias now skips `ams.dialog.header.vi-medium.padding-block`.

`dialog.scss` reads `--ams-dialog-header-medium-padding-block` first. The token CSS always defines it, so the `vi-medium` fallback never applies. On develop that didn’t matter, because the medium token forwarded to `--ams-dialog-header-vi-medium-padding-block`. Now it forwards to the two halves, so an app that overrides `--ams-dialog-header-vi-medium-padding-block` sees no change. Until this PR, that was the token we told people to use.

Could the value keep pointing at the `vi-medium` token? The `$deprecated` message can still name the halves. The resolved value stays the same, and an override anywhere in the chain works again, as step 3 of “Renaming a token” in the tokens README expects. The token build passes with this change.

```suggestion
            "$value": "{ams.dialog.header.vi-medium.padding-block}",
```

packages-proprietary/tokens/src/components/ams/dialog.tokens.json
```diff
@@ -218,8 +263,8 @@
         },
         "medium": {
           "padding-block": {
-            "$deprecated": "Use `ams.dialog.footer.vi-medium.padding-block` instead. Will be removed on or after 2026-10-20.",
-            "$value": "{ams.dialog.footer.vi-medium.padding-block}",
+            "$deprecated": "Use `ams.dialog.footer.vi-medium.padding-block-start` and `ams.dialog.footer.vi-medium.padding-block-end` instead. Will be removed on or after 2026-10-20.",
+            "$value": "{ams.dialog.footer.vi-medium.padding-block-start} {ams.dialog.footer.vi-medium.padding-block-end}",
```
VincentSmedinga: Same as the header: this alias skips `ams.dialog.footer.vi-medium.padding-block`, so an override of `--ams-dialog-footer-vi-medium-padding-block` has no effect.

```suggestion
            "$value": "{ams.dialog.footer.vi-medium.padding-block}",
```

packages-proprietary/tokens/src/components/ams/menu.tokens.json
```diff
@@ -46,8 +46,8 @@
           "$type": "dimension"
         },
         "padding-block": {
-          "$deprecated": "Use `ams.menu.vi-wide.padding-block` instead. Will be removed on or after 2026-10-20.",
-          "$value": "{ams.menu.vi-wide.padding-block}",
+          "$deprecated": "Use `ams.menu.vi-wide.padding-block-start` and `ams.menu.vi-wide.padding-block-end` instead. Will be removed on or after 2026-10-20.",
+          "$value": "{ams.menu.vi-wide.padding-block-start} {ams.menu.vi-wide.padding-block-end}",
```
VincentSmedinga: Same here: `menu.scss` reads `--ams-menu-wide-padding-block` first, so an override of `--ams-menu-vi-wide-padding-block` has no effect while this alias skips it.

```suggestion
          "$value": "{ams.menu.vi-wide.padding-block}",
```

VincentSmedinga (changes requested):
Three aliases still drop an override of the `vi-` token they used to forward to. Suggestions on the lines below.

RubenSibon:
> The Stylelint rule can miss interpolated math violations and falsely report function-like text containing escaped quotes.

I’ve committed some fixes and tests to the issues you raised with some help of your local counterpart.

I do find it hard to understand exactly what these longer, more complicated regexes do. This is an example of: I just have to trust that what AI finds and solves is sound. 🤷🏻

VincentSmedinga (approved):

PR #2958, opened 2026-09-03
feat(Card)!: Display the image in an area of a fixed shape

RubenSibon:
# Display the image in an area of a fixed shape

## Links

- [Card documentation in this branch’s deploy](https://designsystem.amsterdam/demo-DES-1973-card-image-aspect-ratio/?path=/docs/components-navigation-card--docs)
- [The new Image Fit story](https://designsystem.amsterdam/demo-DES-1973-card-image-aspect-ratio/?path=/story/components-navigation-card--image-fit)
- [DES-1973](https://gemeente-amsterdam.atlassian.net/browse/DES-1973)

## What

Card Image now renders a wrapping element around its `img`. That element gets an area of a fixed shape, 16 by 9 to begin with, from the new `ams.card.image.aspect-ratio` token — the same shape an Image has, so a Card that is left alone needs no thought about it.

Two props follow from that:

- `aspectRatio` now shapes the area rather than the image itself.
- `objectFit` is new. It is `contain` by default, which fits the image inside the area and keeps all of it in view. Set it to `cover` to fill the area and crop the image to it.

`CardImageProps` is exported, since Card Image now has props of its own beyond those of Image.

## Why

Card Image drew its image at the full width of the Card, so a portrait aspect ratio made it taller than the Card is wide. In a grid or a list, that left the text of one Card far below that of the next, which is not what a set of Cards is meant to look like.

Giving the image an area of its own caps its height and lets every Card in a set be the same size, whatever the shape of the files behind them.

## How

The wrapping element carries `ams-card__image` and the aspect ratio class; the `img` inside is a plain `ams-image`. The area sets `aspect-ratio` and `display: grid`, and the image fills it rather than its width alone, which is what limits the height. A modifier class flips `object-fit` to `cover`.

The area is a ratio rather than a maximum height so that it scales with the width of the Card, the way every other box in the system does. A length would never bite in a narrow Card and would crop a plain 16 by 9 image in a wide one.

Cropping is a decision rather than an assumption, so `contain` is the default and `cover` is asked for. Where the image does not reach the edges, its own background colour covers the area — the one Image already shows while it loads — so no new colour token was needed.

The area follows the pattern of the Image Slider thumbnail, which also sets an aspect ratio on the container and lets the image do its work inside it.

## Checklist

- [x] Add or update stories

## Additional notes

**This is a breaking change for anyone styling the markup.** `ams-card__image` moves from the `img` to the element around it, and a `className` passed to `Card.Image` now lands on that element too. A `ref` still reaches the `img`, and every Image prop is still passed to it.

Visual regressions are expected and intended. The Image Fit story is new, and the visual test story no longer passes `aspectRatio="4:3"` — that prop reshapes the area now, so leaving it in would have snapshotted a letterboxed image in a story that is about the switch to a horizontal layout.

Two points that could use a second opinion:

- The guideline that every Card in one set shares an `aspectRatio` has no enforcement; it is set per Card. If it needs teeth, the lever is overriding `--ams-card-image-aspect-ratio` once on the container of the set.
- Card Image renders the wrapping element itself, passing the rest of its props to the nested Image. The alternative discussed was a separate subcomponent around it, which would keep the two sets of props apart but adds a component to the API — and would break the container query selector `.ams-card:has(> .ams-card__image)`, which needs Card Image to stay a direct child of Card.

[DES-1973]: https://gemeente-amsterdam.atlassian.net/browse/DES-1973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


storybook/src/components/Card/Card.stories.tsx
```diff
@@ -110,6 +116,38 @@ export const HorizontalLayout: DefaultStory = {
...
+          <Card.Image alt="" objectFit="cover" src="https://picsum.photos/id/122/800/1200" />
+          <Card.Heading level={2}>
+            <Card.Link href="#">Bijgesneden</Card.Link>
+          </Card.Heading>
+          <Paragraph>De afbeelding vult het vlak, waarbij de boven- en onderkant wegvallen.</Paragraph>
+        </Card>
+      </Grid.Cell>
+    </Grid>
+  ),
+}
```
RubenSibon, line 149: The [Code tab for this Story](https://designsystem.amsterdam/demo-DES-1973-card-image-aspect-ratio/?path=/story/components-navigation-card--image-fit) should only show the rendered code and not the Storybook boilerplate around it.

packages/css/src/components/card/card.scss
```diff
@@ -27,8 +27,30 @@ $ams-card-horizontal-breakpoint: 36rem;
     outline-width: 0.0625rem;
   }
 
-  > .ams-card__image:not(:last-child) {
-    margin-block-end: var(--ams-card-image-margin-block-end);
+  /*
+   * Gives the image an area of its own, so every Card in a set is the same size whatever the
+   * shape of the file.
+   */
+  > .ams-card__image {
```
VincentSmedinga: The child combinator `>` is only there to combine with `:last-child`, to guard against unintended hits (even though nested Cards would be an abonimation). Now it’s used for all these selectors. Maybe we should keep the `margin-block-end` declaration in a separate block, and not use the child combinator for the aspect ratio things where it’s unnecessary.

packages/react/src/Card/CardImage.tsx
```diff
@@ -8,17 +8,34 @@ import type { ForwardedRef } from 'react'
...
-import type { ImageProps } from '../Image'
+import type { ImageProps } from '../Image/Image'
 
-import { Image } from '../Image'
+import { generateAspectRatioClass } from '../Image/generateAspectRatioClass'
+import { Image } from '../Image/Image'
+
+export type CardImageProps = {
+  /** How the image fills the area. Contain shows all of it, cover crops it to the edges. */
+  readonly objectFit?: 'contain' | 'cover'
```
VincentSmedinga: We can make this a prop of Image itself. It would make the feature usable for images in all contexts, not just in cards. And spare us a child combinator in CSS, which is at odds with encapsulation – the styles can go into `image.scss`.

packages/css/src/components/card/card.scss
```diff
@@ -27,8 +27,30 @@ $ams-card-horizontal-breakpoint: 36rem;
...
+      margin-block-end: var(--ams-card-image-margin-block-end);
+    }
+
+    /*
+     * Fills the area rather than the width alone, which is what limits the height of the image.
+     * Its own background then covers the area too, wherever the file does not reach the edges.
+     */
+    > .ams-image {
+      block-size: 100%;
+      object-fit: contain;
```
VincentSmedinga: Could `cover` be the default, with `contain` as the option?

Image itself uses `object-fit: cover`, so every Card with an image crops it today. With `contain` as the default, every existing Card whose file doesn’t match its area gets letterboxed after the upgrade, even though the consumer changed nothing. That’s the main reason this PR needs the `!`.

With `cover` as the default, a Card that passes an `aspectRatio` looks the same as before: the area takes the ratio the image had, and the image fills it the same way. Consumers who want the whole image in view can choose `contain`. The upgrade then only changes where `className` lands and adds the wrapping element.

I see the point about cropping being a decision. But consumers already made it when they chose Image, and a new default shouldn’t reverse it for them.

This would mean:

- dropping the `object-fit: contain` rule, since `.ams-image` already covers
- renaming the modifier to `ams-card__image--contain`, and checking for `objectFit === 'contain'` in `CardImage`
- updating the JSDoc of `objectFit`, the token description, and the two lines in `Card.docs.mdx` that mention `cover`

VincentSmedinga (changes requested):
I put you on the wrong path in the ticket, sorry. I wrote that Card Image had to render a `div` around the `img`, but I tried it without one, and I think we don’t need it.

An `img` can be the area itself. Image already sets `aspect-ratio` and `inline-size: 100%`, so the box has a fixed shape, and `object-fit` decides how the file sits inside it. I rendered your markup and CSS next to a version with `object-fit` on the `img` alone. I used a portrait file in both fits, without an aspect ratio and with `4:3` and `9:16`, in the vertical and horizontal layouts. They match in Chromium; Safari even renders incorrectly with the wrapper.

So I’d suggest a smaller change:

- An `objectFit` prop on Image, with `contain` and `cover`, where `cover` is the default. That’s what Image does today, so nothing changes for existing images, and the change isn’t breaking.
- Card Image passes all Image props on, so Cards get the prop without any change to Card.
- The Card docs then say what the ticket is after. Give every Card in a set the same aspect ratio, and use `contain` for a portrait photo instead of a portrait aspect ratio.

Your Image Fit story and most of the docs text would carry over.

Two things to settle first. With `contain`, the loading colour stays visible beside the image after it loads, which is a question for design. (Related, Aram has let me know that the grey background isn’t ideal in our File component. Maybe we should revert that.)

And it’s worth checking whether NL Design System already names this prop.

This changes the PR so much that a new one may be simpler than reworking this branch. What would you prefer?

RubenSibon:
> I put you on the wrong path in the ticket, sorry. I wrote that Card Image had to render a `div` around the `img`, but I tried it without one, and I think we don’t need it.
> 
> [...]
> 
> This changes the PR so much that a new one may be simpler than reworking this branch. What would you prefer?

I’ll investigate for myself if we can do without the `div`. Then I’ll think about making those changes in this PR or start fresh.

RubenSibon:
> I’ll investigate for myself if we can do without the `div`. Then I’ll think about making those changes in this PR or start fresh.

I’ve also come to the conclusion that we do not need the `div` wrapper. So reverting this and removing a breaking-change PR is a major win.

I think I’ll just refactor and/or revert these changes in this PR.

On whether `objectFit` should live on `Image` rather than `Card.Image`: `.ams-image` currently hardcodes `object-fit: cover`. Developer variable fitting should an Image concern. Image would keep `cover` as its default for compatibility and then `Card.Image` would default it to `contain`. That removes Card’s only bespoke prop. So that seems like a sensible change.

Can we smuggle that change into this PR...? No, I guess a seperate ticket and PR for `Image` would be the correct way to go. Thoughts, @VincentSmedinga?

VincentSmedinga:
> `Card.Image` would default it to `contain`

I’m not so sure that’s the best option.

‘Contain’ shows grey (or white) bands at the edges of an image with a different intrinsic aspect ratio. I think that would surprise people in a Card. Cards often come in sets and are expected to all look rectangular and equal. I’d keep `cover` as the default everywhere. A different default in a different context seems confusing as well.

Maybe we have two contexts here:
1. deliberate: a portrait image in a Card → contain (coloured bands are expected)
2. unintentional: an image of say 1634 × 898 pixels → cover (a slight cutoff is accepted)

This lets people set the prop in the deliberate path, instead of finding an error with an image with a slightly deviating aspect ratio. It’s more forgiving.

That also suggests against reusing this branch. This PR is for Card, the new approach targets Image (albeit with some Card docs), that’s quite the difference. And we don’t really need the history of the stray path.

RubenSibon:
Closing this PR for 2 new branches and PRs: one for `Image` and one for `Card[.Image]`.

PR #2959, opened 2026-09-04
fix(Card): Document and demonstrate a Card without a link

RubenSibon:
# Document and demonstrate a Card without a link

## Links

- [DES-2004](https://gemeente-amsterdam.atlassian.net/browse/DES-2004)
- [Card docs, "Without a link" example](https://designsystem.amsterdam/demo-DES-2004-card-without-link-example/?path=/docs/components-navigation-card--docs)
- [Without a link story](https://designsystem.amsterdam/demo-DES-2004-card-without-link-example/?path=/story/components-navigation-card--without-link)

## What

Adds a documented example of a Card without a `Card.Link`: three Cards naming the phases of a project, each with a `Card.HeadingGroup` and a `Metadata` tagline, none of them clickable.

Updates the Card guidelines to say a link is optional rather than implied mandatory, and corrects two TSDoc descriptions (`Card`, `Card.Heading`) that stated the link as a certainty.

Explicitly rules out one adjacent case: a heading with a set of links, such as the blocks a Navigation Page is built from. Those stay Prose; a Card would wrap them in an `article` element they don't need.

## Why

`Card` never required a link in code — `Card.Heading` renders fine without a `Card.Link` inside it — but every story used one, and the docs read as though a link were mandatory ("Give a Card one link and no more"). NLDS's own vision on Cards ([discussion 422](https://github.com/orgs/nl-design-system/discussions/422#discussioncomment-17353679)) treats a Card as a structure that may or may not carry a link, and this PR brings our documentation in line with what the component already does.

This grew out of a ticket asking whether our Card should also cover Navigation Page's Link Sections (heading + Link List, no link). It shouldn't: tested that conversion directly against the real CSS and tokens and it regresses spacing, adds unneeded `article` landmarks, and produces a double focus ring when a `LinkList.Link` sits inside a Card without being a `Card.Link`. This PR only adds the case that *did* hold up: a linkless Card for content that has no page of its own.

## How

- `storybook/src/components/Card/Card.stories.tsx`: new `WithoutLink` story, three Cards in a Grid row.
- `storybook/src/components/Card/Card.docs.mdx`: "When to use", "When not to use", "How to use", "Examples", "Features", and "Accessibility" updated to describe the linkless case; explicitly warns against using it for a heading-plus-links block.
- `packages/react/src/Card/Card.tsx` and `CardHeading.tsx`: TSDoc wording no longer states the link as certain.

No changes to `Card`, `CardHeading`, or `CardLink` behaviour — the linkless case already worked.

## Checklist

- [x] Add or update stories

## Additional notes

- **Draft, experimental.** This is a suggestion for a linkless Card example, not a settled decision — it may change shape or get rejected once the wider "Card as Link vs. Card" question is resolved.
- Feedback wanted specifically on: whether "Without a link" is the right example content (project phases), and whether the "When not to use" line pointing at Navigation Page's Link Sections belongs here or should wait for that decision to land separately.

[DES-2004]: https://gemeente-amsterdam.atlassian.net/browse/DES-2004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storybook/src/components/Card/Card.docs.mdx
+++ b/storybook/src/components/Card/Card.docs.mdx
@@ -75,18 +75,27 @@ Set `linkComponent` to integrate with a routing library.
 A Card needs more content than just a title.
 Add a short text and optionally an image and metadata.
 
+A Card does not need a link.
+Use one without a link for a set of blocks that carry the same kind of content and line up in a row, where each names something that has no page of its own.
+
 ### When not to use
 
 This component is not the best option if the target content does not represent an article-like page.
 In that case, use a Heading, a Paragraph and a [Standalone Link](/docs/components-navigation-standalone-link--docs) instead.
+
+Do not use a Card for a block of a heading and a set of links, such as the ones a [Navigation Page](/docs/pages-public-navigation-page--docs) is built from.
+[Prose](/docs/utilities-css-prose--docs) spaces the parts of such a block, and the `article` element of a Card would announce it as a composition of its own.
 
 ### How to use
 
 Card Heading requires a `level`, because there is no sensible default.
 The guidelines for regular [links](/docs/components-navigation-link--docs) and [headings](/docs/components-text-heading--docs) apply.
 
-Give a Card one link and no more.
+Give a Card that leads to a page one link and no more.
 The Card Link covers the whole Card to make it clickable, so a second link inside it would sit underneath and be impossible to click.
+
+Leave a Card without a link plain.
+Nothing about it should suggest that it can be clicked: no hover effect, no pointer cursor, and no arrow beside its title.
 
 ### How to write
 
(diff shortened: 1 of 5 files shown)


storybook/src/components/Card/Card.stories.tsx
```diff
@@ -155,3 +155,48 @@ export const TopTasks: Story = {
...
   ),
 }
+
+export const WithoutLink: Story = {
+  parameters: {
+    layout: 'fullscreen',
+  },
+  render: (args) => (
+    <Grid paddingVertical="x-large">
+      <Grid.Cell span={{ narrow: 4, medium: 4, wide: 4 }}>
```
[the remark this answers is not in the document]
RubenSibon: Makes sense.

storybook/src/components/Card/Card.docs.mdx
```diff
@@ -36,7 +36,7 @@ The heading provides screen reader users with enough context.
 
 ### Heading
 
-Contains the name and link of the Card.
+Contains the name of the Card, and its link where the Card has one.
```
VincentSmedinga:
```suggestion
Contains the name of the Card, often wrapped in a link.
```

storybook/src/components/Card/Card.docs.mdx
```diff
@@ -75,19 +75,28 @@ Set `linkComponent` to integrate with a routing library.
 A Card needs more content than just a title.
 Add a short text and optionally an image and metadata.
 
+A Card does not need a link.
+Use one without a link for a set of blocks that carry the same kind of content and line up in a row, where each names something that has no page of its own.
```
VincentSmedinga:
```suggestion
It can make a block with one or a few links at the end or even without a link.(Standalone Link, Link List) without a link for a set of blocks that carry the same kind of content and line up in a row, where each names something that has no page of its own.
```

TODO

PR #2966, opened 2026-09-04
feat(Data List): List labels with values, actions and a header

RubenSibon:
# feat(Data List): List labels with values, actions and a header

## Rename of Data Summary to Data List

We’ve renamed Data Summary to Data List during the review process of this PR.

The Figma page, associated git branch and build names have not been renamed.

We’ve also moved this component to the ‘Containers’ group.

## Links

- [DES-1991](https://gemeente-amsterdam.atlassian.net/browse/DES-1991)
- [Data Summary in Figma (not renamed yet)](https://www.figma.com/design/9IGm6IdPUYizBNGsUnueBd/Amsterdam-Design-System?node-id=48983-138)
- [Data List documentation page](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/docs/components-containers-data-list--docs)
- [Orientation](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--orientation) – resize to see a label and its value stack
- [Stacked](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--stacked)
- [Actions](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--actions) – includes an item with two actions
- [Composite value](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--composite-value)
- [Header](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--header)
- [Form review](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/story/components-containers-data-list--form-review) – groups per person, each under a Data List Header
- [Description List](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/docs/components-text-description-list--docs) and [Table](https://designsystem.amsterdam/demo-DES-1991-data-summary/?path=/docs/components-containers-table--docs), whose See also sections now name Data List

## What

Adds Data List: an overview that presents data as a list of labels with their values, separated by a line, where an item can offer one or more actions on its value.
It comes with Data List Header, which introduces a Data List with a heading and at most one action.

Both ship across all four layers.

- Tokens: `ams.data-list.*` and `ams.data-list-header.*`, tracing the values in the Figma design.
- CSS: `.ams-data-list`, with `__item`, `__label`, `__value` and `__actions` elements, `--horizontal` / `--vertical` and `--narrow` / `--medium` / `--wide` modifiers; and `.ams-data-list-header`.
- React: `DataList` with `Item`, `Label`, `Value` and `Actions` subcomponents and `orientation` and `termsWidth` props, and a separate `DataListHeader`, both exported from the package barrel.
- Storybook: one documentation page covering both components, with stories for both orientations, items with one or two actions, a composite value, a header, and a form review with several groups.

Data List Header has no documentation page of its own, like Hint; the Data List page documents it.

Two existing pages gain a link: Description List and Table both describe content Data List suits better, so their See also sections now name it.

## Why

Data Summary (sic) was added to the Figma library and had no counterpart in code.
Services were left building the pattern by hand, which is where the accessible name of a repeated ‘Wijzigen’ link and the relationship between a label and its value tend to get lost.

A summary of the answers in a form is the most common application of this pattern.
Rather than shipping a separate Form Summary component, this PR documents it as an application of Data List, which keeps one component to maintain.

Long or multipage forms summarise several people or subjects on one page.
A Heading without a line above each Data List made those groups hard to tell apart, so Data List Header ships together with the list.

## How

The markup is a description list: `dl` for the list, a `div` per item, `dt` for the label and `dd` for the value.
This ties each label to its value in the markup rather than only through the layout, which is what [WCAG 1.3.1](https://www.w3.org/TR/WCAG22/#info-and-relationships) asks for.

Actions renders a second `dd`, the way the GOV.UK Summary List does it.
HTML offers no better element inside a `dl`, and it means a ‘Wijzigen’ link is announced in the context of the label it acts on.
The text, the icon and the type of each action are the consumer’s to choose; the stories pass Standalone Links, and give each one a unique accessible name with the `ams-visually-hidden` helper.

Several actions sit on one line with a thin separator between each pair, as in the Figma building block: a border on every action after the first, with the same space on either side.
They never wrap, so a separator never starts a row; in a narrow column their labels wrap instead.

An item is a three-column grid: the label, the value, and the actions.
The label column is at most `15rem` by default, or a ratio set through `termsWidth`, and the actions column collapses to nothing for items that have none.
Below a container width of `32rem` the grid does not apply, so every value moves under its label and the component never forces horizontal scroll.

Data List Header renders a `div` rather than a `header`, so it never becomes a banner landmark outside `main`.
It owns the line below its content and, through `:has(+ .ams-data-list)`, the space to a Data List that follows.
The Row, Heading and action inside it are plain compositions and keep their own styles.
A header cannot live inside the `dl`: a `div` there may only wrap `dt` and `dd`, which the W3C validator confirms.

Description List already solved two of these problems.
Both components switch orientation at the same container width, and both level a label with the first line of its value by giving it half the difference in computed line heights.
Rather than duplicate either, this PR moves the width to `common/breakpoint.scss` and the calculation to a mixin in `common/align-first-line.scss`, and points both components at them.
Compiled CSS for Description List is byte-identical, verified by compiling its `description-list.scss` from `develop` and diffing against the new one.

## Checklist

- [x] Add or update stories

## Additional notes

**Decisions to confirm with design.**

- The header line uses `ams.color.text.default` (#202020), the closest token to the raw black in Figma.
- The header Heading is level 2 at its own size, as in Figma; an earlier sketch used `size="level-3"`.
- The space between the header line and the list is `ams.space.l`, measured from the Figma frames (24px narrow, 36px wide).
- A Row wraps only when its content does not fit, so at 360px the heading and the button still sit side by side, where the Figma mock stacks them.
- In the Actions story ‘Verwijderen’ is a Standalone Link, assuming it leads to a confirmation page; a Button works too, but is taller, and so is the separator next to it.

**Naming clash with `datalist`.** The docs note under ‘When not to use’ that Data List is unrelated to the HTML `datalist` element.

**One removed Sass variable, deliberately not marked breaking.** The extraction drops `$ams-description-list-breakpoint`. The `css` package publishes `src/` alongside `dist/`, so a consumer importing the SCSS partial could in principle have referenced it. The documented way to consume the package is the compiled CSS, and Sass variables are not among our conventions to protect, so the title is `feat` rather than `feat!`.

**Extracted below the documented threshold.** We have the convention to extract a shared mixin at three or more occurrences; this is two.

**One stylelint disable added.** `defensive-css/no-fixed-sizes` warns on any size declaration whose value it cannot resolve, and a mixin parameter is such a value. The side effect is that our own `ams/no-fixed-sizes` no longer resolves those two declarations against the built tokens, since the call sites are now `@include` rather than declarations. Both were passing; every value involved is a rem-based clamp times a unitless line height.

**Statistics figure is stale, and not by this PR.** `introduction.docs.mdx` claims 66 components; `develop` already exports 67 modules from the React barrel and this PR adds two.

**Deferred.** For consistency, Description List’s Section could be renamed to Item. That is a breaking change to a released component, so it is left for a separate PR.

[DES-1991]: https://gemeente-amsterdam.atlassian.net/browse/DES-1991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/css/src/common/align-first-line.scss
```diff
@@ -0,0 +1,17 @@
...
+ * Levels the first line of an element with the first line of the element beside it,
+ * where the two use different line heights.
+ * Half the difference in computed line heights goes above the element with the shorter
+ * line box, which centres it against the taller one.
+ * Pass font sizes as tokens in relative units, and line heights as unitless numbers.
+ */
+
+@mixin align-first-line($font-size, $line-height, $reference-font-size, $reference-line-height) {
+  /* stylelint-disable-next-line defensive-css/no-fixed-sizes -- The rule cannot resolve Sass parameters; the tokens passed in are relative units. */
+  padding-block-start: calc(($reference-font-size * $reference-line-height - $font-size * $line-height) / 2);
```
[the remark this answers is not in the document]
RubenSibon: Well observed. I chose to overrule this rule because I felt that abstraction was warranted for this complicated pattern that could easily drift apart when changes happen to one, but not the other.

I’m curious what the other human reviewers think.

packages/css/src/components/index.scss
```diff
@@ -21,6 +21,7 @@
 @use "character-count/character-count";
 @use "checkbox/checkbox";
 @use "column/column";
+@use "data-summary/data-summary";
```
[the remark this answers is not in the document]
RubenSibon: Again a good observation. Our DoD and docs for Agents are out-of-date. We no longer require `README.md`’s for our components with the exception of CSS utilities. We should update those docs.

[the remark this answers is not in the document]
RubenSibon: We’re updating the docs and the DoD with [PR 2967](github.com/Amsterdam/design-system/pull/2967).

storybook/src/components/DataSummary/DataSummary.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <DataSummary.Label>Type</DataSummary.Label>
+        <DataSummary.Value>Ophaalbrug</DataSummary.Value>
+      </DataSummary.Item>,
+      <DataSummary.Item key={3}>
+        <DataSummary.Label>Bouwjaar</DataSummary.Label>
+        <DataSummary.Value>1934</DataSummary.Value>
+      </DataSummary.Item>,
+    ],
+  },
+  decorators: [wrapInInlineSizeQueryContainer('_ams-resize-horizontal', { inlineSize: '18rem' })],
```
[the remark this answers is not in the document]
RubenSibon: This mirrors `DescriptionList.stories.tsx`'s own Orientation story, line-for-line — same helper, same value. Fixing it here alone would make the two siblings inconsistent. Worth a named preset on `wrapInInlineSizeQueryContainer` if we want to fix it everywhere; out of scope for this PR.

storybook/src/components/DataSummary/DataSummary.test.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * Wrapper used by the single Chromatic ‘Test’ story:
+ * - first render example is identical to the original Test story (container context will fallback to Page or none)
+ * - then render example in a narrow container-query context
+ */
+const DataSummaryWithContainerQueryExamples = (props: DataSummaryProps) => (
+  <>
+    <DataSummary {...props} />
+
+    {/* Data Summary in a narrow container */}
+    <div className="ams-query-container-inline-size" style={{ inlineSize: '31.99rem' }}>
```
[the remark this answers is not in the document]
RubenSibon: Same pattern as `DescriptionList.test.stories.tsx`, copied for consistency — the raw `div` carries a class Grid/Column/Row don't expose, and `31.99rem` is the breakpoint minus a hair, not a design value. Two occurrences now, same as the mixin above; a shared test helper is a fair follow-up, not a blocker here.

storybook/src/components/DataSummary/DataSummary.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { PencilIcon } from '@amsterdam/design-system-react-icons'
+import { DataSummary } from '@amsterdam/design-system-react/src'
+import { dataSummaryOrientations } from '@amsterdam/design-system-react/src/DataSummary/DataSummary'
+
+import { wrapInInlineSizeQueryContainer } from '#storybook/_common/decorators'
+
+const meta = {
+  title: 'Components/Text/Data Summary',
+  component: DataSummary,
+  args: {
```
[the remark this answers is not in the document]
RubenSibon: This has been added on [lines 57 to 62](https://github.com/Amsterdam/design-system/pull/2966/changes#diff-68c20bd17cd24ad87e837cafcd60037bca41b051b37ee2e54ab23ffc6a0979ceR57-R62).

packages/react/src/DataSummary/DataSummaryAction.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { forwardRef } from 'react'
+
+export type DataSummaryActionProps = PropsWithChildren<HTMLAttributes<HTMLElement>>
+
+/**
+ * The slot for the link or button that acts on the value of an item of a Data Summary.
+ *
+ * @see {@link https://designsystem.amsterdam/?path=/docs/components-text-data-summary--docs Data Summary docs at Amsterdam Design System}
+ */
+export const DataSummaryAction = forwardRef(
```
VincentSmedinga: Action or Actions?

storybook/src/components/DataSummary/DataSummary.docs.mdx
```diff
@@ -0,0 +1,123 @@
...
+The Item, Label, Value, and Action subcomponents make up a Data Summary.
+An Item groups one label with its value, Label and Value render the `<dt>` and `<dd>` elements, and an Action holds the link or button that changes that value.
+None of them has props of its own; the examples on this page show them in place.
+
+## Usage guidelines
+
+### When to use
+
+- To show a set of data that a reader checks or scans, such as the details of an object, a case, or a person.
+- To let a reader review the answers they gave in a form before they submit it.
```
VincentSmedinga: Using a list here digresses from the convention (since the docs rewrite).

storybook/src/components/DataSummary/DataSummary.docs.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+A label and its value sit side by side, and stack as soon as the container gets too narrow for both.
+Resize the following example to see the Data Summary adapt its layout to the available width.
+
+<Canvas of={DataSummaryStories.Orientation} />
+
+### Stacked
+
+Set the orientation to ‘vertical’ to place every value below its label, however wide the container is.
+Use this when the values are long, or when the summary sits in a narrow column of a wider page.
```
VincentSmedinga: I think we should remove “or when the summary sits in a narrow column of a wider page”, we have the container query to automate this.

storybook/src/components/DataSummary/DataSummary.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+          </StandaloneLink>
+        </DataSummary.Action>
+      </DataSummary.Item>,
+      <DataSummary.Item key={3}>
+        <DataSummary.Label>Onderwerp</DataSummary.Label>
+        <DataSummary.Value>Melding over de openbare ruimte</DataSummary.Value>
+        <DataSummary.Action>
+          <StandaloneLink href="#" icon={PencilIcon}>
+            Wijzigen<span className="ams-visually-hidden"> onderwerp</span>
+          </StandaloneLink>
```
RubenSibon, line 194: This button should be left aligned and not span the entire width of the container.

packages/react/src/DataSummary/DataSummaryAction.tsx
RubenSibon, line 1: - This area’s name should be plural because it may contain multiple actions.
- We’re looking into using an Action Group here.

storybook/src/components/DataSummary/DataSummary.docs.mdx
```diff
@@ -0,0 +1,123 @@
...
+<Canvas of={DataSummaryStories.CompositeValue} />
+
+### Form summary
+
+A summary of the answers in a form is an application of Data Summary rather than a component of its own.
+Combine it with a heading, an action per item that returns to the matching question, and a button that submits the form.
+
+Group data about several people or subjects by placing a heading above each Data Summary.
+
+<Canvas of={DataSummaryStories.FormSummary} />
```
RubenSibon: Do not use Form summary or `FormSummary`. NLDS has a [Community component with that name](https://nldesignsystem.nl/form-summary/). We intend to make Description List usable in the context of summarizing and editing filled-in forms. Those stories should be called something like “Data Summary in a form”.

Side note: we believe NLDS should merge Form Summary into Data Summary and explain better how that differs from a Description List which is very similar in terms of HTML elements and semantics.

RubenSibon: In a later iteration we will add more examples of data summaries in long or multipage forms under Pages -> Internal. The scope of this PR remains just the component with this minimal variant as a single story.

RubenSibon (commented):
We are now considering to rename Data Summary to Data List.

See [this thread](https://github.com/orgs/nl-design-system/discussions/356#discussioncomment-11163878) on NLDS’ discussion page on Data Summary.

packages/react/src/DataSummary/DataSummaryAction.tsx
RubenSibon: - Added an Action Group in the story.
- Did not make the subcomponent plural to keep it consistent with Label and Value.

packages/react/src/DataList/DataList.tsx
```diff
@@ -0,0 +1,63 @@
...
+    </dl>
+  ),
+)
+
+DataListRoot.displayName = 'DataList'
+
+/**
+ * An overview of data, presenting each label with its value on a line of its own.
+ *
+ * @see {@link https://designsystem.amsterdam/?path=/docs/components-containers-data-list--docs Data List docs at Amsterdam Design System}
```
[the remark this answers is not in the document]
RubenSibon: Yeah, I was working on updating the description while you were reviewing. Done!

storybook/src/components/DataList/DataList.docs.mdx
```diff
@@ -0,0 +1,124 @@
...
+### Stacked
+
+Set the orientation to ‘vertical’ to place every value below its label, however wide the container is.
+Use this when the values are long, or when the summary sits in a narrow column of a wider page.
+
+<Canvas of={DataListStories.Stacked} />
+
+### Actions
+
+An item can offer one or more actions on its value, such as a link to the step of a form where that value is entered or a button that opens a dialog to edit the value directly.
```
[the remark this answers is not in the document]
RubenSibon: Nope, we now want to allow for multiple actions. We should still make sure they are seperated properly visually and add this case to a story. This is on the to do list.

packages/css/src/components/data-list/data-list.scss
```diff
@@ -0,0 +1,114 @@
...
+  display: flex;
+
+  // Actions never wrap so that a separator never starts a line. With many actions in a narrow column, their labels wrap instead.
+  flex-wrap: nowrap;
+  margin-block-start: var(--ams-data-list-actions-margin-block-start);
+
+  // A line separates each action from the one before it.
+  > * + * {
+    border-inline-start: var(--ams-data-list-actions-separator-border-inline-start);
+    padding-inline-start: var(--ams-data-list-actions-column-gap);
```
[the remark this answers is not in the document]
RubenSibon: Maybe I should turn off automatic Copilot reviews again because it is annoying that you review before I had the chance to update the description.

And you really seem to take the PR description as the source of truth. It is actually the other way around. The code comes first and the PR description is updated afterwards. Maybe something that we can adjust in the `code-review` skill.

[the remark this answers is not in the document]
RubenSibon: Update: I’ve disabled automatic reviews on every push, only on PR creation.

storybook/src/components/DataList/DataList.docs.mdx
```diff
@@ -0,0 +1,143 @@
+{/* @license CC0-1.0 */}
+
+import { Canvas, Controls, Description, Meta, Primary, Title } from "@storybook/addon-docs/blocks";
+
+import { DesignTokensTable } from "#storybook/_components/DesignTokensTable/DesignTokensTable";
+import tokens from "#tokens/components/ams/data-list.tokens.json";
+
+import * as DataListStories from "./DataList.stories.tsx";
```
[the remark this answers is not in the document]
RubenSibon: Good catch. While it is correct that Data List Header is exported publically, but we really mean for it to be used only together with Data List. For this reason we do not have public documentation for it as a standalone component, but merging the Header’s tokens into that of Data List makes sense.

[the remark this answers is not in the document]
RubenSibon: Your suggestion broke the build process because Storybook doc pages may not contain `const`, only `import` statements. With [this fix commit](https://github.com/Amsterdam/design-system/pull/2966/commits/d960249b9409148bac1e490d32dca736c52f2716) the merging is happening directly in the tokens table.

packages/css/src/components/data-list/data-list.scss
```diff
@@ -0,0 +1,114 @@
...
+  font-size: var(--ams-data-list-label-font-size);
+  font-weight: var(--ams-data-list-label-font-weight);
+  line-height: var(--ams-data-list-label-line-height);
+  margin-block-end: var(--ams-data-list-label-margin-block-end);
+}
+
+.ams-data-list__value {
+  @include reset-dd;
+
+  display: grid;
```
dlnr: Why is the value a grid? It's going to contain inline items and I've not seen and grid rules for children of this component.

packages/css/src/components/data-list/data-list.scss
```diff
@@ -0,0 +1,114 @@
...
+    grid-template-columns: var(--ams-data-list-item-vi-medium-narrow-grid-template-columns);
+  }
+
+  .ams-data-list--horizontal.ams-data-list--medium .ams-data-list__item {
+    grid-template-columns: var(--ams-data-list-item-vi-medium-medium-grid-template-columns);
+  }
+
+  .ams-data-list--horizontal.ams-data-list--wide .ams-data-list__item {
+    grid-template-columns: var(--ams-data-list-item-vi-medium-wide-grid-template-columns);
+  }
```
dlnr, line 69: These grid template setup(s) are not very intuitive or flexible. Take this screen for example:

Image

The value is possibly (in real world usage) a much larger string or maybe even another child component with some form data value. So what I would like to add is `grid-template-areas`, something like this:

```css
.ams-data-list__item:has(.ams-data-list__actions) {
    grid-template-areas:
      "label actions"
      "value value";
    grid-template-columns: 1fr 1fr;
    grid-template-rows: auto auto;

    .ams-data-list__label {
      grid-area: label;
    }

    .ams-data-list__actions {
      grid-area: actions;
      justify-content: end;
    }

    .ams-data-list__value {
      grid-area: value;
      margin-inline-start: 0;
    }
  }
```

Would achieve this:

Image

This is more flexible with less css. There are some overrides like the margin.

packages/css/src/components/data-list/data-list.scss
```diff
@@ -0,0 +1,114 @@
...
+    &:first-of-type {
+      grid-row-start: 1;
+    }
+  }
+
+  .ams-data-list--horizontal .ams-data-list__actions {
+    grid-column-start: 3;
+    grid-row-start: 1;
+    margin-block-start: 0;
+    margin-inline-start: var(--ams-data-list-item-vi-medium-column-gap);
```
dlnr: Thought we didn't use margins to not upset other sibling components?

packages/css/src/components/description-list/description-list.scss
```diff
@@ -3,20 +3,14 @@
...
- * CSS Custom properties cannot be used in container queries.
- * We resort to Sass variables for our breakpoint values.
- */
-
-/** The container width at which Description List toggles its orientation. */
-$ams-description-list-breakpoint: 32rem;
-
 @mixin horizontal-orientation {
-  @container ams-query-container-inline-size (inline-size >= #{$ams-description-list-breakpoint}) {
+  @container ams-query-container-inline-size (inline-size >= #{$ams-container-breakpoint-term-and-value}) {
```
dlnr, line 13: I prefer a query based on `ch` so it's easy to define the line length when it's supposed to break. Based on readability, not a magic number. Our breakpoint should not be a basis for container queries, defeats the purpose.

packages/react/src/DataListHeader/DataListHeader.tsx
dlnr, line 1: Not a big fan of the separate header. I understand a DL doesn't have the option to add a header but DataList could've been a wrapper containing the list. Perhaps even something fancy like:

```html
<section>
  <header>Summary <button>action</button></header>
  <dl>
    <dt>Term</dt><dd>Description</dd><dd><button>action</button></dd>
    <dt>Term</dt><dd>Description</dd><dd><button>action</button></dd>
  </dl>
  <footer>Some footnote or action</footer>
<section>

```

storybook/src/components/DataList/DataList.docs.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+By default the column for the labels is at most `15rem` wide.
+Set `termsWidth` to `narrow`, `medium`, or `wide` when a ratio between labels and values serves the content better.
+Labels and values sit next to each other when their [query container](/docs/utilities-css-query-container--docs) is wide enough (at least `32rem`).
+
+## Accessibility
+
+A Data List renders `dl`, `dt`, and `dd` elements.
+These tie each label to its value in the markup instead of only through the layout, which is the kind of relationship [WCAG 1.3.1](https://www.w3.org/TR/WCAG22/#info-and-relationships) asks to be made available.
+An Item renders a `div` around one pair, which HTML permits inside a `dl`.
```
dlnr, line 129: Is this relevant information?

storybook/src/components/DataList/DataListHeader.test.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </StandaloneLink>
+      </DataList.Actions>
+    </DataList.Item>
+  </DataList>
+)
+
+export const Test: Story = {
+  // Data List Header has no props to vary, so this lists its compositions by hand.
+  render: () => (
+    <Column gap="x-large">
```
dlnr, line 38: This allows you to "play" with the gap value.

packages/css/src/components/data-list-header/data-list-header.scss
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+.ams-data-list-header {
+  border-block-end: var(--ams-data-list-header-border-block-end);
+  padding-block-end: var(--ams-data-list-header-padding-block-end);
+
+  // The header and the Data List below it read as one unit, so the header sets the space between them.
+  &:has(+ .ams-data-list) {
+    margin-block-end: var(--ams-data-list-header-margin-block-end);
+  }
```
dlnr, line 13: So it's possible to use a DataListHeader without the list? What if anybody want's to and it won't add whitespace bellow? They'll tinker with Column gap like in the stories for these components? This might be another good reason to do the child component header rather then the isolated component.

dlnr (commented):
Some of my personal impressions when thinking about using this component.

packages/react/src/DataListHeader/DataListHeader.tsx
RubenSibon: I had discussions with @VincentSmedinga about this and we ended up with the solution for a separate component. In many cases the Data List may not need this particular header with an underline, but the list will always be contained in a wrapper of some kind. Not ideal.

packages/css/src/components/data-list-header/data-list-header.scss
RubenSibon: We do not document the Data List Header as a separate component outside the Data List docs. We instruct developers to only use it in combination with Data List.

This is somewhat similar to how we deal with [Hint](https://github.com/Amsterdam/design-system/tree/develop/packages/react/src/Hint). That component is not publicly and separately documented, but developers could [import it independently](https://github.com/Amsterdam/design-system/blob/a2be66cfe8c7e30f4c7774a59e4eddc0279c6246/packages/react/src/index.ts#L34). 

Admittedly, a difference is that developers do have to import `DataListHeader` themselves according to our docs. That _could_ give them the wrong idea.

I’m curious what @VincentSmedinga thinks about our decision with your concerns in mind.

packages/css/src/components/data-list/data-list.scss
RubenSibon: That seems like a good suggestion. I’ll investigate!

packages/css/src/components/description-list/description-list.scss
RubenSibon: We have been using this breakpoint, specifically created for container queries, for a while now in Description List. We have now renamed it to also be suitable for Data List. So it is not exactly the same as our breakpoints for media queries.

But I’ll investigate on whether we could switch over to queries on `ch` values. But that’s probably a refactor for another ticket.

storybook/src/components/DataList/DataListHeader.test.stories.tsx
RubenSibon: What do you mean?

storybook/src/components/DataList/DataList.docs.mdx
RubenSibon: Maybe not super relevant, but way down in the Design and Accessibility sections we have more room to include “by the way” kind of information that could be useful or interesting for readers interested in the fine details and considerations.

packages/css/src/components/data-list/data-list.scss
```diff
@@ -0,0 +1,114 @@
...
+  column-gap: var(--ams-data-list-actions-column-gap);
+  display: flex;
+
+  // Actions never wrap so that a separator never starts a line. With many actions in a narrow column, their labels wrap instead.
+  flex-wrap: nowrap;
+  margin-block-start: var(--ams-data-list-actions-margin-block-start);
+
+  // A line separates each action from the one before it.
+  > * + * {
+    border-inline-start: var(--ams-data-list-actions-separator-border-inline-start);
```
dlnr: This seems like a break from the general design, we never use vertical separator borders anywhere. Did you get a design for this component? Enough whitespace should allow for enough separation for the buttons.

packages/react/src/DataList/DataListActions.tsx
```diff
@@ -0,0 +1,26 @@
...
+export type DataListActionsProps = PropsWithChildren<HTMLAttributes<HTMLElement>>
+
+/**
+ * Holds the links or buttons that act on the value of an item of a Data List.
+ *
+ * @see {@link https://designsystem.amsterdam/?path=/docs/components-containers-data-list--docs Data List docs at Amsterdam Design System}
+ */
+export const DataListActions = forwardRef(
+  ({ children, className, ...restProps }: DataListActionsProps, ref: ForwardedRef<HTMLElement>) => (
+    <dd {...restProps} className={clsx('ams-data-list__actions', className)} ref={ref}>
```
dlnr, line 22: Didn't we discuss using the ActionGroup component? So that standalone links and tertiary buttons would be aligned properly?

RubenSibon: Indeed, I should’ve used Action Group.

packages/css/src/components/data-list/data-list.scss
RubenSibon: I’ll remove the vertical separators.

We already have the horizontal underline under the header without precedent,  but which does come from the design. But let’s not add even more lines undocumented. :-)

packages/css/src/components/data-list/data-list.scss
RubenSibon: So, I’ve looked into it and my conclusion is that this needs a design decision.

Your suggestion puts label and actions on one row with the value underneath, but only for items that have actions.

In the second screenshot, “Johan” then starts at the left edge while the bsn stays in the value column, so the values no longer line up. So that seems like a downgrade to me.

Named areas would also stack a second value in the same cell as the first, where today it moves to the next row.

But I also do see some improvements for how the layout stacks vertically. We need to discuss this with @bas-kroese and @VincentSmedinga.

packages/css/src/components/data-list/data-list.scss
RubenSibon: You’re right, a value is plain text flow now. Values with several parts wrap them in a Column, as the composite value story shows.

packages/css/src/components/data-list/data-list.scss
RubenSibon: Replaced: the item now uses `column-gap` and the value and actions no longer have margins.

packages/react/src/DataList/DataListActions.tsx
RubenSibon: Done: `DataList.Actions` now puts its children in an Action Group, so links and tertiary buttons line up and wrap, without each item needing its own wrapper.

Wondering if we should also use Action Group for the single action in the Data List Header. Action Group is intended for “one or more actions” after all.

Also, why do we require at most one action in the header? Can’t it be two such as “Wijzigen” en “Verwijderen”?

What do you say @VincentSmedinga?

packages/css/src/components/data-list/data-list.scss
dlnr: I've added the `:has()` to illustrate you could only apply this to items with actions. It's not a must. The current solution doesn't really work for me, but there are many (other) ways to make the layout collapse when space is tight. My solution is consistent with header rows that contain actions, this helps a lot if your interface has a lot of editable parts.

packages/react/src/DataListHeader/DataListHeader.tsx
dlnr: Not necessarily, you can do a lot conditionally if there is no header.

storybook/src/components/DataList/DataListHeader.test.stories.tsx
dlnr: I mean a developer kan randomly pick a gap value.

packages/react/src/DataList/DataListActions.tsx
dlnr: It should of no concern to us how many actions they choose to use, it can be 1 or 12 if you ask me.

storybook/src/components/DataList/DataListHeader.test.stories.tsx
RubenSibon: Yeah so? Why is this relevant?

This is in the test stories for Chromatic... not something that developers will see.

packages/react/src/DataListHeader/DataListHeader.tsx
RubenSibon: My preference would be that components are as “pure” as possible: in principle what element renders as the root should not change based on certain conditions being met.

Also, the name of the component wouldn’t really be appropriate if in some instances you get 2 components wrapped together, one of which is not a list, but a header, and in other instances you get just a `dl`.

storybook/src/components/DataList/DataList.docs.mdx
dlnr: I don't know I thought we only added super relevant information to keep the docs as short as possible. I wouldn't rant about random semantic fact from the mdn or w3c documentation. Maybe you can link relevant documentation about roles and such?

storybook/src/components/DataList/DataListHeader.test.stories.tsx
dlnr: Right the "Test" story is hidden.

storybook/src/components/DataList/DataList.docs.mdx
RubenSibon: Yeah maybe... I’ll wait for what @VincentSmedinga thinks.

RubenSibon (commented):
Change in direction:

- For this PR we only deliver a simple Data List component without a seperate or integrated header component. That pattern might come in a later iteration.
- We do include a story with the header pattern with a Heading and Action Group in a Row above the Data List. There will not be an underline.

PR #2972, opened 2026-09-15
feat(Spinner): Add a component for loading states

Arman-Kokor:
# Describe the pull request

## Links

N/A

## What

The **Spinner** component is a circular loading indicator for loading states where the duration is short or unknown. 

- Three sizes for the spinner: `small`, `medium`, and `large`.
- Each size has its own diameter and border width to keep the proportions of the spinner consistent.

## Why

The design system needs a simple loading indicator for situations where something is loading for a short or unknown amount of time.

A Spinner gives applications a consistent way to communicate that something is still loading without needing to create their own loading indicator.

The three sizes allow the Spinner to be used in different contexts, from inline loading states and buttons to larger page.

## How

- The Spinner uses the ams-spinner BEM class and a size modifier class for each available size: small, medium, and large.

- The spinner is created with a circular border. Three sides use the track colour while the start border uses the current progress colour, making the rotating indicator visible.

- The rotation is handled by the ams-spinner-rotate keyframes. The animation is exposed through the ams-spinner-animation token, allowing the duration, timing function, keyframes or animation itself to be changed.

- Size and border width are controlled through separate tokens for each size. This allows the border to scale with the Spinner instead of using the same border width for every size.

- The Spinner container renders aria-hidden="true" by default. The documentation explains how to announce the loading state with a status message and aria-busy (WCAG 4.1.3).

## Checklist

- [x] Add or update stories

## Additional notes

- The track colour currently uses #d1d1d1, since there is no fitting grey token available for the Spinner.

- The Spinner is animated by default. It runs only for users who have not requested reduced motion and the whole animation sits behind a single --ams-spinner-animation token, so a theme can retime it, point it at its own keyframes or switch it off.

- The border width is defined separately for each Spinner size rather than as one shared value, so the visual weight of the border remains appropriate for small, medium, and large Spinners.


RubenSibon (commented):
Your first PR for ADS! Hooray! 🥳

I’ll go through it carefully when I have the time. We’ve already reviewed together to see if it was at least ready to be a PR. I think it was.

I asked Copilot to review as we often do with PRs and it came with many comments. That might be overwhelming. Sorry for that! Just go through each one-by-one and see if Copilot has a point (their comments are not always relevant or useful).

packages-proprietary/tokens/src/components/ams/spinner.tokens.json
```diff
@@ -0,0 +1,71 @@
+{
+  "ams": {
+    "spinner": {
+      "animation": {
+        "$value": "ams-spinner-rotate 0.8s linear infinite",
+        "$description": "The loading animation."
+      },
+      "track-color": {
+        "$value": "#D4D4D4",
```
VincentSmedinga: This colour isn’t in our palette, let’s update to the closest shade of grey. Not using a token here is correct; we don’t have a semantic representation for this kind of usage yet.

```suggestion
        "$value": "##d1d1d1",
```

packages/css/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,34 @@
...
+
+@use "animation";
+
+.ams-spinner {
+  aspect-ratio: 1;
+  border-block-end-color: var(--ams-spinner-track-color);
+  border-block-start-color: var(--ams-color-progress-current);
+  border-inline-end-color: var(--ams-spinner-track-color);
+  border-inline-start-color: var(--ams-spinner-track-color);
+  border-radius: 50%;
```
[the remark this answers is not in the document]
VincentSmedinga: This has been added in the meantime.

packages/css/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,41 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@use "animation";
+
+.ams-spinner {
+  aspect-ratio: 1;
+  border-block-end-color: var(--ams-spinner-track-color);
+  border-block-start-color: var(--ams-color-progress-current);
```
VincentSmedinga: Even though the linter forced it, it’s pretty inventive to specify all 4 borders to get a quarter circle arc :+1:

packages/css/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,41 @@
...
+  border-inline-end-color: var(--ams-spinner-track-color);
+  border-inline-start-color: var(--ams-spinner-track-color);
+  border-radius: 50%;
+  border-style: solid;
+  box-sizing: border-box;
+  @media (forced-colors: active) {
```
VincentSmedinga: We use an empty line before nested blocks – not sure why the linter didn’t catch this.

```suggestion

    @media (forced-colors: active) {
```

packages/react/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+import type { ForwardedRef, HTMLAttributes } from 'react'
+
+import { clsx } from 'clsx'
+import { forwardRef } from 'react'
+
+export const spinnerSizeOptions = ['small', 'medium', 'large'] as const
```
VincentSmedinga: We use a plural here for other components; it’s simpler: `spinnerSizes`. All 5 occurrences should change.

storybook/src/components/Spinner/Spinner.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    size: 'medium',
+  },
+}
```
VincentSmedinga: The `size` argument isn’t needed here; `medium` is already its default. Because it shows up in the code view, people would unnecessarily copy/paste it along.

```suggestion
export const Default: Story = {}
```

packages/react/src/Spinner/Spinner.tsx
VincentSmedinga: That said, I’m not sure about the sizes yet. We haven’t researched yet when to use a Spinner versus a Skeleton, even though the docs in this PR suggest guidelines for that choice.

One option I can see is a spinner in front of the label of a button, after it has been pressed. The docs also mention it as the first example. That would mean we’d need an option that aligns with the regular size of body text, like Icon Button does through `IconProps`. But introducing all these sizes here feeld a bit premature as well.

Additionally, the grey and blue colours don’t work in a blue primary Button. One more thing to discuss with UX.

packages/react/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,32 @@
...
+} & Readonly<HTMLAttributes<HTMLDivElement>>
+
+/**
+ * A spinning circle that is used for a short or unknown amount of time when something on the page is loading.
+ *
+ * @see {@link https://designsystem.amsterdam/?path=/docs/components-feedback-spinner--docs Spinner docs at Amsterdam Design System}
+ */
+export const Spinner = forwardRef(
+  ({ className, size = 'medium', ...restProps }: SpinnerProps, ref: ForwardedRef<HTMLDivElement>) => (
+    <div {...restProps} aria-hidden className={clsx('ams-spinner', `ams-spinner--${size}`, className)} ref={ref} />
```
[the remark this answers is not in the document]
VincentSmedinga: Yes, let’s do this. It requires updating some types as well.

packages/css/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+@use "animation";
+
+.ams-spinner {
+  aspect-ratio: 1;
+  border-block-end-color: var(--ams-spinner-track-color);
```
VincentSmedinga: Even though the linter forced it, it’s pretty inventive to specify all 4 borders to get a quarter circle arc :+1:

storybook/src/components/Spinner/Spinner.docs.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+## Usage guidelines
+
+### When to use
+
+Use a Spinner when content is loading for a short or unknown amount of time. For example, when a button is submitting a form or a widget is refreshing its data.
+
+### When not to use
+
+Do not use a Spinner when the duration or progress is known, such as with a file upload. Show the actual progress instead.
```
VincentSmedinga: We put every sentence on its own line in Markdown, so that changing a sentence doesn’t mark other sentences in a Git diff, and the update is easier to handle.

packages-proprietary/tokens/src/components/ams/spinner.tokens.json
```diff
@@ -0,0 +1,71 @@
...
+        "$description": "The colour of the ring's resting track.",
+        "$extensions": {
+          "nl.amsterdam.type": "color"
+        }
+      },
+      "small": {
+        "size": {
```
VincentSmedinga: Component tokens end in the property they set, and `size` isn’t one. It gets that name because the CSS sets both `block-size` and `inline-size` from it.

Could you set one dimension and let `aspect-ratio: 1` supply the other, so the tokens become `ams.spinner.small.block-size` and so on?

That is how the Progress List marker draws its circle: `aspect-ratio: 1` with `border-radius: 50%`, sized by `ams.progress-list.step.marker.shape.block-size` alone.

packages/css/src/components/spinner/_animation.scss
```diff
@@ -0,0 +1,17 @@
...
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+// A steady rotation signals that loading is under way. Users who prefer reduced motion keep a static ring.
+@media (prefers-reduced-motion: no-preference) {
+  .ams-spinner {
+    animation: var(--ams-spinner-animation);
+  }
+}
```
VincentSmedinga: I understand how this reduces motion, but I’m unsure whether we should serve people who set this preference a static ring 😁

packages/css/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,41 @@
...
+    border-block-start-color: Highlight;
+    border-inline-end-color: GrayText;
+    border-inline-start-color: GrayText;
+  }
+}
+
+.ams-spinner--small {
+  block-size: var(--ams-spinner-small-size);
+  border-width: var(--ams-spinner-small-border-width);
+  inline-size: var(--ams-spinner-small-size);
```
VincentSmedinga: We only need `block-size` in these three; `aspect-ratio` takes care of the width.

packages/css/src/components/spinner/_animation.scss
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license EUPL-1.2+
+ * Copyright Gemeente Amsterdam
+ */
+
+// A steady rotation signals that loading is under way. Users who prefer reduced motion keep a static ring.
```
VincentSmedinga: A static ring can read as broken rather than loading. [WCAG’s understanding of 2.2.2](https://www.w3.org/WAI/WCAG22/Understanding/pause-stop-hide.html) treats a preload animation as essential partly for that reason: an indicator that looks frozen makes people think the content is stuck. And `prefers-reduced-motion` is defined around non-essential motion, where `reduce` means an interface that removes or replaces it.

Let’s slow the rotation for these users rather than stop it, behind a second token? A ring this small is at the low-risk end of the vestibular triggers, so there is room to let some motion through. This would require a new token.

storybook/src/components/Spinner/Spinner.docs.mdx
VincentSmedinga: I’ve skipped reviewing the content of the docs for now; let’s await the discussion with UX as they might change the component’s appearance or behaviour.

storybook/src/components/Spinner/Spinner.stories.tsx
VincentSmedinga: What could be an representative example of where the spinner could be used? If we can find one, let’s add that as well.

VincentSmedinga (changes requested):
@Arman-Kokor Well done! I’m sure we can get this to a state where we can merge it.

Most of my comments are about things you could hardly know: naming conventions, empty lines etc.

Some others are more fundamental: how we should approach this components. That’s mostly because we didn’t write a spec first, so totally understandable as well. Let’s settle those together with UX.

PR #2973, opened 2026-09-17
fix(Description List): Add usage guidelines

RubenSibon:
# Add usage guidelines to the Description List documentation

## Links

- [Description List documentation](https://designsystem.amsterdam/?path=/docs/components-text-description-list--docs#usage-guidelines) – the page with the empty section
- [DES-2005](https://gemeente-amsterdam.atlassian.net/browse/DES-2005)

## What

- Add a ‘When to use’ subsection that names the kinds of content a Description List suits: properties of a subject, a glossary, or a summary of submitted data.
- Add a ‘When not to use’ subsection that points to Table, Unordered List, and Ordered List for adjacent scenarios.
- Add a ‘How to use’ subsection about vertical whitespace, term length, and the width of the terms column.

## Why

The ‘Usage guidelines’ heading on the Description List page had no content.
It was introduced empty in #2670, which applied the shared content model to all component pages.
There were no earlier guidelines to restore: the CSS README it replaced only contained Design and References sections, and no version of the documentation since #1139 described usage.

## How

The subsections follow the content model in `documentation/component-docs.md`.

- ‘When not to use’ separates the Description List from its closest alternatives by content shape: a Table compares several items by the same properties, and a plain list has items without a term.
- ‘How to use’ reuses the sentence about Margin and Prose from Ordered List and Unordered List, so the list components give the same spacing advice.
- The advice on term length and column width connects to the `termsWidth` prop and to the layout described in the Design section.

Deliberately left out:

- No ‘How to write’ subsection, since the component has no text conventions of its own beyond those of the content it holds.
- No new stories: the existing examples already cover the configurations the guidelines mention.
- The guidance under ‘Multiple terms and descriptions’ in Examples is really a usage rule, but moving it is left for a separate change to keep this one focused.

## Checklist

- [x] Add or update stories

## Additional notes

These guidelines are newly written, not restored, so please review the content rather than just the formatting.

Feedback is especially welcome on:

- whether the examples in ‘When to use’ match how teams use the component in practice
- whether other alternatives belong in ‘When not to use’

[DES-2005]: https://gemeente-amsterdam.atlassian.net/browse/DES-2005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


storybook/src/components/DescriptionList/DescriptionList.docs.mdx
```diff
@@ -25,6 +25,22 @@ None of them has props of its own; the examples on this page show them in place.
 
 ## Usage guidelines
 
+### When to use
+
+Use a Description List for pairs of terms and descriptions, such as the properties of a subject, a glossary, or a summary of submitted data.
```
VincentSmedinga: Timing is not ideal here; both the properties of a subject and a summary of submitted data require a Data List as soon as #2966 has been merged.

VincentSmedinga (commented):
Let’s wait for #2966 so we can write these guidelines in one pass.

storybook/src/components/DescriptionList/DescriptionList.docs.mdx
RubenSibon: I do not see an issue with merging this small fix before the Data List PR. Wouldn’t it actually be better to merge this one first so that we can add Data List to the When not to use section in Description List?

RubenSibon: Maybe it is better to remove the “, or a summary of submitted data” from this PR and then merge this one first.

RubenSibon:
```suggestion
Use a Description List for pairs of terms and descriptions, such as a glossary.
```

VincentSmedinga (approved):

